PR #5, opened 2016-02-26
Merge the new-proposal branch from yast2-storage

ancorgs:
* Import the code in that branch keeping the history
* Some small fixes needed for `rake osc:build` to work


src/lib/expert-partitioner/dialogs/create-partition.rb
```diff
@@ -62,7 +82,7 @@ def doit
 
       storage = Yast::Storage::StorageManager.instance
 
-      staging = storage.staging()
+      storage.staging()
```
aschnell: If the variable is not used this is a nop.

src/examples/fake-devicegraphs-yaml-format.md
```diff
@@ -0,0 +1,304 @@
+# YAML File Format For Fake Device Graphs
```
shundhammer: With this repo restructuring, we might want to move this document to a toplevel doc/ directory.

shundhammer:
Apart from moving the .md doc to a doc/ directory and getting rid of storage-manager.rb vs. storage_manager.rb, LGTM.

We might still want to update the toplevel README.md which is not very much fitting any more.

PR #8, opened 2016-03-01
[Do not merge yet] Rubocop

ancorgs:
Still a work in progress


ancorgs:
Replaced by #10 (the same but without deleting `return`).

PR #10, opened 2016-03-03
Tuned Rubocop

ancorgs:


.rubocop.yml
```diff
@@ -0,0 +1,45 @@
...
+  /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml
+
+# Redundant returns add legibility for developers used to other languages
+Style/RedundantReturn:
+  Enabled: false
+
+# Don't complain on definitions like this. We like question marks!
+# def some_boolean?; some_boolean; end
+TrivialAccessors:
+  ExactNameMatch: true
```
jreidinger: it is wrong one, it is actually option https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L895 but I also use this, but its goal is to allow non exact match so something like
```
def name
  @intenal_name
end
```
AllowPredicates is about question mark at the end, so it is subset of exact match

src/lib/expert_partitioner/dialogs/format.rb
```diff
@@ -61,27 +57,21 @@ def run
       end
     end
 
-
-  private
+    private
```
jreidinger: to be honest I actually very dislike this indentation, but if storage subteam agreed on it, no problem.
For me main reason why I do not like is that it is global modificator and when inlined into methods indentation it can be overlooked so

```
class C
  trash
  ....
  ...

  private
  
  ....
  trash
  ...
end
```
is easier to overlook what is private then
```
class C
  trash
  ....
  ...

private
  
  ....
  trash
  ...
end
```

so my project enforce style with outer indentation of private

src/lib/expert_partitioner/dialogs/format.rb
```diff
@@ -90,30 +80,32 @@ def create_dialog
...
+    def filesystem_items
+      [
+        Item(Id(Storage::FsType_EXT4), "Ext4"),
+        Item(Id(Storage::FsType_XFS), "XFS"),
+        Item(Id(Storage::FsType_BTRFS), "Btrfs"),
+        Item(Id(Storage::FsType_SWAP), "Swap"),
+        Item(Id(Storage::FsType_NTFS), "NTFS"),
+        Item(Id(Storage::FsType_VFAT), "VFAT")
+      ]
+    end
```
jreidinger: we can be even more strict and have it as constant as there is no translation.

src/lib/expert_partitioner/dialogs/format.rb
```diff
@@ -90,30 +80,32 @@ def create_dialog
...
     def doit
-      begin
-
-        log.info "doit #{@blk_device.name}"
+      log.info "doit #{@blk_device.name}"
 
-        if !RemoveDescendantsPopup.new(@blk_device).run()
-          return
-        end
+      return if !RemoveDescendantsPopup.new(@blk_device).run
```
jreidinger: `unless`? I do not like unless in multiline if but for trailing one it is encouraged even by suse style guide.

src/lib/expert_partitioner/dialogs/format.rb
```diff
@@ -90,30 +80,32 @@ def create_dialog
...
-      rescue Storage::DeviceHasWrongType
-        log.error "doit on non blk device"
+      mount_point = Yast::UI.QueryWidget(:mount_point, :Value)
+      if !mount_point.empty?
+        log.info "doit mount-point #{mount_point}"
+        filesystem.add_mountpoint(mount_point)
       end
 
+    rescue Storage::DeviceHasWrongType
+      log.error "doit on non blk device"
```
jreidinger: not related to this pr, but probably something for tuesday. But using exception for common work-flow like block devices is the most discouraged think I know about using exception.

src/lib/expert_partitioner/tree_views/disks.rb
```diff
@@ -108,70 +97,57 @@ def items
...
         Yast::Popup::Error("Only partitions can be deleted.")
         return
       end
 
-      if RemoveDescendantsPopup.new(partition).run()
-        staging.remove_device(partition)
-        update(true)
-      end
-
+      return unless RemoveDescendantsPopup.new(partition).run
```
jreidinger: just my matter of taste but I prefer for better readability empty line between early exit and rest of method to more emphasise it.

src/lib/expert_partitioner/dialogs/format.rb
lslezak: JFYI: you can have translations even for constants, see https://github.com/webyast/webyast/wiki/Localization#translating-constants

*(I should move some parts this doc to YaST...)*

src/lib/storage/abstract_device_factory.rb
```diff
@@ -131,12 +133,12 @@ def build_tree_recursive(parent, name, content)
           check_param(name, content.keys)
 
           # Split up pure parameters and sub-product descriptions
-          sub_prod = content.select{ |k,v|  factory_products.include?(k) }
-          param    = content.select{ |k,v| !factory_products.include?(k) }
+          sub_prod = content.select { |k, _v|  factory_products.include?(k) }
+          param    = content.select { |k, _v| !factory_products.include?(k) }
```
jreidinger: not related to this pr, but when I see these two lines I immediatelly get idea using `partition` like
```
sub_prod, param = content.partition { |k, _v|  factory_products.include?(k) }
```

src/lib/storage/abstract_device_factory.rb
```diff
@@ -163,8 +165,12 @@ def build_tree_recursive(parent, name, content)
         end
       end
 
+      # rubocop:disable Lint/UselessAccessModifier
+
       protected
 
+      # rubocop:enable Lint/UselessAccessModifier
+
```
jreidinger: actually what is a goal of this protected? Also I think that documentation is at wrong place. Where yardoc will render it? I think for such stuff documentation have to be in class documentation itself so when you subclass you usually read top level documentation. Another way how to do it, is to make real methods and use e.g. abstract_method gem that is already used by yast2 to mark methods as purely abstract.

src/lib/storage/abstract_device_factory.rb
```diff
@@ -242,7 +248,7 @@ def factory_methods
       # @return [Array<String>] product names
       #
       def factory_products
-        if @factory_products_cache == nil
+        if @factory_products_cache.nil?
```
jreidinger: as cache is not boolean then simple `if !@factory_products_cache`

src/lib/storage/extensions.rb
```diff
@@ -20,26 +20,21 @@
...
+    silencer = ::Storage::Silencer.new
     yield
   ensure
     silencer.turn_off
   end
 
-
   class Device
-
+    # rubocop:disable Metrics/CyclomaticComplexity
```
jreidinger: actually it make sense to say that this method too complex. Unless there is `Right` call to size, which looks a bit inconsistent to rest you can use constant with `FIELD_ID_TO_NAME = [ sid: N_("Storage ID") ]` and then here `fields.map { |f| _(FIELD_ID_TO_NAME[f]) }`. If right still have to be used, still for me it is easier to have here this hash defined as variable (having one value containing Right) and then just look at it. It is more effective and also easier to read and extend.

src/lib/storage/proposal/devicegraph_generator.rb
```diff
@@ -54,12 +54,12 @@ def devicegraph(volumes, initial_graph: ::Storage::Devicegraph.probed)
...
         end
 
-      protected
+        protected
 
         attr_writer :got_desired_space
 
         def got_desired_space?
-          !!@got_desired_space
+          @got_desired_space
```
jreidinger: this is actually not equivalent, as `!!` ensure boolean value.

src/lib/storage/proposal/partition_creator.rb
```diff
@@ -205,19 +172,35 @@ def distribute_extra_space(volumes, extra_size)
...
+          extra_size = available_size * (volume.weight / total_weight)
+          new_size = extra_size + volume.size
+          new_size > volume.max_size ? volume.max_size : extra_size
+        end
+
         # Create partitions without LVM in the complex case: There are multiple
         # slots of free disk space, so we need to fit the volumes as good as
         # possible.
+        def create_non_lvm_complex
+          raise "Not implemented yet"
```
jreidinger: there is actually specialized exception for it..something like NotImplemented

src/lib/storage/proposal/partition_creator.rb
```diff
@@ -350,7 +345,10 @@ def create_physical_volumes(volume_group)
         # @param strategy     [Symbol] :desired or :min_size
         #
         def create_logical_volume(volume_group, vol, strategy)
-          log.info("Creating LVM logical volume #{vol.logical_volume_name} with strategy \"#{strategy}\"")
+          log.info(
+            "Creating LVM logical volume #{vol.logical_volume_name} at #{volume_group} "\
+            "with strategy \"#{strategy}\""
+          )
```
jreidinger: a lot of changes in this file. Is it covered by tests?

src/lib/storage/yaml_writer.rb
```diff
@@ -119,30 +107,52 @@ def yaml_disk(disk)
...
       # @return [Hash]
       #
+      # rubocop:disable Metrics/AbcSize
       def yaml_partition(partition)
         content = {}
         content["size"] = DiskSize.new(partition.region.to_kb(partition.region.length)).to_s
         content["name"] = partition.name
         content["type"] = @partition_types[partition.type]
-        content["id"  ] = @partition_ids[partition.id]
+        content["id"] = @partition_ids[partition.id]
```
jreidinger: well, this break intention of aligned assignment, but it also reveals fact, that there is not reason to not direct add it to hash like:
```
content = {
  "size"  => DiskSize.new(partition.region.to_kb(partition.region.length)).to_s,
  "name => partition.name,
  ...
}
```
which itself is proper aligned and it is ensured by rubocop

jreidinger:
in general I like it, just few minor stuff

src/lib/expert_partitioner/dialogs/format.rb
aschnell: Calling the format dialog for a non-block-device is plain wrong so the exception here is the right thing to do.

src/lib/storage/extensions.rb
aschnell: That function is of the simplest kind.

src/lib/expert_partitioner/dialogs/format.rb
jreidinger: @aschnell Ah, OK. I do not get it from log msg. Maybe improve a log message a bit? And is it safe at all to just ignore exception?

src/lib/storage/extensions.rb
jreidinger: I disagree it do
1) create Header term
2) remap fields from ids to its name
3) align Size element to right ( I do not get why, some comment would be helpful )

BTW it also have so bit cyclomatic complexity because it do mapping inside of Header call, which is also quite discouraged to have multiline map inside method call arguments. I think even simple
```
names = fields.map do |field|
  ...
end

return Header(*names)
```
will help with cyclomatic complain.

src/lib/expert_partitioner/dialogs/format.rb
kobliha: I'm for the second option

src/lib/expert_partitioner/dialogs/format.rb
ancorgs: Sure there are a lot of things we can do... but let's not overdo in this pull request.

src/lib/expert_partitioner/dialogs/format.rb
ancorgs: As stated in the Scrum call, I also prefer the alignment you mention. But this is actually the YaST team rubocop configuration (not something agreed by the storage subteam). We should not add a per-repository difference.

.rubocop.yml
ancorgs: cool. Will fix it.

src/lib/expert_partitioner/dialogs/format.rb
ancorgs: Yep, my fault. I'll fix it.

src/lib/storage/abstract_device_factory.rb
ancorgs: I agree with you. But if I remember properly, the convenience of using abstract_method vs these comments was already discussed. Let's not restart that discussion (at least here).

src/lib/storage/proposal/devicegraph_generator.rb
ancorgs: Then we should add another exception to the general YaST team rubocop configuration. Another setting we should not override locally.

src/lib/storage/proposal/partition_creator.rb
ancorgs: Yes, I corrected that in many other places. I missed this one. Will fix it.

src/lib/storage/proposal/partition_creator.rb
ancorgs: Nothing is currently covered by tests. There is a separate PBI for that. At this point in time, in code review we trust.

ancorgs: Well, in code review and in manual testing. The proposal behavior is exactly the same in two systems before and after the changes.

src/lib/expert_partitioner/dialogs/format.rb
jreidinger: ok, just expect to fix it if my pr will be merged. Actually this one is already in two yast repositories e.g.
https://github.com/yast/yast-yast2/blob/master/.rubocop.yml#L33
https://github.com/yast/yast-bootloader/blob/master/.rubocop.yml#L95

jreidinger:
LGTM

PR #12, opened 2016-03-08
Adapted to new OBS location

ancorgs:
@aschnell maybe it would be a good idea to delete `home:aschnell:storage-redesign` to avoid future confusion.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -20,5 +20,7 @@ require "yast/rake"
 
 Yast::Tasks.configuration do |conf|
   # The package does not live in the official YaST:Head OBS project
-  conf.obs_project = "home:aschnell:storage-redesign"
+  conf.obs_project = "YaST:storage-ng"
+  # Non-existent project to prevent accidental submission to Tumbleweed
+  conf.obs_sr_project = "NONE"
 end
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -20,5 +20,7 @@ require "yast/rake"
 
 Yast::Tasks.configuration do |conf|
   # The package does not live in the official YaST:Head OBS project
-  conf.obs_project = "home:aschnell:storage-redesign"
+  conf.obs_project = "YaST:storage-ng"
+  # Non-existent project to prevent accidental submission to Tumbleweed
+  conf.obs_sr_project = "NONE"
```
lslezak: Um, why not `nil`? It does not work?

ancorgs: Our tasks are not prepared to skip the request in case of `nil`. In the best case, setting it to `nil` is equivalent to setting it to the default value (i.e. Factory). In the worst case is equivalent to setting it to the empty string... thus invalid osc commands.

ancorgs: Well, not sure which case is worst or better actually :wink: But none of them are desirable.

lslezak: I see, I thought `nil` could mean none, but it's the default...

lslezak:
LGTM

PR #14, opened 2016-03-09
Proof of concept of some RSpec test for Proposal

ancorgs:
* This pull request mixes fixes and test. It's just to discuss
 * Too much code is needed in the RSpec file in order to fake and
   to compare devicegraphs. Let's think about it.
 * The tests are actually failing right now, because the generated
   tree has a wrong "free" entry
 * Turns out our proposal is quite destructive with windows :)


shundhammer:
I agree that we might want to move much of that checking code out from the RSpec file. Maybe it's time for some "devicegraph-helpers.rb" module?

Other than that, LGTM; we'll keep working on that for a while anyway, so this will see changes in the future anyway.

ancorgs:
Updated (`push -f`) Now it even works :-)

shundhammer:
LGTM

src/lib/storage/disk_analyzer.rb
```diff
@@ -228,7 +228,7 @@ def find_windows_partitions
             disk = ::Storage::Disk.find(devicegraph, disk_name)
             disk.partition_table.partitions.each do |partition|
               if WINDOWS_PARTITION_IDS.include?(partition.id)
-                windows_partitions << partition.name if windows_partition?(partition.name)
+                windows_partitions << partition.name if windows_partition?(partition)
```
jreidinger: when I see this code just small suggestion. What about putting check for `WINDOWS_PARTITION_IDS.include?` also to `windows_partition?` method? Now it is already quite deep nesting and it helps a bit.

ancorgs:
Removed all the workarounds (like `256` instead of `range`), split in meaningful commits and pleased Rubocop.

src/lib/storage/proposal/partition_creator.rb
```diff
@@ -230,6 +230,8 @@ def create_partition(vol, partition_id, free_slot)
             partition
           # FIXME: rescue ::Storage::Exception when SWIG bindings are fixed
           rescue RuntimeError => ex
+            # Don't hide our own exceptions
+            raise if ex.is_a?(Yast::Storage::Proposal::Error)
```
jreidinger: maybe less tricky to let ruby exception handler to recognize it, so something like
```
# Don't hide our own exceptions
rescue Yast::Storage::Proposal::Error
  raise
rescue RuntimeError => ex
  ...
```

.rubocop.yml
```diff
@@ -38,3 +38,12 @@ Lint/HandleExceptions:
 Style/MethodName:
   Exclude:
     - 'src/lib/expert_partitioner/ui_extensions.rb'
+
+# This two are not equivalent if using refinements (and we use them to add
+# functionality to the classes coming from libstorage)
+#   using RefinementAddingFooToThing
+#   things.map { |s| s.foo } # it works
+#   things.map(&:foo) # it does not work
+# See https://bugs.ruby-lang.org/issues/9786
```
mvidner: It is a RuboCop bug, not a Ruby bug.
It is already reported: https://github.com/bbatsov/rubocop/issues/2720

ancorgs: I didn't want to imply it was a ruby bug, just wanted to link to a good explanation. But I will add a link to the rubocop bug.

src/lib/storage/proposal/partition_creator.rb
ancorgs: Definitely. I don't know what I was thinking :smile: Will fix it.

src/lib/storage/disk_analyzer.rb
```diff
@@ -146,18 +141,25 @@ def find_installation_disks
...
+        dev_names(candidate_disk_objects)
+      end
+
+      # Array with all disks in the devicegraph
+      #
+      # @return [Array<::Storage::Disk>]
+      def all_disks
+        # FIXME: to_a should not be necessary:
+        # libstorage should return something that Ruby can handle.
+        # This is very likely a problem of the Swig bindings.
```
aschnell: Swig does report something that ruby can handle. But it is not a ruby array.

ancorgs: I just wanted to keep the comment added by @shundhammer, since it contains a `FIXME`. For me it's fine in the current state.

ancorgs:
Travis didn't passed because the libstorage-ng package for ubuntu need an update

ancorgs:
@shundhammer Mmmm, just to be sure. Imagine I have 10 cylinders (enough for my example).

A primary partition from cyl0 to cyl1 (both inclusive). And then an extended one from cyl2 to cyl7. Finally cyl8 and cyl9 are free. Inside that extended partition I have one logical from cyl2 to 3 and another one from cyl4 to cyl5. cyl6 and cyl7 are assigned to the extended partition but not to any logical partition.

The Yaml should be something like this (with many poetic licenses), isn't it?
```yml
- partition:
  type: primary
  size: two cylinders
- partition:
  type: extended
  size: 6 cylinders
- partition:
  type: logical
  size: 2 cylinders
- partition:
  type: logical
  size: 2 cylinders
- free:
  size: 2 cylinders
```
I mean, the 2 cylinders that are part of the extended partition but not assigned to any logical are not expected to be present in the Yaml (they are not free). Right?

ancorgs:
On the other hand, if the second logical partition is in cyl6-7 instead of cyl4-5 the yaml version would look exactly the same, which probably is not a good idea.

shundhammer:
Right: The YAML writer only creates an empty slot if there is a partition after it. For empty space after the last partition, no slot is created. But this is somewhat arbitrary behaviour that should simply make it easier to write those things manually.

ancorgs:
Mmm, broken travis. I'll check

src/lib/storage/fake_device_factory.rb
```diff
@@ -162,6 +162,9 @@ def create_disk(_parent, args)
         @disks << name
         disk = ::Storage::Disk.create(@devicegraph, name)
         disk.size_k = size.size_k
+        # range (number of partitions that the kernel can handle) used to be
+        # 16 for scsi and 64 for ide. Now it's 256 for most of them.
+        disk.range = args["range"] || 256
```
jreidinger: I prefer having 256 as constant as it is more descriptive, but it is used only here and with comment, it is fine for me.

src/lib/storage/proposal/partition_creator.rb
```diff
@@ -217,7 +217,7 @@ def create_partition(vol, partition_id, free_slot)
             disk = ::Storage::Disk.find(devicegraph, free_slot.disk_name)
             ptable = disk.partition_table
             if logical_partition_preferred?(ptable)
-              create_extended_partition(disk, free_slot.region) unless ptable.has_extended
+              create_extended_partition(disk, free_slot.slot.region) unless ptable.has_extended
```
jreidinger: NP: as can be seen from this bug it is quite confusing to have free_slot which is actually not slot and you have to call slot to get it.

src/lib/storage/yaml_writer.rb
```diff
@@ -139,7 +158,7 @@ def yaml_partition(partition)
           "size" => DiskSize.new(partition.region.to_kb(partition.region.length)).to_s,
           "name" => partition.name,
           "type" => @partition_types[partition.type],
-          "id"   => @partition_ids[partition.id]
+          "id"   => @partition_ids[partition.id] || "0x#{partition.id.to_s(16)}"
```
jreidinger: NP: this code is quite confusing, why partition.id is not id and need additional mapping? and what means default here? hexa code of partition_id

test/test_refinements.rb
```diff
@@ -0,0 +1,76 @@
...
+        end
+
+        def device_tree
+          writer = Yast::Storage::YamlWriter.new
+          writer.yaml_device_tree(self)
+        end
+      end
+
+      refine ::Storage::Devicegraph.singleton_class do
+        def new_from_file(filename)
```
jreidinger: is it really new_from_file or load_from_file?

jreidinger:
few notes, but none blocking, so LGTM

test/test_refinements.rb
ancorgs: I used `new_from_file` because it's actually kind of an equivalent to `Devicegraph.new` because it creates a new Devicegraph and returns it. With the code style we are using in yast2-storage-ng, I would expect a method called `load_from_file` to receive an empty devicegraph than then it will populate (quite un-Ruby, I know).

test/proposal_test.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    def input_file_for(name)
+      File.join(DATA_PATH, "input", "#{name}.yml")
+    end
+
+    def output_file_for(name)
+      File.join(DATA_PATH, "output", "#{name}.yml")
+    end
+
+    def fake_scenario(scenario)
```
ancorgs: Let's hijack this (already merged) PR to discuss something about `FakeProbing` - it brings back the concept of a catalog of devicegraphs indexed by name.

test/proposal_test.rb
```diff
@@ -0,0 +1,135 @@
...
+      fake_probing = Yast::Storage::FakeProbing.new
```
ancorgs: This line, apart from doing the obvious thing, creates the "fake" entry in the catalog. I have already explained how bad I believe is this. Moreover, creating an empty devicegraph doesn't seem to be a responsibility of `FakeProbing`.

test/proposal_test.rb
```diff
@@ -0,0 +1,135 @@
...
+      devicegraph = fake_probing.devicegraph
+      Yast::Storage::FakeDeviceFactory.load_yaml_file(devicegraph, input_file_for(scenario))
+      fake_probing.to_probed
```
ancorgs: This copies a devicegraph from the well-know position "fake" to the well-known position "probed". All my usual fears about global stuff arise.

test/proposal_test.rb
```diff
@@ -0,0 +1,135 @@
...
+    end
```
ancorgs: I would prefer `fake_scenario`to look like this:
```ruby
def fake_scenario(scenario)
  fake_probing = Yast::Storage::FakeProbing.new
  devicegraph = ::Storage::Devicegraph.new_from_file(input_file_for(scenario))
  fake_probing.to_probed(devicegraph)
end
```

I think it's:
 - more consistent with the rest of the file (now the initial empty devicegraph for probed is created by `FakeProbing` and the initial empty devicegraph for the results are created "by hand").
 - reduces the responsibility of `FakeProbing`
 - Makes `#to_probed` more flexible, not only working from "fake" to "probed"
 - The most important one - No catalog i.e. no global registration of `StorageManager.instance["fake"]`

src/lib/storage/yaml_writer.rb
shundhammer: We have an enum for those IDs; that's what fdisk calls "partition type", as in 0x83==Linux or 0x82==swap etc.; symbolic names are very much preferred here for better readability. Hex numbers are also allowed, but we should really have an enum entry for it. But then, we can never know what some weird system is using that we don't have an enum value for yet, so it makes sense to have a hex fallback. Remember we might also use that to write probed values to disk, and there it might actually happen.

test/proposal_test.rb
shundhammer: "fake" is not a well-known device graph. It exists only when created as FakeProbing.

"probed" does have a well-defined meaning, as the name implies. At some point we do have to trigger hardware probing in all normal cases (except in a testing environment), so it makes perfect sense to have something like this with a predefined meaning. Otherwise that probing stuff would have to be scattered all over the place.

test/proposal_test.rb
shundhammer: Frankly, I don't like this approach. Previously, the responsibility was distributed: FakeProbing took care about inhibiting the default behaviour, i.e. it prevented the lib from doing real hardware probing, and returned a device graph to be treated as the faked "probed" device graph.

This new fake_scenario() function silently assumes that it is safe to create just any new device graph and then move it to "probed". IMHO this assumption is not safe. This is the current behaviour of libstorage-ng, but this is really an implementation detail that is better shielded by FakeProbing; that's what FakeProbing is for.

src/lib/storage/yaml_writer.rb
jreidinger: ok, make sense, in such case I really think it is better to have different map name like `@partition_symbolic_names`

PR #16, opened 2016-03-10
Moved env vars from StorageManager to expert partitioner client

shundhammer:


ancorgs:
LGTM

aschnell:
That should work but the lines setting the filename of the devicegraph and arch have been lost some time ago.

src/clients/expert_partitioner.rb
```diff
@@ -18,4 +18,26 @@
...
+      true,
+      ::Storage::ProbeMode_READ_DEVICEGRAPH,
+      ::Storage::TargetMode_DIRECT
+    )
+  # probe
+  else
+    ::Storage::Environment.new(true)
+  end
+
+Yast::Storage::StorageManager.create_instance(storage_environment) if storage_environment
```
aschnell: Isn't the "if storage_environment" useless?

PR #18, opened 2016-03-11
Several agreed small code reorganizations

ancorgs:


ancorgs:
The github view is doing quite interesting wrong assumptions like
`lib/storage/refined_devicegraph.rb → /examples/proposal_dump_yaml.rb`

Obviously, that's not a moved file. They are completely different files committed in different commits. So please review in a per-commit basis to avoid those artifacts.

src/lib/storage/planned_volumes_list.rb
```diff
@@ -29,7 +29,7 @@ module Storage
     # Set of PlannedVolume elements
     #
     # Implements Enumerable and provides some extra methods to query the set
```
aschnell: The set of what? And why set if the class is called List?

aschnell:
Concerning refining the Disk class. Do you know of the downcast pitfall (see https://github.com/openSUSE/libstorage-ng/blob/master/bindings/storage-downcast.i)?

aschnell:
Otherwise LGTM.

src/lib/storage/planned_volumes_list.rb
ancorgs: It comes from the previous line "Set of PlannedVolume elements". But I can clarify if needed. Using "set" just as a synonym for "list" or "collection" with no particular reason. I will rephrase it a bit.

PR #21, opened 2016-03-14
add boot partition requirements

wfeldt:


shundhammer:
LGTM, but of course this will remain work in process for quite a while.

wfeldt:
Ok, could you actually  merge it?

PR #23, opened 2016-03-17
Firsts test for PartitionCreator and first prototype of DevicegraphQuery

ancorgs:


src/lib/storage/devicegraph_query.rb
```diff
@@ -0,0 +1,64 @@
...
+    #
+    # While writing the tests, I found myself needing to do similar checks, but
+    # not necessarily restricted to the candidate disks
+    #
+    # This class (together with the Disk refinement it relies on) should allow
+    # to do those "high level" queries, setting a filter (only disk_names for
+    # the time being) in advance to set the scope
+    #
+    # The class is now used in the RSpec tests and in the proposal code
+    class DevicegraphQuery
```
shundhammer: I like the idea of having a DevicegraphQuery class.

src/lib/storage/devicegraph_query.rb
```diff
@@ -0,0 +1,64 @@
...
+      using Refinements::Disk
+
+      # Free disk space below this size will be disregarded
+      TINY_FREE_CHUNK = DiskSize.MiB(30)
+
+      attr_reader :devicegraph
+
+      def initialize(devicegraph, disk_names: nil)
```
shundhammer: I was about to suggest to add a block parameter for filtering for most of the query methods, but that should already be covered by this disk_names parameter. OTOH we might still consider passing a block parameter as an alternative (not sure if it's a good approach to pass it to the constructor, though).

test/proposal/partition_creator_test.rb
```diff
@@ -0,0 +1,103 @@
...
+        # FIXME: actually I wouldn't have expected these values, but 23GiB and 26GiB
+        expect(query.partitions).to contain_exactly(
+          an_object_with_fields(mountpoint: "/", size_k: 25015296),
+          an_object_with_fields(mountpoint: "/home", size_k: 26363904),
+          an_object_with_fields(mountpoint: "swap", size: 1.GiB)
+        )
+      end
+    end
+  end
+end
```
shundhammer: Very well readable now. I like it.

src/lib/storage/devicegraph_query.rb
ancorgs: The idea of doing it in the constructor is that you can set the scope once and then you can perform several subsequent queries. It would be more obvious with a name meaning "Querier" (ugly word) instead of "Query". I'm open to suggestions.

I like the idea of passing a block to set more complex scopes. That's something we can do in addition to some predefined and more clear named filters like `disk_names` once we need it.

shundhammer:
Looks very good to me. I like the way the tests are readable now, and I like the idea of the query class.

src/lib/storage/devicegraph_query.rb
shundhammer: Yes, "Querier" is indeed an ugly word. Probably it doesn't even exist. I strongly suggest to stick with "Query" which is also a noun, it's well-known, widely used, and it wouldn't stick out like a sore thumb.

shundhammer:
Still LGTM after all those changes

PR #25, opened 2016-03-21
Added missing copyright header

ancorgs:


lslezak:
LGTM

<sub>NP: IIRC `encoding: utf-8` is not needed anymore as it is the default (already for some time).</sub>

lslezak:
Still LGTM :wink:

PR #26, opened 2016-03-21
Devices lists

ancorgs:
Inline documentation is still missing

Some usage examples
```ruby
using DevicegraphLists

disks = a_devicegraph.disks.with(name: ["/dev/sda", "/dev/sdb"])
a_devicegraph.partitions.map {|p| p.name } # ["/dev/sda1", "/dev/sdb1", "/dev/sdb2", "/dev/sdc1"]
disks.partitions.map {|p| p.name } # ["/dev/sda1", "/dev/sdb1", "/dev/sdb2"]

primary_parts = disks.partitions.with(type: ::Storage::PartitionType_PRIMARY)
a_device_graph.filesystems.to_a # all file systems in the devicegraph
disks.filesystems.to_a # Filesystems in the chosen disks
primary_parts.filesystems.to_a # fs in primary parts of chosen disks

# Look ma! also with blocks
partitioned_disks = a_devicegraph.disks.with { |disk| disk.partition_table }
partitioned_disks.filesystems.size
```

### DIFF
--- a/src/lib/storage/devices_list.rb
+++ b/src/lib/storage/devices_list.rb
@@ -23,6 +23,11 @@ require "storage"
 
 module Yast
   module Storage
+    # Base class to implement lists of devices.
+    #
+    # @see DisksList
+    # @see PartitionsList
+    # @see FilesystemsList
     class DevicesList
       include Enumerable
       extend Forwardable
@@ -30,9 +35,12 @@ module Yast
       class << self
         attr_reader :device_class
 
+        # rubocop:disable TrivialAccessors
+        # Macro-style method to specify the class of the elements
         def list_of(klass)
           @device_class = klass
         end
+        # rubocop:enable TrivialAccessors
       end
 
       def_delegators :@list, :each, :empty?, :length, :size
@@ -42,13 +50,26 @@ module Yast
         @list = list || full_list
       end
 
+      # Subset of the list matching some conditions
+      #
+      # Returns a list containing only the elements that meets some conditions.
+      # Those conditions can be expressed as a list of attributes and values or
+      # as a block that will be evaluated for every element in the list.
+      #
+      # It returns a list of the same time to allow chaining several calls
+      # @example
+      #   filtered = a_list.with(type: [type1, type2], name: nil)
+      #   filtered.with { |element| element.id.start_with? "a" }
+      #
+      # @param attrs [Hash] attributes to filter by. The keys of the hash are
+      #     method names, the values of the hash can be the desired value for
+      #     the attribute or an Enumerable with several accepted values.
+      # @return [DevicesList] subtype-preserving
       def with(attrs = {})
         new_list = list.select do |element|
           attrs.all? { |attr, value| match?(element, attr, value) }
         end
-        if block_given?
-          new_list.select!(&Proc.new)
-        end
+        new_list.select!(&Proc.new) if block_given?
         self.class.new(devicegraph, list: new_list)
       end
 
@@ -57,10 +78,17 @@ module Yast
       end
 
     protected
-    
+
       attr_reader :devicegraph
       attr_accessor :list
-      
+
+      # Default collection of devices
+      #
+      # In many cases it can inferred by asking libstorage for the class
+      # specified via .list_of. Sometimes is not possible and this method needs
+      # to be redefined in the child class.
+      #
+      # return [Array]
       def full_list
         self.class.device_class.all(devicegraph).to_a
       end
--- a/src/lib/storage/devices_list.rb
+++ b/src/lib/storage/devices_list.rb
@@ -19,25 +19,23 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "storage"
+
 module Yast
   module Storage
     class DevicesList
       include Enumerable
+      extend Forwardable
 
       class << self
-        attr_accessor :device_class
-        attr_accessor :default_delegate
+        attr_reader :device_class
 
         def list_of(klass)
-          self.device_class = klass
-        end
-
-        def by_default_delegate_to(list)
-          self.default_delegate = list
+          @device_class = klass
         end
       end
 
-      attr_reader :devicegraph
+      def_delegators :@list, :each, :empty?, :length, :size
 
       def initialize(devicegraph, list: nil)
         @devicegraph = devicegraph
@@ -54,40 +52,13 @@ module Yast
         self.class.new(devicegraph, list: new_list)
       end
 
-      def each(&block)
-        @list.each(&block)
-      end
-
       def dup
         self.class.new(devicegraph, @list.dup)
       end
 
-      # Returns true if the list contains no elements
-      #
-      # @return [Boolean]
-      def empty?
-        list.empty?
-      end
-
-      # Number of elements in the list
-      #
-      # @return [Fixnum]
-      def length
-        list.length
-      end
-      alias_method :size, :length
-
-      def method_missing(meth, *args, &block)
-        delegate_list = self.class.default_delegate
-        if delegate_list
-          delegate_list.send(meth, *args, &block)
-        else
-          super
-        end
-      end
-
     protected
     
+      attr_reader :devicegraph
       attr_accessor :list
       
       def full_list
@@ -95,9 +66,20 @@ module Yast
       end
 
       def match?(element, attr, value)
-        real_value = element.send(attr)
-        return true if real_value == value
-        value.is_a?(Array) && value.include?(real_value)
+        begin
+          real_value = element.send(attr)
+        rescue ::Storage::WrongNumberOfChildren
+          # Checking for something that is not there
+          return false
+        end
+
+        begin
+          return true if real_value == value
+        rescue TypeError
+          # Collections coming from SWIG perform strict type check for ==
+          raise unless value.is_a?(Enumerable)
+        end
+        value.is_a?(Enumerable) && value.include?(real_value)
       end
     end
   end
--- a/src/lib/storage/devices_list.rb
+++ b/src/lib/storage/devices_list.rb
@@ -1,0 +1,104 @@
+# encoding: utf-8
+
+# Copyright (c) [2016] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Yast
+  module Storage
+    class DevicesList
+      include Enumerable
+
+      class << self
+        attr_accessor :device_class
+        attr_accessor :default_delegate
+
+        def list_of(klass)
+          self.device_class = klass
+        end
+
+        def by_default_delegate_to(list)
+          self.default_delegate = list
+        end
+      end
+
+      attr_reader :devicegraph
+
+      def initialize(devicegraph, list: nil)
+        @devicegraph = devicegraph
+        @list = list || full_list
+      end
+
+      def with(attrs = {})
+        new_list = list.select do |element|
+          attrs.all? { |attr, value| match?(element, attr, value) }
+        end
+        if block_given?
+          new_list.select!(&Proc.new)
+        end
+        self.class.new(devicegraph, list: new_list)
+      end
+
+      def each(&block)
+        @list.each(&block)
+      end
+
+      def dup
+        self.class.new(devicegraph, @list.dup)
+      end
+
+      # Returns true if the list contains no elements
+      #
+      # @return [Boolean]
+      def empty?
+        list.empty?
+      end
+
+      # Number of elements in the list
+      #
+      # @return [Fixnum]
+      def length
+        list.length
+      end
+      alias_method :size, :length
+
+      def method_missing(meth, *args, &block)
+        delegate_list = self.class.default_delegate
+        if delegate_list
+          delegate_list.send(meth, *args, &block)
+        else
+          super
+        end
+      end
+
+    protected
+    
+      attr_accessor :list
+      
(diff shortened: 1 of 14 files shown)


src/lib/storage/devices_list.rb
```diff
@@ -0,0 +1,104 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Yast
+  module Storage
+    class DevicesList
+      include Enumerable
```
jreidinger: I would like just to point out one of very bad design of using enamurable - if you use e.g. select or grep or map result is array and not `DeviceList` class, so be aware of it. I already considered to improve it by new better module having specific container constructor, but never realized.

src/lib/storage/devices_list.rb
```diff
@@ -0,0 +1,104 @@
...
+      def initialize(devicegraph, list: nil)
+        @devicegraph = devicegraph
+        @list = list || full_list
+      end
+
+      def with(attrs = {})
+        new_list = list.select do |element|
+          attrs.all? { |attr, value| match?(element, attr, value) }
+        end
+        if block_given?
```
shundhammer: +1

Okay, so we can now specify a block to select which ones to include. This allows us to use more complex conditions, too.

src/lib/storage/devices_list.rb
```diff
@@ -0,0 +1,104 @@
...
+module Yast
+  module Storage
+    class DevicesList
+      include Enumerable
+
+      class << self
+        attr_accessor :device_class
+        attr_accessor :default_delegate
+
+        def list_of(klass)
```
shundhammer: +1

Using introspection will avoid a lot of code duplication.

src/lib/storage/devices_list.rb
```diff
@@ -0,0 +1,104 @@
...
+          self.device_class = klass
```
jreidinger: looks a bit strange to have accessor and then implement alias this way. Is there reason for it?

jreidinger: I expect to have device_class reader only. or having only accessor.

src/lib/storage/devices_list.rb
```diff
@@ -0,0 +1,104 @@
...
+        end
+
+        def by_default_delegate_to(list)
+          self.default_delegate = list
+        end
+      end
+
+      attr_reader :devicegraph
```
jreidinger: do we need to have it publicly exposed?

src/lib/storage/devices_list.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      def initialize(devicegraph, list: nil)
+        @devicegraph = devicegraph
+        @list = list || full_list
```
jreidinger: why not use full_list as default value in header above?

src/lib/storage/devices_list.rb
```diff
@@ -0,0 +1,104 @@
...
+        list.empty?
+      end
+
+      # Number of elements in the list
+      #
+      # @return [Fixnum]
+      def length
+        list.length
+      end
+      alias_method :size, :length
```
jreidinger: what about using forwardable? http://ruby-doc.org/stdlib-2.0.0/libdoc/forwardable/rdoc/Forwardable.html

src/lib/storage/devices_list.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      def method_missing(meth, *args, &block)
+        delegate_list = self.class.default_delegate
+        if delegate_list
+          delegate_list.send(meth, *args, &block)
```
jreidinger: what about querying if it understand such method? Result is quite confusing result like `A.new.method_foo => Exception unknown method method_foo for class XYZ`

Also you use unsafe send which allows to call private methods of delegated list. So I suggest to use respond_to? on delegate list at first

src/lib/storage/devices_list.rb
ancorgs: It needs `@devicegraph` to be already assigned, it cannot be evaluated in advance.

src/lib/storage/devices_list.rb
jreidinger: BTW you can also have this behavior with forwardable module used above with `def_delegators self.class.default_delegate, *self.class.default_delegate.public_instance_methods`

src/lib/storage/devices_list.rb
ancorgs: No particular reason, I can change `attr_accessor` into `attr_reader`.

src/lib/storage/devices_list.rb
ancorgs: Yep, Ruby collections are not type-preserving. Not sure if it can be called a "bad design". Collection in most languages are either not type-preserving either not DRY.

src/lib/storage/devices_list.rb
ancorgs: No so far... but it will not hurt. I can make it private if you prefer so.

src/lib/storage/disks_list.rb
```diff
@@ -0,0 +1,58 @@
...
+require "storage"
+require "storage/devices_list"
+require "storage/partitions_list"
+require "storage/filesystems_list"
+#require "storage/free_disk_spaces_list"
+require "storage/refinements/disk"
+
+module Yast
+  module Storage
+    class DisksList < DevicesList
```
jreidinger: :+1: for not using Query in class name. I actually think we can omit List word and have simple `Disks < Devices`

src/lib/storage/devices_list.rb
```diff
@@ -0,0 +1,104 @@
...
+      attr_accessor :list
+      
+      def full_list
+        self.class.device_class.all(devicegraph).to_a
+      end
+
+      def match?(element, attr, value)
+        real_value = element.send(attr)
+        return true if real_value == value
+        value.is_a?(Array) && value.include?(real_value)
```
jreidinger: Why limit it only for Arrays? Can be possible something like `Disks.with(partitions: partition_list_instance)`?

src/lib/storage/devices_list.rb
jreidinger: ah, interdepedencies. It will be more visible if it took device graph as parameter ;)

src/lib/storage/devices_list.rb
jreidinger: in general it is good to have small API ane increase if needed. Moving it to private is backward incompatible change. good api is close to change and open to enhancement.

src/lib/storage/devices_list.rb
jreidinger: I often when I design API try to create instance in irb and call `methods` on it, to see if all methods make sense and if it is not overloaded.

src/lib/storage/devices_list.rb
jreidinger: What is problem for is that it is not collection but enumerable mixin which can be mixed into any object and then its behavior is a bit confusing for me and I often need to redefine some methods to return that object instead of default one.

src/lib/storage/disks_list.rb
```diff
@@ -0,0 +1,58 @@
...
+      list_of ::Storage::Disk
+      by_default_delegate_to :partitions
+
+      using Refinements::Disk
+
+      def partitions
+        part_list = list.reduce([]) { |sum, disk| sum + disk.all_partitions }
```
jreidinger: this can be performance problem if there are thousands of disks and each have some partitions ( yes, we have such case in past and it took 10 hours to install it ). so better use `sum.concat(disk.all_partitions)`

src/lib/storage/disks_list.rb
```diff
@@ -0,0 +1,58 @@
...
+        PartitionsList.new(devicegraph, list: part_list)
```
jreidinger: as you use delegation for it? and if I see correctly this class is immutable, so maybe make sense to cache result?

src/lib/storage/disks_list.rb
```diff
@@ -0,0 +1,58 @@
...
+        # Add filesystems not included in #partitions (directly on disk)
+        list.each do |disk|
+          fs_list << disk.filesystem if !disk.partition_table && disk.filesystem
+        end
+        FilesystemsList.new(devicegraph, list: fs_list)
+      end
+
+      def free_disk_spaces
+        spaces_list = list.reduce([]) { |sum, disk| sum + disk.free_spaces }
+        FreeDiskSpacesList.new(devicegraph, list: spaces_list)
```
jreidinger: same note about immutable.

src/lib/storage/disks_list.rb
```diff
@@ -0,0 +1,58 @@
...
+require "storage/partitions_list"
+require "storage/filesystems_list"
+#require "storage/free_disk_spaces_list"
+require "storage/refinements/disk"
+
+module Yast
+  module Storage
+    class DisksList < DevicesList
+      list_of ::Storage::Disk
+      by_default_delegate_to :partitions
```
jreidinger: this basically mean for me, that disk is a partition as it this method looks to me like multiple inheritance. I prefer to have something like `disks.partitions.partition_specific_method` then `disks.partition_specific_method`

ancorgs: I agree that `by_default_delegate_to` opens the door to that, which is worse than the benefit it provides (delegating the hierarchy in a DRY way). I will remove it.

src/lib/storage/free_disk_spaces_list.rb
```diff
@@ -20,54 +20,36 @@
...
-        @disk_names = disk_names
-      end
-
-      def disks
-        if @disk_names
-          @disk_names.map { |disk_name| ::Storage::Disk.find(@devicegraph, disk_name) }
-        else
-          devicegraph.all_disks.to_a
-        end
+      def useful
```
jreidinger: I think name is not nice, as also smaller ones can be useful. What about reasonable_big?

src/lib/storage/devices_list.rb
ancorgs: Not really, because `disk.partitions` (assuming it would exists, which is not true[1]) would be some kind of collection (hopefully an array), so you would be comparing two collections, i.e. using line 99, with `==`.

A better example of what I believe you wanted to say would be, for example, 
```ruby
disks.partitions.with(filesystem: another_list.filesystems)
```
Still, what would be the criteria instead of `value.is_a?(Array)`. Responding to `include?` is not a good idea, since strings respond to that and it could lead to tricky situations. Maybe responding to `to_a` but in that case I prefer to do everything more explicit and require the caller to do:
```ruby
disks.partitions.with(filesystem: another_list.filesystems.to_a)
```

I prefer "it expects a value or an array of possible values" to a more loosely definition like "it expects a value or something that could include the dialog".

[1] With libstorage would be need to do something like `disk.partition_table.partitions`

src/lib/storage/disks_list.rb
ancorgs: Not sure what you mean with delegation here.

No so sure about the caching. Maybe it makes sense but imagine this situation:
```ruby
disks = my_devicegraph.disks # One disk with 2 partitions and 1 empty disk
disk.partitions # => 2 partitions
# create another partition in the empty disk
disks.partitions # => With caching 2 partitions. Without it, 3
```

src/lib/storage/free_disk_spaces_list.rb
ancorgs: I agree `useful` is not a nice name. But not sure if `reasonable_big` is better. :wink: I'll try to look for a better one.

src/lib/storage/disks_list.rb
ancorgs: I dislike class names in plural. It's easy to confuse them with the singular class (Disk vs Disks) and they challenge you to think in unnatural ways like "create a devices" or "compare two devices..es..ermm, wait, what's the plural of 'devices'?". I prefer to "create a list of devices" and "compare two list of devices".

jreidinger: ok

src/lib/storage/devices_list.rb
jreidinger: what about `is_a?(Enumerable)` as String is not enumerable, but other collecion is and include? is in enumerable, so you are on safe side http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-include-3F

src/lib/storage/disks_list.rb
jreidinger: ok, we will see in reality how much performance it took.
By delegation I mean that every call that goes to delegated object ( here it is partitions ) using this object creation.

src/lib/storage/free_disk_spaces_list.rb
jreidinger: at least below when it is used for me reasonable_big look there better, but I welcome any better name.

src/lib/storage/disks_list.rb
ancorgs: In places needing extra performance you can always use libstorage API directly. It's less convenient, but has 0 overhead.

src/lib/storage/proposal/settings.rb
```diff
@@ -82,6 +84,7 @@ def initialize
           # Not yet in control.xml
           @home_min_size                 = DiskSize.GiB(10)
           @home_max_size                 = DiskSize.unlimited
+          @useful_free_space_min_size    = DiskSize.MiB(30)
```
shundhammer: Not sure if we want to move this responsibility to control.xml that is typically written by a PM.

Things might easily break internally if we leave this open for outside configuration to somebody who doesn't know the implications of this.

ancorgs: Adding the value to `Proposal::Settings` looked like the cheapest way to share it between different proposal sub-components. We don't need to put it in control.xml right now, but having it in `Settings` allows us to do it configurable in the future. I like that flexibility and I expect responsibility from PMs if we document which options are dangerous.

Would it be enough if I make more obvious (using constants, for example) the fact that we don't plan to add `settings.useful_free_space_min_size` to control.xml for the time being?

Of course, there are many other options. Another equally cheap alternative would be to extract `TINY_FREE_CHUNK` (most likely together with other constants like `VOLUME_GROUP_SYSTEM`) to `/lib/storage/proposal/constants.rb` to make easy to share it between proposal subcomponents.

shundhammer: Being able to easily share it is a good argument. But it should be clear that this may never become a parameter for PMs to configure; this is strictly technical.

In the not-so-far future this might even become a value that will be calculated from our own partition alignment logic: It will be something like one sector less than the alignment size.

But I can live with this for the time being.

shundhammer:
Aside from the above discussion, LGTM. I can live with the current status.

PR #29, opened 2016-03-23
DevicesList: better handling of SWIG corner cases

ancorgs:


src/lib/storage/devices_list.rb
```diff
@@ -98,15 +98,19 @@ def match?(element, attr, value)
           real_value = element.send(attr)
         rescue ::Storage::WrongNumberOfChildren
           # Checking for something that is not there
-          return false
+          return value.nil?
```
jreidinger: I think it deserve comment this logic. So if it is nil, then it match to all values? looks a bit strange

src/lib/storage/devices_list.rb
```diff
@@ -98,15 +98,19 @@ def match?(element, attr, value)
...
         end
 
+        # First of all, check for exact match
         begin
           return true if real_value == value
-        rescue TypeError
-          # Collections coming from SWIG perform strict type check for ==
-          raise unless value.is_a?(Enumerable)
+        # Objects coming from SWIG perform strict type check for ==. Thus they
+        # raise exceptions on type mismatch, instead of simply returning false
```
jreidinger: OMG, that is very bad SWIG behavior.

test/devices_list_test.rb
```diff
@@ -58,6 +58,14 @@
         )
       end
 
+      it "filters by nil values" do
```
jreidinger: sorry, I do not get from description what it do.

src/lib/storage/devices_list.rb
ancorgs: The comment is there. It says "checking for something that is not there". And "something that is not there" only matches your criteria if you were looking for nil. I can extend the comment, though

jreidinger: ah, then I really think it need extend comment, it is not much obvious.

test/devices_list_test.rb
ancorgs: It makes more sense if you read the spec as a whole (with all the variants of "it filters by xxx"). Anyway, it will be better like this?

```it "filters by missing values" do```

or maybe

```it "filters checking for nil attributes"```

or something in between?

ancorgs: Or

```it "filters by attributes that are nil" do```

jreidinger: `filters by missing values` sounds good, as it really search for missing value and not for value that are nil.

jreidinger:
LGTM

PR #37, opened 2016-03-30
finalzied x86 boot partition setup doc

wfeldt:


doc/boot-partition.md
```diff
@@ -130,21 +134,34 @@ i.e. valid for all architectures (s390, see below)
...
-if not create:
-- /boot/efi
-- fat32
-- size: 33 MB, 200MB
-- limit: below 2TB
+## x86-legacy
+
+*[this assumes grub will be installed into mbr]*
+
+- required size for grub embedding (currently): 84 kB; to be safe, check for >= 256 kB
```
WenhuaChang: The size of core.img would be bigger if boot partition is on mdadm and lvm. But I think 256kB is sufficient for that also.

doc/boot-partition.md
```diff
@@ -130,21 +134,34 @@ i.e. valid for all architectures (s390, see below)
...
+- if dos partition table
+    - embed in unused space before 1st partition
+    - fail if not enough space
```
WenhuaChang: Here I'd like to add that grub2-install would not fail immediately but will seek other possibilities. In most filesystems that the blocklist will be used to locate core.img on that filesystem. If the filesystem cannot support blocklist, for eg the btrfs and zfs, then it will really fail. For that we also carried a patch in grub2-install for btrfs specifically for this issue, it will fallback to use the bootloader pad to embed the image if the mbr gap is too small and that's even better fallback than blocklist imho. :)

doc/boot-partition.md
```diff
@@ -130,21 +134,34 @@ i.e. valid for all architectures (s390, see below)
...
-- fat32
-- size: 33 MB, 200MB
-- limit: below 2TB
+## x86-legacy
+
+*[this assumes grub will be installed into mbr]*
+
+- required size for grub embedding (currently): 84 kB; to be safe, check for >= 256 kB
+- if dos partition table
+    - embed in unused space before 1st partition
```
WenhuaChang: I think in upstream doc they used the term "mbr gap" to that unused space between mbr and 1st partition.

doc/boot-partition.md
```diff
@@ -130,21 +134,34 @@ i.e. valid for all architectures (s390, see below)
...
+    - fail if not enough space
+    - note: grub-install may later fail if it detects the area is used by some other weird software
```
WenhuaChang: Yes, there's seems to be detection there already.

doc/boot-partition.md
```diff
@@ -130,21 +134,34 @@ i.e. valid for all architectures (s390, see below)
...
+- if gpt partition table
+    - BIOS boot partition must exist (gpt type 21686148-6449-6e6f-744e-656564454649); parted: set flag `bios_grub` to `on`
+    - if it doesn't exist, create:
+        - size: [1MB, 2MB, 100MB]
+        - no fs
+        - no mount point
+  - if existing BIOS boot partition is too small, delete and re-create
+  - grub will use 1st BIOS boot partition it finds, co-op with other grub instances on same disk not possible
```
WenhuaChang: Yes, it just erased by new installed one. OTOH, maintaining a per distro bios_grub partition would be in a bigger mess imho.

doc/boot-partition.md
```diff
@@ -130,21 +134,34 @@ i.e. valid for all architectures (s390, see below)
...
+- note: `yast bootloader` will be responsible for boot flag handling, if necessary
+
```
WenhuaChang: I'd like to add that.
1. Don't use gpt-sync-mbr or hybrid mbr. Only standard "protective mbr".
2. Make sure the boot flag is set in protective mbr and also in normal mbr, because some bios is stubborn to fix themselves and insist to detect the active/boot flag before handing over control to mbr boot code. It seems to me YaST has fixed this already, but still can be reproduced in our hybrid iso images.

doc/boot-partition.md
```diff
@@ -130,21 +134,34 @@ i.e. valid for all architectures (s390, see below)
...
+  
+## x86-efi
+
+- efi system partition required; dos type 0xef, gpt type c12a7328-f81f-11d2-ba4b-00a0c93ec93b
+- reuse existing partition
+- if it doesn't exist, create:
+    - size: [33MB, 200MB, 1GB]
+    - fat32
+    - mount at /boot/efi
+- if existing efi system partition is too small, delete and re-create
```
WenhuaChang: Yes, in addition on EFI we still need that boot flag set for gpt(protective mbr) and also mbr.
Thanks.

wfeldt: No. Why?

doc/boot-partition.md
wfeldt: tried with our grub2 and that doesn't seem to work

shundhammer:
LGTM

PR #40, opened 2016-04-07
Added root_device and candidate_devices to Proposal::Settings

ancorgs:
* Allows the caller to decide where / (and related partitions) is going to be installed.
* Knowing the root_device in advance allows BootRequirementsChecker to properly do its job.
* If root_device and candidate_device are not provided, they are calculated in a new additional first step.

### DIFF
--- a/src/lib/storage/proposal.rb
+++ b/src/lib/storage/proposal.rb
@@ -26,6 +26,7 @@ require "storage/proposal/exceptions"
 require "storage/proposal/settings"
 require "storage/proposal/volumes_generator"
 require "storage/proposal/devicegraph_generator"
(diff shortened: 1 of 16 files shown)


shundhammer:
Looks reasonable to me. Actually, I had expected a lot more impact with this change.

gabi2:
For me the changes also make sense.

ancorgs:
Updated. Only documentation is missing now.

ancorgs:
Updated. Ready for review

wfeldt:
I think it doesn't do what it intends to. (Or I'm missing some point.)

Anyway, I'll merge everything, so we can go on.

PR #41, opened 2016-04-07
add fake_from_yaml method to StorageManger

wfeldt:


ancorgs:
Currently this adds some code duplication with `FakeProbing`, so I would like to see this code coming in and `FakeProbing` dying as part of the pull request.

shundhammer:
Looks good so far. But the examples and clients seem to be missing.

shundhammer:
LGTM now.

ancorgs:
Travis and jenkins don't look happy. Even coveralls is complaining (but not sure if coveralls claims are always legitimate).

test/boot_requirements_checker_test.rb
```diff
@@ -40,7 +40,7 @@ def find_vol(mount_point, volumes)
     let(:storage_arch) { instance_double("::Storage::Arch") }
 
     before do
-      Yast::Storage::FakeProbing.new
+      Yast::Storage::StorageManager.fake_from_yaml
```
ancorgs: Hmm, calling `fake_from_yaml` without yaml is a quite hacky and unexpected usage. `fake_from_yaml` actually does two things - creating a fake instance of the StorageManager and populating `probed` with yaml.

For cases like this, in which we only want to do the first part but not the second one, we should have another method.

Like already suggested here, for example, https://github.com/yast/yast-storage/pull/193#discussion_r53643452

src/lib/storage/storage_manager.rb
```diff
@@ -58,6 +59,26 @@ def instance
...
         end
 
+        # Use this as an alternative to the instance method.
+        # Probing is skipped and the device tree is initialized from yaml_file.
+        # Any existing probed device tree is replaced.
+        #
+        # @return [::Storage::Storage] libstorage object
+        #
+        def fake_from_yaml(yaml_file = nil)
+          create_instance(
```
ancorgs: This call to `create_instance` should be in a separated public method IMHO. See below for cases in which is needed.

test/boot_requirements_checker_test.rb
wfeldt: calling fake_from_yaml(nil) seems an obvious semantic to me, but rubocop won't let me

my first thought was to call the method just 'fake'; maybe that is more understandable after all

wfeldt: > For cases like this, in which we only want to do the first part but not the second one,
> we should have another method

I can't follow you here. That's like arguing against constructors with arguments.

fake() resp. fake_from_yaml() creates a fake instance (and populates it with the arg)

src/lib/storage/storage_manager.rb
wfeldt: I think not. The current semantics works quite well; this patch shows that the code gets much simpler.

wfeldt:
merging, you can shoot me later...

Actually, I think renaming fake_from_yaml -> fake might be a good idea and address part of your concerns.

PR #43, opened 2016-04-11
Missing parts of new algorithm

ancorgs:
Missing documentation and tests.

Now the previous (already merged) PR should make more sense. The plan is that some volumes can request to be in a particular disk (the only use-case now is that they request to be in `settings.root_device`).

### DIFF
--- a/src/lib/storage/planned_volumes_list.rb
+++ b/src/lib/storage/planned_volumes_list.rb
@@ -112,28 +112,34 @@ module Yast
       # @param descending [Boolean] whether to use descending order
       # @return [Array]
       def sort_by_attr(*attrs, nils_first: false, descending: false)
-        @volumes.sort do |one, other|
+        @volumes.each_with_index.sort do |one, other|
           compare(one, other, attrs, nils_first, descending)
-        end
+        end.map(&:first)
       end
 
     private
 
+      # @param one [Array] first element: the volume, second: its original index
+      # @param other [Array] same structure than previous one
       def compare(one, other, attrs, nils_first, descending)
-        result = compare_attr(one, other, attrs.first, nils_first, descending)
+        one_vol = one.first
+        other_vol = other.first
+        result = compare_attr(one_vol, other_vol, attrs.first, nils_first, descending)
         if result.zero?
           if attrs.size > 1
             # Try next attribute
             compare(one, other, attrs[1..-1], nils_first, descending)
           else
-            # Keep original order
-            -1
+            # Keep original order by checking the indexes
+            one.last <=> other.last
           end
         else
           result
         end
       end
 
+      # @param one [PlannedVolume]
+      # @param other [PlannedVolume]
       def compare_attr(one, other, attr, nils_first, descending)
         one_value = one.send(attr)
         other_value = other.send(attr)
@@ -144,6 +150,8 @@ module Yast
         end
       end
 
+      # @param one [PlannedVolume]
+      # @param other [PlannedVolume]
       def compare_values(one, other, descending)
         if descending
           other <=> one
@@ -152,6 +160,8 @@ module Yast
         end
       end
 
+      # @param one [PlannedVolume]
+      # @param other [PlannedVolume]
       def compare_with_nil(one, other, nils_first)
         if one.nil? && other.nil?
           0
--- a/src/lib/storage/planned_volumes_list.rb
+++ b/src/lib/storage/planned_volumes_list.rb
@@ -98,28 +98,51 @@ module Yast
       end
       alias_method :<<, :push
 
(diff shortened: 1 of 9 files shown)


src/lib/storage/proposal/partition_creator.rb
```diff
@@ -211,17 +217,27 @@ def create_non_lvm_complex
...
+        #
         # It tries to honor the value of #max_start_offset for each volume, but
         # it does not raise an exception if that particular requirement is
         # impossible to fulfill, since it's usually more a recommendation than a
         # hard limit.
         #
         # @param volumes [Array<PlannedVolume>]
         def create_volumes_partitions(volumes)
-          volumes.sort_by_attr(:max_start_offset).each do |vol|
+          volumes.sort_by_attr(:disk, :max_start_offset).each do |vol|
```
wfeldt: The problem with this approach (already before the patch) is that it basically iterates volumes in random order.
Which is a bit bad when it comes to reproducibility. (Ruby's sort seems to keep the order if all elements are equal
but if a single one is not, you're out of luck.)

src/lib/storage/proposal/partition_creator.rb
```diff
@@ -87,9 +87,15 @@ def total_free_size
 
         # List of free spaces in the devicegraph
         #
+        # @param TODO: documentation
         # @return [FreeDiskSpacesList]
-        def free_spaces
-          candidate_disks.free_disk_spaces.with do |space|
+        def free_spaces(disk_name = nil)
```
wfeldt: SCNR, but isn't __this__ surprising behavior? free_spaces(string = device)  vs. free_spaces(nil) iterating over candidate_disks?

src/lib/storage/proposal/partition_creator.rb
ancorgs: Well, now that we can sort by several attributes we could do something like this if you want:
```ruby
# :disk, and :max_start_offset because are relevant for the algorithm,
# :mount_point and :filesystem_type just to ensure consistency of order
volumes.sort_by_attr(:disk, :max_start_offset, :mount_point, :filesystem_type).each do |vol|
```

wfeldt: That doesn't solve the problem; you need at least one (varying) attribute that every volume has.

I was actually hoping that volumes already represent some kind of 'natural' order.

wfeldt:
generally, though, this lgtm

src/lib/storage/proposal/partition_creator.rb
ancorgs: Does not look surprising to me. Without parameters it lookups all the devices. With an optional parameter you can restrict the scope of the search.

Any suggestion?

PS.- Yes, I got the reference :wink: But this case is sooo different... :smile:

src/lib/storage/proposal/partition_creator.rb
ancorgs: I don't think currently `volumes` is intended to represent any kind of natural order. But nothing stops us from doing `sort_by_attr` more deterministic than Ruby's `sort` if we feel more comfortable. I will change it to always preserve the original order if all the specified criteria are the same.

src/lib/storage/proposal/partition_creator.rb
wfeldt: This line killed me:

> free_spaces(vol.disk).first

because (without having looked up the implementation of free_spaces) my thought was: Isn't vol.disk usually nil? That's certainly wrong code...

Maybe you have hidden the binding to certain disks a bit too low-level.

But that's just my thoughts. I'm fine with the code as it is.

ancorgs: Makes sense. I'll try to make it more obvious.

src/lib/storage/proposal/partition_creator.rb
wfeldt: The actual order doesn't matter but we should stick to __some__. Else you could get completely different proposals with only slight config changes.

src/lib/storage/planned_volumes_list.rb
```diff
@@ -98,28 +98,51 @@ def push(volume)
...
+      def compare(one, other, attrs, nils_first, descending)
+        result = compare_attr(one, other, attrs.first, nils_first, descending)
+        if result.zero?
+          if attrs.size > 1
+            # Try next attribute
+            compare(one, other, attrs[1..-1], nils_first, descending)
           else
-            compare_values(one_value, other_value, descending)
+            # Keep original order
+            -1
```
wfeldt: I think not; the compare value can't be independent of the two values compared to make a consistent sort.

wfeldt:
lgtm

PR #44, opened 2016-04-12
added reference to LoPAPR spec

wfeldt:
LoPAPR is basically a modernized CHRP


kobliha:
You have a nice LGTM from @k0da already

PR #45, opened 2016-04-12
Improvements and tests for PReP boot strategy

ancorgs:
IMPORTANT: it relies on a `::Storage::Arch#open_nv?` that is still not there

### DIFF
--- a/src/lib/storage/boot_requirements_strategies/prep.rb
+++ b/src/lib/storage/boot_requirements_strategies/prep.rb
@@ -34,26 +34,41 @@ module Yast
       class PReP < Base
         def needed_partitions
           volumes = super
-          volumes << prep_volume if prep_partition_missing?
+          volumes << prep_volume if prep_partition_required? && prep_partition_missing?
           volumes
         end
 
       protected
 
         def prep_partition_missing?
-          true	# #prep_partitions not implemented yet
-          # disk_analyzer.prep_partitions.empty?
+          partitions = disk_analyzer.prep_partitions[settings.root_device]
+          partitions.nil? || partitions.empty?
+        end
+
+        def prep_partition_required?
+          # no need of PReP partition in OPAL/PowerNV/Bare metal
+          !arch.power_nv?
         end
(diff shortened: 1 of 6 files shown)


src/lib/storage/boot_requirements_strategies/prep.rb
```diff
@@ -34,27 +34,42 @@ module BootRequirementsStrategies
...
-          vol.min_size = DiskSize.MiB(1)
+          vol = PlannedVolume.new(nil)
+          # So far we are always using msdos partition ids
+          vol.partition_id = ::Storage::ID_PPC_PREP
+          vol.min_size = DiskSize.kiB(256)
           vol.max_size = DiskSize.MiB(8)
           vol.desired_size = DiskSize.MiB(1)
+          # Make sure that alignment does not result in a too big partition
+          vol.align = :keep_size
+          vol.bootable = true
```
wfeldt: only with dos partition table, not with gpt

ancorgs: I expect `PartitionCreator` to be smart enough to use `bootable` only in ms-dos partition tables. So far, all our code works only with ms-dos partition tables.

ancorgs: To avoid that kind of uncertainty, I will document the attributes of `PlannedVolume`, something due a long time ago.

wfeldt: Note there technically _is_ a (legacy) boot flag for gpt partitions; we're just not needing it in our grub2 setup.

shundhammer:
LGTM, but can we test it?

ancorgs:
Let's wait for merge until the new libstorage is available in the repos, so travis can pass.

ancorgs:
@shundhammer It's probably time to start generating libstorage binaries for all the architectures we plan to support. Then, testing it would be a matter of installing 2 packages and running the proposal_demo client.

PR #46, opened 2016-04-14
Reusing partitions

ancorgs:
General mechanism to reuse partitions including working scenarios with swap and EFI.

### DIFF
--- a/src/lib/storage/disk_analyzer.rb
+++ b/src/lib/storage/disk_analyzer.rb
@@ -72,25 +72,58 @@ module Yast
 
       DEFAULT_DISK_CHECK_LIMIT = 10
 
-      # TODO: document
+      # Representation of a partition from DiskAnalyzer point of view,
+      # consisting on 2 attributes: name (device name) and size (DiskSize).
+      # @!parse class Partition; end
       Partition = Struct.new(:name, :size)
 
-      # TODO: document
-      attr_reader :installation_disks, :candidate_disks
-      attr_reader :windows_partitions, :linux_partitions, :efi_partitions
-      attr_reader :prep_partitions, :swap_partitions
-      attr_reader :devicegraph
+      # @return [Array<String>] device names of installation media.
+      #       Filled by #analyze.
+      attr_reader :installation_disks
+
+      # @return [Array<String>] device name of disks to install on.
+      #       Filled by #analyze.
+      attr_reader :candidate_disks
+
+      # @return [Hash] Linux partitions found in each candidate disk.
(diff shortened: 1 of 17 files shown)


ancorgs:
Now all methods DiskAnalyzer#xxx_partitions return lists of `DiskAnalyzer::Partition` objects, indexed by disk. A `DiskAnalyzer::Partition` object contains both the name and the size.

src/lib/storage/disk_analyzer.rb
```diff
@@ -438,6 +415,36 @@ def remove_installation_disks(disks)
...
+      # @param ids [::Storage::ID, Array<::Storage::ID>]
+      # @return [Hash]
+      def partitions_with_id(ids)
+        pairs = candidate_disks.map do |disk_name|
+          partitions = disks.with(name: disk_name).partitions.with(id: ids).to_a
+          partitions.map! { |part| Partition.new(part.name, DiskSize.kiB(part.size_k)) }
+          [disk_name, partitions ]
+        end
+        Hash[pairs]
+      end
```
wfeldt: I'm all for returning not just the device names; but why not use the existing Storage objects directly? Having
two different Partition classes doesn't help readability. And the same question comes up for \@root_device, etc...

ancorgs: Let's discuss it on Monday with @shundhammer

src/lib/storage/disk_analyzer.rb
```diff
@@ -71,21 +72,53 @@ class DiskAnalyzer
...
-      attr_reader :windows_partitions, :linux_partitions, :efi_partitions
+      # @return [Array<String>] device names of installation media.
+      #       Filled by #analyze.
+      attr_reader :installation_disks
+
+      # @return [Array<String>] device name of disks to install on.
+      #       Filled by #analyze.
+      attr_reader :candidate_disks
+
+      # @return [Hash] Linux partitions found in each candidate disk.
```
shundhammer: +1 for improving documentation for those accessors, but
maybe add [Hash<storage::Partition>] (or whatever it is) for clarification?

wfeldt:
lgtm

src/lib/storage/disk_analyzer.rb
ancorgs: It would be something like... let me think...
```
@return [Hash{String => Array<::Storage::Partition>}] blah
```
I will add it

shundhammer:
Besides that small documentation issue LGTM.

shundhammer:
LGTM

PR #47, opened 2016-04-14
Legacy

wfeldt:


wfeldt:
this is WIP

src/lib/storage/boot_requirements_strategies/base.rb
```diff
@@ -22,6 +22,13 @@
 # find current contact information at www.suse.com.
 
 require "yast"
+require "storage/refinements"
+require "storage/disk_size"
+require "storage/planned_volume"
+require "storage/planned_volumes_list"
+require "storage/proposal/exceptions"
```
ancorgs: It looks strange that a class that lives out of the Proposal namespace contains this include.

src/lib/storage/boot_requirements_strategies/base.rb
```diff
@@ -31,9 +38,14 @@ module BootRequirementsStrategies
       class Base
         include Yast::Logger
 
+        using Refinements::Disk
+        using Refinements::DevicegraphLists
+
         def initialize(settings, disk_analyzer)
           @settings = settings
           @disk_analyzer = disk_analyzer
+          @root_disk = @disk_analyzer.devicegraph.disks.with(name: settings.root_device).first
```
ancorgs: Hmmm, AFAIK one of the goals of the `DiskAnalyzer` class was to hide the existence and details of `Devicegraph` (you know, "we are not in the business of doing devicegraphs")[1]. Not sure if `@root_disk` is a consistent idea with the original design.

[1] From that point of view, maybe even fact that `DiskAnalyzer#devicegraph` is public can be considered kind of a bug.

src/lib/storage/boot_requirements_strategies/legacy.rb
```diff
@@ -21,18 +21,73 @@
...
+        using Refinements::Disk
+
+        def needed_partitions
+          volumes = super
+          volumes << grub_volume if grub_partition_required? && grub_partition_missing?
+          if mbr_gap_required?
+            mbr_gap = disk_analyzer.mbr_gap[settings.root_device]
+            # fail if gap is too small
+            if mbr_gap < DiskSize.kiB(256)
+              raise Proposal::Error
```
ancorgs: Same than before. `BootRequirementsChecker` does not live inside the `Proposal` namespace. It should not raise `Proposal::Error`

I'm not saying that we should not move `BootRequirementsChecker` into `Proposal` if we decide so.

src/lib/storage/boot_requirements_strategies/legacy.rb
```diff
@@ -21,18 +21,73 @@
...
+      protected
+
+        def grub_partition_missing?
+          partitions = disk_analyzer.grub_partitions[settings.root_device]
+          partitions.nil? || partitions.empty?
+        end
+
+        # only needed on GPT partition table
+        # FIXME: currently nobody creates a partition table on an empty disk...
+        def grub_partition_required?
```
ancorgs: For me it makes sense that `DiskAnalyzer` also returns the type of partition present in each candidate disk, instead of having this check here.

Actually, in #46 I propose to index by disk most information returned by `DiskAnalyzer`

src/lib/storage/refinements.rb
```diff
@@ -0,0 +1,42 @@
...
+
+#
+# add some useful methods to libstorage-ng objects
+#
+
+require "storage"
+
+module Storage
+ class Disk
+      def inspect
```
ancorgs: Maybe it makes sense to implement this kind of things directly in libstorage?
http://www.swig.org/Doc1.3/Ruby.html#Ruby_nn54

src/lib/storage/boot_requirements_strategies/base.rb
wfeldt: I also think this is clumsy. IMO it would help if we'd pass more the disk objects around instead of just device names. Then you don't constantly have to look up the actual object in some strange collection thing.

src/lib/storage/boot_requirements_strategies/legacy.rb
wfeldt: Create a Checker Error and raise that?

src/lib/storage/refinements.rb
wfeldt: Ultimately, I think you are right. But that applies to most of the things currently in refinements. Could we compromise
to keep them _until_ they are implemented in libstorage?

ancorgs: For things that will eventually appear in libstorage in a future (like #inspect or #partition_table?), I agree that monkey patching is the way to go. No reason to create a refinement that you need to explicitly activate everywhere. Just patch the class for the time being.

src/lib/storage/boot_requirements_strategies/legacy.rb
ancorgs: Sounds better, yes.

On the other hand, raising an exception here looks like the right thing to do right now. But maybe we should discuss later if we can do something better in general. I mean, maybe the fact that the current gap is not big enough is fixable in next steps of the proposal (there is a change that we end up moving or deleting the first partition).

Anyways, something to write down and discuss later. Probably not worth considering in the end.

wfeldt:
replaced by https://github.com/yast/yast-storage-ng/pull/48

PR #48, opened 2016-04-19
implement legacy strategy

wfeldt:
- I had to add #device_by_name to DiskAnalyzer to get the root device object
- there's still the unsolved problem where to create the partition table for a blank disk

### DIFF
--- a/src/lib/storage/boot_requirements_strategies/base.rb
+++ b/src/lib/storage/boot_requirements_strategies/base.rb
@@ -22,7 +22,7 @@
 # find current contact information at www.suse.com.
 
 require "yast"
-require "storage/refinements"
+require "storage/patches"
 require "storage/disk_size"
 require "storage/planned_volume"
 require "storage/planned_volumes_list"
--- a/src/lib/storage/boot_requirements_strategies/base.rb
+++ b/src/lib/storage/boot_requirements_strategies/base.rb
@@ -51,6 +51,7 @@ module Yast
 
         attr_reader :settings
         attr_reader :disk_analyzer
+        attr_reader :root_disk
 
         def boot_partition_needed?
           settings.use_lvm # || settings.encrypted
--- a/src/lib/storage/boot_requirements_strategies/base.rb
+++ b/src/lib/storage/boot_requirements_strategies/base.rb
@@ -22,6 +22,10 @@
 # find current contact information at www.suse.com.
 
 require "yast"
+require "storage/refinements"
+require "storage/disk_size"
+require "storage/planned_volume"
+require "storage/planned_volumes_list"
 
 module Yast
   module Storage
@@ -34,6 +38,7 @@ module Yast
         def initialize(settings, disk_analyzer)
           @settings = settings
           @disk_analyzer = disk_analyzer
+          @root_disk = @disk_analyzer.device_by_name(settings.root_device)
         end
 
         def needed_partitions
--- a/src/lib/storage/boot_requirements_strategies/legacy.rb
+++ b/src/lib/storage/boot_requirements_strategies/legacy.rb
@@ -21,11 +21,7 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "yast"
 require "storage/boot_requirements_strategies/base"
-require "storage/planned_volume"
-require "storage/planned_volumes_list"
-require "storage/disk_size"
 
 module Yast
   module Storage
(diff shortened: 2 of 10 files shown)


src/lib/storage/boot_requirements_strategies/base.rb
```diff
@@ -34,6 +38,7 @@ class Base
         def initialize(settings, disk_analyzer)
           @settings = settings
           @disk_analyzer = disk_analyzer
+          @root_disk = @disk_analyzer.device_by_name(settings.root_device)
```
ancorgs: NP: can we please add a protected `attr_reader :root_disk`?

src/lib/storage/disk_analyzer.rb
```diff
@@ -155,9 +169,18 @@ def analyze(devicegraph)
...
+        log.info("GRUB    partitions: #{@grub_partitions}")
         log.info("Swap    partitions: #{@swap_partitions}")
+        log.info("MBR gap: #{@mbr_gap}")
       end
 
+    # Look up devicegraph element by device name.
+    #
+    # @return [<::Storage::Device}>]
+    def device_by_name(name)
+      devicegraph.disks.with(name: name).first
```
ancorgs: Looks like it only find disks, not devices in general.

src/lib/storage/disk_analyzer.rb
```diff
@@ -261,6 +292,33 @@ def find_linux_partitions
...
+      #
+      # The result is a Hash in which each key is the name of a candidate disk
+      # and the value is the DiskSize of the MBR gap.
+      #
+      # Note: the gap sizes on non-DOS partition tables are 0 (by definition).
+      #
+      # FIXME: sizes in Region are more or less useless atm, Arvin will fix this.
+      # If that's done switch from kb to byte units.
+      #
+      # @return [Hash{String => Array<Yast::Storage::DiskSize>}]
```
ancorgs: This would actually be
```
@return [Hash{String => Yast::Storage::DiskSize}]
```

src/lib/storage/disk_analyzer.rb
```diff
@@ -101,11 +101,21 @@ class DiskAnalyzer
...
+      #     found in each candidate disk. Filled by #analyze.
+      #     @see #find_grub_partitions
+      attr_reader :grub_partitions
+
       # @return [Hash{String => Array<::Storage::Partition>}] Swap partitions
       #     found in each candidate disk. Filled by #analyze.
       #     @see #find_swap_partitions
       attr_reader :swap_partitions
 
+      # @return [Hash{String => Array<Yast::Storage::DiskSize>}] MBR gap sizes
```
ancorgs: This would actually be
```
@return [Hash{String => Yast::Storage::DiskSize}]
```

ancorgs:
Just a couple of minor comments. Other than that (and beside the lack of tests and rubocop complains), it looks good so far.

src/lib/storage/refinements.rb
```diff
@@ -0,0 +1,55 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+#
+# add some useful methods to libstorage-ng objects
+#
+
```
ancorgs: "Refinements" is the name of a known concept in Ruby. We should not abuse the word.

wfeldt: "patches"?

ancorgs: ok

wfeldt:
Ancor, any objection to merging this so far? I promise to add test cases later but the L3 distracts me
for the moment.

test/boot_requirements_checker_test.rb
```diff
@@ -44,6 +44,9 @@ def find_vol(mount_point, volumes)
     end
     let(:analyzer) { instance_double("Yast::Storage::DiskAnalyzer") }
     let(:storage_arch) { instance_double("::Storage::Arch") }
+    let(:dev_sda) { instance_double("::Storage::Disk") }
+    let(:pt_gpt) { instance_double("::Storage::PartitionTable") }
```
ancorgs: Mmm, why do we need `pt_gpt`? It does not seem to be used.

wfeldt: the test cases will be extended and it will be used there

ancorgs:
LGTM so far

src/lib/storage/boot_requirements_strategies/legacy.rb
```diff
@@ -21,18 +21,67 @@
...
       # Strategy to calculate the boot requirements in a legacy system (x86
       # without EFI)
       class Legacy < Base
+        def needed_partitions
+          volumes = super
+          volumes << grub_volume if grub_partition_required? && grub_partition_missing?
+          if mbr_gap_required?
+            mbr_gap = disk_analyzer.mbr_gap[settings.root_device]
+            # fail if gap is too small
+            raise BootRequirementsChecker::Error if mbr_gap < DiskSize.kiB(256)
```
ancorgs: I would tend to expect that you need to require the file in which `BootRequirementsChecker::Error` is defined in order to use it here... which would become quite messy since that file already requires the strategies.

ancorgs: In other words, so far the visibility goes from BootRequirementsChecker to the strategies. This introduces visibility the other way around, which is unexpected.

ancorgs: I would suggest something like https://github.com/yast/yast-storage-ng/pull/52

wfeldt: I agree, it looked somewhat strange.
So, in general, "thou shalt not raise another classes exception"?

PR #49, opened 2016-04-19
Reusing of swap's UUIDs and labels

ancorgs:


src/lib/storage/proposal/devicegraph_generator.rb
```diff
@@ -92,6 +93,19 @@ def provide_space(volumes, initial_graph, disk_analyzer)
...
+          deleted_swap = space_maker.deleted_partitions.detect do |part|
+            part.id == ::Storage::ID_SWAP
+          end
+          swap_volume = volumes.detect { |vol| vol.mount_point == "swap" }
+          if deleted_swap && swap_volume
+            swap_volume.uuid = deleted_swap.uuid
+            swap_volume.label = deleted_swap.label
+          end
+          refined
+        end
```
wfeldt: And that works when there are several swap partitions?

ancorgs: So far we only generate one swap volume. But it's true this part should be agnostic from that fact. I will modify it so it tries to reuse as many swaps UUIDs as possible.

wfeldt:
this could work

src/lib/storage/proposal/devicegraph_generator.rb
```diff
@@ -92,6 +94,35 @@ def provide_space(volumes, initial_graph, disk_analyzer)
...
+            deleted_swap = deleted_swaps[idx]
+            break unless deleted_swap
+
+            swap_volume.uuid = deleted_swap.filesystem.uuid
+            swap_volume.label = deleted_swap.filesystem.label
+          end
+
+          refined
+        end
+
```
wfeldt: So, you basically keep a collection of old names around and pick from that pool if you need some. Just one question: can it happen that a label you resurrect is already in use?

ancorgs: Mmm, not in the proposal. It only proposes empty labels or reused labels. Of course, there is a chance that the user runs the expert partitioner after the proposal and break things. But I think there is no chance the proposal can break things on its own.

And for the UUIDs, they are generated during filesystem creation and the UUID mechanism is supposed to be robust enough against collisions.

wfeldt:
ok, lgtm

PR #50, opened 2016-04-20
summarize s390 requirements

gabi2:


wfeldt:
It's a start. If I only knew how to detect all those different types...

ancorgs:
Same here. It's a start, but I still don't feel that I could really write the code right from that. I guess I need some investigation before.

gabi2:
I only could talk shortly to Ihno yesterday, he was off-site and today/tomorrow he will be in meetings. I have sent him this summary, he will have a look and we can talk again on Monday.
For the format/types: LDL and FBA are already detected by libstorage. It shouldn't be a problem  to detect DIAG (info Arvin). Ihno also offers to provide a script which detects DASD types.

gabi2:
Ancorgs: what questions do you have in mind? I can ask Ihno to get missing information.

ancorgs:
@gabi2 Just confirm that I understood it correctly:

 * We will always propose (at least) 2 partitions: `/`and `/boot/zipl`
 * Those two partitions has to be in the same disk (even if it's not 100% mandatory, looks like we should enforce this)
 * That disk can be any DASD or zfcp, as long as it's not formated as LDL, FBA or DIAG

ancorgs:
Just two noob questions

* The /boot/zipl partition is just a regular linux partition (I mean, no special partition id). Isn't it?
* Right now, we always propose a separate boot if using RAID or LVM. Is that still true for s390? i.e. Using RAID in s390 should result in three partitions (/, /boot and /boot/zip)?

wfeldt:
1) yes, and it must be ext2, I think
2) as I understand it, /boot is not required as they copy kernel/initrd into the zipl partition anyway but
better lets get s390 expert confirmation here

gabi2:
@ancorgs: I have talked to Ihno (Steffen also was here); the remaining questions should be answered. In short: an extra /boot partition (on RAID, LVM) is not required, /boot/zipl is enough (it's a regular linux partition), / and /boot/zipl should be on same disk.

ancorgs:
Much more clear now. Thanks.

wfeldt:
clear enough now, imo

ancorgs:
I agree (and the best part is that is exactly what we already implemented in #53). So let's merge this!

wfeldt:
@ancorgs: /boot partition  is created in base.rb - this should probably be changed now...

ancorgs:
@wfeldt that creation is overriden in the `Zipl`subclass because https://github.com/yast/yast-storage-ng/pull/53/files#diff-e654811e79058eb5bae83cc4b1092172L35 doesn't call `super`.

For me, that's fine. The general case is to create it so for me it's fine to have it in the base clase (so most subclasses simply call `super`).

wfeldt:
@ancorgs: I see. Then that's fine.

PR #51, opened 2016-04-20
Sketch of an idea to fill several free spaces up reusing existent code

ancorgs:
Only PartitionCreator is fully working, the rest is temporary broken (but it's not a big deal)

 * The main idea is to reuse the already working algorithm we have for filling a single FreeDiskSpace (`#create_non_lvm_simple`)
 * We introduce a previous step (implemented in a separate class) that decides in which free space should we allocate every volume (not sure about the algorithm yet)
 * Then, the original `#create_non_lvm_simple` is executed for each free space (only with its particular subset of pre-selected volumes)
 * SpaceMaker also uses the new class to decide if the freed space is enough (is not longer just a matter of size)

With this idea, we kinda reduce the problem to write a sensible `VolumesDispatcher#distribute` method.

That method should solve a problem very similar to the well-known one called "variable sized bin packing problem", a.k.a. VSBPP ([info about bin packing in wikipedia](https://en.wikipedia.org/wiki/Bin_packing_problem)). Of course, we don't need a complete solution for it, although it seems to be easily solved using linear programming: [paper](http://ac.els-cdn.com/S0305054806002747/1-s2.0-S0305054806002747-main.pdf?_tid=a7c2ef26-07b3-11e6-ab1d-00000aacb35f&acdnat=1461238067_0741fd8f2be1ee4b5c236adbcfd3a84a), [one example](https://github.com/fdabrandao/vpsolver/blob/0e8d41df4525798b2239b099b4d6e0a5146a13e3/examples/vpsolver/example_vsbpp.py).

The pull request also includes RSpec tests for several scenarios. The tests pass and proper devicegraphs are generated, probing that all the pieces actually fit together. The bad side if that the tests actually fake `VolumesDispatcher#distribute` returning always the best distribution for each scenario. We need to write a proper algorithm achieving that without faking... in next sprint.

The current output of those new tests look like this.
```
Yast::Storage::Proposal::PartitionCreator
  #create_partitions
    when the exact space is available
      creates partitions matching the volume sizes
    when some extra space is available
      distributes the extra space
    when there is no enough space to allocate start of all partitions
      raises an error
    when some volume is marked as 'reuse'
      does not create the reused volumes
      distributes extra space between the new (not reused) volumes
    when a ms-dos type partition is used
      when the only available space is in an extended partition
        creates all partitions as logical
      when the only available space is completely unassigned
        creates primary/extended/logical partitions as needed
    when there are several free spaces
      if all the volumes fit in one space
        creates all the partitions in the same space
        uses the biggest space it can fill completely
      if no single space is big enough
        and all the spaces are inside an extended partition
          creates all partitions as logical
        and the spaces are unassigned (ms-dos partition table)
          creates primary/extended/logical partitions as needed
        if it's possible to avoid gaps
          completely fills all the used spaces
        if it's not possible to fill all the spaces
          creates the smallest possible gap
      if disk restrictions apply to some volume
        if a proper distribution is possible
          honors the disk restrictions
          completely fills all the used spaces
        if the only way to avoid gaps is breaking the disk restrictions
          honors the disk restrictions
          creates the smallest possible gap
        if is not possible to honor the disk restrictions
          raises an error
```

RSpec tests for `SpaceMaker` are not included... yet.


ancorgs:
I still think the general approach of distributing the volumes between the spaces and then running `#create_non_lvm_simple` for each space is worth exploring (and this pull request proves that it would be very easy to implement). BUT the algorithm implemented in `#distributed_volumes` absolutely sucks. I will try to come up with something better.

ancorgs:
I'm considering to add `target_size` (`:min` or `:desired`) as an attribute to `PlannedVolumesList`. 90% of the times we pass a list of volumes to a method, we also pass the target size. I'll do if nobody is against.

src/lib/storage/proposal/volumes_dispatcher.rb
```diff
@@ -0,0 +1,89 @@
...
+        # different approaches.
+        #
+        # @raise NoDiskSpaceError if it's not unable to do the matching
+        #
+        # @param volumes [PlannedVolumesList]
+        # @param free_spaces [Array<FreeDiskSpace>]
+        # @param target_size [Symbol] :desired or :min
+        #
+        # @return [Hash{FreeDiskSpace => PlannedVolumesList]
+        def distribution(volumes, free_spaces, target_size)
```
shundhammer: 'distribute' might be a better name for this; we are using verbs in the imperative form everywhere else. A noun would be more fitting for a class IMHO.

ancorgs:
With this idea, we reduce the problem to a well-known one - the [multiple knapsack problem](https://en.wikipedia.org/wiki/Knapsack_problem#Multiple_knapsack_problem) (distributing N parcels or partitions into M knapsacks or spaces).

ancorgs:
Mmm, not really. The bin packing problem assumes all the bins has the same capacity.

ancorgs:
It's actually the "variable sized bin packing problem"... this is starting to turn complicated :smile:

ancorgs:
And looks like the variable sized bin packing problem VSBPP is quickly and easily solved using linear programming: http://ac.els-cdn.com/S0305054806002747/1-s2.0-S0305054806002747-main.pdf?_tid=a7c2ef26-07b3-11e6-ab1d-00000aacb35f&acdnat=1461238067_0741fd8f2be1ee4b5c236adbcfd3a84a https://github.com/fdabrandao/vpsolver/blob/0e8d41df4525798b2239b099b4d6e0a5146a13e3/examples/vpsolver/example_vsbpp.py

ancorgs:
Closed by #62

PR #54, opened 2016-04-26
- fixed unit symbol for Kibibyte (see https://en.wikipedia.org/wiki/Kibibyte)

aschnell:


wfeldt:
lgtm

aschnell:
The Jenkins failure surprise me since locally the testsuite runs fine.

PR #56, opened 2016-04-26
Better handling of logical partitions in PartitionCreator

ancorgs:
This fixes two bugs:

 * In some cases it created logical partitions out of the extended one (we even had a test that considered that to be a valid expected result)
 * In some cases, it created primary partitions inside the extended one (resulting in inconsistent devicegraphs). New test to prevent this included.


src/lib/storage/proposal/partition_creator.rb
```diff
@@ -281,14 +281,27 @@ def create_partition(vol, partition_id, free_slot)
...
+          if space_inside_extended?(free_slot)
+            logical = true
+          elsif logical_partition_preferred?(ptable)
+            # It's not possible to have more than one extended partition,
+            # even if logical partitions are preferred
+            if !ptable.has_extended
+              create_extended_partition(disk, free_slot.slot.region)
+              logical = true
+            end
+          end
```
wfeldt: I'm wondering if this can happen:

[ small_gap | p1 | p2 | p3 | big_gap ]

Then create a new partition -> fits in small_gap -> needs extended -> small_gap becomes extended, big_gap is lost.

Hm...

wfeldt:
lgtm; I think the case mentioned above can't happen starting from a valid pt

src/lib/storage/proposal/partition_creator.rb
ancorgs: @wfeldt Actually, the current code simply can't handle more than one space. See PR in https://trello.com/c/Yvl7G5tB/ for initial experiments about it... and that case is indeed something worth adding to the list of scenarios in that PBI,

wfeldt: Uhm, it really **can** happen:
[ gap1, p1, p2, gap2 ] ->  add small partition ->
[p0, gap1, p1, p2, gap2] -> add small partition -> boom

wfeldt: I would, in the strategy above, convert the biggest free space into extended and then try again.

PR #57, opened 2016-04-26
added tests for legacy boot requirements

wfeldt:


test/boot_requirements_checker_test.rb
```diff
@@ -183,8 +184,65 @@ def find_vol(mount_point, volumes)
...
+                expect(checker.needed_partitions).to include(
+                  an_object_with_fields(partition_id: ::Storage::ID_GPT_BIOS, reuse: nil)
+                )
+              end
+            end
+
+            context "if there is already a GRUB partition" do
+              let(:grub_partitions) { { dev_sda.name => [analyzer_part(dev_sda.name + "2")] } }
+
+              it "does not propose a new GRUB partition" do
```
ancorgs: NP: I expected this to be equal to the equivalent ms-dos part, i.e.
```
it "does not require any particular volume" do
  expect(checker.needed_partitions).to be_empty
end
```

ancorgs:
Looks good (just a minor comment), although I still miss the lvm path.

test/boot_requirements_checker_test.rb
wfeldt: I was thinking that we probably should also do this reuse stuff here and then we'd have to look for a partition with reuse flag. But the checker code atm doesn't do this and I didn't want to change it in this pr, so I coded it this way.

wfeldt:
The lvm path is identical and use_lvm is currently not touched in legacy.rb; we are currently at the point where the number of test cases will explode. Doesn't rspec have a way to avoid copy-and-paste here? And, btw, should we split boot_requirements_checker_test.rb into separate files?

ancorgs:
@wfeldt In this particular case, I actually wanted the test to be quite redundant, with the goal of getting a very exhaustive output presenting the full picture to the experts. Something like http://paste.opensuse.org/31210067

Looking at that output, it seems obvious that the x86 part should by DRY-able without loosing relevant information. That's true and we should do it. But I would like the PPC and S390 part to remain verbose and comprehensive, even at the cost of some RSpec code duplication.

wfeldt:
Good point. But then I would suggest to unify the wording and/or be more precise. For example:

"requires only a new /boot/efi partition" -> "adds a 100 MiB vfat /boot/efi partition"

ancorgs:
It actually contains a full description of the partition. It has a part describing which partitions it proposes and then another part describing each partition in detail

ancorgs:
@wfeldt #61 should address all your concerns while keeping the comprehensive and explicit output

PR #59, opened 2016-04-27
New new libstorage

wfeldt:
this is WIP

- there are still size_k uses, currently covered up by patching a size_k method
- disk sizes in test yaml files must be checked, partitions might no longer fit
- DiskSize class doesn't use bytes internally but should
- new Topology class not yet used; instead fake partitions a 1MiB aligned
- ruby segfaults ATM when running unit tests


src/examples/x9.rb
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env ruby
```
ancorgs: A script called `x9`, without comments and with variables like `sm, dg, r_all, dd, sl, r1, dp`.... To be honest, I would expect the content of `src/examples' to be more didactic :wink:

src/examples/yaml_writer_demo.rb
```diff
@@ -20,13 +20,16 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "yast"  # changes $LOAD_PATH
+
 $LOAD_PATH.unshift(File.expand_path("../../lib", __FILE__))
 
 require "storage/storage_manager.rb"
 require "storage/yaml_writer.rb"
+require "pp"
```
ancorgs: Why?

src/lib/storage/disk_size.rb
```diff
@@ -135,9 +172,9 @@ def unit_multiplier(unit)
       def +(other)
         return DiskSize.unlimited if any_operand_unlimited?(other)
         if other.is_a?(Numeric)
-          DiskSize.new(@size_k + other)
-        elsif other.respond_to?(:size_k)
-          DiskSize.new(@size_k + other.size_k)
+          DiskSize.new(@size_b + other)
```
ancorgs: This may look NP, but you use `@size_b` directly everywhere in the file. I you would use `size_b` (the accessor instead of directly the instance variable), it would be more robust to changes in the future. Hopefully we will not change the internal unit again but in general I like to avoid direct access to instance variables, even within the class (uniform access principle).

src/examples/x9.rb
wfeldt: Sorry, the script was temporary, will remove.

src/examples/yaml_writer_demo.rb
wfeldt: There are a few probably, will check which can be removed again.

src/lib/storage/disk_size.rb
```diff
@@ -190,23 +227,41 @@ def <=>(other)
...
       # Return numeric size and unit ("MiB", "GiB", ...) in human-readable form
       # @return [Array] [size, unit]
       def to_human_readable
-        return [UNLIMITED, ""] if size_k == -1
+        return [UNLIMITED, ""] if @size_b == -1
 
         unit_index = 0
-        size = @size_k.to_f
+        # prefer, 0.50 MiB over 512 KiB
+        size2 = @size_b * 2
```
ancorgs: Maybe `size2 = size_b * 2.0`?

src/lib/storage/disk_size.rb
wfeldt: Ok, would you agree to redo such things in a new pr?

src/lib/storage/disk_size.rb
wfeldt: No, both are integers.

src/lib/storage/fake_device_factory.rb
```diff
@@ -223,13 +257,14 @@ def str_to_ptable_type(string)
       def create_partition(parent, args)
         log.info("#{__method__}( #{parent}, #{args} )")
         disk_name = parent
-        size      = args["size"] || DiskSize.zero
+        size      = args["size"] || DiskSize.unlimited
```
ancorgs: What does it mean to have a partition with unlimited size?

wfeldt: -> fill all (the remaining) space

src/lib/storage/fake_device_factory.rb
```diff
@@ -86,8 +84,9 @@ def initialize(devicegraph)
         super(devicegraph)
         @partitions     = {}
         @disks          = Set.new
-        @first_free_cyl = {}
-        @cyl_count      = {}
+        @free_blob      = nil
+        @free_regions   = []
+        @mbr_gap        = nil
```
ancorgs: Hmmm... is ok to have this kind of initializations in the factory constructor? Shouldn't be them executed in every call to `load_yaml_file`? According to the documentation is ok to have something like:
```ruby
factory = FakeDeviceFactory.new(devicegraph)
yaml_files.each do |file|
  factory.load_yaml_file(file)
end
```
Would that work?

src/lib/storage/patches.rb
```diff
@@ -63,12 +63,97 @@ def dasd_type
...
+      flags += "e" if self.extended_possible
+      flags += "L" if self.logical_slot
+      flags += "l" if self.logical_possible
+      nice_size = Yast::Storage::DiskSize.B(region.length * region.block_size)
+      "<PartitionSlot #{self.nr} #{self.name} #{flags} #{nice_size}, #{self.region.show_range}>"
+    end
+
+    alias to_s inspect
+
+    def type?(type)
```
ancorgs: It is planned to add these methods to libstorage-ng any time soon? That's the goal of `patches.rb`, to temporarily provide stuff that will be eventually provided by the library.

wfeldt: Arvin wanted to add some introspection on the libstorage side; until then, we should keep at least something.

src/lib/storage/refinements/partition_attributes.rb
```diff
@@ -42,9 +42,9 @@ def label
           end
 
           # @return [DiskSize]
-          def size
-            DiskSize.KiB(size_k)
-          end
+#          def size
+#            DiskSize.KiB(size_k)
+#          end
```
ancorgs: Now `partition.size` returns a Fixnum (bytes) isn't it? In that case, instead of commenting this method I would prefer to rename it to `disk_size` or something similar.

src/lib/storage/disk_size.rb
ancorgs: Sure

src/lib/storage/fake_device_factory.rb
wfeldt: Actually, yes (but for a bug I just noticed).

But this is in fact something I wanted to ask you about. As something like
`@mbr_gap        = nil`
looks pretty weird, or at least highly un-ruby, to me.

src/lib/storage/refinements/partition_attributes.rb
wfeldt: This should go away completely. size is provided in libstorage. There might be some size_k's left to kill, though. But let's do this later.

src/lib/storage/fake_device_factory.rb
wfeldt: That is, I want to have the internal class vars shown in one place without having to add some dummy initialization.

doc/fake-devicegraphs-yaml-format.md
```diff
@@ -136,12 +153,28 @@ Example:
   beginning of that extended partition, so the last one of those logical
   partitions might have size "unlimited" again.
 
+  size may be missing and defaults to 'unlimited'.
+
+  Note: all size values are exact. So, for example, on a 100 GiB disk you
+  can't create a 100 GiB partition but only one with 1023 MiB.
```
ancorgs: I got lost. Didn't you mean "on a 1 GiB disk you can't create a 1 GiB partition but only one with 1023 MiB"? 100GiB would be 102400 MiB. Isn't it?

On the other hand, I'm not sure if it's obvious for all readers why the maximum size is 1023.

src/lib/storage/fake_device_factory.rb
ancorgs: We have that kind of initializations to nil in several classes. It has already stopped looking weird to me. :wink:

doc/fake-devicegraphs-yaml-format.md
wfeldt: true enough; adjusted

src/lib/storage/fake_device_factory.rb
wfeldt: Ok, in this case see my last blob resetting most of them in #create_disk.

ancorgs:
The fact that libstorage switched from `size_k` to `size` instead of `size_b` really destroys all the consistency in our APIs. :disappointed: 
Before the change you could trust that any method called `#size` or `#whatever_size` returned a `DiskSize` objects and only methods with more explicit names like `size_k` returned Fixnums. Now `Partition#size` returns a Fixnum and `FreeDiskSpace#size` returns a `DiskSpace` (just two examples). That sucks.

ancorgs:
This is still a WIP, but I'm not against merging since is not more broken than before

PR #62, opened 2016-06-02
Working code to use multiple spaces

ancorgs:
This closes #51 

This includes a working implementation of the idea sketched in #51. With tests for many situations including all kind of crazy combinations of extended and primary partitions.

There are still some TODOs about updating documentation of some methods.

We need to rethink SpaceMaker a little bit to make it slightly smarter. After that, we definitely need to rethink its unit tests.

It would also be nice to add more and better unit tests for DevicegraphGenerator... but we need to stop at some point. :smile: 

It's recommended to review it by individual commits. Taking a look to the tests for the new/changed classes in advance to checking the code can also be useful.

### DIFF
--- a/src/lib/storage/free_disk_space.rb
+++ b/src/lib/storage/free_disk_space.rb
@@ -58,6 +58,18 @@ module Yast
       def size
         DiskSize.new(@slot.region.to_kb(@slot.region.length))
       end
+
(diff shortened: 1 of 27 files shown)


wfeldt:
looks reasonable; tests will break horribly, though, but that can't be helped for the moment
so, go ahead

PR #63, opened 2016-06-03
Smarter SpaceMaker with enhanced tests 

ancorgs:
* It honors volume's disk restrictions (don't waste time freeing space in the wrong disk)
 * It tries to resize Windows partitions before deleting them
 * Much better tests

### DIFF
--- a/src/lib/storage/proposal/space_maker.rb
+++ b/src/lib/storage/proposal/space_maker.rb
@@ -68,9 +68,8 @@ module Yast
         #     volumes
         #
         def provide_space(volumes)
-          new_graph = original_graph.copy
+          @new_graph = original_graph.copy
           @deleted_names = []
-          @distribution = nil
 
           # Partitions that should not be deleted
           keep = []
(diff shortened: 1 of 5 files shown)


src/lib/storage/proposal/space_maker.rb
```diff
@@ -211,71 +273,93 @@ def shrink_partition(partition, shrink_size)
...
+        # @see #deletion_candidate_partitions for supported types
         #
-        # @param devicegraph [DeviceGraph] devicegraph to update
         # @param volumes [PlannedVolumesList]
+        # @param type [Symbol] type of partition to delete
         # @param keep [Array<String>] device names of partitions that should not
         #       be deleted
-        def delete_partitions!(devicegraph, volumes, keep)
+        # @param disk [String] optional disk name to restrict operations to
+        def delete_partitions!(volumes, type, keep, disk)
```
wfeldt: Would it make sense to have an early 'if success?' here? If only to avoid repeatedly building of the partition list and not filling the log unnecessarily?

wfeldt:
You are deleting partitions starting with the first; would it make sense to reverse the order per disk so that the last partition on a disk is deleted first?

src/lib/storage/proposal/space_maker.rb
ancorgs: Yes, let's make it more consistent with `resize_windows!`

wfeldt:
lgtm - if you are fine with the failing tests...

ancorgs:
The failing tests should be a consequence of outdated libstorage-ng. So I will wait until tomorrow for OBS to publish everything and will re-run travis then.

ancorgs:
OBS did it job. Tests are finally green.

PR #70, opened 2016-06-14
Turn all unit tests green again

ancorgs:

PR #71, opened 2016-06-14
rubocup fixes

wfeldt:
fix a number of rubocup complaints and adjust config a bit

### DIFF
--- a/src/lib/storage/fake_device_factory.rb
+++ b/src/lib/storage/fake_device_factory.rb
@@ -54,12 +54,18 @@ module Yast
       # Sub-products are not listed here.
       VALID_PARAM =
         {
-          "disk"            => ["name", "size", "block_size", "io_size", "min_grain", "align_ofs", "mbr_gap"],
+          "disk"            => [
+            "name", "size", "block_size", "io_size", "min_grain", "align_ofs", "mbr_gap"
+          ],
           "partition_table" => [],
           "partitions"      => [],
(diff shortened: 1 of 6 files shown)


src/lib/storage/fake_device_factory.rb
```diff
@@ -281,9 +287,9 @@ def create_partition(parent, args)
 
         # partitions are created in order, so first suitable slot should be fine
         # note: skip areas we marked as empty
-        slot = slots.find { |slot|
-          slot.possible?(type) && !@free_regions.member?(slot.region.start)
-        }
+        slot = slots.find do |s|
```
shundhammer: I liked the old version better. Was that something Rubocop wanted to be changed? Or was it a line length limitation?

IHMO that only shows how absurd many of the Rubocop settings are.

shundhammer:
Other than that one issue I commented above, LGTM.

src/lib/storage/fake_device_factory.rb
wfeldt: rubo wanted {} -> do ... end

shundhammer: Well, yes, but then maybe simply keep "slot" as the variable name?

wfeldt: Ah, you mean that. It wanted that also ('slot' shadows the outside 'slot').

ancorgs:
#70 already solves all these offenses (and any other)

PR #72, opened 2016-06-17
Rename size methods according to the returned type

ancorgs:


ancorgs:
This was already LGTMed by @shundhammer in IRC

PR #74, opened 2016-06-17
automatically create boot requirements document

wfeldt:
- added rspec formatter creating markdown
- added rake task creating boot requirements as MD file


ancorgs:
Rubocop is offended. After fixing that, it LGTM

PR #83, opened 2016-07-15
rubocop update

jreidinger:


src/lib/storage/fake_device_factory.rb
```diff
@@ -212,9 +212,9 @@ def set_topology_attributes!(disk, args)
         io_size = args["io_size"]
         min_grain = args["min_grain"]
         align_ofs = args["align_ofs"]
-        disk.topology.optimal_io_size = io_size.size if io_size && io_size.size > 0
+        disk.topology.optimal_io_size = io_size.size if io_size && !io_size.empty?
```
aschnell: Is that really readable? What does empty mean for a size?

src/lib/storage/fake_device_factory.rb
```diff
@@ -242,7 +242,7 @@ def create_partition_table(parent, args)
       # @return [Fixnum]
       def str_to_ptable_type(string)
         # Allow different spelling
-        string = "msdos" if string.downcase == "ms-dos"
+        string = "msdos" if string.casecmp("ms-dos").zero?
```
aschnell: I find the old version much better readable. Now you have to lookup casecmp to know that it returns for equal strings, e.g. could be true or 0.

src/lib/storage/fake_device_factory.rb
lslezak: "empty for a size" ? And what does "size for size" mean as in the original?

Anyway, if something has size greater than zero than it's not empty... 
For me it's better readable.

lslezak:
LGTM

aschnell:
LGTM with all checks failing and coverage descresing by 18%?

lslezak:
Ouch sorry, overlooked that failures, canceling my LGTM... :wink:

src/lib/storage/fake_device_factory.rb
jreidinger: I agree, that size looks strange here, but also as @lslezak mention, `io_size.size` It looks like io_size is not good name...maybe it is number of available blocks? what it actually strands for?

src/lib/storage/fake_device_factory.rb
jreidinger: agreed, it is one of new performance check and maybe we should disable it. ( it is faster this way as casecmp is faster then creating new modified string ). @lslezak what do you think?

jreidinger:
yep, I know about failing tests. Decrease of coverage is caused by failing tests, but from my initial look, I do not see any obvious reason why it failing, so I plan to wait for storage-ng experts for hints, what that exception and missing methods means as from diff I do not see something wrong.

src/lib/storage/fake_device_factory.rb
aschnell: io_size is the optimal input/output size of the disk. So it really represents a size. Technically it is a ```DiskSize``` object. The ```DiskSize``` object has a ```zero?``` method but not an ```empty?``` method (which is correct).

AFAIS the change breaks the testsuite:

> undefined method `empty?' for <DiskSize 2.00 GiB (2147483648)>:Yast::Storage::DiskSize

aschnell:
Explanation given above, DiskSize has no ```empty?``` method. Was that change does automatically by rubocop?

lslezak:
I guess it was changed automatically by Rubocop.

So we have two options:

- Add `empty?` method
- Disable that check

I do not know which one is better in this case...

jreidinger:
yes, automatic one as rubocop expects that when object have `size` method, it also have `empty?`. 
Still even when explained, I found quite strange to have size method for Size object. Maybe method name like to_i ( which indicate conversion to plan integer value ) is better name?

aschnell:
Assuming ```size == 0``` is equal to ```empty?``` is terrible wrong. E.g. you could define ```size``` for a volume group to be the size in bytes and ```empty?``` to mean that it has no logical volumes. Those definitions are totally valid.

I vote to disable that check.

lslezak:
Um yes, you could define `empty?` with a completely different meaning, unrelated to `size`. But I'd avoid that, that would be quite confusing. In that case you could define a separate method e.g. `volumes` and call `empty` on it (`volumes.empty?`).  BTW even C++ STL defines `empty()` related to the `size()`...

So to get forward - if you really want to use `empty?` with a different meaning or do not want to define it at all then let's disable the check.

src/lib/storage/fake_device_factory.rb
lslezak: I do not have a strong option here:

- The old way is probably a bit more readable (but once you learn this idiom it's OK IMHO)
- If you cannot remember it you can still write the old `downcase` style and let `rubocop -a` fix it for you, the change is quick and cheap
- The original is a bit slower (but this is irrelevant in YaST)
- We should not diverge from the upstream Ruby style without any good reason

I'm not sure which item is the most important here...

jreidinger:
I disabled zero predicate checker and revert change. But it still failing with strange StorageException. Not very helpful to find what exactly failing and why. @ancorgs any idea how to get more info from failed tests?

aschnell:
The ```size``` -> ```empty?``` is still done many times:

> undefined method `empty?' for <DiskSize 2.00 GiB (2147483648)>:Yast::Storage::DiskSize

jreidinger:
@aschnell ok, I will check all changes for the size

src/lib/storage/fake_device_factory.rb
jreidinger: when I am thinking about it, you actually would like to do case compare, and downcasing string is just workaround for it. So casecmp do case insensitive comparison, but its value is not so nice. Maybe some helper for String like `case_match?` would be the best.

lslezak:
@jreidinger thanks for updating it.

The last unresolved thing is that `casecmp("ms-dos").zero?` part... 

How should we proceed? I already [commented this change above](https://github.com/yast/yast-storage-ng/pull/83#discussion_r70993443) and I do not have a strong opinion here.

But I'd probably slightly more prefer the new style with `casecmp` to less diverge from the upstream style 
although it's `zero?` part is a bit less readable. On the other hand the `casecmp` part is quite self descriptive and you can guess what it does...

@jreidinger, @aschnell, others... :question::question:

aschnell:
Since I'm not the original author of the code I let @shundhammer and @ancorgs have the last word on the ```casecmp.zero?``` issue.

jreidinger:
same for me as long as it is consistent

src/lib/storage/fake_device_factory.rb
mvidner: String#`casecmp` is consistent with String#`<=>` and also with (libc)  `strcmp`

The original upstream discussion for the cop (and other ones) is https://github.com/bbatsov/rubocop/issues/1677

ancorgs: I don't have a strong opinion either. For me `casecmp` is acceptable, although obviously less readable. In this case the performance gain is marginal but maybe in some other scenario with a big loop or something...

@jreidinger suggests to write a generic helper. For that kind of things we should use add another extension to `Yast::CoreExt`[1]. Let's say `Yast::CoreExt::ComparableString` for example. But let's not forget that this change is suggested because of performance reasons and overloading String is not free performance-wise. So we could end up with a whole mechanism to bring readability back that actually ends up killing the original reason to decrease readability in the first place. :wink:

So in short, I'm fine with the change to start using `casecmp`.  I'm also fine with adding a generic helper to abstract that, but only if we can prove that is worth it from the performance POV.

[1] https://github.com/yast/yast-ruby-bindings/tree/master/src/ruby/yast/core_ext

wfeldt:
BTW, there's a `DiskSize#to_i` (aliased to `#size`) if `#size` irks you.

Whatever the rationale behind appending `.freeze` may be, I will not be tempted to comment on that.

jreidinger:
@wfeldt rationale behind `.freeze`, is that it make constant really constant, so if something modify it, it raise exception and not silently modify it.

jreidinger:
@wfeldt yast, `DiskSize#to_i` make much more sense to me, as size method for me indicate that it returns Size object.

wfeldt:
@jreidinger, I said I will not comment on `.freeze` (because one should not use bad language publicly), not that I can't imagine why it's used.

wfeldt:
**lngtm**, I stongly object to the `.freeze`.

ancorgs:
I don't like the `freeze` either.

wfeldt:
@jreidinger, yes `DiskSize#size`, was not a good choice, that's why we came up with the alias.

ancorgs:
We are in time for dropping `DiskSize#size`. I think only the alias is used currently in the code (that's what I tried as part of https://github.com/yast/yast-storage-ng/pull/72 )

jreidinger:
@ancorgs actually freeze helps me to find some bugs in our testsuite when doing this rubocop update. I found that some constants in tests are passed to methods, that modify it. So in fact, I find it quite good technique to prevent unexpected behavior.

src/lib/storage/fake_device_factory.rb
lslezak: I agree with @ancor, writing a helper for String to make it readable again is actually against the original reason why Rubocop complains... In that case I'd rather keep the current style and disable the Rubocop check.

But there is one important thing: if you cannot remember the exact syntax for the new style comparison you can still use the old way (`foo.downcase == "bar"`) and let `rubocop -a` fix it for you.

(BTW the case insensitive comparison is not used much in YaST so we are actually solving a minor case...)

So to move forward I'm voting for the new syntax :wink: 

Let's use the "reaction" button to vote here:
-  :+1: for the new style
-  :-1: for the old style

lslezak:
IIRC the reason for `freeze` is that in some future Ruby version the constants will be frozen by default. This change makes sure the code is forward compatible.

mvidner:
About `freeze`, yes, it is arguably a bug in the language that a Ruby constant only makes the name-object binding constant, and the object is mutable by default. I think it is an acceptable price to append `.freeze` and get stronger guarantees about program behavior.

wfeldt:
Adding `.freeze` is like optimizing gcc code with `__attribute__`. It's ok when there are one or two places where it's needed. Adding this to all kinds of constant expressions is just nonsense. At one point we strive to make code more readable and on the other hand we add some weird blobs?

If that's really a valuable addition, the ruby language is seriously broken. And if, like @lslezak mentions, this might be made superfluous in the future - all the more reason **not** to change the code now.

rubocop has all kinds of rules for varying purposes. It says nowhere you should aim add enabling them **all**. And this does not look like a useful rule to me.

lslezak:
I believe that Rubocop will add a new check in the future which will remove the `.freeze` modifiers when they become obsolete. I guess simple `rubocop -a` will get rid all of them...

Frankly, I'm not much happy about this change either, but as @jreidinger pointed out he found some bugs in our code which modified the constants (ouch...).

And I do not think it makes the code less readable, the meaning of the `.freeze` modifier is pretty self-descriptive, it's just a bit annoying. I think we can live with that for some time...

jreidinger:
@wfeldt actually for me it is more like `const` then `__attribute__` as it helps to prevent misuse.

lslezak:
Yes, it's like `void * const` (old) vs. `const void * const` (new).

src/lib/storage/abstract_device_factory.rb
```diff
@@ -300,7 +296,6 @@ def check_param(name, param)
       # @param child  [String] name of child  factory product
       #
       def check_hierarchy(parent, child)
-        # rubocop:disable Style/GuardClause
```
ancorgs: For consistency, remove also this four lines below
```
# rubocop:enable Style/GuardClause
```

src/lib/storage/devices_list.rb
```diff
@@ -34,8 +34,6 @@ class DevicesList
 
       class << self
         attr_reader :device_class
-
-        # rubocop:disable TrivialAccessors
```
ancorgs: again, remove also the `enable` below.

src/lib/storage/disk_size.rb
```diff
@@ -145,7 +145,6 @@ def parse(str)
         # "MiB", ...). The base of this exponent is 1024. The base unit is KiB.
         #
         def unit_exponent(unit)
-          # rubocop:disable Style/AndOr
```
ancorgs: One more time, deleting `disable` and leaving `enable` back.

src/lib/storage/fake_device_factory.rb
```diff
@@ -174,7 +174,7 @@ def fixup_param(name, param)
       #
       # FIXME: this method is too complex. It offends three different cops
       # related to complexity.
-      # rubocop:disable Metrics/PerceivedComplexity, Metrics/AbcSize, Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
```
ancorgs: another case of leftover `enable`

src/lib/storage/fake_device_factory.rb
```diff
@@ -270,7 +270,7 @@ def str_to_ptable_type(string)
       #
       # FIXME: this method is too complex. It offends four different cops
       # related to complexity.
-      # rubocop:disable Metrics/PerceivedComplexity, Metrics/AbcSize, Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
```
ancorgs: Guess what? :wink:

src/lib/storage/abstract_device_factory.rb
jreidinger: it is actually auto correction from rubocop, probably only one half is implemented

src/lib/storage/fake_device_factory.rb
jreidinger: really? I see there `rubocop:enable all`

src/lib/storage/proposal/space_maker.rb
```diff
@@ -205,15 +205,17 @@ def resize_windows!(volumes, disk, force: false)
...
               missing_required_size(volumes, disk)
             ].min
             shrink_partition(res[:partition], shrink_size)
-            return if success?(volumes)
+
+            success?(volumes)
           end
-          log.info "Didn't manage to free enough space by resizing Windows"
+
+          log.info "Didn't manage to free enough space by resizing Windows" unless success
```
ancorgs: In this method, I think the original version makes the intention of the code more obvious when compared to the modified version. It would be nice if somebody else than @shundhammer or me do the exercise of reading both versions to confirm my feelings.

src/lib/storage/fake_device_factory.rb
ancorgs: Ah, then no. :wink:  I didn't check. I just saw 'disable' and assumed the same problem-

src/lib/storage/proposal/space_maker.rb
jreidinger: well, the most readable would be to move whole shrinking to own method, to my taste this method is too long, but I do not like to do big refactoring. BTW I agree with rubocop, that having return inside block is quite tricky workflow, that prevents future refactoring, as moving this block to own method then need special handling for such "long jump".

jreidinger: something like
```
success = sorted_resizables(parts_by_disk.values.flatten).any?(&:try_shrink)

log.info "Didn't manage to free enough space by resizing Windows" unless success
```

jreidinger: btw also `sorted_resizables(parts_by_disk.values.flatten)` is not nice api, I expect something like `disks.partitions_with_windows.reduce(:+).sort_by(&:size).any?` which is more readable for me what it actually probably do ( give me disks, their windows partitions and sort it by size )

ancorgs: Well, it's true that removing the `return` from the inner block makes refactoring much easier. You convinced me, let's do the change (we can do the refactoring in a separate PR).

src/lib/storage/fake_device_factory.rb
jreidinger: same as above, enable all, but your comments helps me to check all enable

lslezak:
The new syntax won in the [case insensitive poll](https://github.com/yast/yast-storage-ng/pull/83#discussion_r71327745) 5:0, so let's use the new syntax.

I'm giving my LGTM.

The discussion is quite long but it seems everything is resolved now and we can merge it. If I overlooked something then react ASAP please.

lslezak:
Maybe one more question: Do we agree on the `freeze` usage? We described the reasons in [this and the following comments](https://github.com/yast/yast-storage-ng/pull/83#issuecomment-233618514), is it OK for you? Do you agree with using `freeze` or do you still think we need to discuss it more?

wfeldt:
Sorry, there was nothing putting me in favor of `.freeze`.

If you see code where it helps you, fine with me. But I'm against adding it to every constant unconditionally.

lslezak:
@wfeldt um, it's hard to say where it helps until you actually use it.

It's like a `foo(const string &bar)` method, it would work exactly the same way also after removing that `const`. 

But it is still a good practice to have it there as it prevents from programming mistakes when somebody modifies `bar` by mistake. It's also used unconditionally everywhere.

It's the same situation here in Ruby. It's just a bit more tricky because constants are actually not real constants as someone would expect.

ancorgs:
It may not be the most well-argued opinion but... I simply hate that `freeze` change. :wink:

wfeldt:
@ancorgs: I really like your argument ;-)

ancorgs:
We need an agreement here. I want to merge new code this morning. Should we vote?

lslezak:
OK, so let's vote,  use the *reaction* button to vote here this way:

-  :+1: for the new style (with `freeze`)
-  :-1: for the old style (without `freeze`)

lslezak:
I'm not much happy about adding the `freeze` attribute, but given that it's for the future compatibility and one day (Ruby 3.0?) we can remove it possibly with a simple `rubocop -a` I voted for the new style.

JFYI in the other YaST modules we already switched to the new style, if the old style wins here we might need to update several packages to be consistent...

ancorgs:
To be honest, Ruby constants work as I would have expected (defining an object as a constant does not make it immutable).

I can see situations in which, in addition to defining something like a constant, the programmer would want to make explicit that the object is immutable. But making it for every single constant looks overkill (not to say ugly) to me.

On the other hand, I don't like to diverge from main style guide. So I'm voting against the change, but in case of draw I could change my vote to break the tie.

jreidinger:
one example where I have to disable it, as I need to modify constant with definining default handler - https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/stage1_proposal.rb#L256
but in general I find it useful as constants should be really constant to avoid surprise for people comming from other languages

wfeldt:
This aims at changing the ruby language semantics. If and when ruby itself changes, that's another story. But here it's like using `#define` or `__attribute__` to change C.

And sorry, saying that implementing a change **now** is ok because it can be easily undone **later** seems to be a strange kind of argument to me.

Moreover, it wouldn't be the first time that people used an anticipated future ruby feature in advance only to be hit by the slightly different real implementation in ruby later. So saying that you can painlessly remove the `.freeze` later and be done is not justified.

wfeldt:
Really, don't we have anything better to do?

lslezak:
OK, I have disabled the `Style/MutableConstant` cop check here and reverted the autocorrection done by Rubocop.

ancorgs:
lgtm

shundhammer:
On 19.07.2016 13:49, Steffen Winterfeldt wrote:
> **lngtm**, I stongly object to the `.freeze`.

+1

If this is needed, the interpreter should to it automatically. If it can't do 
that, well, that's just too bad. But that's certainly not something the 
developers should be bothered with.


Kind regards
-- 
Stefan Hundhammer  <shundhammer@suse.de>
YaST Developer

SUSE Linux GmbH
GF: Felix Imendörffer, Jane Smithard, Graham Norton; HRB 21284 (AG Nürnberg)
Maxfeldstr. 5, 90409 Nürnberg, Germany

src/lib/storage/fake_device_factory.rb
shundhammer: On 19.07.2016 14:30, Ladislav Slezák wrote:
>> -        string = "msdos" if string.downcase == "ms-dos"
>> > +        string = "msdos" if string.casecmp("ms-dos").zero?
>
> But there is one important thing: if you cannot remember the exact
 > syntax for the new style comparison you can still use the old way
 > (`foo.downcase == "bar"`) and let `rubocop -a` fix it for you.

I would not consider that a "fix": It auto-replaces something readable with 
something that is hard to understand.


Kind regards
-- 
Stefan Hundhammer  <shundhammer@suse.de>
YaST Developer

SUSE Linux GmbH
GF: Felix Imendörffer, Jane Smithard, Graham Norton; HRB 21284 (AG Nürnberg)
Maxfeldstr. 5, 90409 Nürnberg, Germany

jreidinger:
@shundhammer interpret will do it automatic from ruby3, but now it is not done due to backward compatibility.

PR #85, opened 2016-07-22
Adapt boot requirements to future Grub usage

ancorgs:
It was agreed with Michael Chang that the way we use Grub will change in the near future, embedding grubenv in a raw block writable by grub-once. This adapts the partitioning proposal to that change.


doc/boot-requirements.md
```diff
@@ -63,42 +63,49 @@
...
-		- with GPT partition table
-			- if there is no GRUB partition
-				- in a partitions-based proposal
-					- **only requires a new GRUB partition**
-				- in a LVM-based proposal
-					- **requires /boot and a GRUB partitions**
-			- if there is already a GRUB partition
-				- in a partitions-based proposal
+			- in a partitions-based proposal
+				- if proposing root (/) as Btrfs
```
wfeldt: But surely that does **not** work?

src/lib/storage/boot_requirements_strategies/legacy.rb
```diff
@@ -29,45 +29,45 @@ module BootRequirementsStrategies
       # Strategy to calculate the boot requirements in a legacy system (x86
       # without EFI)
       class Legacy < Base
+        GRUB_SIZE = DiskSize.KiB(256)
+        GRUBENV_SIZE = DiskSize.KiB(2)
```
wfeldt: It said 1 kib in the doc above.

ancorgs: Well, I wanted to stay safe for future. We also have 256 kib for grub even it is currently ~85 kib. We could revisit both values in the future, after the initial mchang's experiments of the new approach.

test/boot_requirements_checker_ppc_test.rb
```diff
@@ -91,7 +94,7 @@
         end
 
         context "if the existent PReP partition is not in the target disk" do
-          let(:prep_partitions) { { "/dev/sdb" => [analyzer_part("/dev/sdb1")] } }
+          let(:prep_partitions) { { "/dev/sdb" => [analyzer_part("/dev/sdb")] } }
```
wfeldt: why `sdb` instead of `sdb1`?

doc/boot-requirements.md
ancorgs: The spec output was validated by Michael Chang. Here the version he "approved" with a couple of explanatory notes (including fragile setups) https://gist.github.com/ancorgs/db671cf6c0c12c5f2c62bae36a7d68c0

test/boot_requirements_checker_ppc_test.rb
ancorgs: That's an error. I probably deleted it by mistake.

doc/boot-requirements.md
wfeldt: Just tested and it did not work. btrfs has only 64k embedding space which is not enough for grub.
Maybe grub has advanced meanwhile - but please check back with mchang.

src/lib/storage/boot_requirements_strategies/legacy.rb
wfeldt: Yes, but in that case the doc says 'check for 256' and we do. Here the docs says we need 1 but the code goes for 2.

I'm not arguing about the value, only for consistency.

doc/boot-requirements.md
ancorgs: He confirmed it twice. First in irc:

> <mchang> We did a similar hack for btrfs .. the environment block is "chained" from file system block to btrfs's reserved area for bootloader.

And also in a mail

> with too small MBR gap or with no MBR gap at all
>             in a partitions-based proposal
>                 if proposing root (/) as Btrfs
>                     does not require any particular volume
> 	mchang: OK

wfeldt: It may be that you can put the env block there but grub itself does not install.

wfeldt:
lgtm but for the boot requirements. I'll only believe when I've seen it work.

doc/boot-requirements.md
WenhuaChang: Hi Steffen,

Thanks for your review. (I think we are talking msdos table on leagacy bios systems)

We fallback to use btrfs bootloader area if mbr gap was too small, due to many laptop/desktop systems shipped with MS Windows are pre-partitioned without enough mbr gap size and our default installation can't work on it. See bsc#879271.

Please note it only works for plain btrfs installation without any lvm and mdadm so that the 64kB size is sufficient. If you have lvm or mdadm, we cannot embed the btrfs area at all because it requires "raw" sectors to work, so even if btrfs offers enough space it still not helpful.

So it's obvious to me. if "map gap too small" error has happened for btrfs on top of lvm/mdadm, we have no way out unless propose a /boot partition for it (with other fs like ext4). 

Hi ancorgs,
I am very sorry for missing out this aspect and the plan should be changed to.

> with too small MBR gap or with no MBR gap at all
>             in a partitions-based proposal
>                 if proposing root (/) as Btrfs
>                     without lvm or mdadm: does not require any particular volume
>                     with lvm or mdadm: require /boot partition with ext4 that supports blocklist.

Is the explanation and plan OK for you and Steffen ?

ancorgs: That's exactly what is implemented - raw Btrfs without LVM. The whole scenario is:
```
not using UEFI (legacy PC)
   with a MS-DOS partition table
      with too small MBR gap
         in a partitions-based proposal <-- i.e. no LVM
            if proposing root (/) as Btrfs
               does not require any particular volume
            if proposing root (/) as non-Btrfs
               raises an exception
         in a LVM-based proposal
            raises an exception
```
Everything clear for me. Exactly as I had understood it.

wfeldt: @WenhuaChang Michael, it's ok when you say you'll implement it this way. But ATM it does **not** work.

WenhuaChang: Sorry but I did not clear on what exactly is "does not work" you meant ? Thanks.

WenhuaChang: Did you see any issue booting from plain btrfs partition that 64kB reserved area isn't enough ?

wfeldt: I've tried with current sle12-sp2 build (no mbr-gap, sda1=1GB swap, sda2=19GB btrfs) and it did not work, grub-install complains with some btrfs something errors. How exactly should grub-install be run in this setup?

WenhuaChang: I think the grub2-install /dev/sda should work for the setup, even the mbr gap is too small it will use btrfs root partition as fallback device which can offer the embedded area of stage2. If not then I think it's a bug. :(

wfeldt: Yes, looks like a bug, then.

PR #89, opened 2016-08-01
Adapt to new namespace conventions

ancorgs:
Replaces #88. This one have explicit commits only moving the files, so git doesn't loose track.


wfeldt:
lgtm

PR #90, opened 2016-09-15
- added new minimalistic inst_disk_proposal client

aschnell:
For https://trello.com/c/9Uj9DmG6/384-3-storageng-installation-storage-proposal.

### DIFF
--- a/src/clients/inst_disk_proposal.rb
+++ b/src/clients/inst_disk_proposal.rb
@@ -1,0 +1,85 @@
+# encoding: utf-8
+
+# Copyright (c) [2016] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact Novell, Inc.
+#
+# To contact Novell about this file by physical or electronic mail, you may
+# find current contact information at www.novell.com.
+
+require "yast"
+require "y2storage"
+
+Yast.import "UI"
+Yast.import "Wizard"
+Yast.import "HTML"
+
+include Yast::I18n
+include Yast::Logger
+
+module Yast
+  #
+  # client to calculate the storage proposal during installation and provide
+  # the user a summary of the storage proposal
+  #
+  class InstDiskProposalClient < Client
+    using Y2Storage::Refinements::Devicegraph
+
+    def main
+      textdomain "storage"
+
+      begin
+        proposal = Y2Storage::Proposal.new(settings: settings)
+        proposal.propose
+        actiongraph = proposal.devices.actiongraph
+        summary = summary(actiongraph)
+      rescue Y2Storage::Proposal::Error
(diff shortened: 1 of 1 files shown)


shundhammer:
LGTM, but then I would not consider myself a Ruby expert. Let's wait for one of these.

src/clients/inst_disk_proposal.rb
```diff
@@ -0,0 +1,85 @@
...
+
+require "yast"
+require "y2storage"
+
+Yast.import "UI"
+Yast.import "Wizard"
+Yast.import "HTML"
+
+include Yast::I18n
+include Yast::Logger
```
mvidner: Including modules into the global namespace should be avoided, please move it into the class.

src/clients/inst_disk_proposal.rb
```diff
@@ -0,0 +1,85 @@
...
+      textdomain "storage"
+
+      begin
+        proposal = Y2Storage::Proposal.new(settings: settings)
+        proposal.propose
+        actiongraph = proposal.devices.actiongraph
+        summary = summary(actiongraph)
+      rescue Y2Storage::Proposal::Error
+        # error message
+        summary = HTML.Para(HTML.Colorize("No proposal possible.", "red"))
```
mvidner: Log the rescued exception?
Mark the string for translation?

src/clients/inst_disk_proposal.rb
mvidner: I noticed that this should be fixed in many files across yast-storage-ng, it is not specific to this PR.

mvidner:
LGTM

src/clients/inst_disk_proposal.rb
aschnell: Moved includes to class.

src/clients/inst_disk_proposal.rb
aschnell: Added both.

ancorgs:
I know the pull request is already merged, but still...

This does not follow the recently agreed code organization.

 - Classes defining the clients/dialogs behavior should live in `src/lib/y2storage`.
 - In most cases, files directly placed in `src/clients` (which are run directly as WFM "scripts") should only contain a call to the appropriate client/dialog class defined in `lib`. Having the definition of the class and its execution in the same file harms testability[1].

On the other hand, this looks like a good opportunity to use the InstallationDialog class[2].

[1] That's why the test coverage remained the same even after adding a whole new class without unit test. We exclude `src/clients` from test coverage because it's impossible to write a good unit test for such files. That's why they should be almost empty, containing just that call to the class placed in `src/lib`, which is accounted for test coverage.
[2] http://www.rubydoc.info/github/yast/yast-yast2/UI/InstallationDialog

src/clients/inst_disk_proposal.rb
```diff
@@ -0,0 +1,86 @@
...
+    def main
+      textdomain "storage"
+
+      begin
+        proposal = Y2Storage::Proposal.new(settings: settings)
+        proposal.propose
+        actiongraph = proposal.devices.actiongraph
+        summary = summary(actiongraph)
+      rescue Y2Storage::Proposal::Error
+        log.error("generating proposal failed")
```
mvidner: Logging something is better than logging nothing, but I meant to log the exact exception rescued, otherwise we will be left in the dark, won't we?

mvidner:
One more thing: this is named a "proposal" but it is different from the many other "proposal" classes with this http://www.rubydoc.info/github/yast/yast-yast2/Installation/ProposalClient interface. 
```rb
# This proposal gets an installation step all for itself
# so it does not conform to the {Installation::ProposalClient} interface.
class InstDiskProposalClient
```

PR #91, opened 2016-09-21
Move inst_disk_proposal to y2storage and add unit test

ancorgs:
And use InstallationDialog instead of a completely custom client.

### DIFF
--- a/src/lib/y2storage/dialogs/inst_disk_proposal.rb
+++ b/src/lib/y2storage/dialogs/inst_disk_proposal.rb
@@ -40,12 +40,19 @@ module Y2Storage
 
     protected
 
+      # For the time being, it always returns the same proposal settings.
+      # To be connected to control.xml and the UI in the future
+      #
+      # @return [Y2Storage::ProposalSettings]
       def settings
         settings = Y2Storage::ProposalSettings.new
         settings.use_separate_home = true
-        return settings
+        settings
       end
 
+      # HTML-formatted text to display in the dialog
+      #
+      # @return [String]
       def summary
         formatted_actiongraph
       rescue Y2Storage::Proposal::Error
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/dialogs/inst_disk_proposal.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      def initialize
+        super
+        textdomain "storage"
+      end
+
+    protected
+
+      def settings
+        settings = Y2Storage::ProposalSettings.new
```
imobachgs: Just curiosity: are settings always the same?

src/lib/y2storage/dialogs/inst_disk_proposal.rb
```diff
@@ -0,0 +1,87 @@
...
+        # error message
+        Yast::HTML.Para(Yast::HTML.Colorize(_("No proposal possible."), "red"))
+      end
+
+      def formatted_actiongraph
+        proposal = Y2Storage::Proposal.new(settings: settings)
+        proposal.propose
+        actiongraph = proposal.devices.actiongraph
+        texts = actiongraph.commit_actions_as_strings.to_a
+        return Yast::HTML.Para(Yast::HTML.List(texts))
```
imobachgs: I woud remove the `return` statement (it's not the Ruby way). Or I would use it everywhere. Just be consistent.

src/lib/y2storage/dialogs/inst_disk_proposal.rb
ancorgs: For the time being, yes. https://trello.com/c/9Uj9DmG6/384-3-storageng-installation-storage-proposal

We will process the "connection" of settings with control.xml and with the UI in an upcoming PBI.

src/lib/y2storage/dialogs/inst_disk_proposal.rb
ancorgs: I just wanted to avoid a religious war by removing the `return` calls that were already there. But if I have to choose between removing them or adding more, I have no doubt. :wink:

imobachgs:
LGTM, thanks.

PR #92, opened 2016-09-26
More reusable DiskAnalyzer

ancorgs:
Make DiskAnalyzer reusable in other contexts beyond partitioning proposal, since it includes quite some valuable logic (like checking for Windows partitions).

### Why

Here (https://github.com/yast/yast-country/pull/89#discussion_r80223751) we are discussing the exact code to check for Windows partitions in yast2-country. Checking for Windows implies:
 - checking for the architecture (something that DiskAnalyzer does but our draft in yast2-country forgets)
 - checking for the partition ID (DiskAnalyzer checks for 4 ids, our draft only NTFS)
 - checking for a primary partition (only done by yast2-country)
 - checking for the content of the filesystem

Obviously we should not have two pieces of code  to check that, and `DiskAnalyzer` looks like the perfect candidate to host the one and only code. We just need to generalize it a little bit so it's useful in more contexts.

### What

- Allow operation on all disks (while retaining the "only installation candidates" mode, useful for proposal).
- Don't calculate everything at a time, do it on demand.
- Remove proposal-specific logic, like skipping Windows check if Linux is found... which was actually an error.[1]

## Bonus track

DiskAnalyzer#windows_partitions rewritten to rely on new libstorage windows check (see the above-mentioned discussion in yast2-country).

[1]  We slightly changed the proposal logic (we found cases in which resizing Windows still made sense) and DiskAnalyzer was left behind.

### DIFF
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -27,6 +27,8 @@ require "storage"
 require "y2storage/disk_size"
 require "y2storage/refinements"
 
+Yast.import "Arch"
+
 module Y2Storage
   #
   # Class to analyze the disks (the storage setup) of the existing system:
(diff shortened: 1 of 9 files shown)


ancorgs:
Don't worry about code coverage. I just wanted to make sure we agree on this (see also the linked discussion). After the code unification (using the new libstorage methods), coverage will increase.

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -322,27 +289,15 @@ def find_mbr_gap
...
-    # found on any of the candidate disks (i.e., on any disk except the
-    # installation medium). This can be called independently from the
-    # outside, though.
-    #
-    # @return [Array <::Storage::Partition>]
-    #
+    # @see #windows_partitions
     def find_windows_partitions
-      return [] unless Arch.x86_64 || Arch.i386
+      return {} unless windows_architecture?
```
shundhammer: The old code returned [], the new one returns {}. Is that intentional?

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -355,31 +310,38 @@ def find_windows_partitions
...
     # possibly be resized.
     #
-    # @param partition [string| device name of the partition to check
+    # @param partition [::Storage::Partition] partition to check
     #
     # @return [Boolean] 'true' if it is a Windows partition, 'false' if not.
     #
     def windows_partition?(partition)
-      return false unless WINDOWS_PARTITION_IDS.include?(partition.id)
-      return false unless Arch.x86_64 || Arch.i386
```
shundhammer: Sure this shortcut is no longer desired for non-PC architectures? How expensive are the subsequent checks? The idea of checking for PC arch here was to avoid them when they clearly don't make any sense (ARM, S/390).

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -521,34 +483,24 @@ def remove_installation_disks(disks)
     # Find partitions from any of the candidate disks that have a given (set
     # of) partition id(s).
     #
-    # The result is a Hash in which each key is the name of a candidate disk
+    # The result is a Hash in which each key is the name of a disk
     # and the value is an Array of ::Storage::Partition objects
     # representing the matching partitions in that disk.
     #
     # @param ids [::Storage::ID, Array<::Storage::ID>]
```
shundhammer: Please document parameter log_label.

shundhammer:
Besides the minor things I commented, LGTM.

Since there is a testsuite for this, I assume the code provides the same functionality as before; since the diff is quite large, I couldn't make sure of that just by looking at it. For that reason, please let somebody else have an additional look at it.

src/lib/y2storage/disk_analyzer.rb
ancorgs: Yes. The old one was a bug. The method returns `Hash{String => Array<::Storage::Partition>}`

src/lib/y2storage/disk_analyzer.rb
ancorgs: Good point. I overlooked that.

ancorgs:
The shorcuts are still desired, but they were redundant. The
#windows_partition? method is protected and currently only called for
partitions in which both checks has already being performed.

mvidner:
So now `DiskAnalyzer` has a mandatory constructor argument. Then we should adapt all users. It seems one place has been left behind:
```
$ grep -rn DiskAnalyzer src/ test/
...
src/examples/boot_req_checker_demo.rb:51:disk_analyzer = Y2Storage::DiskAnalyzer.new
```

imobachgs:
Apart from the issue found by Martin, it looks good.

ancorgs:
Pull request updated with several tests and small fixes to DiskAnalyzer.

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -171,52 +103,108 @@ def device_by_name(name)
...
-    attr_reader :devicegraph
+    # Partitions that can be used as EFI system partitions.
+    #
+    # Checks for the partition id to return all potential partitions.
+    # Checking for content_info.efi? would only detect partitions that are
+    # going to be effectively used.
+    #
+    # @see #scope
+    #
+    # @return [Hash{String => Array<::Storage::Partition>}] @see #partitions_by_id
```
mvidner: `@see #partitions_with_id`

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -70,98 +72,28 @@ class DiskAnalyzer
...
-
-      # Maximum number of disks to check. This might be important on
-      # architectures that tend to have very many disks (s390).
-      @disk_check_limit = DEFAULT_DISK_CHECK_LIMIT
-    end
-
-    # Analyze disks and partitions. Make sure to call this before querying
-    # any member variables.
+    # @return [Fixnum] Maximum number of disks to check with "expensive"
+    # operations.
```
mvidner: Indent "operations" by 2 spaces, otherwise yardoc misparses it and considers it independent from the "@return"

mvidner: I think I am going to fix things like this by myself, it will be easier.

mvidner:
@ancorgs please take my Yardoc fixes: https://github.com/yast/yast-storage-ng/commit/0a32c881ed322e522b8cd04a9ad14a6b13f1a474
With those included it LGTM!

PR #94, opened 2016-09-30
inst_disk_proposal: copy proposed devicegraph to staging if user accepts

ancorgs:


src/lib/y2storage/dialogs/inst_disk_proposal.rb
```diff
@@ -36,6 +36,20 @@ class InstDiskProposal < ::UI::InstallationDialog
       def initialize
         super
         textdomain "storage"
+        propose
+      end
+
+      def next_handler
+        if @devicegraph
```
imobachgs: np: Maybe you should define an accessor (writer/reader or whatever). And don't forget to set the private.

src/lib/y2storage/dialogs/inst_disk_proposal.rb
```diff
@@ -36,6 +36,20 @@ class InstDiskProposal < ::UI::InstallationDialog
       def initialize
         super
         textdomain "storage"
+        propose
```
imobachgs: np: I would rename it to `propose!` as the only effect of this method is to change the object's internal state.

imobachgs:
LGTM, thanks!

PR #96, opened 2016-10-04
document testing iSCSI and iBFT setups with qemu

wfeldt:


teclator:
I maybe would prefer to move it to [yast2-iscsi-lio-server](http://github.com/yast/yast-iscsi-lio-server) as it is iscsi related. We could discuss also if it is better there or in [yast2-iscsi-client](https://github.com/yast/yast-iscsi-client).

In general the doc LGTM just decide where to put it. We could maintain it here and I could link to this doc from the one I wrote, but maybe is better to have it in just one place.

wfeldt:
moved to iscsi-lio-server: https://github.com/yast/yast-iscsi-lio-server/pull/53

PR #101, opened 2016-10-17
libstorage-ng does not longer use id linux for extended partitions

ancorgs:
Adapt the tests to the fix introduced in https://github.com/openSUSE/libstorage-ng/pull/160


aschnell:
LGTM - assuming the failed tests vanish with new libstorage-ng.

PR #102, opened 2016-10-17
SpaceMaker: workaround libstorage restriction

ancorgs:
Without this, our image in OBS cannot perform the installation if there are previous partitions.


shundhammer:
LGTM

PR #107, opened 2016-10-27
Devices list for LVM classes

ancorgs:
Just to have the same query interface that we have for disks, partitions, etc. That is, for example
```ruby
vgs = devicegraph.volume_groups.with { |vg| vg.vg_name.starts_with?("system") }
vgs.lvs.with(lv_name: "swap")
```

### DIFF
--- a/src/lib/y2storage/devices_lists/lvm_vgs_list.rb
+++ b/src/lib/y2storage/devices_lists/lvm_vgs_list.rb
@@ -1,0 +1,57 @@
+#
+# encoding: utf-8
+
+# Copyright (c) [2016] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "storage"
+require "y2storage/devices_lists/base"
+require "y2storage/devices_lists/lvm_pvs_list"
+require "y2storage/devices_lists/lvm_lvs_list"
+
+module Y2Storage
+  module DevicesLists
+    # List of LVM volume groups from a devicegraph
+    class LvmVgsList < Base
+      list_of ::Storage::LvmVg
+
+      # Physical volumes included in any of the volume groups
+      #
+      # @return [LvmPvsList]
+      def lvm_pvs
+        pvs_list = list.reduce([]) { |sum, vg| sum + vg.lvm_pvs.to_a }
+        LvmPvsList.new(devicegraph, list: pvs_list)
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/devices_lists/lvm_vgs_list.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      alias_method :pvs, :lvm_pvs
+      alias_method :physical_volumes, :lvm_pvs
+
+      # Logical volumes included in any of the volume groups
+      #
+      # @return [LvmLvList]
+      def lvm_lvs
+        lvs_list = list.reduce([]) { |sum, vg| sum + vg.lvm_lvs.to_a }
+        LvmLvsList.new(devicegraph, list: lvs_list)
```
jreidinger: looks almost identical as above, only difference is method name and list name..but as long as it is just two methods, it is fine for me.

src/lib/y2storage/refinements/devicegraph_lists.rb
```diff
@@ -34,14 +34,25 @@ module DevicegraphLists
...
             klass.new(self)
           end
         end
+
+        alias_method :vgs, :lvm_vgs
+        alias_method :volume_groups, :lvm_vgs
+        alias_method :pvs, :lvm_pvs
+        alias_method :physical_volumes, :lvm_pvs
+        alias_method :lvs, :lvm_lvs
+        alias_method :logical_volumes, :lvm_lvs
```
jreidinger: to be honest I am not such fun of having such aliases from beginning as it makes reading of code harder, because it allows writer to write any variant and for reader it is harder to remember that it is all same methods. Remember e.g. `map` and `collect` or `reduce` and `inject` or even more famous like `File.exist?` and `File.exists?`. So I suggest to pick the best name from beginning and stick to it and use alias_method only if needed for backward compatibility.

test/devices_list_test.rb
```diff
@@ -234,4 +235,56 @@ def partition_name
...
     end
   end
+
+  describe Y2Storage::DevicesLists::LvmVgsList do
+    let(:scenario) { "lvm-two-vgs" }
+    let(:vgs) { fake_devicegraph.vgs }
+
+    it "contains all volume groups by default" do
+      expect(vgs.size).to eq 2
+      expect(described_class.new(fake_devicegraph).size).to eq 2
```
jreidinger: To be honest I see this `described_class.new(fake_devicegraph)` too many times in test. What about defining `subject { described_class.new(fake_devicegraph) }` ?

src/lib/y2storage/refinements/devicegraph_lists.rb
jreidinger: OK, it is matter of taste. But I still believe that if you see such lines, it is not obvious what it do unless you are very familiar with API:

```
vgs = devicegraph.volume_groups.with { |vg| vg.vg_name.starts_with?("system") }
vgs.lvs.with(lv_name: "swap")
vgs.lvm_lvs.with(lv_name: "root")
vgs.logical_volumes.with(lv_name: "boot")
```

test/devices_list_test.rb
```diff
@@ -151,7 +153,7 @@ def partition_name
 
     it "contains all disks by default" do
       expect(disks.size).to eq 3
-      expect(described_class.new(fake_devicegraph).size).to eq 3
+      expect(.size).to eq 3
```
jreidinger: looks like bug

ancorgs: Fixed

jreidinger:
LGTM, aliases are not blockers for me

PR #108, opened 2016-10-27
- mount special filesystems in target during installation

aschnell:
For https://trello.com/c/NQ524dss/459-8-storageng-bootloader-proposal-efi.

### DIFF
--- a/src/lib/y2storage/clients/inst_prepdisk.rb
+++ b/src/lib/y2storage/clients/inst_prepdisk.rb
@@ -2,7 +2,7 @@
 #
 # encoding: utf-8
 
-# Copyright (c) [2015] SUSE LLC
+# Copyright (c) [2015-2016] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -26,6 +26,8 @@ require "y2storage"
 
 Yast.import "SlideShow"
 Yast.import "Installation"
+Yast.import "FileUtils"
+Yast.import "String"
 
 module Y2Storage
   module Clients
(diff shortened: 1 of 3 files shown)


shundhammer:
LGTM

src/lib/y2storage/clients/inst_prepdisk.rb
```diff
@@ -51,6 +54,24 @@ def commit
...
+        mount_in_target("/dev", "devtmpfs", "-t devtmpfs")
+        mount_in_target("/proc", "proc", "-t proc")
+        mount_in_target("/sys", "sysfs", "-t sysfs")
+      end
+
+      def mount_in_target(path, device, options)
+        storage = Y2Storage::StorageManager.instance
+        target_path = storage.prepend_rootprefix(path)
+
+        if !Yast::FileUtils.Exists(target_path)
```
ancorgs: Is not enough with Ruby's `FileUtils`? At some point we would like to see `Yast::Fileutils` die.

src/lib/y2storage/clients/inst_prepdisk.rb
```diff
@@ -26,13 +26,16 @@
 
 Yast.import "SlideShow"
 Yast.import "Installation"
+Yast.import "FileUtils"
+Yast.import "String"
```
ancorgs: Why is `Yast::String` needed?

src/lib/y2storage/clients/inst_prepdisk.rb
aschnell: No idea, I didn't try it. I asked on IRC what is the best way now but didn't get any reply.

src/lib/y2storage/clients/inst_prepdisk.rb
aschnell: Indeed, it's not needed.

PR #110, opened 2016-10-27
Better management of LVM partitions in SpaceMaker

ancorgs:
When deleting a partition that holds a PV, it now also deletes all the other partitions in the same VG.

Rationale: when deleting a partition that holds a PV of a given VG, we are effectively killing the whole VG (not only implicitly, but even explicitly thanks to libstorage). It makes no sense to leave the other PVs alive. So let's reclaim all the space.

### DIFF
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -543,7 +543,12 @@ module Y2Storage
       lvm_pv = ::Storage::LvmPv.all(devicegraph).to_a.detect { |pv| pv.blk_device.name == name }
       return nil unless lvm_pv
 
-      lvm_pv.lvm_vg.vg_name
+      begin
+        lvm_pv.lvm_vg.vg_name
+      rescue ::Storage::WrongNumberOfChildren
+        # Unassigned PV
+        nil
+      end
     end
   end
 end
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -26,6 +26,7 @@ require "fileutils"
 require "storage"
 require "y2storage/disk_size"
 require "y2storage/refinements"
+require "y2storage/existing_filesystem"
 
 Yast.import "Arch"
 
@@ -420,7 +421,8 @@ module Y2Storage
     #
     def installation_volume?(vol_name)
       log.info("Checking if #{vol_name} is an installation volume")
-      is_inst = mount_and_check(vol_name) { |mp| installation_volume_check(mp) }
+      fs = ExistingFilesystem.new(vol_name)
+      is_inst = fs.mount_and_check { |mp| installation_volume_check(mp) }
       log.info("#{vol_name} is installation volume") if is_inst
       is_inst
     end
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/disk_analyzer.rb
```diff
@@ -507,5 +533,17 @@ def partitions_with_id(ids, log_label)
...
+    # Name of the LVM volume group to which the partition belongs
+    #
+    # @param partition [::Storage::Partition]
+    # @return [String] nil if the partition does not hold a physical volume
+    def vg_for(partition)
+      name = partition.name
+      lvm_pv = ::Storage::LvmPv.all(devicegraph).to_a.detect { |pv| pv.blk_device.name == name }
+      return nil unless lvm_pv
+
+      lvm_pv.lvm_vg.vg_name
```
shundhammer: Might `lvm_pv.lvm_vg` be nil in weird corner cases?
What if there is a PV that is not assigned to a VG yet?

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -301,7 +301,16 @@ def delete_partitions!(volumes, type, keep, disk)
           end
           part = find_partition(part_name)
           next unless part
-          delete_partition(part)
+
+          if lvm_pv?(part)
+            # Strictly speaking, this could lead to deletion of a partition
+            # included in the keep array. In practice it doesn't matter because
+            # PVs are never marked to be reused as a PlannedVolume.
```
shundhammer: Not even if the PV is not part of any VG?

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -326,6 +335,26 @@ def delete_partition(partition)
         end
       end
 
+      # Deletes the given partition and all other partitions in the candidate
+      # disks that are part of the same LVM volume group
+      #
+      # Rational: when deleting a partition that holds a PV of a given VG, we
```
shundhammer: Typo: Rationale

shundhammer:
Besides the typo and the two (possible patholocial) corner cases, LGTM.

src/lib/y2storage/disk_analyzer.rb
ancorgs: I though I had that situation into account, but looking twice I didn't. I only considered the situation of a LVM partition without PV, but that's different from unassigned PVs. I will improve it.

src/lib/y2storage/proposal/space_maker.rb
ancorgs: Not even

src/lib/y2storage/disk_analyzer.rb
ancorgs: @aschnell I was trying to improve this but in a system with unassigned PVs `::Storage::Storage.new(...)` raise a `Storage::Exception`

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -507,5 +533,22 @@ def partitions_with_id(ids, log_label)
...
+    # @param partition [::Storage::Partition]
+    # @return [String] nil if the partition does not hold a physical volume
+    def vg_for(partition)
+      name = partition.name
+      lvm_pv = ::Storage::LvmPv.all(devicegraph).to_a.detect { |pv| pv.blk_device.name == name }
+      return nil unless lvm_pv
+
+      begin
+        lvm_pv.lvm_vg.vg_name
+      rescue ::Storage::WrongNumberOfChildren
```
ancorgs: Due to a crash in libstorage-ng, I cannot verify this would be the correct way of checking whether a PV is unassigned. @aschnell, is correct?

aschnell: Yes, that looks correct.

The exception in libstorage-ng is due to the parser of the lvm tools not being capable of the situation, see TODO at https://github.com/openSUSE/libstorage-ng/blob/master/storage/SystemInfo/CmdLvm.cc#L37.

ancorgs:
Pull request updated to consider the scenario of unassigned PV. Also extracted some responsibility to a new class in order to slim DiskAnalyzer, which was too long according to Rubocop.

shundhammer:
LGTM

PR #111, opened 2016-11-02
LVM-based proposal

ancorgs:
https://trello.com/c/WbcaGFAo

Better review commit by commit.

Takes into account LVM PE rounding and metadata overhead

It does not reuse swap partitions within the volume group. That's intentional. I plan to do it in a separate pull request


test/disk_size_test.rb
```diff
@@ -277,4 +281,67 @@
...
+
+    it "returns the same value if any of the operands is unlimited" do
+      expect(unlimited.ceil(8.GiB)).to eq unlimited
+      expect(8.GiB.ceil(unlimited)).to eq 8.GiB
+      expect(unlimited.ceil(zero)).to eq unlimited
+      expect(unlimited.ceil(unlimited)).to eq unlimited
+    end
+
+    it "returns the same value when rounding to 1 byte" do
+      expect(4.KiB.ceil(one_byte)).to eq 4.KiB
```
mvidner: This could accidentally succeed if you rounded to something else. Better test with 31337 bytes (which is prime)

src/lib/y2storage/planned_volume.rb
```diff
@@ -148,5 +148,30 @@ def to_s
...
+      return nil unless filesystem_type
+      filesystem = partition.create_filesystem(filesystem_type)
+      filesystem.add_mountpoint(mount_point) if mount_point && !mount_point.empty?
+      filesystem.label = label if label
+      filesystem.uuid = uuid if uuid
+      filesystem
+    end
+
+    def ==(other)
+      other.class == self.class && other.internal_state == internal_state
```
mvidner: NP: `other.is_a?(self.class)` will work even for subclasses

ancorgs:
> This is much denser than I expected. I'm only able to comment on the trivial pieces, sorry.

Still useful. It's important to follow the advise of reviewing in a per-commit basis. Although still in that case it's important to know how the classes that are being modified work an interact. That takes quite some time.

src/lib/y2storage/planned_volumes_list.rb
```diff
@@ -66,16 +67,32 @@ def deep_dup
...
 
     # Total sum of all current max sizes of volumes
     #
+    # If the optional argument "rounding" is used, the size of every volume will
+    # be rounded up. # @see DiskSize#ceil
+    #
+    # @param rounding [DiskSize, nil]
     # @return [DiskSize]
-    def max_disk_size
-      @volumes.reduce(DiskSize.zero) { |sum, vol| sum + vol.max_disk_size }
```
mvidner: This can be kept simple `{|sum, vol| sum + vol.max_disk_size.ceil(rounding) }` if you use a default `rounding` that returns the input. That is, `def ...(rounding: 1)` should do.

src/lib/y2storage/disk_size.rb
```diff
@@ -221,6 +221,32 @@ def <=>(other)
...
+      bytes = modulo.zero? ? @size : @size + unit_bytes - modulo
+      DiskSize.new(bytes)
+    end
+
+    # Result of rounding down the size to the previous value that is divisible
+    # by a given size. Returns the same value if it's already divisible.
+    #
+    # @param unit_size [DiskSize]
+    # @return [DiskSize]
+    def floor(unit_size)
```
wfeldt: Would it make sense to allow unit_size to be nil here? There's a lot of if(rounding) in the code below.

src/lib/y2storage/disk_size.rb
```diff
@@ -221,6 +221,32 @@ def <=>(other)
       raise TypeError, "Unexpected #{other.class}; expected DiskSize"
     end
 
+    # Result of rounding up the size to the next value that is divisible by
+    # a given size. Returns the same value if it's already divisible.
+    #
+    # @param unit_size [DiskSize]
+    # @return [DiskSize]
+    def ceil(unit_size)
```
wfeldt: Redefine using floor()? Like
```
def ceil(unit_size)
  new_size = floor(unit_size)
  new_size += unit_size if new_size != self
  new_size
end

File: src/lib/y2storage/disk_size.rb (hunk 4 above)
Comment (ancorgs) [inline src/lib/y2storage/disk_size.rb]: Some of those "if" would remain even if we accept nil here. In total, it would mean removing three `if rounding` at the cost of adding two (one in `floor` and one in `ceil`). Not sure whether is worth it right now.

On the other hand, I think I prefer to keep all the `if rounding` in one class (`PlannedVolumesList`).

**Comment by ancorgs** [comment]
BTW, I have found another rounding problem related to extended partitions. Since it's not strictly related to LVM (we simply had overlooked it so far), I will fix it in a separate pull request on top of this.

File: src/lib/y2storage/proposal/lvm_helper.rb [hunk 6]
```diff
@@ -0,0 +1,320 @@
...
+      # @param part_names [Array<String>] device names of the partitions
+      # @param devicegraph [Storage::Devicegraph] to fetch the partitions
+      def assign_physical_volumes!(volume_group, part_names, devicegraph)
+        devicegraph.partitions.with(name: part_names).each do |partition|
+          volume_group.add_lvm_pv(partition)
+        end
+      end
+
+      # Makes sure the given volume group has enough free extends to allocate
+      # all the planned volumes, by deleting the existing volume groups.
```
wfeldt: deleting logical volumes?

src/lib/y2storage/proposal/lvm_helper.rb
```diff
@@ -0,0 +1,320 @@
...
+      #
+      # This method modifies the volume group received as first argument.
+      #
+      # FIXME: the current implementation does not guarantee than the freed
+      # space is the minimum valid one.
```
wfeldt: I'm wondering if this strategy is necessary at all or we'd be better of just deleting all LVs when reusing an existing VG. At least in the case when we don't fit into any free remaining space.

Imagine for example the VG was created by us and so probably has root+swap/boot+home. This strategy will free home (since root will be too small). I fear this might more or less not do what the user expects. At lease I'm having a hard time finding  a scenario where this strategy gives good results. :-)

src/lib/y2storage/proposal/lvm_helper.rb
ancorgs: oops

src/lib/y2storage/proposal/lvm_helper.rb
ancorgs: I was just trying to be consistent to what we do with partitions. We don't delete all linux partitions in the disk we are going to use. We start low and delete partitions until we have space to place ours, not destructing more than strictly needed. I wanted to follow the same approach with LVs.

Of course, I'm fine with a more destructive approach if everybody agrees (less code to maintain) :-).

test/disk_size_test.rb
ancorgs: done

src/lib/y2storage/proposal/lvm_helper.rb
ancorgs: Since I don't know which strategy is better, I raised the question in yast-devel.

wfeldt:
Ok, I think I've seen it all. :-)

lgtm; even if there are minor open issues, I'd merge this and do any improvements in a new pr

PR #113, opened 2016-11-07
Backported fate#321737: Subvolumes configurable in control.xml

shundhammer:
This is just a rough backport to ensure that feature isn't completely forgotten.

The glue code to put that into is still missing, as well as subvolumes in libstorage-ng.


ancorgs:
I assume the code was originally reviewed as part of the previous pull request, so no need to re-verify. The status is good enough for the purpose of not forgetting.

LGTM.

PR #114, opened 2016-11-15
Proposal: extended partitions overhead (EBR)

ancorgs:
I found out that so far we have been creating partitions that were 2 MiB smaller than planned in some situations.

With this change, when calculating a proposal we now take into account the overhead produced by creating an extended partition and count with it when resizing Windows.

I also removed some "unlimited" from the YAML of the tests, to ensure we don't overlook that any more.

### DIFF
--- a/src/lib/y2storage/proposal/space_distribution.rb
+++ b/src/lib/y2storage/proposal/space_distribution.rb
@@ -36,6 +36,12 @@ module Y2Storage
 
       include Yast::Logger
 
+      # Space that will be consumed by the EBR if we create a new extended
+      # partition. Value based on libstorage-ng behavior
+      # See https://en.wikipedia.org/wiki/Extended_boot_record
+      OVERHEAD_OF_NEW_EXTENDED = DiskSize.MiB(2)
+      private_constant :OVERHEAD_OF_NEW_EXTENDED
+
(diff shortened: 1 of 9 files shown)


src/lib/y2storage/proposal/space_distribution.rb
```diff
@@ -36,6 +36,12 @@ class SpaceDistribution
 
       include Yast::Logger
 
+      # Space that will be consumed by the EBR if we create a new extended
+      # partition. Value based on libstorage-ng behavior
+      # See https://en.wikipedia.org/wiki/Extended_boot_record
+      OVERHEAD_OF_NEW_EXTENDED = DiskSize.MiB(2)
```
wfeldt: Shouldn't this number come out of libstorage, like based on the current alignment policy?

ancorgs: Yes, it should. I wanted to open the pull request to ask @aschnell... but I forgot to do it :-)

ancorgs: @aschnell What we need there is a (pessimistic) estimation.

wfeldt: and isn't the overhead just 1 MiB?

PR #115, opened 2016-11-17
Proposal: logical partitions overhead (EBR)

ancorgs:
replaces #114 

I found out that so far we have been creating partitions that were some MiBs smaller than planned in some situations. 

With these changes, when calculating a proposal we now take into account the overhead produced by creating logical partitions and count with it when resizing windows and when distributing the volumes into the spaces.

I also removed some "unlimited" from the YAML of the tests, to ensure we don't overlook the problem any more.

As bonus, SpaceDistribution now decides in advance which partitions should be logical, as a result, PartitionCreator does not longer creates extended partitions with just one logical inside.

### DIFF
--- a/src/lib/y2storage/proposal/assigned_space.rb
+++ b/src/lib/y2storage/proposal/assigned_space.rb
@@ -43,8 +43,9 @@ module Y2Storage
       def_delegators :@disk_space, :disk_name, :disk_size, :slot, :disk
 
       def initialize(disk_space, volumes)
-        @disk_space = disk_space
-        @volumes    = volumes
+        @disk_space  = disk_space
+        @volumes     = volumes
+        @num_logical = 0
       end
 
       # Checks if the volumes really fit into the assigned space
@@ -76,12 +77,15 @@ module Y2Storage
       #
       # Substracts from the total the space that will be used by new data
       # structures, like the EBRs of the planned logical partitions
+      # See https://en.wikipedia.org/wiki/Extended_boot_record
       #
       # @return [DiskSize]
       def usable_size
+        return disk_space.disk_size if num_logical.zero?
+
         logical = num_logical
-        # If the extended partition already exists, the overhead of our first
-        # logical partition is already counted
+        # If this space is inside an already existing extended partition,
+        # libstorage has already substracted the the overhead of the first EBR.
         logical -= 1 if partition_type == :logical
         disk_space.disk_size - overhead_of_logical * logical
       end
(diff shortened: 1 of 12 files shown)


src/lib/y2storage/proposal/assigned_space.rb
```diff
@@ -64,6 +65,43 @@ def unused
...
+      # Substracts from the total the space that will be used by new data
+      # structures, like the EBRs of the planned logical partitions
+      #
+      # @return [DiskSize]
+      def usable_size
+        logical = num_logical
+        # If the extended partition already exists, the overhead of our first
+        # logical partition is already counted
+        logical -= 1 if partition_type == :logical
+        disk_space.disk_size - overhead_of_logical * logical
```
shundhammer: So, if the extended partition already exists, the overhead is not only not subtracted again, but added (- overhead_of_logical * -1). Is that really intended?

src/lib/y2storage/proposal/partition_creator.rb
```diff
@@ -53,10 +53,11 @@ def create_partitions(distribution)
         self.devicegraph = original_graph.duplicate
 
         distribution.spaces.each do |space|
-          type = space.partition_type
           vols = space.volumes
           disk_space = space.disk_space
-          process_space(vols, disk_space, type)
+          num_logical = space.num_logical
+          usable_size = space.usable_size
+          process_space(vols, disk_space, usable_size, num_logical)
```
shundhammer: Why not simply pass space.usable_size and space.num_logical here and be two lines more concise?
The additional variables don't contribute very much to readability IMHO, and they don't seem to be used anywhere else.

shundhammer:
It looks quite impressive, but I couldn't claim that I understand the majority of it all.

I fear that we'll need some higher-level explanation of the approach and the algorithm; in the optimum case with one or more pictures.

From my limited understanding of this, I'd give it an LGTM after the small comments are answered - with the caveat that I don't quite get the big picture.

src/lib/y2storage/proposal/assigned_space.rb
ancorgs: Actually it never happens because, by definition, the num_logical is always > 0 if the partition is :logical. Ahem... obviously, I need to do that more obvious in the code/comment. :smile:

src/lib/y2storage/proposal/partition_creator.rb
ancorgs: ok

shundhammer:
LGTM now (with the same caveat about not understanding the big picture).

PR #117, opened 2016-11-17
Proposal: improvement in SpaceDistribution

ancorgs:
Now it checks that things still make sense after calculating the number of logical partitions in each assigned space.


shundhammer:
LGTM

PR #119, opened 2016-11-17
More cases (unusual MBR gap sizes) in proposal_test

ancorgs:
Proves that the proposal creates boot partitions when needed (not needed nowadays in most cases)


wfeldt:
lgtm

PR #120, opened 2016-11-22
Improved API for other modules

ancorgs:
This improves the current query API to allow yast2-bootloader to get all the requested information for legacy boot... and beyond!

### DIFF
--- a/src/lib/y2storage/devices_lists/base.rb
+++ b/src/lib/y2storage/devices_lists/base.rb
@@ -76,10 +76,17 @@ module Y2Storage
         self.class.new(devicegraph, @list.dup)
       end
 
+      def +(other)
+        if self.class != other.class
+          raise TypeError, "#{other.class} is not a #{self.class}"
+        end
+        self.class.new(devicegraph, list: list + other.to_a)
+      end
+
     protected
 
       attr_reader :devicegraph
-      attr_accessor :list
+      attr_reader :list
 
       # Default collection of devices
       #
--- a/src/lib/y2storage/devices_lists/base.rb
+++ b/src/lib/y2storage/devices_lists/base.rb
@@ -76,6 +76,13 @@ module Y2Storage
         self.class.new(devicegraph, @list.dup)
       end
 
+      # Returns a new list concatenating both operands.
+      #
+      # Only lists of the same type can be concatenated. Due to the nature of
+      # the objects coming from Storage and the nature of DevicesLists,
+      # heterogeneous lists make no sense.
+      #
+      # @raises TypeError if the types do not match
       def +(other)
         if self.class != other.class
           raise TypeError, "#{other.class} is not a #{self.class}"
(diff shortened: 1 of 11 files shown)


src/lib/y2storage/devices_lists/base.rb
```diff
@@ -76,10 +76,17 @@ def dup
         self.class.new(devicegraph, @list.dup)
       end
 
+      def +(other)
+        if self.class != other.class
```
shundhammer: No support for subclasses?

ancorgs: Intentional. We always want our lists to contain elements of a given Storage class. Classes coming from Storage can't deal with duck typing, so we don't want to mix them.

src/lib/y2storage/devices_lists/lvm_lvs_list.rb
```diff
@@ -28,6 +28,33 @@ module DevicesLists
...
+          end
+        end
+        FilesystemsList.new(devicegraph, list: fs_list.compact)
+      end
+
+      # Volume groups containing the logical volumes
+      #
+      # @return [LvmVgsList]
+      def lvm_vgs
+        vgs = list.map(&:lvm_vg)
```
shundhammer: Can this become 'nil'?

src/lib/y2storage/devices_lists/partitions_list.rb
```diff
@@ -37,14 +37,23 @@ def filesystems
         fs_list = list.map do |partition|
           begin
             partition.filesystem
-          rescue ::Storage::WrongNumberOfChildren
+          rescue Storage::WrongNumberOfChildren
             # No filesystem in the partition
             nil
           end
         end
```
shundhammer: add
    fs_list.compact! 
?

src/lib/y2storage/devices_lists/base.rb
shundhammer: OK

src/lib/y2storage/devices_lists/lvm_lvs_list.rb
shundhammer: On second thought, the overall result can't become nil AFAICS, just the result of individual loop iterations.

shundhammer:
Besides that `fs_list.compact!` that might be a good idea, LGTM.

src/lib/y2storage/devices_lists/lvm_lvs_list.rb
ancorgs: I would say lvs should always be in a vg. If for some weird reason we have a lv floating around, I would expect `unassigned_lv.lvm_vg`to raise an exception, instead of returning nil.

src/lib/y2storage/devices_lists/partitions_list.rb
ancorgs: is already there, in the line below (where the list is created)

shundhammer:
OK, LGTM then.

shundhammer:
LGTM

PR #123, opened 2016-11-24
Some refactoring in SpaceDistribution and friends

ancorgs:
This remove unneeded stuff (like passing the devicegraph), distributes the responsibilities better (the AssignedSpace is responsible for calculating and caching its partition_type) and simplifies code.


ancorgs:
The coverage decreased... because the code decreased and that somehow screwed up the percentages. The exact same scenarios are tested.

shundhammer:
LGTM

PR #124, opened 2016-11-25
Improved LVM proposal

ancorgs:
- It now correctly handles situations when overheads coming from LVM and from logical partitions are combined.
- Better distribution of responsibilities - SpaceDistributionCalculator now knows almost nothing about LVM physical volumes.
- Less hashes floating around, use SpaceDistribution objects.

## For the reviewer

- Review commit by commit. It should be much easier.
- Make sure you have read this before https://github.com/yast/yast-storage-ng/blob/master/doc/proposal.md


src/lib/y2storage/proposal/phys_vol_calculator.rb
```diff
@@ -0,0 +1,213 @@
...
+      # This method explores all the options and finds the best one.
+      #
+      # @param distribution [Proposal::SpaceDistribution] initial distribution
+      #     containing only the non-LVM volumes
+      #
+      # @return [Proposal::SpaceDistribution, nil] nil if it's impossible to
+      #     allocate all the needed physical volumes
+      def add_physical_volumes(distribution)
+        best = nil
+        all_spaces.permutation.each do |sorted_spaces|
```
wfeldt: I'm wondering: does it scale? Can all_spaces provide so many spaces (say on machines with thousands of disks) that `#permutation` explodes?

src/lib/y2storage/proposal/phys_vol_calculator.rb
```diff
@@ -0,0 +1,213 @@
...
+      #
+      # @return [Proposal::SpaceDistribution, nil]
+      def processed(distribution, sorted_spaces)
+        volumes = {}
+        missing_size = lvm_helper.missing_space
+        result = nil
+
+        sorted_spaces.each do |space|
+          available_size = estimated_available_size(space, distribution)
+          next unless available_size > lvm_helper.min_pv_size
```
wfeldt: `>=`  as it seems that `#min_pv_size` includes at least `#extent_size` usable data?
Or even `if ... <  ...`?

wfeldt:
lgtm; just not sure about the permutation thing

src/lib/y2storage/proposal/phys_vol_calculator.rb
ancorgs: I'm not sure either :-) But I expect that to be solved in a upper layer. I mean. We search for all the spaces in `ProposalSettings#candidate_disks`. I hope we will have some pre-processing of the settings to make sure that when calling `DiskAnalyzer`, `PhysVolCalculator` and friends, `ProposalSettings#candidate_disks` does not contain hundreds of disks.

And we can even add some heuristic in the future to limit the options. For example, let's say that for GPT disks we don't need to explore all the possibilities because the distribution will be less tricky. All that can be added afterward.

wfeldt: Ok, but lets remember that more than a dozen free spaces will freeze the code. :-)

wfeldt:
lgtm

PR #126, opened 2016-11-28
More methods for DevicesLists

ancorgs:
I forgot to add these during https://github.com/yast/yast-storage-ng/pull/120 and they are needed for https://github.com/yast/yast-bootloader/pull/384 to really work with LVM.


shundhammer:
I am not sure what those new functions will return when we add devicemapper support, for example for encrypted volumes. We should think about that.

Other than that, LGTM.

PR #128, opened 2016-12-13
Do not set boot flag if not supported

ancorgs:
Part of https://trello.com/c/okn67JXC/465-5-proposalng-be-able-to-make-a-proposal-when-gpt-part-table-exists

It also includes some extra tests to check the "fix the wrong GUID..." task.

Travis will probably fail in a first attempt. This depends on https://github.com/openSUSE/libstorage-ng/pull/176

### DIFF
--- a/src/lib/y2storage/proposal/partition_creator.rb
+++ b/src/lib/y2storage/proposal/partition_creator.rb
@@ -157,7 +157,7 @@ module Y2Storage
         region = new_region_with_size(free_space.slot, vol.disk_size)
         partition = ptable.create_partition(dev_name, region, partition_type)
         partition.id = partition_id
-        partition.boot = !!vol.bootable
+        partition.boot = !!vol.bootable if ptable.partition_boot_flag_supported?
         partition
       end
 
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/proposal/partition_creator.rb
```diff
@@ -157,7 +157,7 @@ def create_partition(vol, partition_id, free_space, primary)
         region = new_region_with_size(free_space.slot, vol.disk_size)
         partition = ptable.create_partition(dev_name, region, partition_type)
         partition.id = partition_id
-        partition.boot = !!vol.bootable
+        partition.boot = !!vol.bootable if ptable.partition_boot_flag_supported?
```
shundhammer: `!!variable` is supposed to recover from `variable.nil?`, right? is it a good idea not to do that in that special case?

shundhammer: My mistake; I had thought it's the same variable on the right side.

Disregard this.

shundhammer:
LGTM

PR #129, opened 2016-12-13
Basic (high level) tests for proposal with GPT

ancorgs:
It includes some refactoring of the tests, review by commit to see that.

### DIFF
--- a/test/proposal_test.rb
+++ b/test/proposal_test.rb
@@ -21,326 +21,74 @@
 # find current contact information at www.suse.com.
 
 require_relative "spec_helper"
-require "storage"
-require "y2storage"
-
-# To avoid repeating this four lines of code all over the file
-RSpec.shared_examples "proposed layout" do
-  using Y2Storage::Refinements::TestDevicegraph
-
-  it "proposes the expected layout" do
-    proposal.propose
-    expect(proposal.devices.to_str).to eq expected.to_str
-  end
-end
+require_relative "support/proposal_examples"
+require_relative "support/proposal_context"
 
 describe Y2Storage::Proposal do
-  describe "#propose" do
-    using Y2Storage::Refinements::TestDevicegraph
+  include_context "proposal"
+
+  describe "#propose (with pre-existing swap partitions)" do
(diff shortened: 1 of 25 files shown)


aschnell:
The rspec files repeat identical blocks multiple times, e.g. ```context "with a separate home" ... context "without separate home" ... ```.

test/proposal_test.rb
```diff
@@ -21,326 +21,74 @@
...
-          let(:separate_home) { false }
-          include_examples "proposed layout"
-        end
-      end
+    it "reuses suitable swap partitions" do
+      proposal.propose
+      sda3 = proposal.devices.partitions.with(name: "/dev/sda3").first
+      expect(sda3).to match_fields(
+        mountpoint: "swap",
+        uuid:       "33333333-3333-3333-3333-33333333",
```
aschnell: I assume you made up that UUID: In general UUIDs follow versions and rules, see https://en.wikipedia.org/wiki/Universally_unique_identifier. The one here might be invalid.

test/support/proposal_context.rb
```diff
@@ -0,0 +1,65 @@
...
+  before do
+    fake_scenario(scenario)
+
+    allow(Y2Storage::DiskAnalyzer).to receive(:new).and_return disk_analyzer
+    allow(disk_analyzer).to receive(:windows_partition?) do |partition|
+      !!(partition.filesystem.label =~ /indows/)
+    end
+
+    allow_any_instance_of(::Storage::Filesystem).to receive(:detect_resize_info)
+      .and_return(resize_info)
```
aschnell: Setting the resize info in the library would increase the usefulness of the tests in that the real API is used.

test/proposal_test.rb
ancorgs: Well, this test is not new, just re-indented. That being said, the idea of the test if to prove than the UUIDS are reused, whatever their are. IMHO, using obviously made-up UUIDs for that purpose makes it more obvious and more readable.

ancorgs:
I will try to reduce repetition of contexts

ancorgs:
Replaced by #130

PR #130, opened 2016-12-14
Added GPT scenarios to proposal_test.rb

ancorgs:
It includes small cleanup to avoid too much repetition in the tests, review by commit to see that.

Replaces #129


imobachgs:
LGTM

PR #131, opened 2016-12-14
Make DiskSize#to_s precise

ancorgs:
The fact that `DiskSize#to_s` returned imprecise numbers is unexpected and can be a problem for log-based debugging. Lines like these were actually logging imprecise information:
```ruby
log.info "This is a size: #{disk_size}"
log.info "Content of the volume: #{volume}"
```

This PR aims to make `to_s` more "programmer friendly" (a.k.a. precise). We still have `to_human_string` for user friendliness (although is not much used so far).

### DIFF
--- a/src/lib/y2storage/disk_size.rb
+++ b/src/lib/y2storage/disk_size.rb
@@ -126,7 +126,7 @@ module Y2Storage
       #   42 GB    (supporting binary units only)
       #
       def parse(str)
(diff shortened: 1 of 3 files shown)


ancorgs:
`#to_human_readable` and `#to_human_readable_ex` were not modified. Just moved down to be private (they are not used externally so far) and renamed to `#human_string_components` and `#string_components`.

PR #133, opened 2016-12-15
Bugfix in PlannedVolumesList#distribute_space

ancorgs:


src/lib/y2storage/planned_volumes_list.rb
```diff
@@ -271,16 +271,19 @@ def extra_space_candidates
...
+    # @param total_size [DiskSize] free space to be distributed among
     #    involved volumes
     # @param total_weight [Float] sum of the weights of all involved volumes
+    # @param assigned_size [DiskSize] space already distributed to other volumes
     # @param rounding [DiskSize] size to round up
     #
     # @return [DiskSize]
-    def volume_extra_size(volume, available_size, total_weight, rounding)
-      extra_size = available_size * (volume.weight / total_weight)
+    def volume_extra_size(volume, total_size, total_weight, assigned_size, rounding)
```
jreidinger: hmm, with this change you exceed 4 parameters per method, which is my inner bell for checking that maybe we need some refactoring?

test/proposal/lvm_helper_test.rb
```diff
@@ -307,7 +307,7 @@
 
         expect(lvs).to contain_exactly(
           an_object_with_fields(lv_name: "one", size: (9.GiB - 4.MiB).to_i),
-          an_object_with_fields(lv_name: "two", size: 9.GiB.to_i),
+          an_object_with_fields(lv_name: "two", size: (9.GiB - 4.MiB).to_i),
```
jreidinger: so previously size was bug? now both "one" and "two" looks same.  So previously there was a bug? To be honest I am bit confused by this and I do not feel confident about change.

src/lib/y2storage/planned_volumes_list.rb
ancorgs: Also mine (in fact, I was expecting rubocop to complain). But since it's just a private method to do same calculations I decided it was ok in this case.

test/proposal/lvm_helper_test.rb
ancorgs: Yes, there was a bug (that was, in fact, the bug I wanted to correct). The sum of the LVs was bigger than the VG size. I have added another commit to make it more obvious in the test. Obviously, better checks in libstorage-ng would have prevented us to introduce the bug. But that's a WIP.

PR #134, opened 2016-12-16
Fix proposal with small partitions

ancorgs:
In GPT + legacy boot the proposal wanted to create very small bios_boot but after creating other bigger partitions, there was no suitable slot for it due to alignment. This prevent such situations.


ancorgs:
@wfeldt I updated the pull request to also remove the intermediate gaps you detected in some situations in the master branch. Now we only have one kind of gap: 1MiB at the end of GPT disks. But that's another story. :-)

ancorgs:
This is actually not a good solution. There is no need to enforce 2MiB so often and there are better ways to deal with the 16.5KiB at the end of GPT partition tables. Working in a better approach.

PR #135, opened 2016-12-19
Manage small partitions and GPT final space

ancorgs:
Correctly manage small partitions (like bios_boot) and the infamous 16.5KiB at the end of GPT partitions

Before this change, in GPT + legacy boot, the proposal wanted to create very small bios_boot but after creating other bigger partitions there was no suitable slot because we didn't round to min_grain. This prevent such situations (tests included).

This also reduce the number of produced gaps.

### DIFF
--- a/test/proposal/partition_creator_test.rb
+++ b/test/proposal/partition_creator_test.rb
@@ -96,6 +96,7 @@ describe Y2Storage::Proposal::PartitionCreator do
           home_vol.weight = 2
           swap_vol.desired = 1.GiB - 1.MiB
           swap_vol.max = 1.GiB - 1.MiB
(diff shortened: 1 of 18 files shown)


test/proposal/partition_creator_test.rb
```diff
@@ -132,6 +132,35 @@
           end
         end
       end
+
+      context "when the space is not divisible by the minimal grain" do
+        # The last 16.5KiB of GPT are not usable, what makes the space not
```
wfeldt: what -> which?

wfeldt:
lgtm, but for two typos

wfeldt:
lgtm

PR #136, opened 2016-12-21
Proposal unit tests for mixed GPT/MBR disks

ancorgs:


test/proposal_test.rb
```diff
@@ -232,5 +232,57 @@
...
+          parts = proposal.devices.partitions
+          fs = parts.filesystems
+          expect(fs.with_mountpoint("/").disks.first.name).to eq "/dev/sdb"
+          expect(fs.with_mountpoint("/home").disks.first.name).to eq "/dev/sdb"
+          expect(fs.with_mountpoint("swap").disks.first.name).to eq "/dev/sdb"
+          expect(parts.with(id: Storage::ID_BIOS_BOOT).disks.first.name).to eq "/dev/sdb"
+        end
+      end
+
+      context "if a disk without free space is chosen for '/'" do
```
jreidinger: I miss here complementary case what happen when there is a free space as I expect it will be more common.

PR #137, opened 2016-12-21
DisksList#with_name_or_partition

ancorgs:
New method very similar to the old `Yast::Storage.GetDisk`, added to avoid repeating the same pattern in yast2-bootloader all over the code.

### DIFF
--- a/test/devices_list_test.rb
+++ b/test/devices_list_test.rb
@@ -234,6 +234,36 @@ describe "devices lists" do
         expect(spaces_sdc.size).to eq 1
       end
     end
+
+    describe "#with_name_or_partition" do
+      it "returns a list of disks" do
+        result = disks.with_name_or_partition("/dev/sda2")
+        expect(result).to be_a(Y2Storage::DevicesLists::DisksList)
+      end
+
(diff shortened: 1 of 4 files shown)


test/devices_list_test.rb
```diff
@@ -234,6 +234,36 @@ def partition_name
...
+      end
+
+      it "filters by a single partition name" do
+        list = disks.with_name_or_partition("/dev/sda2")
+        expect(list.size).to eq 1
+        expect(list.first.name).to eq "/dev/sda"
+      end
+
+      it "filters by a set of names" do
+        list = disks.with_name_or_partition(["/dev/sda1", "/dev/sda2", "/dev/sdb", "/dev/sdb1"])
```
jreidinger: NP, when both syntax is allowed, I think it is more common to use `*param` so it can look like `disks.with_name_or_partition("/dev/sda1", "/dev/sda2", "/dev/sdb", "/dev/sdb1")`

ancorgs: I know and I though about it, but I believe it's better to keep the interface as close to `#with` as possible.

PR #138, opened 2016-12-23
Removed unused ProposalDemo client (kind of obsoleted by Dialogs::InstDiskProposal)

ancorgs:
We introduced `ProposalDemo` to follow the Scrum principle of delivering something executable. Now is not longer needed. We have real code that executes the proposal at `InstDiskProposal`.


aschnell:
LGTM

PR #141, opened 2017-01-16
WIP: make StorageManager a real singleton and add revision

ancorgs:
Pull request opened just for discussion.

I have never liked the fact that (in master)
```ruby
manager = Y2Storage::StorageManager.instance
manager.class != Y2Storage::StorageManager
manager.class == Storage::Storage
```
That's not how a singleton class should behave. It's highly unexpected.

With this proposed change:
```ruby
manager = Y2Storage::StorageManager.instance
manager.class == Y2Storage::StorageManager
manager.storage.class == Storage::Storage
```
Which is more consistent.

In addition, it makes sense to me to store the revision (see [1]) in the same singleton object used to (indirectly) contain the devicegraphs. So...
```ruby
manager = Y2Storage::StorageManager.instance
if manager.revision > 0
  puts "This devicegraph is not the original one #{manager.staging}"
end
```
Notice that, as explained in [1], the revision will be handled manually.

[1] https://github.com/yast/yast-storage-ng/blob/master/doc/software-requirements.md

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jan 17 12:07:18 UTC 2017 - ancor@suse.com
+
+- Improved StorageManager that complies to the Singleton pattern
+  and includes a revision counter for the staging devicegraph.
+- 0.1.8
+
+-------------------------------------------------------------------
 Mon Jan 16 12:27:03 UTC 2017 - ancor@suse.com
 
 - Added 'fstab_options' key to the YAML representation of the
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:        0.1.7
+Version:        0.1.8
 Release:	0
 BuildArch:	noarch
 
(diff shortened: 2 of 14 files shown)


shundhammer:
The singleton part is perfectly fine with me. LGTM for that.

I am not convinced, though, that that singleton should also keep track of the revisions of devicegraphs. Isn't that a completely different responsibility? IMHO that could and should be handled in the devicegraph class itself, either on the C++ side or in some Ruby refinement.

ancorgs:
I don't think we can add state (i.e. a new attribute that needs to be stored) to a SWIG object using refinements.

Adding the revision to the C++ part looks more natural from the encapsulation POV. But it's kind of weird because:
 - The revision is manually updated by the "client" code using the library
 - It's YaST specific (updated when leaving the expert partitioner)

So it's like the library is just offering an integer field for the "client" to do whatever it wants to it. Probably not its responsibility.

wfeldt:
lgtm; but I also think that `#manager.revision` is not intuitively associated with device graphs.

shundhammer:
Yes, having another int field with some accessors (get(), increment(), reset() ) in the C++ class should be easy enough.

ancorgs:
If swig wouldn't be involved, I would do a wrapper/decorator class around Storage::Devicegraph or something like that. But since swig performs strict type checking, all we can do to transparently wrap/decorate a class is using refinements. And as said, with refinements we cannot add state.

shundhammer:
Well, we do have access to the C++ code, so let's take the easy path. ;-)

PR #142, opened 2017-01-17
Make StorageManager a real singleton and add revision

ancorgs:
Installation tested manually. The change does not break anything new.

This makes `StorageManager`behave like a real singleton class (see details in the description of #141)

Also adds a `staging_revision` method to check for changes (needed for https://trello.com/c/rKVpR8ul/391-3-storageng-software-selection)

Created as a separate pull request (from #141, which was RFC) to keep the discussion focused.

### DIFF
--- a/src/lib/y2storage/storage_manager.rb
+++ b/src/lib/y2storage/storage_manager.rb
@@ -27,75 +27,130 @@ require "y2storage/fake_device_factory"
 
 module Y2Storage
   #
-  # Singleton class for the libstorage object.
-  #
-  # You can simply use StorageManager.instance to use it and create a
-  # libstorage instance if there isn't one yet; this will use common default
-  # parameters for creating it.
-  #
-  # If you need special parameters for creating it, you can use
-  # StorageManager.create_instance with a custom ::Storage::Environment
-  #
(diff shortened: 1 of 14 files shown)


src/lib/y2storage/storage_manager.rb
```diff
@@ -27,76 +27,131 @@
...
+      #   libstorage instance
+      # @see .create_test_instance if you just need to create an instance
+      #   without hardware probing
+      # @see .fake_from_yaml for easy mocking
       #
-      # @return [::Storage::Storage] libstorage object
+      # @return [StorageManager]
       #
       def instance
         create_instance unless @instance
```
wfeldt: Isn't this were people jump to write `@instance || create_instance`?

wfeldt:
lgtm; just a minor comment

src/lib/y2storage/storage_manager.rb
ancorgs: Well, we already had a lot of discussions about that when the class was originally created. I don't want to open that box again. :-)

wfeldt: I'm fine with it, just wondering.

PR #144, opened 2017-01-18
Install storage-related packages as needed

shundhammer:
https://trello.com/c/TWFjvWmb/484-2-storageng-install-packages-based-on-used-storage-features

Successfully tested with the latest StorageNG ISO - see Trello.


src/lib/y2storage/used_storage_features.rb
```diff
@@ -0,0 +1,208 @@
...
+    FEATURE_PACKAGES =
+      {
+        # SUSE standard technologies
+        UF_LVM:           "lvm2",
+        # Btrfs needs e2fsprogs for 'lsattr' and 'chattr' to check for CoW
+        UF_BTRFS:         ["btrfsprogs", "e2fsprogs"],
+        UF_SNAPSHOTS:     "snapper",
+
+        # RAID technologies and related
+        UF_DM:            ["device-mapper"],
```
ancorgs: NP: any reason to use an array of one element here? Just curious.

src/lib/y2storage/used_storage_features.rb
```diff
@@ -0,0 +1,208 @@
...
+    end
+
+    # Return a list of software packages required for the storage features.
+    # Uses 'features' if non-nil, otherwise collects the features with
+    # collect_features.
+    #
+    # @param [Array<Symbol>] feature list or nil
+    # @return [Array<Symbol>] package list
+    #
+    def feature_packages(features = nil)
```
ancorgs: So if a list of features is passed, this acts as some kind of class method (the internal state of the object is irrelevant). But if the parameter is nil, then the devicegraph pointed by the object becomes relevant. This looks like two different methods to me.

```ruby
# Class method. As usual, I suck at naming methods
def self.packages_for(features)
  # Pretty much what we have
end
# instance method
def feature_packages
  self.class.packages_for(collect_features)
end
```

test/package_handler_test.rb
```diff
@@ -0,0 +1,101 @@
...
+    subject { described_class.new }
+
+    it "has an empty package list" do
+      expect(subject.pkg_list).to be_empty
+    end
+  end
+
+  context("using a devicegraph with Btrfs and LVM") do
+    let(:dg_features) { ::Storage::UF_BTRFS | ::Storage::UF_LVM }
+    let(:devicegraph) { instance_double("::Storage::devicegraph", used_features: dg_features) }
```
ancorgs: Does this work with `instance_double("::Storage::devicegraph")` instead of `instance_double("::Storage::Devicegraph")` (beware capitalization of Devicegraph)? Good to know

src/lib/y2storage/package_handler.rb
```diff
@@ -0,0 +1,168 @@
...
+    # @return true on success, false on error
+    #
+    def install!
+      compact
+      return if @pkg_list.empty?
+      log.info("Installing #{pkg_list}")
+      success = Yast::Package.DoInstall(@pkg_list)
+      if !success
+        log.error("ERROR: Some packages could not be installed")
+        install_error_popup
```
shundhammer: You are right, but this is taken literally from the old code. It also didn't check the return value. OTOH that means that this might also be a simpler error popup with just an "OK" button. There is not much the user can do at that point anyway.

src/lib/y2storage/package_handler.rb
```diff
@@ -0,0 +1,168 @@
...
+      end
+      success
+    end
+
+    # Set the proposal packages for storage. This marks the packages for
+    # installation; it does not install them yet.
+    #
+    def set_proposal_packages
```
shundhammer: I thought about that, but decided against it. Right now, the integration into the proposal for example explicitly calls set_proposal_packages rather than commit just because that's more explicit, and it fits better in the environment of the proposal.

src/lib/y2storage/package_handler.rb
```diff
@@ -0,0 +1,168 @@
...
+      end
+    end
+
+    # Execute package installation. This will install the stored package list
+    # immediately, so it is not advisable to do this during the OS
+    # installation. In the latter case, use 'set_proposal_packages' instead.
+    #
+    # @return true on success, false on error
+    #
+    def install!
```
shundhammer: See below (set_proposal_packages)

src/lib/y2storage/package_handler.rb
```diff
@@ -0,0 +1,168 @@
...
+      solve
+      nil
+    end
+
+    # Remove duplicates from the package list and those packages that are
+    # already installed.
+    #
+    # @return [Array<String>] compacted list
+    #
+    def compact
```
shundhammer: I chose `install!` because that changes not only the state of this object, but (as a side effect) the state of the target system - immediately. This is in contrast to this `compact` method that has less drastic consequences. If we'd use an exclamation mark for every method that changes this object, we'd end up with most methods ending with an exclamation mark since most of them modify object state somehow.

src/lib/y2storage/used_storage_features.rb
shundhammer: This might be a leftover from times when that feature maybe required multiple packages. Or it might be just an accident.

src/lib/y2storage/used_storage_features.rb
shundhammer: Not sure if we even realistically need that in the forseeable future.

test/package_handler_test.rb
shundhammer: It works in the unit tests. AFAIK that name is only used for debugging anyway; from what I read you can also just omit it. Anyway, I didn't do the capitalization intentionally; this is more an accident.

src/lib/y2storage/package_handler.rb
shundhammer: If anybody from the outside starts messing with packages during all this, the outcome is completely undefined.

OTOH we might want to tell and advanced user beforehand which packages actually get installed, and then you'd call `compact` manually and display the resulting package list. Our logging already does this. `compact` also removes duplicates; that might easily happen if we get more places we need to check what additional packages to install, such as general hardware dependent packages. We had that in the old code for driver packages for disk controllers etc.; Steffen said that this is right now obsolete and done in another place, but we might get that back pretty quickly. Similar for architecture specific packages. And then duplicates might be very common.

src/lib/y2storage/package_handler.rb
ancorgs: If we want to make it explicit I would go all the way down - removing `#commit` and adding a check to alert the programmer if `#set_proposal_packages` is used in an installed system (and the complementary check for `#install`).

src/lib/y2storage/package_handler.rb
ancorgs: Well, IMO the name `#install` is already descriptive enough about the consequences.* So the exclamation mark adds no information there and simply makes Ruby developers brains spin. :-)

(*) Even more if you want to make explicit the difference between `#install` and `#set_proposal_packages`

src/lib/y2storage/used_storage_features.rb
ancorgs: Do you mean we always act based on the stored devicegraph? Then there is no need to receive an argument in `#feature_packages`.

src/lib/y2storage/package_handler.rb
ancorgs: I'm not against removing duplicates since that looks safe to me (although I would do it while adding the packages to the list). I'm worried about the second line in `#compact`, not about the first one.

I just feel there is something intrinsically wrong in conditioning the behavior if the object to the state of the system the first time `#compact` was called (even subsequent calls to `#compact` are affected). Specially since the method is called indirectly in several places. I think this is more correct than having a `#compact` method.
```ruby
def compact_pkg_list
  list = @pkg_list.uniq # To be honest, I would do this in #add_packages, but I don't care much
  list.reject {|pkg| Yast::Package.Installed(pkg) }
end
```

An this is YaST, we don't need "anybody from outside" to create insane workflows to shoot our feet. Just think about
```ruby
# Assume this adds package "xx"
pkg_handler.add_feature_packages(devicegraph)
if !pkg_handler.commit
   log.info("Commit of packages failed. Let's try to fix it")
   do_something
   # The following call has the side effect of uninstalling "xx". This kind of side-effects are quite common in YaST
   Yast::SomeModule.DoAnotherThing()
   do_something_else
   log.info("Retrying commit.")
   # "xx" not installed, although now it makes sense
   pkg_handler.commit
end
```

I will not be that surprised if we end up having code like that in the future. And I won't be surprised of the programmer using this API overlooks the mentioned problem (I myself would assume that every call to `#commit` is just based in the current system state).

shundhammer: Changed to `install`

src/lib/y2storage/package_handler.rb
shundhammer: Now only using simple error popup with just one button. We can't make good use of the user's response to the "Continue" / "Abort" dialog anyway.

src/lib/y2storage/package_handler.rb
shundhammer: Let's please have the API flexible and usable for users who prefer different approaches. I really don't want to force just one specific way of using that class upon everbody using it when there is exactly **zero** cost for getting it both ways. `commit` also serves to document how it should be used in different contexts. You can't get any more concise description for that.

src/lib/y2storage/package_handler.rb
shundhammer: If the application code chooses to do stupid things in the meantime, the result will be surprising. That's a normal fact of life. The life time of an instance of this class is not meant to stretch over all kinds of weird things. There is more that could change in the mean time: The device graph might change, the installed packages on the system might change, whatever. If you want to be extra sure, you will simply have to call `add_feature_packages` again.

Calling any of `commit`, `install`, `set_proposal_packages` will implicitly call `compact` and even handle such cases as packages have already been installed, so an `install` followed by an `install` will do no harm.

Having `compact` available to the public will also do no harm; it just removes packages that not or no longer need to be installed. It's the defensive approach, and that defensive approach is not subject to negotiation from my side.

test/package_handler_test.rb
shundhammer: Changed to "::Storage::Devicegraph" (no idea if that has any semantic effect at all)

.rubocop.yml
```diff
@@ -14,7 +14,7 @@ SingleLineBlockParams:
 # resolutions and Github's pull request view. But let's open the threshold a
 # little bit
 Metrics/LineLength:
-  Max: 105
+  Max: 110
```
ancorgs: To be honest, if changing the offending code is not possible I would prefer to disable the check just for the affected piece of code instead of opening the door for the whole repository.

src/lib/y2storage/used_storage_features.rb
```diff
@@ -0,0 +1,214 @@
...
+          end
+        end
+      end
+
+      log.info("Storage features used: #{features}")
+      features
+    end
+
+    # Return a list of software packages required for the storage features.
+    # Uses 'features' if non-nil, otherwise collects the features with
```
ancorgs: The inline documentation is now outdated.

shundhammer: Ah, right.

shundhammer: Fixed.

.rubocop.yml
shundhammer: Now only excluding the one affected unit test.

src/lib/y2storage/used_storage_features.rb
```diff
@@ -0,0 +1,215 @@
...
+      end
+
+      log.info("Storage features used: #{features}")
+      features
+    end
+
+    # Return a list of software packages required for the storage features in
+    # 'features'.
+    #
+    # @param [Array<Symbol>] feature list or nil
```
ancorgs: "or nil"

PR #153, opened 2017-01-25
Proposal working with empty disks (no partition table)

ancorgs:
But what is there is ready for review. It works. Tested manually and general unit tests included.

Review by commit for better understanding.


src/lib/y2storage/proposal/assigned_space.rb
```diff
@@ -56,12 +56,14 @@ def partition_type
         @partition_type if @partition_type_calculated
 
         @partition_type_calculated = true
-        @partition_type = if disk.partition_table.extended_possible
-          if disk.partition_table.has_extended
-            inside_extended? ? :logical : :primary
+        disk.as_not_empty do
```
shundhammer: Hm... in the old code, the difference was clearly visible. Now it's pretty mixed up IMHO; it is not immediately obvious that one code branch is only for the "completely empty disk" case.

src/lib/y2storage/proposal/space_distribution.rb
```diff
@@ -47,7 +49,9 @@ def initialize(volumes_by_disk_space)
         end
         @spaces.freeze
         spaces_by_disk.each do |disk, spaces|
-          set_num_logical_for(spaces, disk.partition_table)
+          disk.as_not_empty do
```
shundhammer: What happens if the disk is completely empty? AFAICS that means that its whole free space is disregarded, right?

ancorgs: No.

First, we already have support for FreeDiskSpace in empty disks. See https://github.com/yast/yast-storage-ng/pull/152

Second, even if we wouldn't have that, the usage of `as_not_empty` ensures we treat the empty and unused disks as disks with an empty GPT table (that's what they will become in the end if we end up creating partitions there). So we will count all the space except the one used by the (future) data structures.

src/lib/y2storage/proposal/assigned_space.rb
ancorgs: Actually, there is no different branch for completely empty disks. We are still doing the same. But now instead of raising an exception if the partition table is not there we assume that, even if it's not there now, there would be a GPT partition as soon as it's needed.

That's the general idea of using `as_not_empty`. It dresses up completely empty disks as disks with an empty GPT table, so they are considered like so during the calculations (free spaces, logical vs primary, etc.)

shundhammer: OK then.

src/lib/y2storage/proposal/space_distribution.rb
shundhammer: OK

src/lib/y2storage/refinements/disk.rb
```diff
@@ -77,6 +77,51 @@ def free_spaces
...
+          remove_descendants if fake_ptable
+        end
+
+        # Default partition type for newly created partitions
+        #
+        # This method is needed because YaST criteria does not necessarily match
+        # the one followed by Storage::Disk#default_partition_table_type (which
+        # defaults to MBR partition tables in many cases)
+        def preferred_ptable_type
+          # TODO: so far, DASD is not supported, so we always suggest GPT
```
wfeldt: If it's empty, we can partition it any way we like - the s390x corner-case is for disks with 'hard wired' partitions.
afaiui

ancorgs: Well, in the PBI card it can be read "_Based on situation it chooses the right type (DASD, otherwise GPT)_". So assume it was discussed and there are cases were DASD is mandatory. I can remove the `TODO` anyways if you prefer so.

shundhammer:
LGTM

shundhammer:
LGTM on the version bump and change log, too.

src/lib/y2storage/refinements/disk.rb
wfeldt: Yes, just re-checked. You are right. DASDs are also special. Remind me when I forget again. ;-)

wfeldt:
lgtm

PR #157, opened 2017-02-08
Prototype of the guided setup

ancorgs:
* The dialog class is now less smart. Logic moved to a client class.
* The client class supports going back and forward in the installation workflow (the accepted proposal, if any, is now stored in the StorageManager instance) and calls the appropiate dialog for each task.
* Prototype of the guided setup wizard (just one step so far).

The new dialog is not covered by tests, but that's expected in the PBI https://trello.com/c/kMwkVir8/ Tests added as "pending" to make sure we do not forget.

Initial screen
guided1

Change to LVM
guided2

Result
guided3


wfeldt:
I still think that two checkboxes

[x] use lvm
[x] use encryption

are better than tree radio buttons () partition based / () lvm / () encryption

For one, it 's shorter and covers all four cases. And secondly, it doesn't leave the user wonder what 'partition-based' is.

src/lib/y2storage/storage_manager.rb
```diff
@@ -51,30 +51,69 @@ class StorageManager
...
     def initialize(storage_environment)
       @storage = Storage::Storage.new(storage_environment)
       @staging_revision = 0
+      @proposal = nil
     end
 
-    # Sets the devicegraph as the staging one, updating all the associated
-    # information like #staging_revision
+    # Checks whether the staging devicegraph has been previously set, either
+    # manually either through a proposal.
```
wfeldt: or?

ancorgs:
I agree... in the long term.

Having three radio buttons instead of two checkboxes was a conscious decision taken by Ken as long as we don't support encryption without LVM (something that is not planned in the very short term). In the past we had two checkboxes with the second one being disabled if "lvm" was not marked and it was reported to be an usability problem and fixed in that way by Ken (I was involved in the discussion and the implementation).

We can change it to two checkboxes as soon as we support encryption without LVM.

Anyway, this kind of details were explicitly left out of the scope of the PBI. https://trello.com/c/kMwkVir8/

src/lib/y2storage/storage_manager.rb
ancorgs: Ooops

wfeldt:
Ok, agreed.

PR #159, opened 2017-02-09
Encryption in YAML reader

shundhammer:
https://trello.com/c/Tt1a2Pu7/336-5-yast2-storage-ng-encrypted-lvm-proposal

 - Added encryption layer to the YAML reader
- Added unit test for encryption in the YAML reader
- Added encryption to YAML documentation
- Added dependencies between subproducts to AbstractDeviceFactory so the order of file_system vs. encryption in a partition is handled automatically

### DIFF
--- a/src/lib/y2storage/fake_device_factory.rb
+++ b/src/lib/y2storage/fake_device_factory.rb
@@ -168,7 +168,6 @@ module Y2Storage
       DEPENDENCIES
     end
 
-
     #
     # Factory methods
     #
@@ -353,8 +352,6 @@ module Y2Storage
       partition = ptable.create_partition(part_name, region, type)
       partition.id = id
 
-      log.info("#{__method__} finish")
-
       part_name
     end
     # rubocop:enable all
@@ -457,12 +454,11 @@ module Y2Storage
       raise ArgumentError, "\"name\" missing for encryption on #{parent}" if name.nil?
       password = args["password"]
       type_name = args["type"] || "luks"
-      type = fetch(ENCRYPTION_TYPES, type_name, "encryption type", name)
       # We only support creating LUKS so far
       raise ArgumentError, "Unsupported encryption type #{type_name}" unless type_name == "luks"
       encryption = ::Storage::Luks.create(@devicegraph, name)
       encryption.password = password unless password.nil?
-      if @partitions.has_key?(parent)
+      if @partitions.key?(parent)
         # Notify create_file_system that this partition is encrypted
         @partitions[parent]["encryption"] = encryption.name
       end
@@ -555,7 +551,6 @@ module Y2Storage
       log.info("#{__method__}( #{parent}, #{args} )")
 
       blk_device_name = args["blk_device"]
-
       blk_device = ::Storage::BlkDevice.find_by_name(devicegraph, blk_device_name)
 
       parent.add_lvm_pv(blk_device)
--- a/src/lib/y2storage/fake_device_factory.rb
+++ b/src/lib/y2storage/fake_device_factory.rb
@@ -34,6 +34,8 @@ module Y2Storage
   # This is typically used with a YAML file.
   # Use the inherited load_yaml_file() to start the process.
   #
+  # rubocop:disable Metrics/ClassLength
+  #
   class FakeDeviceFactory < AbstractDeviceFactory
     include EnumMappings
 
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/fake_device_factory.rb
```diff
@@ -413,6 +437,34 @@ def create_free(parent, args)
...
+      type_name = args["type"] || "luks"
+      # We only support creating LUKS so far
+      raise ArgumentError, "Unsupported encryption type #{type_name}" unless type_name == "luks"
+      encryption = ::Storage::Luks.create(@devicegraph, name)
+      encryption.password = password unless password.nil?
+      if @partitions.key?(parent)
+        # Notify create_file_system that this partition is encrypted
+        @partitions[parent]["encryption"] = encryption.name
+      end
+      encryption
```
ancorgs: It's not clear to me where the parent and the encryption are linked together. For example, if you do `a_partition.create_encryption(name)` you can then do `a_partition.encryption`. So it's pretty clear than the partition is encrypted. Here, we create the Luks device by and and I guess that link is done in another step, but I'm not sure where/when.

test/fake_device_factory_test.rb
```diff
@@ -137,4 +137,56 @@
...
+    # rubocop:enable all
+
+    io = StringIO.new(input.join("\n"))
+    Y2Storage::FakeDeviceFactory.load_yaml_file(staging, io)
+
+    expect(staging.num_devices).to eq 7
+    expect(staging.num_holders).to eq 5
+    enc = Storage.to_encryption(Storage::BlkDevice.find_by_name(staging, "/dev/mapper/cr_root"))
+    expect(enc.name).to eq "/dev/mapper/cr_root"
+    expect(enc.password).to eq "s3cr3t"
```
ancorgs: Connected to my comment above, I would like to see things like these here, to make sure the encrypted device is created in the proper place.
```ruby
partition = Storage::Partition.find_by_name("/dev/sda2/")
expect(partition.encryption.name).to eq "/dev/mapper/cr_root"
expect(partition.encryption.filesystem.mount_point).to eq "/"
```
So we ensure the right relationships between the partition, the encryption and the filesystem, beyond the encryption just being in the devicegraph.

src/lib/y2storage/fake_device_factory.rb
```diff
@@ -46,12 +46,13 @@ class FakeDeviceFactory < AbstractDeviceFactory
...
-        "partition"  => ["file_system"],
+        "partition"  => ["file_system", "encryption"],
+        "encryption" => ["file_system"],
         "lvm_vg"     => ["lvm_lvs", "lvm_pvs"],
         "lvm_lvs"    => ["lvm_lv"],
-        "lvm_lv"     => ["file_system"],
+        "lvm_lv"     => ["file_system", "encryption"],
         "lvm_pvs"    => ["lvm_pv"],
-        "lvm_pv"     => []
+        "lvm_pv"     => ["encryption"]
```
ancorgs: Is this right? Documentation says "For LVM PVs, add this to the corresponding partition"

shundhammer: Right, that shouldn't be there any more. Thanks for spotting this!

src/lib/y2storage/fake_device_factory.rb
shundhammer: See `create_filesystem`. The problem is that since `encryption` is a sub-product, it is not available as an attribute anymore; the abstract factory level filters those out. So we cannot simply pass it to `create_filesystem` like the other attributes that really belong to the filesystem level (mount_point etc.).

test/fake_device_factory_test.rb
shundhammer: Yes, the unit test is still very simplistic, I know. Let me try to improve this particular bit right now.

test/fake_device_factory_test.rb
```diff
@@ -23,6 +23,7 @@
 require_relative "spec_helper"
 require "storage"
 require "y2storage/fake_device_factory"
+require "pp"
```
ancorgs: leftover

test/fake_device_factory_test.rb
```diff
@@ -181,11 +182,9 @@
...
     Y2Storage::FakeDeviceFactory.load_yaml_file(staging, io)
 
-    expect(staging.num_devices).to eq 7
-    expect(staging.num_holders).to eq 5
-    enc = Storage.to_encryption(Storage::BlkDevice.find_by_name(staging, "/dev/mapper/cr_root"))
-    expect(enc.name).to eq "/dev/mapper/cr_root"
-    expect(enc.password).to eq "s3cr3t"
+    root_part = Storage.to_partition(Storage::BlkDevice.find_by_name(staging, "/dev/sda2"))
+    expect(root_part.encryption.name).to eq "/dev/mapper/cr_root"
+    expect(root_part.encryption.filesystem.mountpoints.first).to eq "/"
```
ancorgs: No need to remove the check for `enc.password`. Don't take my previous suggestion too literally. I just wanted to make sure the test checked the relationship between the Luks and Partition objects. Didn't mean to replace the whole test or to stop using the `enc` variable for readability.

PR #161, opened 2017-02-14
Ask for LVM encryption password

joseivanlopez:
First attempt for asking LVM password. Only a simple password field without checking anything.


src/lib/y2storage/dialogs/guided_setup.rb
```diff
@@ -60,7 +60,9 @@ def dialog_content
               VSpacing(1),
               Left(RadioButton(Id(:mode_lvm), _("LVM-based"), lvm_selected?)),
               VSpacing(1),
-              Left(RadioButton(Id(:mode_encrypted), _("Encrypted LVM-based"), encrypted_selected?))
+              Left(RadioButton(Id(:mode_encrypted), _("Encrypted LVM-based"), encrypted_selected?)),
+              VSpacing(1),
+              Left(Password(Id(:encryption_password), _("Enter encrytion password")))
```
ancorgs: typo: encrytion.

While you fix it, maybe it would be nice to indent this a little bit to the right, so it looks to be "inside" the Encrypted LVM-based. I know the dialog is just temporary, but...

ancorgs: And the value should be initialized to `settings.encryption_password`. Otherwise this would be lost if the user re-enters the dialog.

src/lib/y2storage/dialogs/guided_setup.rb
```diff
@@ -53,19 +53,39 @@ def dialog_title
...
+                Left(RadioButton(Id(:mode_lvm), _("LVM-based"), lvm_selected?)),
+                VSpacing(1),
+                Left(RadioButton(Id(:mode_encrypted), _("Encrypted LVM-based"), encrypted_selected?))
+              )
+            ),
+            VSpacing(0.2),
+            Left(
+              HBox(
+                HSpacing(8),
+                Password(Id(:encryption_password), _("Enter encryption password"))
```
shundhammer: This field should be there twice to make sure the user typed the same thing consistently; since the widget is in password mode, the user can't see what he types. One typo, and he will never be able to decrypt that LVM again.

ancorgs:
Coverage decrease was expected as the dialog is temporary, so LGTM.

src/lib/y2storage/dialogs/guided_setup.rb
```diff
@@ -53,19 +53,39 @@ def dialog_title
...
       end
 
+      def create_dialog
+        super
+        init_widgets
+        true
+      end
+
+      def init_widgets
+        # Remember introduced password
```
shundhammer: "entered", not "introduced" ;-)

src/lib/y2storage/dialogs/guided_setup.rb
ancorgs: The whole dialog is temporary just to have something for openQA (even the unit tests are an empty "pending"). See review section of https://trello.com/c/kMwkVir8/502-2-storageng-proposal-make-it-configurable Specially the follow-up PBIs about making it really usable.

src/lib/y2storage/dialogs/guided_setup.rb
joseivanlopez: Yes ;)

PR #162, opened 2017-02-14
Encrypted LVM tests

ancorgs:


src/lib/y2storage/fake_device_factory.rb
```diff
@@ -452,7 +452,7 @@ def create_free(parent, args)
     #
     def create_encryption(parent, args)
       log.info("#{__method__}( #{parent}, #{args} )")
-      name = args["name"]
+      name = args["name"].split("/").last
```
shundhammer: If we use it like this, maybe we should also check for `start_with("/dev/mapper")

PR #167, opened 2017-02-16
Tests for EncryptionsList class

joseivanlopez:


ancorgs:
Closing with no review because the whole idea of `DevicesList`is becoming obsolete with the new Ruby wrapper. So all the related unit test should be gone in a couple of sprints anyway.

PR #168, opened 2017-02-20
- adapted to libstorage-ng change

aschnell:


aschnell:
The checks fails because they still use the previous version of libstorage-ng.

PR #169, opened 2017-02-23
WIP: Y2Storage classes wrapping libstorage

ancorgs:
WIP: don't look to much at the code, I'm more interested now in discussing the result exposed in this description.

For background and details about the motivation, see the description of [this Hack Week project](https://hackweek.suse.com/15/projects/yast2-storage-ng-as-a-libstorage-ng-wrapper-poc).

Accessing directly to the libstorage-ng API (the `Storage` namespace) from the YaST code has some drawbacks. The API is not very Ruby-like and sometimes it does not completely fit the YaST use-cases. This pull request adds several classes to `Y2Storage` (the yast-storage-ng namespace) to be used by YaST instead of accessing libstorage directly. Of course, these classes rely on libstorage-ng to do the job, but that's transparent to the rest of YaST. A comparison of the usage of these classes vs direct use of libstorage-ng follows.

## Arrays instead of strictly-typed vectors

The `Storage` namespace (libstorage-ng).

```ruby
a_storage_filesystem.mountpoints.class #=> Storage::VectorString 
a_storage_filesystem.blk_devices.class #=> Storage::VectorConstBlkDevicePtr 
Storage::Disk.all(devicegraph).class #=> Storage::VectorDiskPtr
```

Those vectors don't act as usual Ruby collection, so the code directly using storage is **full** of `to_a`.

On the other hand, with the new equivalent classes in `Y2Storage`.

```ruby
a_y2storage_filesystem.mountpoints.class #=> Array
a_y2storage_filesystem.block_devices.class #=> Array
Y2Storage::Disk.all(devicegraph).class #=> Array
```

## No more downcasting

Consider a devicegraph that contains a formatted partition, a formatted LVM logical volume and a directly formatted disk. This is how it looks like in storage.

```ruby
formatted = all_storage_filesystems.map {|fs| fs.blk_devices[0] }
formatted.map {|dev| dev.class } #=> [Storage::BlkDevice, Storage::BlkDevice, Storage::BlkDevice]
```

On the other hand, with y2storage.

```ruby
formatted = all_y2storage_filesystems.map {|fs| fs.blk_devices[0] }
formatted.map {|dev| dev.class } #=> [Y2Storage::Partition, Y2Storage::LvmLv, Y2Storage::Disk]
```

That means the code using directly storage is **full** of downcasting. In the example above with storage, before being able to actually use the elements in the array with all their methods, you would need to check what kind of device is every one of them and do the corresponding cast. Let's see it with a much simpler example.
```ruby
partitionable = a_storage_partition.partitionable
disk = Storage.disk?(partitionable) ? Storage.to_disk(partitionable) : nil
```
That's quite un-natural for a language based on the concept of duck-typing. Specially since the methods to check the real type and to perform the downcasts live directly in the `Storage` namespace. 

On the other hand, with the new classes it would be:
```ruby
partitionable = a_y2storage_partition.partitionable
disk = partitionable.is_a?(Y2Storage::Disk) ? partitionable : nil

# Or simply
disk = a_y2storage_partition.disk
```

Several extra example of downcasting can be found in the examples below, since they are everywhere using libstorage-ng.

## Nil instead of exceptions

With storage
```ruby
a_storage_unused_partition.filesystem #=> raises Storage::WrongNumberOfChildren
a_storage_pv_partition.filesystem #=> raises Storage::DeviceHasWrongType
```

The code using libstorage contains a `rescue` basically every time you need to check for something that may or may not be there. That's typically resolved using `nil` in Ruby, so.
```ruby
a_y2storage_unused_partition.filesystem #=> nil
a_y2storage_pv_partition.filesystem #=> nil
```

## Classes instead of constants and enums

In libstorage-ng enums are used to define the possible file-system types, partition ids, partition types, possible options for mount_by and many many other things. So in the end, they become just numbers with an associated constant name:
```ruby
a_storage_filesystem.type.class #=> Fixnum
a_storage_filesystem.mount_by.class #=> Fixnum
a_storage_partition.type.class #=> Fixnum
a_storage_partition.id.class #=> Fixnum
```
As a result, the code using those types look like this.
```ruby
puts "yes" if a_storage_partition.type == Storage::PartitionType_PRIMARY
puts "Ptable type #{a_storage_msdos_ptable.type}" #=> Prints "Ptable type 1"
puts "extended" if [Storage::FsType_EXT2, Storage::FsType_EXT3, Storage::FsType_EXT4].include?(a_storage_filesystem.type)
```
And the only way of knowing all the possible types is including [enum_mappings.rb](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/enum_mappings.rb) which defines more constants (that contain hashes pointing to the original constants). But, even though, for printing a more descriptive name for the type of a partition, something like this is needed.
```ruby
include Y2Storage::EnumMappings
puts "Type is #{PARTITION_TYPES.invert[a_storage_partition.type]}"
```
For some types there are workarounds, like these
```ruby
Storage.ext3?(a_storage_filesystem.type)
Storage.pt_type_name(a_storage_part_table.type)
puts "Yes" if "ntfs" == a_storage_filesystem.to_s
puts "Also yes" if "ntfs" == a_different_storage_fs.to_s
```
For other enums there are no workarounds and the constants must be used.

In this pull requests all those lists of ids and types are "promoted" to real classes. So
```ruby
a_y2storage_filesystem.type.class #=> Y2Storage::Filesystems::Type
a_y2storage_filesystem.mount_by.class #=> Y2Storage::Filesystems::MountByType
a_y2storage_partition.type.class #=> Y2Storage::PartitionType
a_y2storage_partition.id.class #=> Y2Storage::PartitionId
```
So they are used like this
```ruby
puts "yes" if a_y2storage_partition.type.to_sym == :primary
# or better
puts "yes" if a_y2storage_partition.type.is?(:primary)
puts "Ptable type #{a_y2storage_msdos_ptable.type}" #=> Prints "Ptable type msdos"
```
The `#is?` method can be specialized in some classes to allow things like
```ruby
puts "extended" if a_y2storage_filesystem.type.is?(:ext) #=> Returns true for ext2, ext3 and ext4
```
No hashes are needed to get the whole list or to get a printable "human" name instead of a Fixnum.
```ruby
Y2Storage::PartitionType.all.each do |type|
  puts "Name: #{type.to_s} - Internal libstorage value: #{type.to_i}"
end
```
Last but not least, creating these objects works both by name and by the internal value, so there are equivalent
```ruby
Y2Storage::Filesystems::Type.find(:btrfs)
Y2Storage::Filesystems::Type.find(5)
Y2Storage::Filesystems::Type.find(Storage::FsType_BTRFS)
```

## All sizes as DiskSize objects

Our Ruby classes use `Y2Storage::DiskSize`, but sizes coming from libstorage-ng are always Fixnum (bytes). The new classes unify that.
```ruby
a_storage_partition.size.class # Fixnum (bytes)
a_y2storage_partition.size.class # Y2Storage::DiskSize
a_y2storage_partition.size.to_i.class # Fixnum (bytes)
```

## Adding methods needed by YaST

Right now, when a libstorage-ng class does not fully fulfill the YaST needs we are adding methods to it either using [refinements](https://github.com/yast/yast-storage-ng/tree/master/src/lib/y2storage/refinements) or directly opening the `Storage` namespace and [monkey-patching](https://github.com/yast/yast-storage-ng/tree/master/src/lib/storage). And as you can see in the links, we already have quite some methods in both places.

The new `Y2Storage` classes can be extended without tricks.

Some examples of useful methods (things usually queried by other YaST modules) with their `Storage` counterpart:
```ruby
a_y2storage_disk.gpt?
# That would be...
ptable = a_storage_disk.partition_table if a_storage_disk.has_partition_table # Without this check we risk an exception
!!(ptable && ptable.type == Storage::PtType_GPT)
```

```ruby
a_y2storage_filesystem.all_disks
# That would be (leaving LVM out, that would mean several extra lines of code)
devices = a_storage_fs.block_devices
devices.each_with_object([]) do |dev, disks|
  if Storage.partition?(dev)
    part = Storage.to_partition(dev)
    partitionable = part.partition_table.partitionable
    next unless Storage.disk?(partitionable)
    disks << Storage.to_disk(partitionable)
  elsif Storage.disk?(dev)
    disks << Storage.to_disk(dev)
  end
end
```

## Less surprising API (for Rubyists)

```ruby
Storage::BlkDevice.all(graph) #=> Returns all the block devices as Storage::BlkDevice
Storage::Disk.all(graph) #=> Returns all the disks as Storage::Disk
Storage::Partition.all(graph) #=> Exactly the same than Storage::BlkDevice.all(graph)
```

On the other hand
```ruby
Y2Storage::BlkDevice.all(graph) #=> Returns all the block devices alreay downcasted
Y2Storage::Disk.all(graph) #=> Returns all the disks
Y2Storage::Partition.all(graph) #=> Returns only the partitions (not all the block devices)
```

## Possibility to rearrange/rename classes

Disclaimer: I recognize this point is heavily opinionated, it represents my own _taste_ when organizing code, that may not be shared.

In libstorage, all the classes are directly under the Storage namespace, like
```ruby
Storage::Devicegraph
Storage::Actiongraph
Storage::Disk < Storage::BlkDevice
Storage::Partition < Storage::BlkDevice
Storage::Filesystem < Storage::Device
Storage::BlkFilesytem < Storage::Filesystem
Storage::PartitionTable < Storage::Device
Storage::Msdos < Storage::PartitionTable
Storage::Gpt < Storage::PartitionTable
Storage::PartitionSlot
Storage::Region
```
And, in addition, we have all the constants/enums like
```ruby
Storage::ID_DOS32, Storage::ID_LINUX, Storage::ID_NTFS...
Storage::PartitionType_PRIMARY, Storage::PartitionType_EXTENDED...
Storage::PtType_MSDOS, Storage::Pt_Type_GPT...
Storage::FsType_EXT3...
Storage::MountByType_DEVICE, Storage::MountByType_UUID...
```

In y2storage, we have the opportunity to rearrange that, including the new classes that "substitute" the many constants/enum. So now it looks like

```ruby
Y2Storage::Devicegraph
Y2Storage::Actiongraph
Y2Storage::Disk < Y2Storage::BlkDevice
Y2Storage::Partition < Y2Storage::BlkDevice
Y2Storage::Filesystems::Base < Y2Storage::Device
Y2Storage::Filesystems::BlkFilesytem < Y2Storage::Filesystems:Base
Y2Storage::Filesystems::Type
Y2Storage::Filesystems::MountByType
Y2Storage::PartitionTables::Base
Y2Storage::PartitionTables::Msdos < Y2Storage::PartitionTables::Base
Y2Storage::PartitionTables::Gpt < Y2Storage::PartitionTables::Base
Y2Storage::PartitionTables::Type
```

That is, filesystems and partition tables are namespaced.

## No more methods directly in `Storage`

Many of the above examples show the heavy usage of methods directly living under `Storage` for checking the real underlying class of each device, for casting, for converting from the enums to text, etc. Using the classes in `Y2Storage`, the usage of those methods is not longer needed. There are equivalent methods (when needed) in the class really holding that responsibility.

## How do you like it?

This pull request includes quite some working code, it's already possible to use the `Y2Storage` classes to fake a devicegraph without LVM and to run most of the existing code. So the principle works and the whole thing can be completed during Hack Week for sure. Now I want to know if the YaST team believes it's worth it.


shundhammer:
I admit that I am in a bit of a hurry right now and thus I only glanced through the second half of this, but I really like it. It will make the life of the Ruby-side y2storage hackers much easier, I believe.

wfeldt:
[Sorry, didn't check during Hackweek.]

I also like this. I'd even go a bit further: if this is intended to be a universal 'rubyfying' wrapper around libstorage you might consider changing the namespace from Y2Storage to something more neutral like
'RStorage' or 'StorageNG' to make it clear that this can be used independently of yast.

ancorgs:
Replaced by #184

PR #171, opened 2017-03-01
Proposal PowerPC tests

joseivanlopez:
PBI https://trello.com/c/DRGWzT1s/504-2-storageng-make-proposal-work-with-power.

* Add tests to check proposal matches bootloader requirements for ppc.
* File `proposal_test` was too long, so it is splitted.
* Created files `proposal_scenarios_x86_test` and `proposal_scenarios_ppc_test`.


test/support/proposal_context.rb
```diff
@@ -0,0 +1,77 @@
...
+    allow(Yast::Arch).to receive(:x86_64).and_return(architecture == :x86)
+    allow(storage_arch).to receive(:ppc_power_nv?).and_return(ppc_power_nv)
+    allow(Y2Storage::StorageManager.instance.storage).to receive(:arch).and_return(storage_arch)
+    allow(storage_arch).to receive(:efiboot?).and_return(false)
+    allow(storage_arch).to receive(:x86?).and_return(architecture == :x86)
+    allow(storage_arch).to receive(:ppc?).and_return(architecture == :ppc)
+    allow(storage_arch).to receive(:s390?).and_return(architecture == :s390)
+  end
+
+  subject(:proposal) { described_class.new(settings: settings) }
```
ancorgs: I wouldn't place `subject` in a shared context

test/proposal_scenarios_ppc_test.rb
```diff
@@ -0,0 +1,76 @@
...
+require_relative "spec_helper"
+require "storage"
+require "y2storage"
+require_relative "support/proposal_context"
+
+describe Y2Storage::Proposal do
+  describe "#propose" do
+    using Y2Storage::Refinements::TestDevicegraph
+    using Y2Storage::Refinements::SizeCasts
+    using Y2Storage::Refinements::DevicegraphLists
```
ancorgs: I would say none of these three refinements are really used here

test/proposal_scenarios_ppc_test.rb
```diff
@@ -0,0 +1,76 @@
...
+    let(:scenario) { "empty_hard_disk_50GiB" }
+
+    context "in a PPC64 bare metal (PowerNV)" do
+      let(:ppc_power_nv) { true }
+      let(:expected_scenario) { "ppc_power_nv" }
+
+      context "if requires a /boot partition" do
+        let(:lvm) { true }
+
+        it "creates a /boot partition" do
```
ancorgs: The description does not match the `expect`. This does not check specifically for the `/boot` partition. So I would use,
```ruby
context "if requires a /boot partition" do
  let(:lvm) { true }
  include_examples "proposed layout"
end
```
which would print "it proposes the expected layout". That's what the test actually does.

test/proposal_scenarios_ppc_test.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    context "in a PPC64 non-PowerNV" do
+      let(:ppc_power_nv) { false }
+      let(:expected_scenario) { "ppc_non_power_nv" }
+
+      context "if requires a PReP partition" do
+        let(:lvm) { false }
+
+        it "creates a PReP partition" do
```
ancorgs: Same than above

test/proposal_scenarios_ppc_test.rb
```diff
@@ -0,0 +1,76 @@
...
+          proposal.propose
+          expect(proposal.devices.to_str).to eq expected.to_str
+        end
+      end
+
+      context "if requires /boot and PReP partitions" do
+        let(:lvm) { true }
+
+        it "creates /boot and PReP partitions" do
```
ancorgs: Same than above, again

test/proposal_scenarios_x86_test.rb
```diff
@@ -0,0 +1,137 @@
...
+require "storage"
+require "y2storage"
+require_relative "support/proposal_examples"
+require_relative "support/proposal_context"
+
+describe Y2Storage::Proposal do
+  describe "#propose" do
+    using Y2Storage::Refinements::TestDevicegraph
+    using Y2Storage::Refinements::SizeCasts
+    using Y2Storage::Refinements::DevicegraphLists
```
ancorgs: Same than above, these refinements don't look necessary at first sight.

test/proposal_test.rb
```diff
@@ -24,156 +24,17 @@
 require "storage"
 require "y2storage"
 require_relative "support/proposal_examples"
```
ancorgs: `proposal_examples` is not used here, I believe. Which makes me thing it should be called `proposal_scenario_examples`, BTW.

test/proposal_scenarios_x86_test.rb
joseivanlopez: You are right, they are not used here.

test/proposal_test.rb
joseivanlopez: Some tests here use it:

```
      context "if no disk is enforced for '/'" do
        let(:root_device) { nil }
        let(:yaml_suffix) { "sdb_root_device" }

        include_examples "proposed layout"
        ...
```
so, we have to include it.

test/proposal_scenarios_ppc_test.rb
joseivanlopez: Really, only `using Y2Storage::Refinements::TestDevicegraph` was necessary, but we can remove it if shared proposal examples are used.

test/support/proposal_context.rb
joseivanlopez: Ok

test/proposal_scenarios_ppc_test.rb
joseivanlopez: Ok

test/proposal_test.rb
ancorgs: Oh, I overlooked that one

PR #173, opened 2017-03-03
Encrypted non-LVM

joseivanlopez:
PBI https://trello.com/c/bhwu3b5W/333-3-yast-storage-ng-encrypted-non-lvm-proposal.

* Add encryption for partition-based proposal.
* Swap partitions are not reused when using encryption.
* Add Yaml-based tests to cover new scenarios.

### DIFF
--- a/src/lib/y2storage/proposal/devicegraph_generator.rb
+++ b/src/lib/y2storage/proposal/devicegraph_generator.rb
@@ -55,6 +55,7 @@ module Y2Storage
         # We are going to alter the volumes in several ways, so let's be a
         # good citizen and do it in our own copy
         volumes = volumes.deep_dup
+        add_encryption_password(volumes)
 
         if settings.use_lvm
           lvm_vols, volumes = volumes.split_by(&:can_live_on_logical_volume)
@@ -86,6 +87,17 @@ module Y2Storage
       end
 
     protected
+
+      # Modifies planned volumes adding an encryption password.
+      # Only volumes that can_live_on_logical_volume could have
+      # an encryption password.
+      # @param volumes [PlannedVolumesList]
+      def add_encryption_password(volumes)
+        return volumes unless settings.use_encryption
+        volumes.select(&:can_live_on_logical_volume).map do |volume|
+          volume.encryption_password = settings.encryption_password
+        end
+      end
 
       # Provides free disk space in the proposal devicegraph to fit the
       # volumes in.
--- a/src/lib/y2storage/proposal/devicegraph_generator.rb
+++ b/src/lib/y2storage/proposal/devicegraph_generator.rb
@@ -55,10 +55,10 @@ module Y2Storage
         # We are going to alter the volumes in several ways, so let's be a
         # good citizen and do it in our own copy
         volumes = volumes.deep_dup
-        add_encryption_password(volumes)
+        add_encryption_password!(volumes)
 
         if settings.use_lvm
-          lvm_vols, volumes = volumes.split_by(&:can_live_on_logical_volume)
+          volumes, lvm_vols = volumes.split_by(&:plain_partition?)
         else
           lvm_vols = PlannedVolumesList.new
         end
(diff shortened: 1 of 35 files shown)


src/lib/y2storage/proposal_settings.rb
```diff
@@ -50,6 +50,10 @@ def initialize
       @home_filesystem_type     = ::Storage::FsType_XFS
       @enlarge_swap_for_suspend = false
     end
+
+    def use_encryption
```
ancorgs: This should be `use_encryption?` (with question mark). Although I would prefer `encrypt?` for consistency with the equivalent methods already present in `LvmHelper` and `PlannedVolume`

src/lib/y2storage/proposal/devicegraph_generator.rb
```diff
@@ -87,6 +88,17 @@ def devicegraph(volumes, initial_graph, disk_analyzer)
 
     protected
 
+      # Modifies planned volumes adding an encryption password.
+      # Only volumes that can_live_on_logical_volume could have
+      # an encryption password.
+      # @param volumes [PlannedVolumesList]
+      def add_encryption_password(volumes)
+        return volumes unless settings.use_encryption
+        volumes.select(&:can_live_on_logical_volume).map do |volume|
```
ancorgs: That would mean, of course, adapting the wording in the corresponding tests and, as a result, adapting https://github.com/yast/yast-storage-ng/blob/master/doc/boot-requirements.md

I would probably go one step further than adapting the wording. Right now we usually have

*  With a partitions-based proposal
*  With a LVM-based proposal

I would add a third branch all over those tests, leaving it like

* With a plain partitions-based proposal
* With a LVM-based proposal
* With an encrypted proposal

I know the expectations for the last two cases will be almost identical everywhere, but I believe it's worth having explicit separate contexts from the good testing point of view.

ancorgs:
If I can ask for an extra favor, I would like the new yaml files to follow the same formatting rules (indentation and all that) than the equivalent files without the "enc-" part. That makes comparison MUCH easier for human beings.

src/lib/y2storage/proposal/devicegraph_generator.rb
```diff
@@ -87,6 +88,17 @@ def devicegraph(volumes, initial_graph, disk_analyzer)
 
     protected
 
+      # Modifies planned volumes adding an encryption password.
+      # Only volumes that can_live_on_logical_volume could have
+      # an encryption password.
+      # @param volumes [PlannedVolumesList]
+      def add_encryption_password(volumes)
```
joseivanlopez: It is a bit rare to use bang method here due to it does not modify the state of the object. However, it would be consistent if we use that in other places. 

And about returning the list of volumes, I agree with you, in this case it is better to return whatever other thing.

src/lib/y2storage/proposal/devicegraph_generator.rb
joseivanlopez: And what about encrypted LVM? It would be also necessary ?

src/lib/y2storage/proposal_settings.rb
joseivanlopez: In fact, I used this name for consistency with method ProposalSetting#use_lvm.

src/lib/y2storage/proposal/devicegraph_generator.rb
ancorgs: Strictly speaking, I guess so. But that would probably mean three contexts with the same expectations. A little bit too much. :-)  I guess we can skip it in order to get shorter output. If LVM and encrypted are both equal, I guess everybody can guess that LVM+encrypt is also de same.

src/lib/y2storage/proposal_settings.rb
ancorgs: Which, IMO, should be called `use_lvm?` :wink:

joseivanlopez: Yes, then what is it better? Rename use_encryption as encrypt? or rename both, use_lvm and use_encryption to question mark versions?

src/lib/y2storage/proposal/devicegraph_generator.rb
joseivanlopez: Ok, perfect.

test/support/proposed_partitions_examples.rb
```diff
@@ -32,7 +32,7 @@
 
   it "requires /boot to be in the system disk out of LVM" do
```
ancorgs: I would also adapt the message here. Something
```ruby
it "requires /boot to be in the system disk out of LVM and not encrypted" do
```
or
```ruby
it "requires /boot to be a non-encrypted partition in the system disk" do
```

test/support/proposed_partitions_examples.rb
```diff
@@ -48,7 +48,7 @@
   end
 
   it "requires it to be out of LVM" do
```
ancorgs: Same than above, I would adapt the text to mention encryption.

PR #175, opened 2017-03-07
Allow IS units for disk size and without spaces

joseivanlopez:
PBI https://trello.com/c/o9LkrKGc/525-3-read-setting-from-control-file

Disk size values read from control file can be formatted as "10MB". 

This PR extends DiskSize#parse to allow IS units and without spaces.

### DIFF
--- a/src/lib/y2storage/disk_size.rb
+++ b/src/lib/y2storage/disk_size.rb
@@ -37,6 +37,7 @@ module Y2Storage
     include Comparable
 
     UNITS = ["B", "KiB", "MiB", "GiB", "TiB", "PiB", "EiB", "ZiB", "YiB"]
+    # International System units
     IS_UNITS = ["KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"]
     UNLIMITED = "unlimited"
 
@@ -80,8 +81,8 @@ module Y2Storage
 
       # Create a DiskSize from a parsed string.
       # @param str [String]
-      # @param legacy_units [Boolean] if true, IS units are considered
-      # base 2 units, that is MB is the same than MiB.
+      # @param legacy_units [Boolean] if true, International System units
+      # are considered as base 2 units, that is, MB is the same than MiB.
       #
       # Valid format:
       #
@@ -106,7 +107,8 @@ module Y2Storage
       def parse(str, legacy_units: false)
         str = sanitize(str)
         return DiskSize.unlimited if str == UNLIMITED
-        DiskSize.new str_to_bytes(str, legacy_units: legacy_units)
+        bytes = str_to_bytes(str, legacy_units: legacy_units)
+        DiskSize.new(bytes)
       end
 
       alias_method :from_s, :parse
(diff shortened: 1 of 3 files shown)


test/disk_size_test.rb
```diff
@@ -208,78 +208,67 @@
...
+
     it "should accept \"unlimited\"" do
       expect(described_class.parse("unlimited").to_i).to be == -1
     end
+
     it "should accept \"unlimited\" with surrounding whitespace" do
       expect(described_class.parse("  unlimited ").to_i).to be == -1
     end
+
+    it "should accept IS units" do
```
ancorgs: Please, be more explicit about what "IS" means.

test/disk_size_test.rb
```diff
@@ -208,78 +208,67 @@
...
+      expect(described_class.parse("10 MB").to_i).to eq(10 * 1000**2)
+    end
+
+    it "should work with legacy units" do
```
ancorgs: In general, this RSpec file is not very descriptive (is not very RSpec-like), but this test is even worse. I would have something like
```ruby
context "when using the legacy_unit flag" do
  it "considers international system units to be power of two" do
  end

  it "reads units that are power of two in the usual way" do
  end
end
```

src/lib/y2storage/disk_size.rb
```diff
@@ -105,6 +79,10 @@ def zero
       end
 
       # Create a DiskSize from a parsed string.
+      # @param str [String]
+      # @param legacy_units [Boolean] if true, IS units are considered
+      # base 2 units, that is MB is the same than MiB.
```
ancorgs: please use "international system" instead of "IS"

src/lib/y2storage/disk_size.rb
```diff
@@ -118,40 +96,62 @@ def zero
...
-        str.strip!
+      def parse(str, legacy_units: false)
+        str = sanitize(str)
         return DiskSize.unlimited if str == UNLIMITED
-        size_str, unit = str.split(/\s+/)
-        raise ArgumentError, "Bad number: #{size_str}" if size_str !~ /^\d+\.?\d*$/
-        size = size_str.to_f
-        return DiskSize.new(size) if unit.nil?
-        DiskSize.new(size * unit_multiplier(unit))
+        DiskSize.new str_to_bytes(str, legacy_units: legacy_units)
```
ancorgs: According to SUSE's style guide, the parenthesis should not be omitted. https://github.com/SUSE/style-guides/blob/master/Ruby.md#methods-1

test/disk_size_test.rb
```diff
@@ -208,78 +208,67 @@
...
     end
+
     it "should work with just an integer" do
       expect(described_class.parse("0").to_i).to be == 0
     end
+
+    it "should work without spaces" do
+      expect(described_class.parse("10MiB").to_i).to eq(10 * 1024**2)
+    end
+
```
ancorgs: Only three of these four possible combinations are tested:
 - With/without dot (decimals)
 - With/without spaces between number and unit

NP: I would prefer to see the four "it" blocks all together.

aschnell:
It is "SI", not "IS" (see https://en.wikipedia.org/wiki/International_System_of_Units).

PR #176, opened 2017-03-08
Rename IS_UNITS as SI_UNITS

joseivanlopez:


jreidinger:
sorry, but this is wrong. It is really SI units see https://en.wikipedia.org/wiki/International_System_of_Units

src/lib/y2storage/disk_size.rb
```diff
@@ -38,7 +38,7 @@ class DiskSize
 
     UNITS = ["B", "KiB", "MiB", "GiB", "TiB", "PiB", "EiB", "ZiB", "YiB"]
     # International System units
```
imobachgs: np: I would fix this comment to read like `International System of Units (SI)` so the renaming is more clear.

jreidinger:
my fault i see it other way :)

src/lib/y2storage/disk_size.rb
joseivanlopez: Ok, thanks.

PR #179, opened 2017-03-17
Classes for proposed partitions and lvs

joseivanlopez:
WIP. See [here](https://gist.github.com/joseivanlopez/4b1f1091d212b10b0970cdd691ea2f2e) for details.


test/proposal/space_maker_test.rb
```diff
@@ -44,29 +44,30 @@ def probed_partition(name)
...
     subject(:maker) { described_class.new(fake_devicegraph, analyzer, lvm_helper, settings) }
 
     context "if the only disk is not big enough" do
       let(:scenario) { "empty_hard_disk_mbr_50GiB" }
-      let(:vol1) { planned_vol(mount_point: "/1", type: :ext4, desired: 60.GiB) }
+      let(:partition1) { proposed_partition(mount_point: "/1", type: :ext4, disk_size: 60.GiB) }
 
       it "raises a NoDiskSpaceError exception" do
-        expect { maker.provide_space(volumes) }
+        expect { maker.provide_space(proposed_partitions, partitions_to_keep: keep) }
```
ancorgs: Has `partitions_to_keep` become mandatory or something like that? In this new version of the test it looks more relevant than before (it used to be omitted when not relevant).

test/proposal/space_maker_test.rb
```diff
@@ -303,57 +303,58 @@ def probed_partition(name)
...
+      let(:proposed_partitions) do
+        [
+          proposed_partition(mount_point: "/1", type: :ext4, disk_size: 60.GiB),
+          proposed_partition(mount_point: "/2", type: :ext4, disk_size: 300.GiB)
+        ]
+      end
+      let(:keep) { ["/dev/sda6", "/dev/sda2"] }
+
+      before do
+        skip("Do these tests make sense after adding partitions_to_keep parameter? ")
```
ancorgs: I guess many of the tests should be moved to the caller, making sense that it calls `SpaceMaker` with the right `partitions` and `keep` arguments.

src/lib/y2storage/planned_volumes_list.rb
```diff
@@ -23,6 +23,8 @@
 
 require "yast"
 require "y2storage/planned_volume"
+require "y2storage/proposal/proposed_partition"
+require "y2storage/proposal/proposed_lv"
```
ancorgs: I don't see a reason for these two requires. The rest of the diff for this class is all red, so just removing code, no new code using these classes.

src/lib/y2storage/proposal.rb
```diff
@@ -26,6 +26,8 @@
 require "y2storage/proposal/exceptions"
 require "y2storage/proposal/volumes_generator"
 require "y2storage/proposal/devicegraph_generator"
+require "y2storage/proposal/proposed_partition"
+require "y2storage/proposal/proposed_lv"
```
ancorgs: same than above. Not sure why these new requires are here

src/lib/y2storage/proposal/devicegraph_generator.rb
```diff
@@ -110,19 +70,74 @@ def add_encryption_password!(volumes)
       # @param space_maker [SpaceMaker]
       #
       # @return [::Storage::Devicegraph]
-      def provide_space(no_lvm_volumes, space_maker)
-        if settings.use_lvm
-          provide_space_lvm(no_lvm_volumes, space_maker)
+      def provide_space(volumes, initial_graph, disk_analyzer, target: nil)
```
ancorgs: This is a very misleading name for the method, isn't it?

src/lib/y2storage/proposal/devicegraph_generator.rb
```diff
@@ -110,19 +70,74 @@ def add_encryption_password!(volumes)
...
+        proposed_partitions = proposed_partitions(volumes, target: target)
+        proposed_lvs = proposed_lvs(volumes, target: target)
+        reused_partitions = reused_partitions(volumes)
+
+        lvm_helper = LvmHelper.new(proposed_lvs, encryption_password: settings.encryption_password)
+        space_maker = SpaceMaker.new(initial_graph, disk_analyzer, lvm_helper, settings)
+
+        result = if settings.use_lvm
```
ancorgs: When assigning the result of an `if` I find this indentation more clear:
```ruby
result =
  if whatever
    calculate_if
  else
    calculate_if_not
  end
```

ancorgs:
I have to admit that at some point my attention level decreased, but this looks like several steps in the right direction. 👍

ancorgs:
Closed in favor of #245

PR #180, opened 2017-03-21
Added to_s in ProposalSettings

schubi2:


src/lib/y2storage/proposal_settings.rb
```diff
@@ -127,6 +127,27 @@ def self.new_for_current_product
...
+      text += "  root_max_disk_size: #{root_max_disk_size}\n"
+      text += "  root_space_percent: #{root_space_percent}\n"
+      text += "  btrfs_increase_percentage: #{btrfs_increase_percentage}\n"
+      text += "  min_size_to_use_separate_home: #{min_size_to_use_separate_home}\n"
+      text += "  btrfs_default_subvolume: #{btrfs_default_subvolume}\n"
+      text += "  root_subvolume_read_only: #{root_subvolume_read_only}\n"
+      text += "  home_min_disk_size: #{home_min_disk_size}\n"
+      text += "  home_max_disk_size: #{home_max_disk_size}"
+    end
+
```
ancorgs: Using `+=` is highly inefficient here. You are creating a lot of copies of the string than will then be discarded and garbage-collected. I would go for
```ruby
text = "bla bla\n" \
  "ble ble\n" \
  "blu blu"
```

PR #182, opened 2017-03-22
Adapt prepdisk for update

joseivanlopez:
PBI: https://trello.com/c/JrA9u4mp/531-2-adapt-inst-prepdisk-for-upgrade

### DIFF
--- a/test/clients/inst_prepdisk_test.rb
+++ b/test/clients/inst_prepdisk_test.rb
@@ -39,17 +39,40 @@ describe Y2Storage::Clients::InstPrepdisk do
       allow(storage).to receive(:prepend_rootprefix).with("/proc").and_return "/dest/proc"
       allow(storage).to receive(:prepend_rootprefix).with("/sys").and_return "/dest/sys"
       allow(Yast::SCR).to receive(:Execute).and_return(true)
+      allow(Yast::Mode).to receive(:update).and_return(mode == :update)
     end
 
-    it "uses the destination directory to mount and prepare the result" do
-      expect(storage).to receive(:rootprefix=).with("/dest")
-      client.run
+    context "in installation mode" do
+      let(:mode) { :installation }
+
+      it "uses the destination directory to mount and prepare the result" do
+        expect(storage).to receive(:rootprefix=).with("/dest")
+        client.run
+      end
+
+      it "commits all libstorage pending changes" do
+        expect(storage).to receive(:calculate_actiongraph)
+        expect(storage).to receive(:commit)
+        client.run
+      end
     end
 
-    it "commits all libstorage pending changes" do
-      expect(storage).to receive(:calculate_actiongraph)
-      expect(storage).to receive(:commit)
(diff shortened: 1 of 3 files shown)


test/clients/inst_prepdisk_test.rb
```diff
@@ -39,17 +39,36 @@
...
       allow(storage).to receive(:prepend_rootprefix).with("/sys").and_return "/dest/sys"
       allow(Yast::SCR).to receive(:Execute).and_return(true)
+      allow(Yast::Mode).to receive(:update).and_return(mode == :update)
     end
 
-    it "uses the destination directory to mount and prepare the result" do
-      expect(storage).to receive(:rootprefix=).with("/dest")
-      client.run
+    context "in normal mode" do
+      let(:mode) { :normal }
```
ancorgs: Not sure if checking for `normal` mode makes much sense for this client. I would have expected `installation` and `update` to be tested, but `normal`...

test/clients/inst_prepdisk_test.rb
```diff
@@ -39,17 +39,36 @@
...
 
-    it "commits all libstorage pending changes" do
-      expect(storage).to receive(:calculate_actiongraph)
-      expect(storage).to receive(:commit)
-      client.run
+    context "in update mode" do
+      let(:mode) { :update }
+
+      it "does not commit anything" do
+        expect(client).not_to receive(:commit)
```
ancorgs: `InstPrepdisk#commit` is a protected method. I don't see a good reason to mention it in the test. Checking for `StorageManager#commit` should be enough.

test/clients/inst_prepdisk_test.rb
```diff
@@ -39,17 +39,36 @@
...
+      allow(Yast::Mode).to receive(:update).and_return(mode == :update)
     end
 
-    it "uses the destination directory to mount and prepare the result" do
-      expect(storage).to receive(:rootprefix=).with("/dest")
-      client.run
+    context "in normal mode" do
+      let(:mode) { :normal }
+
+      it "uses the destination directory to mount and prepare the result" do
```
ancorgs: I miss a "does not changes libstorage root prefix" as counterpart for this in the update mode.

test/clients/inst_prepdisk_test.rb
joseivanlopez: Ok, I supposed that `normal` means `installation`. Imo just explained me the differences 👍

test/clients/inst_prepdisk_test.rb
joseivanlopez: Ok.

test/clients/inst_prepdisk_test.rb
joseivanlopez: Yes, you are right. We can test it only looking at public methods.

ancorgs:
It would be nice to also have an entry in the changelog. Version bump not needed

PR #184, opened 2017-03-29
Initial implementation of the Ruby wrapper

ancorgs:
First steps for https://trello.com/c/1gvNrbFQ/546-8-wrapper-based-api

Includes a wrapper for all the currently used Storage classes. In those classes, the methods that has been added or have changed are documented, but most of the methods that are a direct forward to libstorage lack documentation at this moment.

Includes temporary hack in `Y2Storage::StorageManager` to enable usage of both APIs.

It's probably easier to understand if reviewed by commit.


src/lib/y2storage/btrfs_subvolume.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    storage_forward :id
+    storage_forward :top_level?
+    storage_forward :top_level_btrfs_subvolume, as: "BtrfsSubvolume"
+    storage_forward :path
+    storage_forward :nocow?
+    storage_forward :nocow=
+    storage_forward :default_btrfs_subvolume?
+    storage_forward :default_btrfs_subvolume=
+    storage_forward :create_btrfs_subvolume, as: "BtrfsSubvolume"
```
jreidinger: I expect it will be documented in future, right?

ancorgs: Right. There is a task for it in the PBI. But that's a work we can split as soon as we merge this.

src/lib/y2storage/device.rb
```diff
@@ -0,0 +1,98 @@
...
+    storage_forward :displayname
+    storage_forward :detect_resize_info
+    storage_forward :remove_descendants
+    storage_forward :userdata
+    storage_forward :userdata=
+
+    # Checks where the device is a disk
+    #
+    # @return [Boolean] true if this is a Y2Storage::Disk
+    def disk?
```
jreidinger: I am not sure if this methods are worth it. What is its advantage? it looks like common ruby methods to get child class.

src/lib/y2storage/devicegraph.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    # @return [Array<Filesystems::BlkFilesystem>]
+    def blk_filesystems
+      Filesystems::BlkFilesystem.all(self)
+    end
+
+    # @return [Array<Filesystem::Nfs>]
+    def nfs_mounts
+      Filesystems::Nfs.all(self)
+    end
```
jreidinger: hmm, this API is a bit inconsistent for me. Why not having something like `filesystems.with(type: Filesystems::Nfs)` I worry this can increase a lot when more filesystem types are added,

src/lib/y2storage/disk.rb
```diff
@@ -0,0 +1,101 @@
...
+module Y2Storage
+  # A physical disk device
+  #
+  # This is a wrapper for Storage::Disk
+  class Disk < Partitionable
+    include StorageClassWrapper
+    wrap_class Storage::Disk
+
+    storage_forward :rotational
+    alias_method :rotational?, :rotational
```
jreidinger: maybe storage_forward can support this in-place? like `storage_forward :rotational, alias: :rotational?`

src/lib/y2storage/disk.rb
```diff
@@ -0,0 +1,101 @@
...
+    #   end
+    #   empty_disk.partition_table # Not longer there
+    def as_not_empty
+      fake_ptable = nil
+      fake_ptable = create_partition_table(preferred_ptable_type) unless has_children
+
+      yield
+    ensure
+      remove_descendants if fake_ptable
+    end
```
jreidinger: I am not sure if this is good idea, as it create gpt table, even if there is none, which can cause strange results. Still I think that we should use non-existing partition table, that raise exception if not used properly. E.g. what happen if I set boot flag on this fake ptable?

src/lib/y2storage/disk.rb
```diff
@@ -0,0 +1,101 @@
...
+    #   end
+    #   lvm_pv_disk.partition_table # Still the same
+    #
+    # @example With a completely empty disk
+    #   empty_disk.as_not_empty do
+    #     empty_disk.partition_table # => a temporary PartitionTable
+    #   end
+    #   empty_disk.partition_table # Not longer there
+    def as_not_empty
+      fake_ptable = nil
```
jreidinger: I think this line is not needed.

src/lib/y2storage/fake_device_factory.rb
```diff
@@ -276,22 +278,22 @@ def create_partition_table(parent, args)
       log.info("#{__method__}( #{parent}, #{args} )")
       disk_name = parent
       ptable_type = str_to_ptable_type(args)
-      disk = ::Storage::Disk.find_by_name(@devicegraph, disk_name)
+      disk = Disk.find_by_name(@devicegraph, disk_name)
       ptable = disk.create_partition_table(ptable_type)
-      if ::Storage.msdos?(ptable) && @mbr_gap
-        ::Storage.to_msdos(ptable).minimal_mbr_gap = @mbr_gap.size
+      if ptable.respond_to?(:minimal_mbr_gap) && @mbr_gap
```
jreidinger: this is in fact wrong, it should be `ptable.respond_to?(:minimal_mbr_gap=)`

src/lib/y2storage/fake_device_factory.rb
```diff
@@ -398,29 +400,29 @@ def create_partition(parent, args)
...
+      blk_device = BlkDevice.find_by_name(@devicegraph, parent)
+      file_system = blk_device.create_blk_filesystem(fs_type)
+      assign_file_system_params(file_system, fs_param)
       parent
     end
 
+    def assign_file_system_params(file_system, fs_param)
+      ["mount_point", "label", "uuid", "fstab_options"].each do |param|
+        value = fs_param[param]
+        file_system.send(:"#{param}=", value) if value
```
jreidinger: in general I suggest to use public_send instead of send, as send is a bit hacky

src/lib/y2storage/fake_device_factory.rb
```diff
@@ -566,7 +556,7 @@ def create_lvm_lv(parent, args)
       size = args["size"] || DiskSize.zero
       raise ArgumentError, "\"size\" missing for lvm_lv #{lv_name}" unless args.key?("size")
 
-      lvm_lv = parent.create_lvm_lv(lv_name, size.size)
+      lvm_lv = parent.create_lvm_lv(lv_name, size)
```
jreidinger: just to be sure, shouldn't it be `size.to_i`?

src/lib/y2storage/fake_device_factory.rb
```diff
@@ -581,7 +571,7 @@ def create_lvm_lv_stripe_parameters(lvm_lv, args)
       lvm_lv.stripes = stripes if stripes > 0
 
       stripe_size = args["stripe_size"] || DiskSize.zero
-      lvm_lv.stripe_size = stripe_size.size if stripe_size.size > 0
+      lvm_lv.stripe_size = stripe_size if stripe_size.to_i > 0
```
jreidinger: same here? or you simply hold object and not integer?

src/lib/y2storage/device.rb
ancorgs: I added them for convenience during queries. To shorten relatively common checks like:
```ruby
a = filesystem.blk_devices.select(&:disk?)
b= another_filesystem.ancestors.select { |a| a.disk? || a.partition? }
```
That's much readable than
```ruby
a = filesystem.blk_devices.select { |a| a.is_a?(Y2Storage::Disk) }
b= another_filesystem.ancestors.select { |a| is_a?(Y2Storage::Disk) || is_a?(Y2Storage::Partition) }
```

src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -0,0 +1,90 @@
...
+      #
+      # @return [BtrfsSubvolume] nil if the subvolume is not defined or makes no
+      #   sense for this filesystem
+      def top_level_btrfs_subvolume
+        return nil unless supports_btrfs_subvolumes?
+
+        begin
+          storage_subvol = Storage.to_btrfs(to_storage_value).top_level_btrfs_subvolume
+          return nil unless storage_subvol
+        rescue Storage::Exception
```
jreidinger: this looks a bit strange, does libstorage really throw such exception and do not allow to query it?

src/lib/y2storage/filesystems/nfs.rb
```diff
@@ -0,0 +1,23 @@
+require "y2storage/storage_class_wrapper"
+require "y2storage/filesystems/base"
+
+module Y2Storage
+  module Filesystems
+    # Class to represent a NFS mount.
+    #
+    # The class does not provide functions to change the server or path since
+    # that would create a completely different filesystem.
```
jreidinger: so it is in fact Immutable or does it allow any action?

src/lib/y2storage/devicegraph.rb
ancorgs: Of course, you can still do this and get exactly the same result
```ruby
filesystems.select {|fs| fs.type.is?(:nfs) }
```
But since the NFS mounts are quite a particular case that usually has its custom logic all along YaST, I though having convenience methods to fetch them would be good.

So far, we only have two subclasses for `Filesystem::Base` - `Filesystem::BlkFilesystem` (for all  local filesystems) and `Filesystem::Nfs`. I don't foresee we will add many more, to be honest. And I find convenient to be able to fetch them via a direct method. They are separate subclasses for good reasons and making it obvious is harmless IMHO.

src/lib/y2storage/disk.rb
ancorgs: Maybe. It sounds interesting. The Ruby hacker within me want to implement it, but I'm not sure about the benefits. For documentation purposes it's probably better to explicitly use `alias_method` and there is no need to add another level of magic just to save a few characters.

src/lib/y2storage/partition_tables/msdos.rb
```diff
@@ -0,0 +1,17 @@
+require "y2storage/storage_class_wrapper"
+require "y2storage/partition_tables/base"
+
+module Y2Storage
+  module PartitionTables
+    # A MBR partition table
+    #
+    # This is a wrapper for Storage::Msdos
+    class Msdos < Base
+      include StorageClassWrapper
```
jreidinger: I see a lot of repetition there for this module, maybe better way is to add it to Base or have new class StorageBased which contain it. Or is it used somewhere without base or vice versa? for me it looks coupled together.

src/lib/y2storage/partition_tables/partition_slot.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      # rubocop:disable Metrics/CyclomaticComplexity
+      def flags_string
+        flags = ""
+        flags << "P" if primary_slot
+        flags << "p" if primary_possible
+        flags << "E" if extended_slot
+        flags << "e" if extended_possible
+        flags << "L" if logical_slot
+        flags << "l" if logical_possible
```
jreidinger: maybe use "?" version here?

src/lib/y2storage/fake_device_factory.rb
ancorgs: No, this code is already using the Ruby wrapper, which means we operate directly with `Y2Storage::DiskSize` objects. So we always consistently use `DiskSize` objects instead of the old mixture of `Fixnum` for `Storage` and `DiskSize` for "high level" code.

Note: `size.to_i` would also work in this particular case, but there is no need to convert disk sizes to bytes any more.

src/lib/y2storage/partition_tables/partition_slot.rb
jreidinger: btw there can be nicer API, as this is not not 6 flags, but in fact only three ternany flags. So it can be something like `primary # => :set, :available, :unavailable` ? or it can be simulated with true, false, nil, where true is set, false not set, but available, and nil for not available.

src/lib/y2storage/disk.rb
jreidinger: well, maybe alias is not good name, maybe ruby_name is better. What I would like to see is, no `rotational` method and only `rotational?` one

src/lib/y2storage/partitionable.rb
```diff
@@ -0,0 +1,43 @@
...
+    storage_forward :range=
+    storage_forward :default_partition_table_type, as: "PartitionTables::Type"
+    storage_forward :possible_partition_table_types, as: "PartitionTables::Type"
+    storage_forward :create_partition_table, as: "PartitionTables::Base"
+    storage_forward :partition_table, as: "PartitionTables::Base"
+    storage_forward :topology
+
+    storage_class_forward :all, as: "Partitionable"
+
+    # Minimal grain of the partitionable
```
jreidinger: this should be extended, as least link what is in fact grain?

src/lib/y2storage/region.rb
```diff
@@ -0,0 +1,60 @@
...
+    wrap_class Storage::Region
+
+    # @raise [Exception] when comparing Regions with different block_sizes
+    storage_forward :==
+    # @raise [Exception] when comparing Regions with different block_sizes
+    storage_forward :!=
+    # @raise [Exception] when comparing Regions with different block_sizes
+    storage_forward :>
+    # @raise [Exception] when comparing Regions with different block_sizes
+    storage_forward :<
```
jreidinger: well, this definitively need more documentation, like what is compared? if size, then exception is quite bogus as it is easy to compute ( blocksize * length )

src/lib/y2storage/region.rb
```diff
@@ -0,0 +1,60 @@
...
+    storage_forward :end
+    storage_forward :start=
+    storage_forward :length=
+    storage_forward :adjust_start
+    storage_forward :adjust_length
+    storage_forward :block_size, as: "DiskSize"
+    storage_forward :block_size=
+
+    def inspect
+      "<Region #{start} - #{self.end}>"
```
jreidinger: I think at least block size would be nice to have, as it can raise exceptions.

jreidinger: and btw it is in fact just `"<Region #{show_range}>"`

src/lib/y2storage/storage_class_wrapper.rb
```diff
@@ -0,0 +1,258 @@
...
+      #
+      #   blk_device = Storage::find_by_name(a_devicegraph, "/dev/sda2")
+      #   blk_device.class #=> Storage::BlkDevice
+      #   # The user should never need to call this constructor
+      #   y2_device = Y2Storage::BlkDevice.downcasted_new(blk_device)
+      #   y2_device.class #=> Y2Storage::Partition
+      #
+      def wrap_class(storage_class, downcast_to: nil)
+        @storage_class = storage_class
+        @downcast_class_names = Array(downcast_to).compact
```
jreidinger: well, this looks a bit strange, why default value of `downcast_to:` is not `[]` and then simply have `Array(downcast_to)` without need to compact?

jreidinger: Especially when documentation do not mention that nil can be passed.

src/lib/y2storage/storage_class_wrapper.rb
```diff
@@ -0,0 +1,258 @@
...
+          StorageClassWrapper.forward(to_storage_value, method, modifiers, *args)
+        end
+      end
+
+      # Equivalent to #storage_forward used to define class methods instead of
+      # instance ones.
+      #
+      # @see #storage_forward
+      def storage_class_forward(method, as: nil, raise_errors: [])
+        modifiers = { as: as, raise_errors: raise_errors }
```
jreidinger: this is a bit inconvinience. It have named params that is later added to hash. Why not have params as hash from start?

like
```
def storage_class_forward(method, modifiers = {})
  modifiers[:raise_errors] ||= []
 ....
```

src/lib/y2storage/storage_class_wrapper.rb
```diff
@@ -0,0 +1,258 @@
...
+      # Alternative constructor used internally by this module in order to get
+      # fully downcasted objects.
+      #
+      # @param storage object to be wrapped
+      def downcasted_new(object)
+        @downcast_class_names.each do |class_name|
+          underscored = StorageClassWrapper.underscore(class_name.split("::").last)
+          check_method = :"#{underscored}?"
+          cast_method = :"to_#{underscored}"
+          next unless Storage.send(check_method, object)
```
jreidinger: same here, public send is better.

src/lib/y2storage/storage_class_wrapper.rb
```diff
@@ -0,0 +1,258 @@
...
+      wrapper_class_name = modifiers[:as]
+      raise_errors = modifiers[:raise_errors] || []
+
+      processed_args = processed_storage_args(*args)
+      result = storage_object.send(method, *processed_args)
+      processed_storage_result(result, class_for(wrapper_class_name))
+    rescue Storage::WrongNumberOfChildren
+      raise_errors.include?("WrongNumberOfChildren") ? raise : nil
+    rescue Storage::DeviceHasWrongType
+      raise_errors.include?("DeviceHasWrongType") ? raise : nil
```
jreidinger: this is not documented that only subset of exceptions are allowed. What about
```
rescue Exception => exception
  # some logging
  exception_name = exception.class.name.split("::").last
  raise_errors.include?(exception_name) ? raise : nil
end
```

This way is support all exceptions. If you want to avoid collision with ruby errors, you can add
```
raise unless exception.class.name.start_with?("Storage::")
```
so it will catch only storage exceptions.
If storage have proper exception tree, then you can even just catch that base exception.

src/lib/y2storage/storage_class_wrapper.rb
```diff
@@ -0,0 +1,258 @@
...
+      # @param storage object to be wrapped
+      def downcasted_new(object)
+        @downcast_class_names.each do |class_name|
+          underscored = StorageClassWrapper.underscore(class_name.split("::").last)
+          check_method = :"#{underscored}?"
+          cast_method = :"to_#{underscored}"
+          next unless Storage.send(check_method, object)
+
+          klass = StorageClassWrapper.class_for(class_name)
+          return klass.downcasted_new(Storage.send(cast_method, object))
```
jreidinger: bumm, class_for can return nil. In general I suggest to raise error if class_for does not find class.

src/lib/y2storage/storage_class_wrapper.rb
```diff
@@ -0,0 +1,258 @@
...
+    end
+
+    def self.object_for(wrapper_class, storage_object)
+      if wrapper_class.respond_to?(:downcasted_new)
+        wrapper_class.downcasted_new(storage_object)
+      else
+        wrapper_class.new(storage_object)
+      end
+    end
+  end
```
jreidinger: In general I do not like number of methods in this module. Problem is that it add bunch of public available methods, which looks wrong for me. I think we need some encapsulation here.

src/lib/y2storage/storage_enum_wrapper.rb
```diff
@@ -0,0 +1,131 @@
...
+      #
+      #   pri1 = Y2Storage::PartitionType.find(:primary)
+      #   pri2 = Y2Storage::PartitionType.find(Storage::PartitionType_PRIMARY)
+      #   pri1 == pri2 # => true
+      #   pri1.is?(:primary) # => true
+      #   pri2.to_sym # => :primary
+      def wrap_enum(storage_enum, names: [])
+        @storage_enum = storage_enum
+
+        mapping = names.map { |s| [Object.const_get("#{storage_enum}_#{s.upcase}"), s.to_sym] }
```
jreidinger: this is one part I do not like. It can very easily go out of sync. I think it pays of if we do here more introspection ourself and of course log it.

So something like
```
def wrap_enum(storage_enum)
  @storage_enum = storage_enum

  constants = Object.constants.select { |c| c.to_s.start_with?(storage_enum+"_") }.map(&:to_s)
  mapping = constants.map { |c| [Object.const_get(c), c.sub(/#{storage_enum}_/, "").to_sym] }

  @storage_symbols = Hash[mapping] # flatten is in fact not needed, hash can be constructed from list of pairs
```

src/lib/y2storage/storage_enum_wrapper.rb
```diff
@@ -0,0 +1,131 @@
...
+      #   end
+      #
+      #   Y2Storage::PartitionType.all.each do |type|
+      #     puts "Name: #{type.to_s} -> Internal libstorage value: #{type.to_i}"
+      #   end
+      #
+      #   pri1 = Y2Storage::PartitionType.find(:primary)
+      #   pri2 = Y2Storage::PartitionType.find(Storage::PartitionType_PRIMARY)
+      #   pri1 == pri2 # => true
+      #   pri1.is?(:primary) # => true
```
jreidinger: it would be nice to also provide methods like `pri1.primary?` it should be quite easy as you know names.

src/lib/y2storage/storage_enum_wrapper.rb
```diff
@@ -0,0 +1,131 @@
...
+      #       include StorageEnumWrapper
+      #       wrap_enum "Storage::PartitionType", names: [:primary, :extended, :logical]
+      #     end
+      #   end
+      #
+      #   Y2Storage::PartitionType.all.each do |type|
+      #     puts "Name: #{type.to_s} -> Internal libstorage value: #{type.to_i}"
+      #   end
+      #
+      #   pri1 = Y2Storage::PartitionType.find(:primary)
```
jreidinger: in fact you can also define class method for it like `Y2Storage::PartitionType.primary` which returns same and is more convenient.

jreidinger: another options is to have it as constant like `Y2Storage::PartitionType::PRIMARY` to it is predefined and shared.

src/lib/y2storage/device.rb
jreidinger: to be honest this from my POV goes against idea of polymorphysm.
for you examples, I prefer to have something like
```
a = filesystem.disks
b = another_filesystem.ancestors.disks_or_partitions
```

I just feel a bit wrong to so heavily depends on kind of object. So if I transfer to common OOP example with animal these examples for me is like:

1. give me all animals, I do not mind its kind and then ask each animal if it is dog

2. give me all animals, I do not mind its kind and then look if it is dog

3. give me all dogs


I see this kind of methods also in past, but it is a bit different implemented. in base it was set to false and then in specialization to true.

What I see wrong in this approach? It is hard to extend. If new object is introduced, that act as BlkDevice, it cannot modify this class easily to be same of other kids.

Another part is that it adds to each child amount of public methods. This make introspection hard ( try now Disk.public__instance_methods.size ) what it returns. And public methods correspond for me also with responsibility. If we go back to animals example. You require that all animals can answers to questions like:
```
# dog
Are you cat? No.
Are you horse? No.
Are you dog? Yes.
Are you bird? No.
```
so for me it is also question about responsibilities. Also it breaks for me principle of "Tell, not ask" https://martinfowler.com/bliki/TellDontAsk.html . my approach is to tell, give me dogs. your approach for me looks like ask if you are dog and then act.

I hope I do not confuse you. Ask if something is not clear.

src/lib/y2storage/devicegraph.rb
jreidinger: ok, fine. If it is prominent kid, it is fine.

src/lib/y2storage/fake_device_factory.rb
jreidinger: ok, fine. I am thinking about it, but change is so big, that I lost overview. So thanks for confirmation that it do what is expected.

src/lib/y2storage/disk.rb
ancorgs: Ah, ok. That makes more sense.

src/lib/y2storage/fake_device_factory.rb
ancorgs: The code as it is now will work in both cases - either if stripe_size is an integer (bytes) or if is a DiskSize object. So I would rather leave it as it is.

jreidinger: fine for me. Btw maybe having method `zero?` make sense here. I do not see better name for "empty size", but it looks like quite common case to check it.

src/lib/y2storage/filesystems/blk_filesystem.rb
ancorgs: I'm not 100% completely sure what the final API for libstorage will be regarding subvolumes and/or whether it's possible to have a Btrfs filesystem with no top level subvolume (not sure if subvolumes can be disabled as a feature, for example). But looking at the current libstorage code, an exception seems to be (as usual) the expected mechanism to report the search was unsuccessful, since it uses internally `get_single_parent_of_type` .

src/lib/y2storage/filesystems/nfs.rb
ancorgs: At least, you can change the mount point, I would say.

src/lib/y2storage/filesystems/blk_filesystem.rb
jreidinger: so it is not even specialized? I feel quite afraid about it as it can result in hiding real problem, I think we should at least log it, so we know what is libstorage trying to say.

src/lib/y2storage/filesystems/nfs.rb
jreidinger: OK, maybe it is worth to mention it as it is not directly seen.

src/lib/y2storage/region.rb
ancorgs: Sure. As said. That's a different task in the PBI, but we need to merge first to start distributing that part.

src/lib/y2storage/storage_class_wrapper.rb
ancorgs: Yardoc already documents the default value (which is nil) so I don't see a reason to state it again. The method is written to support just a single name or a list. That is, both are valid:
```
wrap_class Storage::Whatever, downcast_to: "SubWhatever"
wrap_class Storage::Whatever, downcast_to: ["Sub1Whatever", "Sub2Whatever"]
```
That's pretty much how old macros use to work in Rails, so I guess that's my Rails pass influencing me. :wink: But I will change it to work only with arrays and to not support single elements, so it's less confusing.

src/lib/y2storage/storage_class_wrapper.rb
ancorgs: It's for the sake of better autogenerated Yardoc documentation. Key attributes are way more clear (with their default value and everything) that hashes that must be described with text.

src/lib/y2storage/region.rb
jreidinger: well, now it is a bit hard to review it, as I do not know, what it actually do, so hard to decide if it make sense to just forward or if we should wrap it.

jreidinger: in general, I think it will be much nicer to simply add `def <=>` and include Comparable, to have full set of methods.

src/lib/y2storage/storage_class_wrapper.rb
jreidinger: no, single element is fine and valid approach for me. What I do not like is default `nil` instead of `[]` which makes more sense for me.
Doc I mean was that types are `[String, Array<String>]` without `nil` which is usually indicator for me to not pass nil there otherwise it crash.

src/lib/y2storage/storage_class_wrapper.rb
jreidinger: in fact, it does not need to be described with text. I use in past http://www.rubydoc.info/gems/yard/file/docs/Tags.md#option and it worked well.

jreidinger: how it looks rendered see e.g. http://www.rubydoc.info/github/yast/yast-yast2/Yast/CWMClass#ShowAndRun-instance_method

src/lib/y2storage/storage_enum_wrapper.rb
ancorgs: I don't like to abuse of automatically generated methods. Too much magic. What if one of the labels is `frozen` or `equal`?

src/lib/y2storage/storage_enum_wrapper.rb
ancorgs: Same as above. I don't like too much magic. What if one of the labels is `new`?

ancorgs: Well, maybe the constant would be ok.

src/lib/y2storage/storage_class_wrapper.rb
ancorgs: Well, the point here is that in general we actually don't want to catch all Storage exceptions, only the two ones that mean "there is not such object" (that would mean returning nil in the typical Ruby API). I will change the name and approach of the argument to be less confusing about what can be done regarding exceptions and what not.

jreidinger: Hmm, my concern is that it is again not so easy to extend and require changes at more places. So another idea. What about inverting logic? instead of having raise_errors parameter, we can have `no_object_exceptions: ["DeviceHasWrongType"]` and let live all exception except specified ones?
Maybe I just do not get logic there. From what I see now it looks like we would like to for specified exceptions return nil instead of letting it live, right?

src/lib/y2storage/storage_enum_wrapper.rb
ancorgs: Constants added

jreidinger:
@ancorgs nice improvements in last commits ;) less lines and nicer API 👍

src/lib/y2storage/align_policy.rb
```diff
@@ -7,6 +7,6 @@ module Y2Storage
   class AlignPolicy
     include StorageEnumWrapper
 
-    wrap_enum "Storage::AlignPolicy", names: [:align_end, :keep_end, :keep_size]
+    wrap_enum "AlignPolicy"
```
jreidinger: great, it looks nicer now. The last idea. Maybe StorageEnum can be base class for such kind of classes? so it can be 
```
class AlignPolicy < StorageEnum
   wrap_enum "AlignPolicy"
end
```

src/lib/y2storage/storage_enum_wrapper.rb
jreidinger: thanks, now it looks better and I think also in code it will look nicer.

src/lib/y2storage/disk.rb
ancorgs: Why is not needed? If the line is not there and the disk has children, it will raise `NameError` in the ensure block. Isn't it?

src/lib/y2storage/align_policy.rb
ancorgs: If you mean turning `StorageEnumWrapper` from a mixin into a class, it makes sense if you look at it in isolation. But I like the fact that the implementation is symmetric to `StorageClassWrapper` (both implemented as modules).

If you mean adding a base class in addition to the mixin (so basically a one-liner class), we could do it, but I'm not sure if it's worth it.

src/lib/y2storage/disk.rb
jreidinger: reason is that ruby define it in context even if it is only in single branch. Try it yourself in this this script

```ruby
def test
  a = 5 if false
ensure
  puts a.inspect
end

test
```

tricky part can be if create_partition_table raise if it is still defined, but as ruby do it in advance it is defined. Try yourself:

```ruby
def test
  a = (false || raise("a")) if true
ensure
  puts a.inspect
end

test
```

src/lib/y2storage/disk.rb
```diff
@@ -69,7 +69,10 @@ def usb?
     #   empty_disk.partition_table # Not longer there
     def as_not_empty
       fake_ptable = nil
-      fake_ptable = create_partition_table(preferred_ptable_type) unless has_children
+      if !has_children
+        fake_ptable = create_partition_table(preferred_ptable_type)
+        fake_ptable.freeze
```
jreidinger: 👍

src/lib/y2storage/disk.rb
ancorgs: I find that Ruby behavior to be potentially surprising for the average (non-Ruby) programmer, so I would rather leave the unneeded line there, it reduces confusion.

src/lib/y2storage/lvm_pv.rb
```diff
@@ -26,7 +26,7 @@ class LvmPv < Device
     #
     # @return [BlkDevice]
     def plain_blk_device
-      blk_device.encryption? ? blk_device.blk_device : blk_device
+      blk_device.is_a?(Encryption) ? blk_device.blk_device : blk_device
```
jreidinger: just note to discussion we have about it ( so not need for action, just showing how ask can be turn into say). For this case it make sense to have method like `plain_device` that return itself for base class and for Encryption it return blk_device

so code can look like
```
class BlkDevice
  def plain_device
     self
  end
end

class Encryption
  def plain_device
     blk_device.plain_device # in case if two encryption is possible
  end
end
```

and then this code is not needed, as you call `lvm_pv.blk_device.plain_device` when need it as plain and for above case
```
def plain_blk_devices
  blk_devices.map(&:plain_device)
end
```
or removing, as this call is already easy enough to use it when makes sense.

src/lib/y2storage/fake_device_factory.rb
ancorgs: `DiskSize.zero?` exists, of course. As mentioned, we are using `#to_i` here to ensure compatibility with old and new way of doing things.

ancorgs:
@jreidinger I made some changes. I believe this can be merged as-is to not block others. Summarizing what's fixed/postponed:

 * I removed the methods that were there to simplify queries (`disk?` and so on). Let's have a separate discussion on how do we want the query API to look like (I added explicit tasks for that in the PBI).
* In addition to the task for documenting we already have in the PBI, I added `TODO` in the particular points in which you had questions (like the meaning of "grain"). For better precision, I would like the authors of the original code to document those points.
* For the time being, let's keep `StorageClassWrapper` and `StorageEnumWrapper` to be both mixins, we can re-evaluate it in the future. Changing that will not change offered API, so let's move on.

src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -0,0 +1,87 @@
...
+      storage_forward :blk_devices, as: "BlkDevice"
+
+      # Raw (non encrypted) version of the formatted devices. If the filesystem
+      # is not encrypted, it returns the same collection that #blk_devices,
+      # otherwise it returns the original devices instead of the encryption
+      # ones.
+      #
+      # @return [Array<BlkDevice>]
+      def plain_blk_devices
+        blk_devices.map { |dev| dev.is_a?(Encryption) ? dev.blk_device : dev }
```
jreidinger: well, this is a bit confusing for me. BlkDevice have method encrypted? and plain_blk_device. So why it is not used? Even now when I know code it is not clear for me why it do not look like `blk_devices.map(&:plain_blk_device)`

src/lib/y2storage/lvm_pv.rb
```diff
@@ -0,0 +1,32 @@
...
+    storage_forward :blk_device, as: "BlkDevice"
+
+    # Raw (non encrypted) version of the device hosting the PV.
+    #
+    # If the PV is not encrypted, this is equivalent to #blk_device, otherwise
+    # it returns the original device instead of the encryption one.
+    #
+    # @return [BlkDevice]
+    def plain_blk_device
+      blk_device.is_a?(Encryption) ? blk_device.blk_device : blk_device
```
jreidinger: same here

ancorgs:
@jreidinger after some offline discussion in the Gran Canaria office I decided to take the risk of diverging from libstorage for `BlkDevice.filesystem`.

src/lib/y2storage/blk_device.rb
```diff
@@ -49,13 +52,21 @@ def encrypted?
...
     #
-    # @return [Filesystems::BlkFilesystem] nil if neither the raw device or its
-    #   encrypted version are formatted
-    def plain_blk_filesystem
-      encrypted? ? encryption.blk_filesystem : blk_filesystem
+    # @param [traverse_encryption] if set to false, the method will return nil
+    #   if the device is encrypted, no matter if the encrypted device is
+    #   formatted or not.
+    # @return [Filesystems::BlkFilesystem] nil if the device is not formatted
+    def blk_filesystem(traverse_encryption: true)
```
jreidinger: well, I do not like this name. It will break in future if there will be more devices that can be between filesystem and blk_device. so what about `traverse: true`?

ancorgs: It's intentional. I wanted to be quite explicit about it being just for the very common case of encrypted filesystems. Think about this example:
```
disk.blk_filesystem(traverse: true)
```
I don't want anybody to be under the impression that it returns a filesystem that is placed in one of the partitions. Or that it will return the filesystem directly placed on a mirror RAID using that disk.

The flag is about traversing encryption and nothing else.

jreidinger: good argument. Still I think that it is mainly for people already familiar with storage-ng as others find it not needed. What about changing to to `disk.blk_filesystem(direct_only: true)` and default to false? So it is clear that direct_only do no traverse, but do not emphasize, it traverse everything?

jreidinger: what in fact I worry, that someone really use this argument for low level access `disk.blk_filesystem(traverse_encryption: false)` and then we introduce another `disk.blk_filesystem(traverse_encryption: false, traverse_cache: false)` for low level access and we need to change it everywhere. With my proposal there are just two cases

1. common user, want smart function

2. expert hacker want low-level

ancorgs: Ok. Since direct access is somehow "expert", let's simply expose `#direct_blk_filesystem` which already have a expert-like name and redefine `#blk_filesystem` to simply do what we expect, without arguments or surprises.

PR #187, opened 2017-03-31
Device#is? + basic tests and small fixes

ancorgs:


src/lib/y2storage/disk.rb
```diff
@@ -110,14 +110,11 @@ def preferred_ptable_type
...
-      partitions.any? { |part| part.name == name }
-    end
-
-    def self.find_by_name_or_partition(devicegraph, name)
-      all(devicegraph).detect { |disk| disk.name_or_partition?(name) }
+    # @see Device#is?
+    #
+    # In this case, true if type is or contains :disk
+    def is?(types)
+      super || types_include?(types, :disk)
```
jreidinger: to be honest I do not like much implementation, as it is a lot of repetition and duplication ( in fact only interesting part of this method is `:disk` rest is just sauce. what about defining method types? which return all types of device? like
```
# base
protected def types
  [:device]
end

# mountable
protected def types
  super << :mountable
end

# and snapshot
protected def types
  super << :snapshot
end
```

if you do not mind some magic, it can be done also with constants
```
#base
TYPE = :device

#mountable
TYPE = :mountable

# snapshot
TYPE = :snapshot

# and implementation
def types
  self.class.ancestors.each_with_object([]) do |klass, result|
     klass.const_get("TYPE") if klass.const_defined?("TYPE")
  end
end
```

but in the end it is just style detail, not problem of API itself.

jreidinger:
in general I like change, only if I do, I will probably use constants for defining type as it is much more DRY.

src/lib/y2storage/device.rb
```diff
@@ -46,5 +51,49 @@ class Device
...
+    #   encryption.blk_device.is?(:disk)
+    #
+    # @example Filtering disks and partitions
+    #
+    #   something.ancestors.select { |dev| dev.is?([:disk, "partition"]) }
+    #
+    # @param types [#to_sym, Array<#to_sym>] name (or names) of the device
+    #   type, as defined in each subclass
+    # @return [Boolean]
+    def is?(_types)
```
jreidinger: thinking about it, I do not like much this sym or array and prefer `*types` so usage will be 
```
encryption.blk_device.is?(:disk)

something.ancestors.select { |dev| dev.is?(:disk, "partition") }
```

jreidinger: this way it also always gets you array, so you do not need to use `Array()`

ancorgs: Yes, using `*types` was my initial though. But it raised my "too Ruby/Rails style" alarm. :wink: But if I'm not the only one that sees that approach as the most natural one, let's go for it.

ancorgs:
@jreidinger Updated to use argument list. `StorageEnumWrapper#is?` adapted in the same way for consistency

PR #188, opened 2017-04-03
inst_prepdisk: don't access libstorage directly

ancorgs:
Manually tested (apart from the existing unit tests)

### DIFF
--- a/src/lib/y2storage/storage_manager.rb
+++ b/src/lib/y2storage/storage_manager.rb
@@ -59,25 +59,19 @@ module Y2Storage
     # @return [Proposal, nil]
     attr_reader :proposal
 
-    def_delegators :@storage, :environment, :arch, :rootprefix
+    def_delegators :@storage, :environment, :arch, :rootprefix, :prepend_rootprefix, :rootprefix=
 
     # @!method rootprefix
     #   @return [String] root prefix used by libstorage
 
-    # Sets the root prefix used by libstorage in subsequent operations
-    #
-    # @param path [String]
-    def rootprefix=(path)
-      storage.rootprefix = path
-    end
+    # @!method rootprefix=(path)
+    #   Sets the root prefix used by libstorage in subsequent operations
+    #   @param path [String]
 
-    # Prepends the current libstorage root prefix to a mountpoint if necessary
-    #
-    # @param path [String] original path (without prefix)
-    # @return [String]
-    def prepend_rootprefix(path)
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/storage_manager.rb
```diff
@@ -59,7 +59,25 @@ class StorageManager
...
+
+    # @!method rootprefix
+    #   @return [String] root prefix used by libstorage
+
+    # Sets the root prefix used by libstorage in subsequent operations
+    #
+    # @param path [String]
+    def rootprefix=(path)
+      storage.rootprefix = path
+    end
```
jreidinger: why it is not also in delegator? I do not see difference between `#rootprefix` and `#rootprefix=`

src/lib/y2storage/storage_manager.rb
```diff
@@ -59,7 +59,25 @@ class StorageManager
...
+
+    # Prepends the current libstorage root prefix to a mountpoint if necessary
+    #
+    # @param path [String] original path (without prefix)
+    # @return [String]
+    def prepend_rootprefix(path)
+      storage.prepend_rootprefix(path)
```
jreidinger: same here

src/lib/y2storage/storage_manager.rb
ancorgs: I tried and it didn't work. The tests failed. Something to do with SWIG, I assumed.

src/lib/y2storage/storage_manager.rb
ancorgs: I have to admit I didn't specifically tested this one. I just assumed the same problem than `rootprefix=`. Let me check further.

src/lib/y2storage/storage_manager.rb
jreidinger: well, checking code of delegators it call `respond_to?` so if return false it is quite strange. and method is called with send, so I do not see much ways how it can failed. do you have exact failure which is caused? just for interest.

jreidinger: at least we should document that it can happen as we can face it also in future.

ancorgs: Mmmm, or maybe it works but cannot be tested with doubles.

ancorgs: Found the culprit. It was test's fault. I will fix it.

PR #189, opened 2017-04-03
Subvol proposal

shundhammer:
- Prepend the default subvolume name to the subvol name upon creation: `var/log` -> `@/var/log`

- Don't shadow other volumes with a subvolume: Don't propose a `@/home` subvol to be mounted at `/home` when another dedicated volume (partition, LVM LV) is created for `/home´

- More unit tests

### DIFF
--- a/src/lib/y2storage/planned_subvol.rb
+++ b/src/lib/y2storage/planned_subvol.rb
@@ -127,9 +127,26 @@ module Y2Storage
       use_subvol
     end
 
+    # Create the subvolume as child of 'parent_subvol'.
+    #
+    # @param parent_subvol [::Storage::BtrfsSubvol]
+    # @param default_subvol [String] "@" or ""
+    # @param mount_prefix [String]
+    #
+    # @return [::Storage::BtrfsSubvol]
+    #
+    def create_subvol(parent_subvol, default_subvol, mount_prefix)
+      name = @path
+      name = default_subvol + "/" + path unless default_subvol.empty?
+      subvol = parent_subvol.create_btrfs_subvolume(name)
+      subvol.nocow = true if no_cow?
+      subvol.add_mountpoint(mount_prefix + @path)
+      subvol
+    end
+
     # Factory method: Create one PlannedSubvol from XML data stored as a map.
     #
-    # @return PlannedSubvol or nil if error
+    # @return [PlannedSubvol] or nil if error
     #
     def self.create_from_xml(xml)
       return nil unless xml && xml.key?("path")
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/planned_subvol.rb
```diff
@@ -127,9 +127,26 @@ def matches_arch?(target_arch = nil, &block)
...
+    #
+    # @param parent_subvol [::Storage::BtrfsSubvol]
+    # @param default_subvol [String] "@" or ""
+    # @param mount_prefix [String]
+    #
+    # @return [::Storage::BtrfsSubvol]
+    #
+    def create_subvol(parent_subvol, default_subvol, mount_prefix)
+      name = @path
+      name = default_subvol + "/" + path unless default_subvol.empty?
```
imobachgs: Just a suggestion: you can use `File.join(default_subvol, path)` to avoid problems with more than one *slash*.

src/lib/y2storage/planned_volume.rb
```diff
@@ -187,9 +187,35 @@ def create_filesystem(partition)
...
+    # @param other_mount_points [Array<String>]
+    #
+    # @return nil
+    #
+    def create_subvolumes(filesystem, other_mount_points)
       return unless filesystem.type == ::Storage::FsType_BTRFS
       return unless subvolumes?
+      parent_subvol = get_parent_subvol(filesystem)
+      prefix = filesystem.mountpoints.first
+      prefix += "/" unless prefix == "/"
```
imobachgs: You can use `File.join` and it will take care of not appending an extra `slash` if not needed.

src/lib/y2storage/planned_subvol.rb
shundhammer: Good idea actually, but:
`irb(main):001:0> ["", "foo/bar"].join("/")`
`=> "/foo/bar"`

In this case I need to retain "foo/bar"; the paths are relative to the main volume's mount point.

Otherwise it's a good idea to use join().

imobachgs: `File.join` is different from `String.join`.

```ruby
File.join("foo", "bar") #=> "foo/bar"
File.join("/", "foo", "bar") #=> "/foo/bar"
File.join("foo", "bar", "/") #=> "foo/bar/"
```

shundhammer: Duh. I used the wrong one. But the right one gives the same result:

`irb(main):007:0> File.join("", "foo/bar")`
`=> "/foo/bar"`

imobachgs:
Well, we will need to make Travis happy again :(

src/lib/y2storage/planned_volume.rb
shundhammer: In this case I do want a trailing slash, but not a double one if the main volume's mount point is "/" (which is the common case for our proposed subvolumes).

PR #192, opened 2017-04-04
Fixes and additions to the ruby wrapper

ancorgs:


src/lib/y2storage/blk_device.rb
```diff
@@ -82,6 +82,36 @@ def blk_filesystem
...
+      return nil unless blk_filesystem
+      blk_filesystem.type
+    end
+
+    # Mount point of the filesystem, if any
+    # @return [String, nil]
+    def filesystem_mountpoint
+      return nil unless blk_filesystem
+      blk_filesystem.mountpoint
+    end
```
jreidinger: well, I feel a bit wrong about these methods. Its name and prefix `filesystem_` indicate that it is in fact filesystem methods. If it is planned as shortcut, then please remember that target is SLE13 and I hope there will be new ruby so it can be called as `filesystem&.mountpoint`

src/lib/y2storage/dasd.rb
```diff
@@ -0,0 +1,50 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage/storage_class_wrapper"
+require "y2storage/partitionable"
+
+module Y2Storage
+  # A DASD (direct-access storage device), typically used in mainframes
```
jreidinger: link to dasd documentation willl be great

jreidinger:
what I am missing is test case for that fixes.

src/lib/y2storage/blk_device.rb
ancorgs: And indeed they are filesystem methods. But I don't see anything wrong with defining these delegators as shortcuts. The name is very explicit about their real nature and they are quite convenient in some scenarios. As long as `filesystem&.mountpoint` is still not a reality in all the systems I don't see any strong reason against. Maybe they would be controversial if called directly `#mounpoint`, `uuid`, etc. but...

jreidinger: well, from my POV what is wrong, that it creates high coupling between blk_device and filesystem. Previously blk_device responsibility was to return filesystem or nil if there is no such filesystem. Now it depends on their API. so now blk_device have to know that filesystem have given subset of methods. see e.g. https://thebojan.ninja/2015/04/08/high-cohesion-loose-coupling/ coupling part.

consider e.g. that method mountpoint get argument. in this case you have to change filesystem method, its caller and additionally also this method.

jreidinger: btw this logic we can get to something like `filesystem_mountpoint_label` which do
```
return nil unless blk_filesystem
mp = blk_filesystem.mountpoint
return nil unless mp
mp.label
```

ancorgs: > btw this logic we can get to something like `filesystem_mountpoint_label`

https://en.wikipedia.org/wiki/Reductio_ad_absurdum

shundhammer: The difference is that the mount point is used all over the place, whereas nobody hardly ever uses the label. So IMHO it is well justified to have a convenience function like this.

src/lib/y2storage/dasd.rb
shundhammer: Other than a Wikipedia article with little useful content, there is not much; at IBM web pages, there are various articles that touch the fringes of the topic, but I have not found anything yet that could be called a useful reference documentation or a useful tutorial. It is not the task of a class documentation to magically conjure up documentation that is neglected and absent everywhere else.

jreidinger: so on what is based our knowledge? What we use as authoritative source? notes from IBM guys in fate? or Ihno only? or if we use kernel module for dasd, maybe that module have documentation?
What if someone want to touch this class, is it his task to "magically conjure up documentation"? we should help other, even poor documentation is better then none. E.g. https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.idad400/dasda.htm is at least something.

src/lib/y2storage/blk_device.rb
jreidinger: well, still I do not like it much, as it is inconsistent. filesystem have much more methods, so why for example is not there fstab_options or mount_by or displayname ?

jreidinger: or better, why these method deserve shortcut or others not? Maybe if there is some logic in it, then API make some sense like shortcuts for ...

src/lib/y2storage/dasd.rb
ancorgs: @jreidinger Yes, that's better than nothing... but that's not the authoritative source that was used to write the class (unfortunately, there is no such link), so I see no reason to link to it in the class documentation. 

That link is simply the best you found with a google search today. Maybe tomorrow it will be dead or there would be a better source of documentation. If there is no concrete document we used as a guide, is probably up to the poor developer to search for "DASD direct-access storage device" in Internet. Sad but true, but pretending we extracted the information from a concrete place will not help much.

jreidinger: ok, if there is no such source, then it probably do not make sense.

shundhammer: @jreidinger: Did you have a look at that page you linked? Do you find that documentation useful? It just explains how hard disks work in general; this is very much the same as the documents linked from the DASD Wikepedia page; one is an ancient IBM document from 1975 (!) that also just explains on 27 pages how magnetic disks work and what cylinders, heads and sectors are, the other one from 2015 at least has enough mercy to omit that stuff, but it also does not contain any useful content, in particular no explanation what is the difference between DASDs and other hard disks; much less that even touches the topic of DASDs on Linux.

shundhammer: @jreidinger: In addition to that, I do not find this style of discussion useful, much less in a **pull request**. If we continue like this, I have serious doubts if we can make any deadlines we are facing.

src/lib/y2storage/blk_device.rb
ancorgs: To be honest, just because we use the shorcuts in the tests to do things like

```ruby
expect(proposal.devices.partitions).to contain_exactly(
  an_object_having_attributes(id: Y2Storage::PartitionId::BIOS_BOOT),
  an_object_having_attributes(filesystem_mountpoint: "/"),
  an_object_having_attributes(filesystem_mountpoint: "/home"),
  an_object_having_attributes(filesystem_mountpoint: "swap")
)
```
or
```ruby
expect(sda3).to have_attributes(
          filesystem_mountpoint: "swap",
          filesystem_uuid:       "33333333-3333-3333-3333-33333333",
          filesystem_label:      "swap3",
          size:                  1.GiB - 1.MiB
)
```

Which I admit is a rather arbitrary criteria.

jreidinger: well, I said my points, I will let it up to you to decide.

ancorgs:
@jreidinger added a (quite minimal) test for only fix. The rest are in fact additions to expose libstorage classes that were not there.

test/disk_test.rb
```diff
@@ -275,4 +275,14 @@
...
+
+  describe "#usb?" do
+    let(:disk_name) { "/dev/sda" }
+
+    # Minimum test (we cannot simulate USB disks right now) to ensure it does
+    # not crash (as it used to do at a point in time)
+    it "returns a boolean value" do
+      expect(disk.usb?).to eq false
+    end
+  end
```
jreidinger: :+1:

src/lib/y2storage/blk_device.rb
shundhammer: More of those convenience methods may be added as the need for them arises. But I'd avoid flooding classes with too many of them; that would just obscure the important things.

If we find ourselves going through lengths of downcasting etc. to use certain methods from those classes, we can always add some more convenience methods.

PR #195, opened 2017-04-06
Use new API in yaml-writer unit test

shundhammer:
- Use new API in yaml-writer unit test

- Added alias for create_filesystem() which is more natural to use than create_blk_filesystem()

### DIFF
--- a/src/lib/y2storage/blk_device.rb
+++ b/src/lib/y2storage/blk_device.rb
@@ -48,6 +48,12 @@ module Y2Storage
 
     storage_forward :create_blk_filesystem, as: "Filesystems::BlkFilesystem", raise_errors: true
 
+    # @!method create_filesystem
+    #   Alias for create_blk_filesystem
+    #
+    #   @return [Filesystems::BlkFilesystem]
+    alias_method :create_filesystem, :create_blk_filesystem
+
     storage_forward :create_encryption, as: "Encryption", raise_errors: true
 
     # @!method direct_blk_filesystem
--- a/test/yaml_writer_test.rb
+++ b/test/yaml_writer_test.rb
@@ -23,34 +23,38 @@
 require_relative "spec_helper"
 require "storage"
 require "y2storage/yaml_writer"
+require "y2storage/disk"
+require "y2storage/partition_tables"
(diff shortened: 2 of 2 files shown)


src/lib/y2storage/blk_device.rb
```diff
@@ -48,6 +48,12 @@ class BlkDevice < Device
 
     storage_forward :create_blk_filesystem, as: "Filesystems::BlkFilesystem", raise_errors: true
 
+    # @!method create_filesystem
```
ancorgs: This shouldn't be needed. Yardoc knows how to handle `alias_method`

test/yaml_writer_test.rb
```diff
@@ -23,34 +23,38 @@
...
+require "y2storage/partition_tables"
+require "y2storage/filesystems"
 
 describe Y2Storage::YamlWriter do
 
-  it "produces yaml of a simple disk and partition setup" do
+  let(:test_environment) do
+    Storage::Environment.new(true, Storage::ProbeMode_NONE, Storage::TargetMode_DIRECT)
+  end
+  let(:storage) { Storage::Storage.new(test_environment) }
```
ancorgs: Well, this change is in fact using the new wrapper in the RSpec test, but it does not adapt the YamlWriter itself (neither proves it works without needing adaptation, if that's the case). Is that planned for an upcoming pull request?

I mean, doing something like:
```ruby
describe Y2Storage::YamlWriter do
  before do
    Y2Storage::StorageManager.create_test_instance
  end
  # Using y2storage_staging is temporary to not break compatibility with stuff accessing libstorage directly
  let(:staging) { Y2Storage::StorageManager.instance.y2storage_staging }

  it "produces yaml of a simple disk and partition setup" do
     [...]
     Y2Storage::YamlWriter.write(staging, io)
  end
end
```

Because after this pull request, YamlWriter still expects a `Storage::Devicegraph` object instead of a `Y2Storage::Devicegraph` one.

shundhammer: Yes, this is only the first part. I just meant to check this in before it becomes rotten and irrelevant again due to rebasing.

PR #197, opened 2017-04-06
Missing unit tests for subvolumes

shundhammer:


shundhammer:
For whatever insane reason, the latest rebase caused a crash with alternating segfault and bad_cast exception somewhere in the yaml_writer; casting the proposal.devices from Y2Devicegraph to its storage value Storage::Devicegraph fixed this. But this can only be a temporary solution.

src/lib/y2storage/yaml_writer.rb
```diff
@@ -383,12 +383,14 @@ def yaml_encryption(encryption)
     # @param filesystem [::Storage::BlkFilesystem]
     # @return [Hash{String => Object}]
     #
+    # rubocop:disable Metrics/CyclomaticComplexity
+    #
```
ancorgs: This is the second pull request in a row in which we disable rubocop calls instead of trying to fix the issue. Not sure if using a tool and then ignoring its warning is the way to go.

test/planned_subvol_test.rb
```diff
@@ -207,4 +208,122 @@
     end
   end
 
+  context ".read_from_xml" do
```
ancorgs: Sorry, I was not aware of this pull request and I already added tests for reading from xml in this other PR that fix a problem in that part (duplicated subvolumes).

ancorgs: On the other hand, this looks more like a set of examples than a typical RSpec test to me (more on that later).

test/proposal_subvol_test.rb
```diff
@@ -0,0 +1,120 @@
...
+      end.compact.first
+    end
+
+    let(:subvol_yaml) do
+      root = root_yaml || {}
+      btrfs_yaml = root["btrfs"] || {}
+      btrfs_yaml["subvolumes"] || []
+    end
+
+    # rubocop:disable Metrics/LineLength
```
ancorgs: Rubocop disables ones again. Maybe in this particular case is not that bad (although we have managed to write many tests without disabling it), but I'm worried about the trend.

test/planned_subvol_test.rb
ancorgs: Elaborating what I mean with typical RSpec test vs set of examples.

This is how the output tests for the same functionality look in #200 
```
Y2Storage::ProposalSettings
  #read_product_features!
    [...] 
    sets #subvolumes based on the list at the 'subvolumes' feature
    [...]
    when reading the 'subvolumes' feature
      creates a PlannedSubvol for each subvolume in the list
      reads each 'path' value
      reads each 'copy_on_write' value
      uses true if 'copy_on_write' is omitted
      reads each 'archs' value as an array of strings
      uses nil if 'archs' is omitted
      for an empty list
        sets #subvolumes to an empty list
      if there is no 'subvolumes' feature
        sets #subvolumes to a fallback list
      if there are nil values in the list
        ignores the nil values
      if there are subvolumes without path 
        ignores those subvolumes
      when 'archs' contains just one name
        creates the planned subvolume if the architecture matches the name
        discards the subvolume if the architecture doesn't match
      when 'archs' contains just one name preceded by '!'
        discards the subvolume if the architecture matches the name
        also discards the subvolume if the architecture doesn't match
      when 'archs' contains a list of names
        creates the planned subvolume if the architecture matches any name
        discards the subvolume if none of the names matches the architecture
      when 'archs' contains names with and without '!'
        creates the subvolume if the positive names match and the negated don't
        discards the subvolume if both positive and negated names match
        discards the subvolume if none of the positive names match
      if there are several subvolumes with the same path
        if only one of them apply to the current architecture
          just creates the relevant planned subvolume
        if several of them apply to the current architecture
          just creates the last one to appear in the list
```
They describe the behavior serving as some kind of documentation to understand what's the purpose of the test. That's the goal of using RSpec and BDD.

This is the output of the equivalent test in this pull request.
```
Y2Storage::PlannedSubvol
  .create_from_xml
    Fully specified subvol
      has the correct path 
      is NoCOW
      is tolerant against whitespace in the archs list
      matches arch 'fake'
      matches arch 'ppc'
      does not match arch 'foo'
      does not match arch 'bar'
    Minimalistic subvol
      has the correct path
      is COW 
      is not arch-specific
      matches arch 'fake'
  .read_from_xml
    basic parsing
      returns a non-nil array
      removes architectures other than the current one
    a home subvolume
      exists
      has the correct path
      is COW
      is arch-specific
    a var/cache subvolume
      exists 
      has the correct path
      is COW
      is arch-specific
    a var/lib/mariadb subvolume
      exists
      has the correct path
      is NoCOW
      is arch-specific
    an arch-specific boot/grub2/s390x-emu subvolume
      exists on arch s390
      is COW
      is arch-specific 
    an arch-specific boot/grub2/i386-pc subvolume
      does not exist on s390
```
Things like "a var/cache volume" or "a var /lib/mariadb subvolume" look like random examples to me and give me no information about the intention of the test without reading the code (in the test or in the original class). Sure every example is there for a reason, but that's not clear from the output because it doesn't look like the typical RSpec one (a description of the behavior).

Thinking in terms of contexts and behavior really helps A LOT to find bugs while writing the tests (it forces you to think in width). If we just translate to RSpec the tests we did manually we are loosing the point and it's just a cumbersome syntax for writing traditional example-based unit tests.

src/lib/y2storage/fake_device_factory.rb
```diff
@@ -609,13 +609,15 @@ def create_lvm_pv(parent, args)
...
-      if default_subvolume
-        blk_device = Y2Storage::BlkDevice.find_by_name(@devicegraph, parent)
+      if default_subvolume && !default_subvolume.empty?
+        blk_device = BlkDevice.find_by_name(devicegraph, parent)
         filesystem = blk_device.filesystem
         if !filesystem || !filesystem.type.is?(:btrfs)
           raise HierarchyError, "No btrfs on #{parent}"
         end
-        filesystem.create_btrfs_subvolume(@default_subvolume)
+        btrfs = Storage.to_btrfs(filesystem.to_storage_value)
```
shundhammer: That was the fix I did at 20:30 to bring this whole thing back to life after rebasing when it started to segfault or throw bad_cast exceptions. Can you understand I lost patience after spending **the whole day** trying to use the YAML scenarios for that thing?

src/lib/y2storage/yaml_writer.rb
shundhammer: I hate to repeat myself, but I do not consider rubocop a helpful tool. It is a tool that insists in insane things, yet lets the important issues go unnoticed.

Seriously, if that function is already too complex for us, we should not even think about doing anything as complex as YaST.

test/planned_subvol_test.rb
shundhammer: Rspec lingo calls those things "examples". Surely there must be a reason for that?

Yes, of course I could have covered the entire fallback list verbatim. But then the tests would break every single time somebody requests yet another subvolume or requests a subvolume to be NoCOW. This would be unmaintainable very quickly.

But I forgot, we are not doing this to do any useful testing, we are just blindly fulfilling metrics. That does not strike me as very useful.

test/proposal_subvol_test.rb
shundhammer: The trend is that rubocop **IS A USELESS TOOL**. I keep saying that, and I keep being ignored.

Rubocop doesn't even get us half as far as a plain C or C++ compiler would get us just by compiling. Yet it is considered the holy grail of software engineering here. I wonder why.

shundhammer:
Now rebasing for the 3rd (or 4th?) time in a row. This kind of code review does not work at all. This is breaking our software development process completely. If we keep doing it like this, I don't see a remote chance to meet the upcoming deadline.

test/planned_subvol_test.rb
ancorgs: Don't get me wrong. I didn't ask you to cover the entire fallback list verbatim. I asked you to cover as many scenarios as possible with explanatory descriptions. Scenarios like having a duplicate subvolume in control.xml or all the possible combinations of negated and not negated architectures in the `arch` xml attribute.

I provided the tests at #200 as an example. I don't think those tests are more fragile than these ones. They don't cover the entire fallback list. They were, as a matter of fact, useful to find an existing bug and to fearlessly refactor one method. I would even like to believe they are useful to clarify something that is not completely clear (at least to me) in the documentation regarding negated architectures.

test/proposal_subvol_test.rb
ancorgs: I don't think you are being ignored. It has been discussed several times. As I see it, the result of those discussions has always been that there are around 10 people in the team that consider it to be useful and a couple of people that consider it to be useless and annoying. 

Again as I see it, that has been the reason to not drop Rubocop and to keep encouraging to try to not "offend" it or disable it - team commitment to the tool, beyond individual opinions.

ancorgs:
This has not been a specially long code review. I just started this morning with it. This got ready for review on Thursday at 20:30 CET and on Friday most of the team-S was off (events, illness, vacation).

No discussions about the smell of the flowers or this kind of things you dislike in pull requests so far. :)

test/proposal_subvol_test.rb
shundhammer: So even if we do use a tool that only adds to the workload and does not provide any tangible gain, the values for those metrics are just random. How and why were they chosen? On what grounds? Didn't we just copy them from someplace else because the people there found those values useful?

In particular, with this complexity metrics, we keep shooting ourselves in the foot. There is a tradeoff between the complexity of one function and the number of tiny functions that don't do very much. The overall complexity stays exactly the same, it is just distributed over a larger number of functions. This is just shuffling problems around, not solving them. And that's my major gripe with that.

I cringe whenever I read comments like "make rubocop" happy - because that's ultimately what it is: Making a tool happy for the sake of the tool, not for the project (let alone for the customer / end user).

ancorgs: First of all, the disabled rubocop cops were not, from the beginning, the blocking part of my review. It was the usage of `Storage` (that we had already removed in the YAML-reader and which re-introduction goes against other PBI in the sprint) and the overlap with the RSpec tests already merged as part of #200.

That being said. I think we are discussing this in the wrong place and with the wrong subset of people. I don't think you should convince me here in this pull request. I'm just encouraging you to follow the rules we gave to ourselves as a team. Even if I would agree with you (not the case, but not relevant), I would still push you to follow the team rules and to try to change them in the right scenario, involving everybody (retrospective meeting maybe?).

src/lib/y2storage/fake_device_factory.rb
```diff
@@ -609,13 +609,15 @@ def create_lvm_pv(parent, args)
...
-      if default_subvolume
-        blk_device = Y2Storage::BlkDevice.find_by_name(@devicegraph, parent)
+      if default_subvolume && !default_subvolume.empty?
+        blk_device = BlkDevice.find_by_name(devicegraph, parent)
         filesystem = blk_device.filesystem
         if !filesystem || !filesystem.type.is?(:btrfs)
           raise HierarchyError, "No btrfs on #{parent}"
         end
-        filesystem.create_btrfs_subvolume(@default_subvolume)
+        btrfs = Storage.to_btrfs(filesystem)
```
ancorgs: This is still combining the `Storage` and `Y2Storage` namespaces and downcasting stuff in the former. This should not happen in the YAML-reader. Introducing this is against https://trello.com/c/1gvNrbFQ/546-8-wrapper-based-api  

Maybe there is a bug in the Ruby wrapper, maybe a problem in another place, but this doesn't seem like the way to go.

shundhammer: Now adapted the fake_device_factory completely to the new API; no more Storage:: or Storage. used anywhere.

test/planned_subvol_test.rb
ancorgs: Once again. Reading of subvolumes from control_xml is already unit-tested at https://github.com/yast/yast-storage-ng/blob/master/test/proposal_settings_test.rb#L269  I think we can remove this whole context. It makes no much sense to test exactly the same twice (just with different wording).

shundhammer: Removed

PR #198, opened 2017-04-06
Refac DiskAnalyzer class

joseivanlopez:
WIP

### DIFF
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -73,7 +73,8 @@ module Y2Storage
         PartitionId::LVM
       ]
 
-    DEFAULT_DISKS_LIMIT = 10
+    # Maximum number of disks to check with "expensive" operations.
+    DEFAULT_CHECK_LIMIT = 10
 
     def initialize(devicegraph)
       @devicegraph = devicegraph
@@ -92,42 +93,41 @@ module Y2Storage
     # Checking for content_info.efi? would only detect partitions that are
     # going to be effectively used.
     #
-    # @see #scope
-    #
-    # @return [Hash{String => Array<Partition>}] see {#partitions_with_id}
+    # @param *disks [Disk, String] disks to analyze. All disks by default.
+    # @return [Array<Partition>] see {#partitions_with_id}
     def efi_partitions(*disks)
       data_for(*disks, :efi_partitions) { |d| partitions_with_id(d, PartitionId::ESP) }
     end
 
     # Partitions that can be used as PReP partition
-    # @see #scope
     #
-    # @return [Hash{String => Array<Partition>}] see {#partitions_with_id}
+    # @param *disks [Disk, String] disks to analyze. All disks by default.
+    # @return [Array<Partition>] see {#partitions_with_id}
     def prep_partitions(*disks)
       data_for(*disks, :prep_partitions) { |d| partitions_with_id(d, PartitionId::PREP) }
     end
 
     # GRUB (gpt_bios) partitions
-    # @see #scope
     #
-    # @return [Hash{String => Array<Partition>}] see {#partitions_with_id}
+    # @param *disks [Disk, String] disks to analyze. All disks by default.
+    # @return [Array<Partition>] see {#partitions_with_id}
     def grub_partitions(*disks)
       data_for(*disks, :grup_partitions) { |d| partitions_with_id(d, PartitionId::BIOS_BOOT) }
     end
 
     # Partitions that can be used as swap space
-    # @see #scope
     #
-    # @return [Hash{String => Array<Partition>}] see {#partitions_with_id}
+    # @param *disks [Disk, String] disks to analyze. All disks by default.
+    # @return [Array<Partition>] see {#partitions_with_id}
     def swap_partitions(*disks)
       data_for(*disks, :swap_partitions) { |d| partitions_with_id(d, PartitionId::SWAP) }
     end
 
     # Linux partitions. This may be a normal Linux partition (type 0x83), a
     # Linux swap partition (type 0x82), an LVM partition, or a RAID partition.
-    # @see #scope
     #
-    # @return [Hash{String => Array<Partition>}] see {#partitions_with_id}
+    # @param *disks [Disk, String] disks to analyze. All disks by default.
+    # @return [Array<Partition>] see {#partitions_with_id}
     def linux_partitions(*disks)
       data_for(*disks, :linux_partitions) { |d| partitions_with_id(d, LINUX_PARTITION_IDS) }
     end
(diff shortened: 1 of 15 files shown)


src/lib/y2storage/disk_analyzer.rb
```diff
@@ -223,120 +173,97 @@ def mbr_gap
...
+    def windows_partitions(*disks, limit: DEFAULT_DISKS_LIMIT)
+      data_for(*disks, :windows_partitions, limit: limit) { |d| find_windows_partitions(d) }
     end
 
-  private
+    # Disks that are suitable for installing Linux.
+    #
+    # @return [Array<Storage::Disk>] candidate disks
+    def candidate_disks(limit: DEFAULT_DISKS_LIMIT)
+      return @candidate_disks if @candidate_disks
```
ancorgs: I find this strange. We have a `limit` param but we cache the result. That means
```
analyzer.candidate_disks(limit: 10) # Returns the result of searching with a limit of 10
[...]
analyzer.candidate_disks(limit: 25) # Returns the result of searching with a limit of 10!!!
```

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -122,41 +95,41 @@ def device_by_name(name)
     # @see #scope
```
ancorgs: We should remove all those `@see #scope`

src/lib/y2storage/disk_analyzer.rb
joseivanlopez: Yes, you are right. In this case what would be the best solution? Avoid a limit parameter and always use a value internally?

src/lib/y2storage/disk_analyzer.rb
joseivanlopez: Yes, doc is not yet updated.

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -223,120 +173,97 @@ def mbr_gap
...
     #
     # @return [Hash{String => Array<Partition>}] see {#partitions_with_id}
-    def windows_partitions
-      @windows_partitions ||= begin
-        result = find_windows_partitions
-        log.info("Found Windows partitions: #{result}")
-        result
-      end
+    def windows_partitions(*disks, limit: DEFAULT_DISKS_LIMIT)
```
ancorgs: The API can be somehow weird
```
all_disks.size #=> 15
analyzer.windows_partitions #=> Only analyzes 10, that's fine
analyzer.windows_partitions(all_disks) #=> Only analyzes 10 although I'm explicitly passing 15
analyzer.windows_partitions(all_disks, limit: nil) #=> This is the correct way to analyze them all
```
Should we maybe (just asking, not saying) honor the `limit` only if `*disks` is empty?
```
all_disks.size #=> 15
analyzer.windows_partitions #=> Only analyzes 10, that's fine
analyzer.windows_partitions(all_disks) #=> Analyzes all the 15 passed ones
analyzer.windows_partitions(all_disks, limit: 10) #=> Only analyzes 10, although it made me pass the 10 (default) explicitly
```
Both options are weird at some point. I'm just not sure what is less surprising.

joseivanlopez: Maybe, limit should be nil by default?

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -223,120 +173,97 @@ def mbr_gap
...
+      data_for(*disks, :windows_partitions, limit: limit) { |d| find_windows_partitions(d) }
     end
 
-  private
+    # Disks that are suitable for installing Linux.
+    #
+    # @return [Array<Storage::Disk>] candidate disks
+    def candidate_disks(limit: DEFAULT_DISKS_LIMIT)
```
ancorgs: We should make very clear that this `limit` works in an expected way. I mean, kind of inverse to the argument with the same name in `#windows_partitions`.  In this case, the smaller the `:limit` the bigger the collection.

src/lib/y2storage/disk_analyzer.rb
ancorgs: Calling the argument `limit` suggests than the result will not have more than `limit`. The original constant name `check_limit` or something similar like `max_checks` or `max_mounts` would be more clear.

ancorgs: Nevermind, I actually messed up limit of checks/mounts (that is related to the number of partitions) with the number of passed disks. It's fine as it is, we just need a better name for `:limit`

joseivanlopez: Ok.

src/lib/y2storage/disk_analyzer.rb
ancorgs: Yes, I would go for that. Sounds simple and good enough.

ancorgs: But please use a different internal constant than `DEFAULT_CHECK_LIMIT`

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -203,140 +147,143 @@ def candidate_disks
...
+    # @return [Array<Disk>] candidate disks
+    def candidate_disks(check_limit: DEFAULT_CHECK_LIMIT)
+      return @candidate_disks if @candidate_disks
+      @candidate_disks = find_candidate_disks(check_limit)
+      log.info("Found candidate disks: #{@candidate_disks}")
+      @candidate_disks
     end
 
-    # Array with all disks in the devicegraph
+    # Checks if a partition belongs to any VG or to an specefic one.
```
ancorgs: typo: specefic

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -203,140 +147,143 @@ def candidate_disks
...
     #
-    # @return [Array<Disk>]
-    def all_disks
-      devicegraph.disks
+    # @param partition [Partition]
+    # @param vg [LvmVg] volume group
+    def partition_in_vg?(partition, vg = nil)
```
ancorgs: Maybe some of this logic about a certain device being part of LVM should be moved to the BlkDevice class or similar places.

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -203,140 +147,143 @@ def candidate_disks
...
+  private
+
+    attr_reader :devicegraph
 
-      dev_names(disks.select { |disk| installation_disk?(disk.name) })
+    # Gets data for a set of disks, stores it and returns that data.
+    #
+    # @param *disks [Disk, String] disks to analyze. All disks by default.
+    # @param data [Symbol] data name.
+    # @param limit [Fixnum, nil] maximum number of disks to check.
```
ancorgs: The limit should not be in the number of disks to check. Otherwise we get to the consistency API problems I already pointed. It should be in the number of expensive operations to perform.

For example, checking a disk with 10 partitions is more expensive than checking 100 unformatted and unpartitioned disks.

Of course, we would have to rethink the caching algorithm or to make the limit a soft limit,

ancorgs:
Just to not forget, we need to adapt yast2-country to this new API as soon as we are done with it.

src/lib/y2storage/disk_analyzer.rb
ancorgs: Moreover, does it makes to apply the limit if we already have cached results? Shouldn't it be a limit on how many NEW operations to perform as a maximum (or a soft-maximum)?

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -106,95 +87,58 @@ def initialize(devicegraph, disk_check_limit: DEFAULT_DISK_CHECK_LIMIT, scope: n
       @devicegraph = Devicegraph.new(devicegraph) if devicegraph.is_a?(Storage::Devicegraph)
```
ancorgs: We are breaking the API with these changes anyway, so let's remove this hack.

ancorgs:
You offended rubocop. Other than that, this LGTM for the time being. So fix rubocop, merge and adapt yast2-country

PR #201, opened 2017-04-10
Some documentation for the wrapper classes

ancorgs:
Still to be improved, of course.

### DIFF
--- a/src/lib/y2storage/blk_device.rb
+++ b/src/lib/y2storage/blk_device.rb
@@ -33,27 +33,75 @@ module Y2Storage
     storage_class_forward :all, as: "BlkDevice"
     storage_class_forward :find_by_name, as: "BlkDevice"
 
+    # @!method name
+    #   @return [String] kernel-style device name
+    #     (e.g. "/dev/sda2" or "/dev/vg_name/lv_name")
     storage_forward :name
+
+    # @!method name=
+    #   @see #name
     storage_forward :name=
+
+    # @!method region
+    #   @return [Region]
     storage_forward :region, as: "Region"
+
+    # @!method region=
+    #   @see #region
     storage_forward :region=
+
+    # @!method size
+    #   @return [DiskSize]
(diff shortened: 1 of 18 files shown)


src/lib/y2storage/blk_device.rb
```diff
@@ -33,27 +33,75 @@ class BlkDevice < Device
...
+    # @!method name=
+    #   @see #name
     storage_forward :name=
+
+    # @!method region
+    #   @return [Region]
     storage_forward :region, as: "Region"
+
+    # @!method region=
+    #   @see #region
```
jreidinger: NP: to be honest I do not like much this see stuff. Do you consider document it as attribute? http://www.rubydoc.info/gems/yard/file/docs/Tags.md#attribute

src/lib/y2storage/btrfs_subvolume.rb
```diff
@@ -29,20 +29,54 @@ module Y2Storage
...
     storage_forward :default_btrfs_subvolume?
-    storage_forward :default_btrfs_subvolume=
+
+    # @!method create_btrfs_subvolume(path)
+    #   @param path [String] path of the new subvolume
+    #   @return [BtrfsSubvolume]
     storage_forward :create_btrfs_subvolume, as: "BtrfsSubvolume"
 
+    # Sets this subvolume as the default one
+    def set_default_btrfs_subvolume
```
jreidinger: NP: I feel a bit strange about this method name. What about `assign_as_default_btrfs_subvolume` or `default_btrfs_subvolume!`?

ancorgs: I also kind of dislike the original name in libstorage. But this should become a direct forward to libstorage once the problems with SWIG are fixed. Thus, I would like to keep the original name if there is not a very strong reason to change it, other than our personal preferences.

jreidinger: OK

PR #193, opened 2017-04-05
Proposal guided setup

joseivanlopez:
PBI: https://trello.com/c/crhdffWC/516-5-storageng-guided-setup

PR #202, opened 2017-04-10
Guided setup

joseivanlopez:
PBI: https://trello.com/c/crhdffWC/516-5-storageng-guided-setup

This PR replaces this old one: https://github.com/yast/yast-storage-ng/pull/193

### DIFF
--- a/src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
+++ b/src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
@@ -23,21 +23,23 @@ require "yast"
 require "y2storage"
 require "y2storage/dialogs/guided_setup/base"
 
+Yast.import "InstExtensionImage"
+
 module Y2Storage
   module Dialogs
     class GuidedSetup
+      # Dialog to select partitioning scheme.
       class SelectScheme < Dialogs::GuidedSetup::Base
-
-        def label
-          "Guided Setup - step 3"
-        end
-
         def encryption_handler
           widget_update(:password, widget_value(:encryption), attr: :Enabled)
           widget_update(:repeat_password, widget_value(:encryption), attr: :Enabled)
         end
 
       protected
+
+        def label
+          "Guided Setup - step 3"
+        end
 
         def dialog_title
           _("Partitioning Scheme")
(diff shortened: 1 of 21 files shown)


src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
```diff
@@ -0,0 +1,200 @@
...
+    class GuidedSetup
+      # Dialog to select partitioning scheme.
+      class SelectScheme < Base
+        def encryption_handler
+          widget_update(:password, widget_value(:encryption), attr: :Enabled)
+          widget_update(:repeat_password, widget_value(:encryption), attr: :Enabled)
+        end
+
+        def back_handler
+          unload_cracklib
```
ancorgs: Wouldn't it be better to redefine `#finish_dialog` or `#close_dialog` in order to unload cracklib? It sounds more robust to me than doing it in every possible handler.

src/lib/y2storage/filesystems/type.rb
```diff
@@ -30,6 +30,82 @@ class Type
       include StorageEnumWrapper
 
       wrap_enum "FsType"
+
+      FILESYSTEMS = {
```
ancorgs: nitpicking: not sure if I like the name. What about `PROPERTIES` or something similar?

src/lib/y2storage/filesystems/type.rb
```diff
@@ -30,6 +30,82 @@ class Type
...
+          name: "ISO9660"
+        },
+        udf:      {
+          name: "UDF"
+        }
+      }
+
+      ROOT_FILESYSTEMS = [:ext2, :ext3, :ext4, :btrfs, :reiserfs, :xfs]
+
+      HOME_FILESYSTEMS = [:ext2, :ext3, :ext4, :btrfs, :reiserfs, :xfs]
```
ancorgs: I would use `private_constant` in all these constants.

src/lib/y2storage/filesystems/type.rb
```diff
@@ -30,6 +30,82 @@ class Type
...
+      # @return [Array<Filesystems::Type>]
+      def self.home_filesystems
+        HOME_FILESYSTEMS.map { |f| find(f) }
+      end
+
+      # Human readable text for a filesystem
+      #
+      # @return [String]
+      def to_human
+        FILESYSTEMS[to_sym][:name]
```
ancorgs: Doesn't look robust. What if a new type is added in libstorage but we don't add the corresponding name?

src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
joseivanlopez: Done

src/lib/y2storage/filesystems/type.rb
joseivanlopez: Yes, maybe PROPERTIES is a better name for that.

src/lib/y2storage/filesystems/type.rb
joseivanlopez: Ok

src/lib/y2storage/filesystems/type.rb
joseivanlopez: Yes, you are right.

PR #203, opened 2017-04-10
Adapt YamlWriter to stop using libstorage directly

ancorgs:


ancorgs:
I checked the report and the coverage decrease is not real.

PR #204, opened 2017-04-11
Guided setup improvements

joseivanlopez:
Add some improvements suggested by UX expert.

* Skip first dialog when there is only one candidate disk.
* Align widgets.

PBI: https://trello.com/c/crhdffWC/516-5-storageng-guided-setup

### DIFF
--- a/src/lib/y2storage/dialogs/guided_setup.rb
+++ b/src/lib/y2storage/dialogs/guided_setup.rb
@@ -65,7 +65,24 @@ module Y2Storage
           "select_filesystem" => { next: :next, back: :back,  abort: :abort }
         }
 
+        # Skip disks selection when there is only one candidate
+        if candidate_disks.size == 1
+          settings.candidate_devices = candidate_disks.map(&:name)
+          sequence["ws_start"] = "select_root_disk"
+        end
+
         Yast::Sequencer.Run(aliases, sequence)
+      end
+
+      # Disks that could be selected for installation.
+      #
+      # @return [Array<Disk>]
+      def candidate_disks
+        return @candidate_disks if @candidate_disks
+        candidates = settings.candidate_devices || []
+        candidates = candidates.map { |d| analyzer.device_by_name(d) }
+        candidates = analyzer.candidate_disks if candidates.empty?
+        @candidate_disks = candidates
       end
     end
   end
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/dialogs/guided_setup.rb
```diff
@@ -65,8 +65,25 @@ def run
...
       end
+
+      # Disks that could be selected for installation.
+      #
+      # @return [Array<Disk>]
+      def candidate_disks
+        return @candidate_disks if @candidate_disks
+        candidates = settings.candidate_devices || []
+        candidates = candidates.map { |d| analyzer.device_by_name(d) }
+        candidates = analyzer.candidate_disks if candidates.empty?
```
jreidinger: I am not sure if I get logic there. If settings is set, it will use it. And if settings is not set, it will return result of analyzer. So if you going back in guided setup and you previously pick only one disk, it will skip selecting disks, right?

test/dialogs/guided_setup_test.rb
```diff
@@ -25,54 +25,82 @@
...
+    expect_any_instance_of(dialog).to receive(:run).once
+  end
+
+  def expect_not_run_dialog(dialog)
+    expect_any_instance_of(dialog).not_to receive(:run)
+  end
+
+  def expect_not_run_select_disks
+    expect_not_run_dialog(Y2Storage::Dialogs::GuidedSetup::SelectDisks)
+  end
```
jreidinger: what is goal of this shorthand. Expect is usually used once for specific test. Allow which is there before make sense, as allow is often used in before handlers and so on. So how often is this expect used?

src/lib/y2storage/dialogs/guided_setup.rb
joseivanlopez: Yes, and that is wrong. I guess that should always consider as list of candidates (that is, the list of disks that could be selected) the result of the analyzer.

test/dialogs/guided_setup_test.rb
joseivanlopez: Right now only once, but maybe in next iterations we will skip more steps of the wizard. In these cases, we will expect to skip the same dialog several times.

jreidinger: well, I found this quite unusual. But it is probably just my taste so NP.

PR #205, opened 2017-04-17
Disable uneeded widgets

joseivanlopez:
PBI: https://trello.com/c/b5DSHXf5/567-2-storageng-guided-set-up-disable-unneeded-widgets

### DIFF
--- a/src/lib/y2storage/dialogs/guided_setup/select_root_disk.rb
+++ b/src/lib/y2storage/dialogs/guided_setup/select_root_disk.rb
@@ -28,9 +28,26 @@ module Y2Storage
     class GuidedSetup
       # Dialog for root disk selection.
       class SelectRootDisk < Base
+        def initialize(*params)
+          super
+          define_disks_handlers
+        end
+
+        # This dialog should be skipped when there is only one candidate
+        # disk for installation and there are not installed systems.
+        def skip?
+          candidate_disks.size == 1 &&
+            analyzer.installed_systems(candidate_disks.first).size == 0
+        end
+
+        # Before skipping, settings should be assigned.
+        def before_skip
+          settings.root_device = candidate_disks.first.name
+        end
+
         def root_disk_handler
-          widget_update(:windows_action, activate_windows_actions?, attr: :Enabled)
-          widget_update(:linux_action, activate_linux_actions?, attr: :Enabled)
+          widget_update(:windows_actions, activate_windows_actions?, attr: :Enabled)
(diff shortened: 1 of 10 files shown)


src/lib/y2storage/dialogs/guided_setup/select_root_disk.rb
```diff
@@ -93,31 +119,71 @@ def linux_actions_widget
...
+          if need_to_select_disk?
+            candidate_disks.detect { |d| widget_value(disk_id(d)) }
+          else
+            candidate_disks.first
+          end
+        end
+
+        def activate_windows_actions?
+          root = selected_disk
+          root ? !analyzer.windows_systems(root).empty? : true
```
ancorgs: I don't think this is correct. The removal of systems affects all the candidate disks, not only the root one.

src/lib/y2storage/dialogs/guided_setup/select_root_disk.rb
```diff
@@ -93,31 +119,71 @@ def linux_actions_widget
...
+        end
+
+        def activate_linux_actions?
+          root = selected_disk
+          root ? !analyzer.linux_systems(root).empty? : true
```
ancorgs: Same here, we should check all the disks.

src/lib/y2storage/dialogs/guided_setup/select_root_disk.rb
```diff
@@ -93,31 +119,71 @@ def linux_actions_widget
...
+          options.each do |option|
+            define_singleton_method :"#{option}_handler" do
+              root_disk_handler
+            end
+          end
+        end
+
+        def disk_id(disk)
+          name = disk.is_a?(String) ? disk : disk.name
+          name.split("/").last
```
ancorgs: Maybe I'm being a little bit paranoid but I think we should add an explicit documentation somewhere stating that we use just the last part of the path as widget id. Sure there is no collision risk using regular disk names like `/dev/sda`, but... you know :wink:

test/dialogs/guided_setup/select_root_disk_test.rb
```diff
@@ -27,51 +27,136 @@
...
-
   before do
     settings.candidate_devices = candidate_disks
   end
 
+  describe "#skip?" do
+    context "when there is only one disk" do
+      let(:candidate_disks) { ["/dev/sda"] }
+
+      context "and there is not any intalled system" do
```
ancorgs: typo: intalled

PR #207, opened 2017-04-18
Add disk analyzer to storage manager

joseivanlopez:
Related to PBI https://trello.com/c/b5DSHXf5/567-2-storageng-guided-set-up-disable-unneeded-widgets

### DIFF
--- a/src/lib/y2storage/storage_manager.rb
+++ b/src/lib/y2storage/storage_manager.rb
@@ -148,11 +148,11 @@ module Y2Storage
       storage.commit
     end
 
-    # Disk analyzer used to analyze the initial devicegraph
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/storage_manager.rb
```diff
@@ -147,6 +148,13 @@ def commit
       storage.commit
     end
 
+    # Disk analyzer used to analyze the initial devicegraph
+    #
+    # @return [DiskAnalyzer]
+    def disk_analyzer
```
ancorgs: I would prefer something like `probed_disk_analyzer`, to make the connection with `probed` explicit.

joseivanlopez: Ok

PR #208, opened 2017-04-19
Usability fixes

joseivanlopez:
Related to PBI https://trello.com/c/b5DSHXf5/567-2-storageng-guided-set-up-disable-unneeded-widgets

* Focus password field when encryption is selected.
* Improve password error message and allow to use a simple password.

### DIFF
--- a/src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
+++ b/src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
@@ -32,11 +32,11 @@ module Y2Storage
       # Dialog to select partitioning scheme.
       class SelectScheme < Base
         # Handler for :encryption check box.
-        # @param allow_focus [Boolean] whether password field can be focused
-        def encryption_handler(allow_focus: true)
+        # @param focus [Boolean] whether password field should be focused
+        def encryption_handler(focus: true)
           widget_update(:password, using_encryption?, attr: :Enabled)
           widget_update(:repeat_password, using_encryption?, attr: :Enabled)
-          return unless allow_focus && using_encryption?
+          return unless focus && using_encryption?
           Yast::UI.SetFocus(Id(:password))
         end
 
@@ -77,7 +77,7 @@ module Y2Storage
         def initialize_widgets
           widget_update(:lvm, settings.use_lvm)
           widget_update(:encryption, settings.use_encryption)
-          encryption_handler(allow_focus: false)
+          encryption_handler(focus: false)
           if settings.use_encryption
             widget_update(:password, settings.encryption_password)
             widget_update(:repeat_password, settings.encryption_password)
--- a/src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
+++ b/src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
@@ -31,9 +31,13 @@ module Y2Storage
     class GuidedSetup
       # Dialog to select partitioning scheme.
       class SelectScheme < Base
-        def encryption_handler
-          widget_update(:password, widget_value(:encryption), attr: :Enabled)
-          widget_update(:repeat_password, widget_value(:encryption), attr: :Enabled)
+        # Handler for :encryption check box.
+        # @param allow_focus [Boolean] whether password field can be focused
+        def encryption_handler(allow_focus: true)
+          widget_update(:password, using_encryption?, attr: :Enabled)
+          widget_update(:repeat_password, using_encryption?, attr: :Enabled)
+          return unless allow_focus && using_encryption?
+          Yast::UI.SetFocus(Id(:password))
         end
 
       protected
@@ -73,7 +77,7 @@ module Y2Storage
         def initialize_widgets
           widget_update(:lvm, settings.use_lvm)
           widget_update(:encryption, settings.use_encryption)
-          encryption_handler
+          encryption_handler(allow_focus: false)
           if settings.use_encryption
             widget_update(:password, settings.encryption_password)
             widget_update(:repeat_password, settings.encryption_password)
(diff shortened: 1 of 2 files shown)


src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
```diff
@@ -24,15 +24,20 @@
...
   module Dialogs
     class GuidedSetup
       # Dialog to select partitioning scheme.
       class SelectScheme < Base
-        def encryption_handler
-          widget_update(:password, widget_value(:encryption), attr: :Enabled)
-          widget_update(:repeat_password, widget_value(:encryption), attr: :Enabled)
+        # Handler for :encryption check box.
+        # @param allow_focus [Boolean] whether password field can be focused
+        def encryption_handler(allow_focus: true)
```
jreidinger: why name of of variable `allow_focus`? And also description is misleading "whether password field can be focused" even if it is true as disabled element cannot have focus, this is about something different. It is about setting focus when possible. So maybe use simple name "focus".

src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
```diff
@@ -141,11 +146,25 @@ def password_correct?
...
         # Password is considered strong when cracklib returns an empty message.
+        # User has the last word to decide whether to use a weak password.
         def password_strong?
           message = cracklib_message
           return true if message.empty?
-          Yast::Report.Warning(message)
-          false
+
+          popup_text = [
+            _("The password is too simple:"),
```
jreidinger: I am not sure if problem is always too simple, as it can have different problem. What about "The password is too weak"? which fits better?

test/dialogs/guided_setup/select_scheme_test.rb
```diff
@@ -149,17 +149,40 @@
...
+        end
+
+        context "and password is not accepted" do
+          let(:password_accepted) { false }
+
+          it "does not save password in settings" do
+            subject.run
+            expect(subject.settings.encryption_password).to eq(nil)
+          end
+        end
```
jreidinger: 👍 for tests, but I miss tests for other change that is new focus parameter, to verify that focus SetFocus is properly called

src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
joseivanlopez: I would prefer what you propose: "The password is too weak". But I have used "too simple" to mimic this: https://openqa.opensuse.org/tests/385493#step/user_settings/5

So, the question is whether continue using the same than before or not.

test/dialogs/guided_setup/select_scheme_test.rb
```diff
@@ -149,17 +149,40 @@
...
+          it "saves password in settings" do
+            subject.run
+            expect(subject.settings.encryption_password).to eq(password)
+          end
+        end
+
+        context "and password is not accepted" do
+          let(:password_accepted) { false }
+
+          it "does not save password in settings" do
```
jreidinger: in fact it does not save, but explicitly set no password, which is difference.

test/dialogs/guided_setup/select_scheme_test.rb
joseivanlopez: You are right. I will add that.

src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
jreidinger: well, in such case we should prefer consistency of yast and installation. Even when my proposal is more consistent with the most common web ui wording.

src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
joseivanlopez: Well, I used `allow_focus` instead of `focus` because with `focus` looks like the focus should bed forced, and the idea is only to allow the focus of the field or not.

jreidinger: well, but `allow_focus` looks like something like `SetFocusable`. If you want to explicitely mention that it is not guarantied then maybe something like `focus_if_possible` but simple focus, sounds better for me.

jreidinger: or maybe `focus_if_enabled`

joseivanlopez: Umm, I think we are not understanding the same. The field for password can have three situations:
* (a) not enabled and not focused
* (b) enabled and focused
* (d) enabled and not focused

States (a) and (b) occur when encryption checkbox is checked/unchecked, and state (c) could happen when dialog is opened and a password exists. In this case we do not want focus the field, but it is enabled. So, the param :allow_focus is for controlling this case.

joseivanlopez: In fact, I think that `focusable` could be a good name, isn't it?

jreidinger: I understand that varable, so `focus_if_enabled` means that we want (b) and not (d). Still `allow_focus` indicate for me that we want to allow focusing something, which is not what it do. It focuses or not in b and d. (a) is different case as focus is not possible on not enabled button.

joseivanlopez: You are right. Option (a) is out of the play, so as you mention at first, simply `focus` is a good name.

jreidinger: well, `focusable` indicate that something is focusable. In libyui focusable == enabled. So it is confusing. You do not set something to be focusable, but to change focus under certain conditions.

test/dialogs/guided_setup/select_scheme_test.rb
jreidinger: ah, in fact it is true, as it don't call `update_setttings!`...
So that means, that I miss there test for new behavior in `update_settings!` which explicitely set nil instead of setting it only when pwd is not empty.

PR #209, opened 2017-04-19
Moved many methods from DiskAnalyzer to Partitionable

ancorgs:


src/lib/y2storage/disk_analyzer.rb
```diff
@@ -48,24 +48,6 @@ module Y2Storage
...
-    WINDOWS_PARTITION_IDS =
-      [
-        PartitionId::NTFS,
-        PartitionId::DOS32,
-        PartitionId::DOS16,
-        PartitionId::DOS12,
-        PartitionId::WINDOWS_BASIC_DATA,
-        PartitionId::MICROSOFT_RESERVED
-      ]
-
```
ancorgs: Because the logic for `candidate_disks` remains in DiskAnalyzer. I moved the other two constants because the logic was moved to PartitionId.

src/lib/y2storage/proposal/partition_killer.rb
```diff
@@ -134,8 +136,7 @@ def delete_lvm_partitions(partition)
       # @param partition [Partition]
       # @return [Boolean]
       def lvm_pv?(partition)
-        lvm_pv_names = disk_analyzer.used_lvm_partitions.map(&:name)
-        lvm_pv_names.include?(partition.name)
+        !!(partition.lvm_pv && partition.lvm_pv.lvm_vg)
```
ancorgs: Your suggested code would still return `nil` (instead of `false`) if there is no PV. If you want to avoid double negation you would need to do something like
```ruby
!(partition.lvm_pv.nil? || partition.lvm_pv.lvm_vg.nil?)
# or
!partition.lvm_pv.nil? || !partition.lvm_pv.lvm_vg.nil?
```
IMO, both options are way less readable that using double negation (something we do relatively often, as proved by the fact that we disabled the rubocop check to prevent it https://github.com/yast/yast-devtools/blob/62a49a654a4bd402c478d434808c7ea50024b610/ytools/y2tool/rubocop_yast_style.yml#L202)

Indeed, `lvm_vg?` would be a better name. I didn't add it to `BlkDevice` or `Partition` because I felt it could be confusing and too different from the libstorage approach.

PR #210, opened 2017-04-20
- pass storage object when creating devicegraph objects

aschnell:
Needed for https://trello.com/c/WAGRJLgv/542-5-libstorage-ng-support-of-alternative-device-naming-schemes.

### DIFF
--- a/src/lib/y2storage/devicegraph.rb
+++ b/src/lib/y2storage/devicegraph.rb
@@ -85,7 +85,8 @@ module Y2Storage
     # @param filename [String]
     # @return [Devicegraph]
     def self.new_from_file(filename)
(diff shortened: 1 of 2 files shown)


src/lib/y2storage/devicegraph.rb
```diff
@@ -85,7 +85,8 @@ class Devicegraph
     # @param filename [String]
     # @return [Devicegraph]
     def self.new_from_file(filename)
-      devicegraph = ::Storage::Devicegraph.new
+      storage = Y2Storage::StorageManager.instance.storage
```
ancorgs: Well, strictly speaking the `Y2Storage::` part is not needed here.

PR #211, opened 2017-04-20
- adapted to change in libstorage-ng

aschnell:
Needs https://github.com/openSUSE/libstorage-ng/pull/238.


aschnell:
The failed checks are due to the fact that travis does not recheck with the new version of libstorage-ng.

PR #212, opened 2017-04-21
Set default yes to accept weak password

joseivanlopez:
PBI https://trello.com/c/b5DSHXf5/567-2-storageng-guided-set-up-disable-unneeded-widgets


jreidinger:
I am not sure if security team won't complain

joseivanlopez:
In other dialogs of YaST (user creation) when user introduces a weak password, the popup message sets by default "accept" to yes, so here it is done the same for consistency.

jreidinger:
@joseivanlopez well, if it is same as in user, then OK, but in general security team do not like to see insecure option as default

PR #213, opened 2017-04-21
- save probed and staging devicegraph before commit

aschnell:


src/lib/y2storage/clients/inst_prepdisk.rb
```diff
@@ -41,6 +41,11 @@ class InstPrepdisk
       def run
         return :auto if Mode.update
 
+        storage = manager.storage
+
+        storage.probed.save(Directory.logdir + "/probed-devicegraph.xml")
+        storage.staging.save(Directory.logdir + "/staging-devicegraph.xml")
+
```
joseivanlopez: A better place to put that lines is `StorageManager#commit` to save logs in all commit calls, not only in the client. Moreover, I think that `Directory` is not defined in that context, so it should be `Yast::Directory`. Does it run without failing? Another thing, would it not be better to save the xml version of the graph directly in the log file instead of create a new file?

aschnell: I have tested the code and it works.

Since the XML files can be very huge I prefer to have them as separate files. It must also be easy to load them on another system without first cutting them from the logs to create test cases.

Whether the code is here or in the manager I do not mind.

joseivanlopez: Nice if it works, so please, move it to `StorageManager#commit` to have them with all commit calls 👍

ancorgs: In fact, if we are going to write it to a separate file (not to y2log) with a fixed name, it's better to do it in the client instead of doing it in `StorageManager`. So it does not get overwritten with subsequent calls to `commit` (for example, running the expert partitioner in the installed system).

ancorgs: @joseivanlopez `Directory` works because there is a `include Yast`.

Last but not least, doesn't this create a couple of files in `~` every time the unit tests are executed?

PR #216, opened 2017-04-24
Fix tests that depend on arch

joseivanlopez:
Fix current problems generating iso for other archs.


test/disk_analyzer_test.rb
```diff
@@ -33,8 +33,7 @@
   end
 
   describe "#windows_partitions" do
-    # let(:scenario) { "mixed_disks" }
-    let(:scenario) { "complex-lvm-encrypt" }
+    let(:scenario) { "mixed_disks" }
```
ancorgs: Hmmm, this looks suspicious

PR #217, opened 2017-04-24
Apply settings about what to do with existing partitions

ancorgs:
This implements the main functionality of
https://trello.com/c/PnqN8jgz/549-5-storageng-guided-set-up-apply-the-setting-what-to-do-with-existing-partitions

It should be way easier to review commit by commit.

### DIFF
--- a/src/lib/y2storage/proposal.rb
+++ b/src/lib/y2storage/proposal.rb
@@ -105,7 +105,7 @@ module Y2Storage
 
     # @return [PlannedVolumesList]
     def volumes_list
-      generator = VolumesGenerator.new(populated_settings, initial_graph)
+      generator = VolumesGenerator.new(populated_settings, clean_graph)
       generator.volumes
     end
 
@@ -116,14 +116,24 @@ module Y2Storage
     # @return [Devicegraph]
     def devicegraph(volumes)
       generator = DevicegraphGenerator.new(populated_settings)
-      generator.devicegraph(volumes, initial_graph, disk_analyzer)
+      generator.devicegraph(volumes, clean_graph, space_maker)
     end
 
-    # Disk analyzer used to analyze the initial devigraph
+    def space_maker
+      @space_maker ||= SpaceMaker.new(disk_analyzer, populated_settings)
+    end
+
+    # Disk analyzer used to analyze the initial devicegraph
     #
     # @return [DiskAnalyzer]
     def disk_analyzer
       @disk_analyzer ||= DiskAnalyzer.new(initial_graph)
+    end
+
+    # Copy of #initial_graph without all the partitions that must be wiped out
+    # according to the settings
+    def clean_graph
+      @clean_graph ||= space_maker.delete_unwanted_partitions(initial_graph)
     end
 
     def initial_graph
--- a/src/lib/y2storage/proposal.rb
+++ b/src/lib/y2storage/proposal.rb
@@ -78,28 +78,44 @@ module Y2Storage
       raise UnexpectedCallError if proposed?
       settings.freeze
       @proposed = true
-      @volumes = volumes_list(:all, populated_settings)
-      @devices = devicegraph(@volumes, populated_settings)
+
+      # FIXME: The current implementation, tries :desired and then :min for
+      # every root disk. It would probably make more sense to try first :desired
+      # for all possible root disks and then do the same with :min.
+      exception = nil
+      candidate_roots.each do |disk_name|
+        populated_settings.root_device = disk_name
+        exception = nil
+
+        begin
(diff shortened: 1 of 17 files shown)


src/lib/y2storage/proposal.rb
```diff
@@ -78,91 +78,94 @@ def propose
...
+
+        begin
+          @volumes = volumes_list
+          @devices = devicegraph(@volumes)
+        rescue Error => error
+          log.info "Failed to make a proposal using root device #{disk_name}: #{error.message}"
+          exception = error
+        end
+
+        break unless exception
```
jreidinger: if I get code, then exception is part of workflow to find suitable disk from candidates. I personally find a bit awkward to have exception part of common workflow.

src/lib/y2storage/proposal_settings.rb
```diff
@@ -38,13 +38,59 @@ class ProposalUserSettings
...
+    #   @return [String] password to use when creating new encryption devices
     secret_attr   :encryption_password
 
+    # @return [Boolean] whether to resize Windows systems if needed
+    attr_accessor :resize_windows
+
+    # @return [Symbol] what to do regarding removal of existing partitions
+    #   hosting a Windows system.
+    #
+    #   * :none never Delete a Windows partition.
```
jreidinger: strange case. probably "Never delete"

src/lib/y2storage/proposal_settings.rb
```diff
@@ -53,11 +99,43 @@ def initialize
...
+    # Whether the settings disable deletion of a given type of partitions
+    #
+    # @see #windows_delete_mode
+    # @see #linux_delete_mode
+    # @see #other_delete_mode
+    #
+    # @param type [#to_s] :linux, :windows or :other
+    # @return [Boolean]
+    def delete_forbidden(type)
+      send(:"#{type}_delete_mode").to_sym == :none
```
jreidinger: better use `public_send` so you respect visibility

jreidinger: and why to_sym is needed? From documentation it looks like it should return always symbol

src/lib/y2storage/proposal_settings.rb
```diff
@@ -53,11 +99,43 @@ def initialize
...
+    # Whether the settings enforce deletion of a given type of partitions
+    #
+    # @see #windows_delete_mode
+    # @see #linux_delete_mode
+    # @see #other_delete_mode
+    #
+    # @param type [#to_s] :linux, :windows or :other
+    # @return [Boolean]
+    def delete_forced(type)
+      send(:"#{type}_delete_mode").to_sym == :all
```
jreidinger: same here.

src/lib/y2storage/proposal.rb
ancorgs: If we have to make a proposal and we have several possible candidates, we try with every one of them, capturing the exception if anything went wrong in order to try with the next one. I find it pretty natural.

The calculation of a proposal for a particular disk can go wrong for many reasons and at many different depth levels, so exceptions are the less cumbersome way to handle that. Otherwise, every single component would need an extra mechanism to communicate the possible errors, which sounds like re-inventing exceptions.

src/lib/y2storage/proposal_settings.rb
ancorgs: Well, `public_send` makes sense when sending messages to another object. But we are sending a message to ourselves, so it makes no much sense.

ancorgs: `to_sym` is not strictly needed. But it's harmless and save us from future changes in which somebody can decide to accept both `"none"` and `:none` as valid values for `#whatever_delete_mode`. I guess I have seen enough of that in Rails to become extra-vigilant :-)

src/lib/y2storage/proposal_settings.rb
ancorgs: oops, will fix it

src/lib/y2storage/proposal_settings.rb
jreidinger: OK with public send.
But regarding to_sym, if we want to support to accept both "none" and `:none`, then still I think that internally it should be consistent. so modify `{type}_delete_mode` to accept both, but storage only one. If we declare that method return `Symbol` we should rely on it as also other users of API will rely on it.

src/lib/y2storage/proposal.rb
jreidinger: well, OK. Even I don't find it natural. Especially to return only exception of the last candidate when all candidates failed. But not critical.

PR #219, opened 2017-04-25
Pmbr

jreidinger:


src/lib/y2storage/partition_tables/gpt.rb
```diff
@@ -29,6 +29,14 @@ module PartitionTables
     # This is a wrapper for Storage::Gpt
     class Gpt < Base
       wrap_class Storage::Gpt
+
+      # @!method pmbr_boot?
+      #   @return [Boolean] whether protective MBR flag is set
+      storage_forward :pmbr_boot?, to: :is_pmbr_boot
```
ancorgs: The SWIG bindings already take care of this renaming, so this should simply be:
```ruby
storage_forward :pmbr_boot?
```

src/lib/y2storage/partition_tables/gpt.rb
```diff
@@ -29,6 +29,14 @@ module PartitionTables
...
+
+      # @!method pmbr_boot=(value)
+      #   @attr value [Boolean] set/unset flag
+      storage_forward :pmbr_boot=, to: :set_pmbr_boot
```
ancorgs: Same than above, SWIG already renames setters to match Ruby conventions, so:
```ruby
storage_forward :pmbr_boot=
```

PR #221, opened 2017-04-26
Log graph

joseivanlopez:
* Save devicegraphs xml only when **inst_prepdisk** does a commit.
* Avoid creation lo xml files in tests.
* Add tests.

### DIFF
--- a/src/lib/y2storage/clients/inst_prepdisk.rb
+++ b/src/lib/y2storage/clients/inst_prepdisk.rb
@@ -52,6 +52,9 @@ module Y2Storage
 
       # Commits the actions to disk
       def commit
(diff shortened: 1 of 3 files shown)


aschnell:
We will need the logs also when the expert partitioner is used.

ancorgs:
But we don't want every single call to `commit` to overwrite the previous one. Do we? That's why I think the right approach would be to log all the calls to `commit` at `StorageManager` level, but directly in the log (y2log file), just like we log any other activity. There we have rotation and all the features you want to have in a log. I'm fine with having these two separate files for the installation case, since it's a specially useful one. But at `StorageManager` level I would just do `log.info staging.to_xml`.

PR #227, opened 2017-04-27
Fix bug with resize info

joseivanlopez:
Try to solve problem in openQA when disk has a Windows.

https://yast-openqa.suse.cz/tests/4939

### DIFF
--- a/src/lib/y2storage/device.rb
+++ b/src/lib/y2storage/device.rb
@@ -98,13 +98,11 @@ module Y2Storage
     #   @return [Boolean]
     storage_forward :exists_in_staging?
 
-    # @!method detect_resize_info
-    #   Information about the possibility of resizing a given device.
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/device.rb
```diff
@@ -149,6 +147,21 @@ def siblings
       storage_siblings(itself)
     end
 
+    # @!method detect_resize_info
+    #   Information about the possibility of resizing a given device. Returns
+    #   nil if the device does not exist in the probed devicegraph.
+    #   @see ResizeInfo
+    #   @see exists_in_probed?
+    #   @note This is slightly different from Storage::detect_resize_info, which
+    #     requires to be called in a device that belogs to the probed devicegraph.
```
ancorgs: belogs

PR #229, opened 2017-04-27
- added dummy activate callbacks

aschnell:
For https://trello.com/c/li8xjUAE/538-5-storageng-probe-inactive-luks.

Next step will be to implement popups for the callbacks but without any callbacks set no activation is done (not even for LVM) likely breaking tests.


aschnell:
The failing checks are due to requiring a newer libstorage-ng.

PR #235, opened 2017-05-04
Add callback to open luks

joseivanlopez:
As part of PBI https://trello.com/c/li8xjUAE/538-5-storageng-probe-inactive-luks


test/callbacks/activate_callbacks_test.rb
```diff
@@ -0,0 +1,72 @@
...
+
+require_relative "../spec_helper"
+require "y2storage/callbacks/activate_callbacks"
+
+describe Y2Storage::ActivateCallbacks do
+  subject { described_class.new }
+
+  describe "#luks" do
+    before do
+      allow_any_instance_of(Y2Storage::Dialogs::Callbacks::ActivateLuks)
```
ancorgs: Usage of `allow_any_instance_of` is discouraged. You can do, for example,
```ruby
let(:dialog) { whatever_you_want }

before do
  allow(Y2Storage::Dialogs::Callbacks::ActivateLuks).to receive(:new).and_return dialog
  allow(dialog).to whatever
end
```

src/lib/y2storage/callbacks/activate_callbacks.rb
```diff
@@ -21,18 +21,27 @@
...
 
-require "yast"
 require "storage"
+require "y2storage/dialogs/callbacks/activate_luks"
 
 module Y2Storage
-  # class to implement callbacks used during activate in ruby
+  # class to implement callbacks used during activate
   class ActivateCallbacks < Storage::ActivateCallbacks
+    include Yast::Logger
```
ancorgs: I just realized this file doesn't follow Ruby (and YaST) conventions. Is named
`lib/y2storage/callbacks/activate_callbacks.rb`, so it should define `Y2Storage::Callbacks::ActivateCallbacks` (which, btw, is redundant). Directory structure is supposed to follow the namespace structure.

So I would suggest

`lib/y2storage/callbacks/activate.rb` -> `Y2Storage::Callbacks::Activate` (I would go for this)

or

`lib/y2storage/activate_callbacks.rb` -> `Y2Storage::ActivateCallbacks` (also valid)

Same applies, of course, to the corresponding tests

test/dialogs/callbacks/activate_luks_test.rb
```diff
@@ -0,0 +1,90 @@
...
+        expect(Yast::UI).to receive(:ChangeWidget).once.with(Id(:accept), :Enabled, true)
+        subject.run
+      end
+
+      context "and dialog is accepted" do
+        let(:action) { :accept }
+
+        it "returns that password as encryption password" do
+          subject.run
+          expect(subject.encryption_password).to eq(password)
```
ancorgs: Maybe NP, but I would expect all those `expect(subject.encryption_password)` to be inside a `describe "#encryption_password"`, not inside a `describe "#run"`. I other words, I would expect the output to be something like
```
#run
  when a password is entered
    enables accept button
    and dialog is accepted
      returns :accept
    and dialog is not accepted
      returns :cancel
  when a password is not entered
    does not enable accept button
    and dialog is not accepted
      returns :cancel

#encryption_password
  before running the dialog
    returns nil
  after running the dialog
    if dialog was accepted
      returns the entered password
    if dialog was not accepted
      returns nil
```

src/lib/y2storage/callbacks/activate_callbacks.rb
aschnell: Hmm, I asked about the naming on IRC.

ancorgs: @aschnell Yes, and the only definitive point on that conversation was to follow the Ruby/YaST convention of having 1:1 relationship between namespaces and directories. Something that we finally didn't do:

```
<ancorgs>  I fine with creating a Y2Storage::Callbacks namespace (and, thus, subdirectory) then.
           As you wish
<ancorgs>  as said, without the full picture I don't have a strong opinion, whatever you decide
<ancorgs>  just make sure every namespace has its corresponding directory so the structures of the
           files tree matches the structure of the namespaces, as is standard in the Ruby world
<ancorgs>  CamelCase for the namespaces and snake_case for the equivalent directory. That's all
<arvin>    so e.g. src/lib/y2storage/callbacks/activate.rb
<ancorgs>  yes
<ancorgs>  which will define the Y2Storage::Callbacks::Activate class
```

src/lib/y2storage/dialogs/callbacks/activate_luks.rb
```diff
@@ -0,0 +1,119 @@
...
+                    _("The following device contain an encryption signature but the \n" \
+                        "password is not yet known.")
+                  )
+                ),
+                VSpacing(0.2),
+                Left(Label("UUID: #{uuid}")),
+                VSpacing(0.2),
+                Left(
+                  Label(
+                    _("The password need to be known if the device is needed either \n" \
```
joseivanlopez: Yes, it is a bit strange. Since it is a little different to the original one, which was refereed to several devices, I guess we could rewrite it.

test/callbacks/activate_callbacks_test.rb
joseivanlopez: Yes, much better.

src/lib/y2storage/callbacks/activate_callbacks.rb
aschnell: You also approved the PR. I don't know the Ruby world so good.

ancorgs: Sure, I was not pointing fingers or nothing like that. Don't take it personally. 😄 As I said, "I just realized", which implies somebody (me in this case, but no matter who) overlooked it in the previous PR.

Shit happens. :wink: That's (partially) why we use code review.

src/lib/y2storage/dialogs/callbacks/activate_luks.rb
```diff
@@ -0,0 +1,115 @@
...
+module Y2Storage
+  module Dialogs
+    module Callbacks
+      # Dialog to manage luks activation callbacks
+      class ActivateLuks < UI::Dialog
+        include SecretAttributes
+
+        secret_attr :encryption_password
+
+        def initialize(uuid, attempt)
```
ancorgs: NP: `attempt` is actually never used. Just saying, I'm not against having it here.

src/lib/y2storage/dialogs/callbacks/activate_luks.rb
```diff
@@ -0,0 +1,115 @@
...
+                        "password is not yet known.")
+                  )
+                ),
+                VSpacing(0.2),
+                Left(Label("UUID: #{uuid}")),
+                VSpacing(0.2),
+                Left(
+                  Label(
+                    _("The password needs to be known if the device is needed either \n" \
+                        "during an update or if it contains an encrypted volume.")
```
ancorgs: Just a possible suggestion (I suck at this): "The password is needed if the device contains a system to be updated or a LVM physical volume to be used during installation."

src/lib/y2storage/dialogs/callbacks/activate_luks.rb
joseivanlopez: Right, in current dialog version it is not used, but maybe would be a good idea show the number of failed attempts for the current device.

src/lib/y2storage/dialogs/callbacks/activate_luks.rb
joseivanlopez: Much better thanks!

src/lib/y2storage/dialogs/callbacks/activate_luks.rb
kobliha: Once you start telling the user number of failed attempts, they will start asking for `How much do I have left?` and start feeling the urge, and start feeling the discomfort. We don't want our users to feel bad about good things (we even want them to feel good about bad ones ;)).

src/lib/y2storage/dialogs/callbacks/activate_luks.rb
```diff
@@ -0,0 +1,115 @@
...
+                        "password is not yet known.")
+                  )
+                ),
+                VSpacing(0.2),
+                Left(Label("UUID: #{uuid}")),
+                VSpacing(0.2),
+                Left(
+                  Label(
+                    _("The password is needed if the device contains a system to be\n" \
+                        "updated or belongs to a LVM to be used during installation.")
```
kobliha: BTW, minor, `a LVM` --> `an LVM`

src/lib/y2storage/dialogs/callbacks/activate_luks.rb
joseivanlopez: Yes, maybe it is not so good idea to show the attempts :).

src/lib/y2storage/dialogs/callbacks/activate_luks.rb
joseivanlopez: Ok, I am going to fix it. Thanks!

PR #239, opened 2017-05-11
Fixed a type error in PartitionTreeView#create

mvidner:
Expected argument 0 of type unsigned long long, but got Y2Storage::DiskSize

### DIFF
--- a/src/lib/expert_partitioner/tree_views/partition.rb
+++ b/src/lib/expert_partitioner/tree_views/partition.rb
@@ -37,7 +37,7 @@ module ExpertPartitioner
 
     def create
       tmp = ["Name: #{@partition.name}",
-             "Size: #{::Storage.byte_to_humanstring(@partition.size, false, 2, false)}"]
+             "Size: #{::Storage.byte_to_humanstring(@partition.size.to_i, false, 2, false)}"]
 
       @partition.udev_paths.each_with_index do |udev_path, i|
         tmp << "Device Path #{i + 1}: #{udev_path}"
--- a/src/lib/expert_partitioner/tree_views/partition.rb
+++ b/src/lib/expert_partitioner/tree_views/partition.rb
@@ -37,7 +37,7 @@ module ExpertPartitioner
 
     def create
       tmp = ["Name: #{@partition.name}",
-             "Size: #{::Storage.byte_to_humanstring(@partition.size.to_i, false, 2, false)}"]
+             "Size: #{@partition.size.to_human_string}"]
 
       @partition.udev_paths.each_with_index do |udev_path, i|
         tmp << "Device Path #{i + 1}: #{udev_path}"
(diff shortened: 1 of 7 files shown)


mvidner:
I needed to fix this while working on https://github.com/yast/yast-partitioner/pull/2 (commit https://github.com/yast/yast-partitioner/pull/2/commits/cd0e2b37b5249191909cf4dd54fc70972f7415e9)

src/lib/expert_partitioner/tree_views/partition.rb
```diff
@@ -37,7 +37,7 @@ def initialize(partition)
 
     def create
       tmp = ["Name: #{@partition.name}",
-             "Size: #{::Storage.byte_to_humanstring(@partition.size, false, 2, false)}"]
+             "Size: #{::Storage.byte_to_humanstring(@partition.size.to_i, false, 2, false)}"]
```
ancorgs: `@partition.size` is a DiskSize object. You don't need to convert it back to integer and then use a libstorage function to format it. You can simply do `@partition.size.to_human_string`.

See http://www.rubydoc.info/github/yast/yast-storage-ng/master/Y2Storage/DiskSize

mvidner: Ha! Indeed, thanks.

aschnell: And if ```@partition``` is a libstorage-ng Partition object you can also just use ```@partition.size_string```.

ancorgs:
And BTW, how did we get to the situation in which the expert partitioner received a `DiskSize` object? Shouldn't the expert partitioner be using only pure libstorage (SWIG) objects? We are not adapting it to the wrapped classes. We are writing yast-partitioner instead, isn't it?

mvidner:
@ancorgs good question. The easy answer is "because I am a beginner in Storage", but obviously that's not much helpful :) Maybe I am doing it all wrong by mixing and connecting inappropriate layers. Please see https://github.com/yast/yast-partitioner/pull/2/files and tell me whether the layering makes sense.

If we use "volume group" as example, hmm, are you saying I should use [Storage::LvmVg](http://www.rubydoc.info/github/yast/yast-storage-ng/Storage/LvmVg) instead of [Y2Storage::LvmVg](http://www.rubydoc.info/github/yast/yast-storage-ng/Y2Storage/LvmVg) ?

ancorgs:
The part that doesn't make sense is that you are using `ExpertPartitioner` in yast-partitioner.

The expert partitioner currently living in the yast-storage-ng repository (and defining that namespace) is just a prototype created to play with the libstorage classes.

The goal of yast-partitioner is to REPLACE that prototype with another one that looks like the one in the old yast-storage. Not to USE that prototype. That's what the PBI is about - creating the dialogs from scratch (copying and pasting UI code from the old partitioner is allowed). These new dialogs should work on top of the Y2Storage classes, that's right.

mvidner:
OK! So I will instead copy the fixed code to https://github.com/yast/yast-partitioner

PR #244, opened 2017-05-17
Human string partition table

jreidinger:


ancorgs:
In general, this LGTM. But in [Filesystems::Type](https://github.com/yast/yast-storage-ng/blob/d549d5aed5e950539c88603a7c9086759e13bd94/src/lib/y2storage/filesystems/type.rb) `#to_human` is used instead of `#to_human_string`. We should unify that. What about using always `#to_human_string` (more consistent with libstorage), having `#to_human` like an alias (I assume some people prefer the latter, since it's there)?

jreidinger:
@ancorgs fine for me. In fact I do not like to_human as I expect that it return human object. So prefered one is `to_human_string` or `to_human_s` to be consistent with `to_s`. But whatever name you pick, I think it is good to have it unified and consistent. Btw I pick `to_human_string` because DiskSize use it.

PR #245, opened 2017-05-18
Add Planned::Device to enable level 3 of AutoYaST and fix the shadowed subvolumes bug

ancorgs:
Better review commit by commit. They should all make sense and "tell the story".

This PR addresses two PBIs:

1) https://trello.com/c/pYefPIfV/593-3-finish-pending-pr-classes-for-proposed-partitions-and-lvs-179

This replaces #179 by implementing all the improvements present there. The ideas and a lot of the code was reused, but the commits were not reused because it was very hard to apply then cleanly. So this complies to [Iván's gist](https://gist.github.com/joseivanlopez/4b1f1091d212b10b0970cdd691ea2f2e) and provides the `Planned::Partition`, `Planned::LvmLv` and `Planned::BtrfsSubvolume` that we will need to implement level 3 of AutoYaST in the next sprint.

2) https://trello.com/c/ywolMrUx/578-2-storage-ng-no-shadowed-subvolumes

Now the shadowed subvolumes are removed from the list of planned devices (formerly known as the list of planned volumes) as soon as the list is calculated (in `PlannedDevicesGenerator`, formerly known as `VolumesGenerator`).

Manually tested. This is the summary if no separate home partition is proposed (there is a "home" subvolume).
no-separate-home

And this is the summary with a separate home partition (the "home" subvolume is no proposed).
separate-home


src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -0,0 +1,145 @@
...
+      # @return [Filesystems::BlkFilesystem] filesystem
+      def create_filesystem(blk_dev)
+        return nil unless filesystem_type
+
+        filesystem = blk_dev.create_blk_filesystem(filesystem_type)
+        filesystem.mountpoint = mount_point if mount_point && !mount_point.empty?
+        filesystem.label = label if label
+        filesystem.uuid = uuid if uuid
+
+        btrfs_setup(filesystem)
```
ancorgs: The first line of `btrfs_setup` is already `return unless filesystem.supports_btrfs_subvolumes?` Is there a need to check in both places?

src/lib/y2storage/proposal/planned_devices_generator.rb
```diff
@@ -0,0 +1,223 @@
+#!/usr/bin/env ruby
+#
+# encoding: utf-8
+
+# Copyright (c) [2016] SUSE LLC
```
ancorgs: This is a renamed file, not really a new one. The original is from 2016 and we usually don't update copyright note. Of course, I can do it for this one.

src/lib/y2storage/planned/can_be_formatted.rb
joseivanlopez: Ok, it is not necessary.

PR #247, opened 2017-05-24
enhance documentation of DiskSize class

wfeldt:
There's no code change except for the definition of #to_storage_value which was done so yardoc sees it.

### DIFF
--- a/src/lib/y2storage/disk_size.rb
+++ b/src/lib/y2storage/disk_size.rb
@@ -33,6 +33,16 @@ module Y2Storage
   # allowed in principle but the special value -1 is reserved to mean
   # 'unlimited'.
   #
+  # {DiskSize} objects are used through {Y2Storage} whenever a size has to be specified.
+  # Notable exception here is class {Region} where several methods expect Integer arguments.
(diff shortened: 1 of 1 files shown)


src/lib/y2storage/disk_size.rb
```diff
@@ -33,6 +33,15 @@ module Y2Storage
   # allowed in principle but the special value -1 is reserved to mean
   # 'unlimited'.
   #
+  # {DiskSize} objects are used through {Y2Storage} whenever a size has to be specified.
+  # Notable exceptions here are{Region#start} and {Region#length} in class {Region}, which are +Integer+.
```
jreidinger: what about Region#last?

jreidinger: ah, not last, but end - https://github.com/yast/yast-storage-ng/blob/aacfb73ee25b43c6c6a415c83622ef1f4785d9a6/src/lib/y2storage/region.rb#L59
which should be also integer.

mvidner:
@wfeldt awesome! Thank you!

lslezak:
Wow, great work, thanks a lot @wfeldt ! :medal_sports:

PR #248, opened 2017-06-01
add test for reading of stripes

jreidinger:
it is demonstration of current failure.

Let me also a bit complain about logic in DiskSize[1], which expect that anything that is not DiskSize or String will have method round, which is not much helpful for debugging and also robust. It will be much nicer to simply raise ArgumentError with description what it actually get ( in this case it looks like `nil` ).

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/disk_size.rb#L98


ancorgs:
Replaced by #249

PR #249, opened 2017-06-01
Do not try to convert setter result in the wrapper

ancorgs:
Converting the result of the setter is not a good idea. Assignations should always return the left part of the expression, with no additional conversion.

### DIFF
--- a/src/lib/y2storage/disk_size.rb
+++ b/src/lib/y2storage/disk_size.rb
@@ -71,6 +71,7 @@ module Y2Storage
     # @!scope class
     #
     # Accepts +Numeric+, +Strings+, or {DiskSize} objects as initializers.
+    # @raise [ArgumentError] if anything else is used as initializer
     #
     # @see initialize
     # @see parse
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/disk_size.rb
```diff
@@ -91,13 +92,15 @@ class DiskSize
...
-      end
+      @size =
+        if size.is_a?(Y2Storage::DiskSize)
+          size.to_i
+        elsif size.is_a?(::String)
+          Y2Storage::DiskSize.parse(size).size
+        elsif size.respond_to?(:round)
+          size.round
+        end
+      raise(ArgumentError, "Cannot get the bytes count for #{size.inspect}") unless @size
```
jreidinger: in fact you can have it in else, no need for additional check if size

ancorgs: Somehow, I didn't like to have a `raise` in one of the branches of an assignation. A matter of taste, I guess.

jreidinger: ok

PR #250, opened 2017-06-02
Doc improvements for StorageEnumWrapper, Region, PartitionType.

mvidner:


src/lib/y2storage/partition.rb
```diff
@@ -99,6 +99,8 @@ class Partition < BlkDevice
     storage_forward :legacy_boot=
 
     # @!method self.create(devicegraph, name, region, type)
```
mvidner: Ancor said we could even remove this method as it seems not used

PR #252, opened 2017-06-06
- removed wrong test

aschnell:
The test checks for the name of a partition which cannot be created. In other words checks for not valid data.


ancorgs:
So the behavior of `#inpect` changed in this example because now partition table contain better checks. We should probably then adapt the test instead of removing part of it.

aschnell:
The behaviour changed because inspect outputs invalid data.

ancorgs:
If `#inspect` outputs invalid data, then `#inspect` should be fixed (instead of removing the test in order to ignore the fact that the method is wrong).

aschnell:
Well, inspect outputs all data of the struct which is ok for an inspect function. But checking that output in unit tests looks wrong to me.

PR #254, opened 2017-06-06
Don't insist on installing filesystem support packages that are not available in any repo (bsc#1039830)

shundhammer:
Port to Storage-NG:

https://trello.com/c/yqdIek3N/606-2-handle-optional-fs-packages-correctly

https://bugzilla.suse.com/show_bug.cgi?id=1039830

### DIFF
--- a/src/lib/y2storage/package_handler.rb
+++ b/src/lib/y2storage/package_handler.rb
@@ -74,7 +74,14 @@ module Y2Storage
     def add_feature_packages(devicegraph)
       used_features = UsedStorageFeatures.new(devicegraph)
       packages = used_features.feature_packages
-      packages = packages.delete_if { |pkg| unavailable_optional_package?(pkg) }
+      packages.delete_if do |pkg|
+        if unavailable_optional_package?(pkg)
+          log.warn("WARNING: Skipping unavailable filesystem support package #{pkg}")
+          true
+        else
+          false
+        end
+      end
       add_packages(packages)
     end
 
@@ -85,10 +92,8 @@ module Y2Storage
     # @return [Boolean] true if package is optional and unavailable,
     #                   false if not optional or if available.
     def unavailable_optional_package?(package)
-      return false unless UsedStorageFeatures::OPTIONAL_PACKAGES.include?(package)
-      return false if Yast::Package.Available(package)
-      log.warn("WARNING: Skipping unavailable filesystem support package #{package}")
-      true
+      return false unless UsedStorageFeatures.optional_package?(package)
+      !Yast::Package.Available(package)
     end
 
     # Start a package dependency resolver run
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/package_handler.rb
```diff
@@ -73,7 +73,22 @@ def add_packages(pkg_list)
     #
     def add_feature_packages(devicegraph)
       used_features = UsedStorageFeatures.new(devicegraph)
-      add_packages(used_features.feature_packages)
+      packages = used_features.feature_packages
+      packages = packages.delete_if { |pkg| unavailable_optional_package?(pkg) }
```
ancorgs: NP: `delete_if` already modifies the caller, so there is no need to assign it again. In other words
```ruby
packages = used_features.feature_packages
packages.delete_if { |pkg| unavailable_optional_package?(pkg) }
```

src/lib/y2storage/package_handler.rb
```diff
@@ -73,7 +73,22 @@ def add_packages(pkg_list)
...
+    end
+
+    # Check if a package is an optional package that is unavailable.
+    # See also bsc#1039830
+    #
+    # @param package [String] package name
+    # @return [Boolean] true if package is optional and unavailable,
+    #                   false if not optional or if available.
+    def unavailable_optional_package?(package)
+      return false unless UsedStorageFeatures::OPTIONAL_PACKAGES.include?(package)
```
ancorgs: NP: To fully honor the Uniform Access Principle, this should be a class method, instead of accessing directly to the constant from outside of `UsedStorageFeatures`

src/lib/y2storage/package_handler.rb
```diff
@@ -73,7 +73,22 @@ def add_packages(pkg_list)
...
+      return false if Yast::Package.Available(package)
+      log.warn("WARNING: Skipping unavailable filesystem support package #{package}")
```
ancorgs: NP: IMO it should be a responsibility of the caller to log the fact that this package is being removed from the list (skipped). This method should only return `true` or `false` and make no assumptions about the consequences.

In other words, the code deleting something from the list and the code logging that something was deleted should be in the same place. This method is actually not skipping the package, the `delete` is doing that.

src/lib/y2storage/package_handler.rb
shundhammer: Ah, right. I was unsure about that. Changing.

src/lib/y2storage/package_handler.rb
shundhammer: I thought about that, but having a one-line method for that purpose seemed overkill to me. But then, it doesn't cost us much to do that. Changing.

src/lib/y2storage/package_handler.rb
shundhammer: Right. Changing.

PR #255, opened 2017-06-08
Add `BlkDevice#to_be_formatted?`

ancorgs:
2 lines of code, 169 of RSpec. :smile:

### DIFF
--- a/test/blk_device_test.rb
+++ b/test/blk_device_test.rb
@@ -24,6 +24,8 @@ require_relative "spec_helper"
 require "y2storage"
 
 describe Y2Storage::BlkDevice do
+  using Y2Storage::Refinements::SizeCasts
+
   before do
     fake_scenario("complex-lvm-encrypt")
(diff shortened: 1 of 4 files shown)


test/blk_device_test.rb
```diff
@@ -24,6 +24,8 @@
 require "y2storage"
 
 describe Y2Storage::BlkDevice do
+  using Y2Storage::Refinements::SizeCasts
```
jreidinger: why it is needed?

ancorgs: Just convenience. To be able to write

```ruby
vg1.create_lvm_lv("newlv", 1.GiB)
```
down there (lines 135, 169, 214...), instead of
```ruby
vg1.create_lvm_lv("newlv", Y2Storage::DiskSize.GiB(1))
```

jreidinger: ok

PR #256, opened 2017-06-09
Add partitions proposal client

joseivanlopez:
PBI: https://trello.com/c/Bf7k0xKs/601-5-storageng-showing-proposal-in-the-general-proposal-overview-primarily-autoyast


src/lib/y2storage/dialogs/actions_presenter.rb
```diff
@@ -0,0 +1,127 @@
...
+# find current contact information at www.novell.com.
+
+require "yast"
+require "y2storage"
+
+Yast.import "HTML"
+
+module Y2Storage
+  # Presenter for actions of an actiongraph
+  class ActionsPresenter
```
ancorgs: The namespace does not match the file location.

src/lib/y2storage/clients/partitions_proposal.rb
```diff
@@ -0,0 +1,117 @@
...
+
+Yast.import "Popup"
+
+module Y2Storage
+  module Clients
+    # Proposal client to show the list of storage actions
+    class PartitionsProposal < ::Installation::ProposalClient
+      include Yast::Logger
+
+      def self.state(storage_manager)
```
ancorgs: Passing the storage_manager as argument looks weird. It's a singleton class, after all.

src/lib/y2storage/clients/partitions_proposal.rb
```diff
@@ -0,0 +1,117 @@
...
+        self.staging_revision = storage_manager.staging_revision
+
+        staging = storage_manager.y2storage_staging
+        actiongraph = staging ? staging.actiongraph : nil
+        self.actions_presenter = ActionsPresenter.new(actiongraph)
+      end
+
+      def initialize
+        create_proposal
+        self.class.state(storage_manager)
```
ancorgs: Forcing the call to `.state` here (a method that is supposed to just return something) in order to initialize looks weird.

src/lib/y2storage/clients/partitions_proposal.rb
```diff
@@ -0,0 +1,117 @@
...
+          "preformatted_proposal" => actions_presenter.to_html,
+          "links"                 => actions_presenter.events,
+          "language_changed"      => false
+        }
+      end
+
+      def ask_user(param)
+        event = param["chosen_id"]
+
+        if actions_presenter.events.include?(event)
```
ancorgs: what about adding `ActionsPresenter#can_handle?` instead of having that `.events.include?` in several places?

src/lib/y2storage/clients/partitions_proposal.rb
```diff
@@ -0,0 +1,117 @@
...
+          actions_presenter.update_status(event)
+          result = :again
+        else
+          Yast::Report.Warning(_("This action is not enabled at this moment"))
```
ancorgs: Write something to the logs, including the name of the unhandled event (it would maybe be useful to show the name in the pop-up as well).

src/lib/y2storage/clients/partitions_proposal.rb
ancorgs: Even more weird: `state` (that looks like a method to query something) is called to actually initialize a value, but `create_proposal`(the looks like a method that indeed creates something) in many cases will actually not create a proposal (if there is already one).

src/lib/y2storage/clients/partitions_proposal.rb
```diff
@@ -0,0 +1,140 @@
...
+    #
+    # To solve that, instance class attributes are used. PartitionsProposal
+    # class has an internal state that is updated with each object creation.
+    #
+    # @see PartitionsProposal.update_state
+    class PartitionsProposal < ::Installation::ProposalClient
+      include Yast::Logger
+
+      def initialize
+        if storage_manager.proposal.nil?
```
ancorgs: This implies displaying this proposal always destroys the result of the expert partitioner, so expert partitioner simply cannot be used. Isn't it?

Take a look to how this is solved in `inst_disk_proposal` that, as proved by its unit tests, only proposes something when really needed, respecting a manually set staging devicegraph.
https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/clients/inst_disk_proposal.rb#L41

src/lib/y2storage/clients/partitions_proposal.rb
```diff
@@ -0,0 +1,145 @@
...
+        #
+        # @see ActionsPresenter
+        def update_state
+          storage_manager = StorageManager.instance
+          return if staging_revision == storage_manager.staging_revision
+
+          self.staging_revision = storage_manager.staging_revision
+
+          staging = storage_manager.y2storage_staging
+          actiongraph = staging ? staging.actiongraph : nil
```
ancorgs: First of all, I miss a RSpec example for `ActionsPresenter` explaining what is returned by `#to_html` when the object is initialized with `nil`. I assume `#to_html` returns the empty string or an empty `<p>`.

That being said, is that the expected output of `make_proposal` in that case or show this client show some error message (even an error blocking the installation to continue)?

PR #257, opened 2017-06-12
add root_ok? and home_ok? methods to Y2Storage::Filesystems

wfeldt:
cf. https://github.com/yast/yast-update/pull/74


ancorgs:
Wouldn't it make more sense to have the `root_ok?` and `home_ok?` methods in the `Filesystems::Type` class? Of course, then we can have methods in `Filesystems::Base` delegating the question to its type. 

But the responsibility of knowing if a type is ok or not belongs to the `Type`class (after all, this started as a way of not exposing `root_filesystems` and this implementation actually forces us to expose it).

wfeldt:
replaced by https://github.com/yast/yast-storage-ng/pull/258

PR #259, opened 2017-06-12
WIP: AutoYaST proposal

ancorgs:


src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+  #   proposal = Y2Storage::AutoinstProposal.new(partitioning)
+  #   proposal.proposed?            # => false
+  #   proposal.proposed_devicegraph # => nil
+  #   proposal.planned_devices      # => nil
+  #
+  #   proposal.propose              # => Performs the calculation
+  #
+  #   proposal.proposed?            # => true
+  #   proposal.proposed_devicegraph # => Proposed layout
+  class AutoinstProposal
```
jreidinger: what is reason to not use common API for auto clients? Same as use other modules with common methods like read/write import/export etc.?

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+
+    alias_method :devices, :proposed_devicegraph
+
+    # Constructor
+    #
+    # @param partitioning [Array<Hash>] Partitioning schema from an AutoYaST profile
+    # @param devicegraph  [Devicegraph] starting point. If nil, then probed devicegraph
+    #   will be used
+    # @param disk_analyzer [DiskAnalyzer] if nil, a new one will be created
+    #   based in the initial devicegraph
```
jreidinger: "based on"

src/lib/y2storage/autoinst_proposal.rb
imobachgs: It is an Storage proposal to be used during autoinstallation, not an AutoYaST proposal, so they're not related.

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+    #
+    # @raise [UnexpectedCallError] if called more than once
+    # @raise [NoDiskSpaceError] if there is no enough space to perform the installation
+    def propose
+      raise UnexpectedCallError if proposed?
+
+      drives = Proposal::AutoinstDrivesMap.new(initial_devicegraph, partitioning)
+
+      space_maker = Proposal::AutoinstSpaceMaker.new(disk_analyzer)
+      devicegraph = space_maker.provide_space(initial_devicegraph, drives)
```
jreidinger: API is very confusing. Why `provide_space` method return `devicegraph`? It doesn't make any sense.

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+
+      @proposed_devicegraph = propose_devicegraph(devicegraph, drives)
```
jreidinger: another confusing method. Why propose_devicegraph get as parameter device graph? and if so, why it need pre-processed device graph and not initial one?

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+      @proposed = true
+
+      nil
```
jreidinger: why explicit nil is needed?

src/lib/y2storage/autoinst_proposal.rb
imobachgs: The idea of an SpaceMaker is to free some space (removing partitions) on a devicegraph, returning a new and modified devicegraph.

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+    # This method falls back to #proposed_guided_devicegraph when the device map
+    # does not contain any partition.
+    #
+    # @param devicegraph [Devicegraph]                 Starting point
+    # @param drives      [Proposal::AutoinstDrivesMap] Devices map from an AutoYaST profile
+    # @return [Devicegraph] Devicegraph containing the planned devices
+    def propose_devicegraph(devicegraph, drives)
+      planned_devices = plan_devices(devicegraph, drives)
+      create_devices(devicegraph, planned_devices, drives.disk_names)
+    rescue Error
```
jreidinger: hmm, sorry, but no. This is not correct for more reasons:

1. what actually is wanted here is rescue only if there are no partitions specified. Why simple not check drives, if there is any partition and do not abuse exception for such case?

2. it rescue all Errors and always fallback to guided setup. What if error is in autoyast profile? Simply ignoring error and use different proposal is very from from my POV, as result is something that user do not want.
What's more, Error is also e.g. NoDiskSpaceError, which means, that there is not enough free space and it use instead different formatting then valid user specified, which is bad if user later depends on it e.g. in some of its autoyast script, which then failed which strange error because autoyast do not do what is requested in profile.

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+    # If the :desired target fails, it will retry with the :min target.
+    #
+    # @param devicegraph [Devicegraph]       Starting point
+    # @param drives      [AutoinstDrivesMap] Devices map from an AutoYaST profile
+    # @return [Devicegraph] Proposed devicegraph using the guidede proposal approach
+    #
+    # @raise [Error] No suitable devicegraph was found
+    # @see proposed_guided_devicegraph
+    def propose_guided_devicegraph(devicegraph, drives)
+      guided_devicegraph_for_target(devicegraph, drives, :desired)
```
jreidinger: API question. Why planner by default do not have this behavior? that desired means use ideal and if not fit, use min? Why it is caller responsibility?

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+    rescue Error
```
jreidinger: why generic error? I expect here `NoDiskSpaceError`

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+      guided_devicegraph_for_target(devicegraph, drives, :min)
+    end
+
+    # Calculates list of planned devices
+    #
+    # If the list does not contain any partition, then it follows the same
+    # approach than the guided proposal
```
jreidinger: maybe "as" or "like" instead of "than"?

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+    #
+    # @param devicegraph [Devicegraph]                 Starting point
+    # @param drives      [Proposal::AutoinstDrivesMap] Devices map from an AutoYaST profile
+    # @return [Array<Planned::Device>] Devices to add
+    def plan_devices(devicegraph, drives)
+      planner = Proposal::AutoinstDevicesPlanner.new(devicegraph)
+      planner.planned_devices(drives)
+    end
+
+    # Creates a devicegraph using the same approach than guided partitioning
```
jreidinger: same here

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+    #
+    # @param devicegraph [Devicegraph]       Starting point
+    # @param drives      [AutoinstDrivesMap] Devices map from an AutoYaST profile
+    # @param target      [Symbol] :desired means the sizes of the partitions should
+    #   be the ideal ones, :min for generating the smallest functional partitions
```
jreidinger: it in fact return `Devicegraph` is it intention? create_devices returns it, but this one not. Confusing for me, what is intention.

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+      guided_planner = Proposal::PlannedDevicesGenerator.new(guided_settings, devicegraph)
+      planned_devices = guided_planner.planned_devices(target)
+      create_devices(devicegraph, planned_devices, drives.disk_names)
+    end
+
+    # Creates planned devices on a given devicegraph
+    #
+    # @param devicegraph     [Devicegraph]            Starting point
+    # @param planned_devices [Array<Planned::Device>] Devices to add
+    # @return [Devicegraph]
```
jreidinger: Why it return Devicegraph when it modify passed one?

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+      guided_settings = proposal_settings_for_disks(drives.disk_names)
+      guided_planner = Proposal::PlannedDevicesGenerator.new(guided_settings, devicegraph)
+      planned_devices = guided_planner.planned_devices(target)
+      create_devices(devicegraph, planned_devices, drives.disk_names)
+    end
+
+    # Creates planned devices on a given devicegraph
+    #
+    # @param devicegraph     [Devicegraph]            Starting point
+    # @param planned_devices [Array<Planned::Device>] Devices to add
```
jreidinger: disk names not documented, why it is needed?

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+    # @return [Devicegraph]
+    def create_devices(devicegraph, planned_devices, disk_names)
+      devices_creator = Proposal::AutoinstDevicesCreator.new(devicegraph)
+      devices_creator.devicegraph(planned_devices, disk_names)
```
jreidinger: this API is really confusing. There is `Proposal::AutoinstDevicesCreator` which is fine. It is something create devices. OK. it have argument devicegraph on which it operates. Still fine.
But then there is method `devicegraph`, why? creator do not create devicegraph! it modify passed one.  why simple do not have method `create` that gets arguments what to create. and if devicegraph needs to be accessed, then have it as attr_reader.

So API will look like:
```
devices_creator = Proposal::AutoinstDevicesCreator.new(devicegraph)
devices_creator.create(planned_devices, disk_names)

devices_creator.devicegraph
```

src/lib/y2storage/autoinst_proposal.rb
imobachgs: Take into account that `initial_devicegraph` is not modified. So if you use that one, you will be missing the changes by `AutoinstSpaceMaker`.

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+    # Constructor
+    #
+    # @param partitioning [Array<Hash>] Partitioning schema from an AutoYaST profile
+    # @param devicegraph  [Devicegraph] starting point. If nil, then probed devicegraph
+    #   will be used
+    # @param disk_analyzer [DiskAnalyzer] if nil, a new one will be created
+    #   based in the initial devicegraph
+    def initialize(partitioning: [], devicegraph: nil, disk_analyzer: nil)
+      @partitioning = partitioning
+      @initial_devicegraph = devicegraph
```
jreidinger: if `devicegraph` is nil, where it is assigned to probed one? I cannot find it in code.

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+      devices_creator.devicegraph(planned_devices, disk_names)
+    end
+
+    # Returns the product's proposal settings for a given set of disks
+    #
+    # @param disk_names [Array<String>] Disks names to consider
+    # @return [ProposalSettings] Proposal settings considering only the given disks
+    def proposal_settings_for_disks(disk_names)
+      settings = ProposalSettings.new_for_current_product
+      settings.candidate_devices = disk_names
```
jreidinger: Do I get it correctly that here it limit number of devices to be used from autoyast profile, where it can specify which disk only use?

src/lib/y2storage/boot_requirements_checker.rb
```diff
@@ -61,26 +67,36 @@ def needed_partitions(target = :desired)
...
-      elsif arch.s390?
-        @strategy = BootRequirementsStrategies::ZIPL.new(settings, devicegraph)
-      elsif arch.ppc?
-        @strategy = BootRequirementsStrategies::PReP.new(settings, devicegraph)
-      end
-
-      # Fallback to Legacy as default
-      @strategy ||= BootRequirementsStrategies::Legacy.new(settings, devicegraph)
+      klass =
+        if arch.x86? && arch.efiboot?
```
jreidinger: is there any reason to use libstorage-arch instead of yast2 Arch class? I do not see it documented at https://github.com/ancorgs/yast-storage-ng/blob/272abb79b527f2bf0e869f74b029c81ac68936fa/src/lib/y2storage/storage_manager.rb

src/lib/y2storage/boot_requirements_checker.rb
```diff
@@ -61,26 +67,36 @@ def needed_partitions(target = :desired)
...
-    attr_reader :settings
+    # @return [Devicegraph] starting situation.
     attr_reader :devicegraph
 
+    # @return [Array<Planned::Device>] devices that are already planned to be
+    #   added to the starting devicegraph.
+    attr_reader :planned_devices
+
+    # @return [String, nil] device name of the disk that the system will try to
+    #   boot first. Only useful in some scenarios like legacy boot.
```
jreidinger: please document what happens if there is nil. How it then pick booting disk? auto-detection or simple picking first one?

src/lib/y2storage/boot_requirements_checker.rb
jreidinger: btw this is broken for aarm64, as it use also EFI for booting.

src/lib/y2storage/boot_requirements_checker.rb
```diff
@@ -61,26 +67,36 @@ def needed_partitions(target = :desired)
...
-
-      # Fallback to Legacy as default
-      @strategy ||= BootRequirementsStrategies::Legacy.new(settings, devicegraph)
+      klass =
+        if arch.x86? && arch.efiboot?
+          BootRequirementsStrategies::UEFI
+        elsif arch.s390?
+          BootRequirementsStrategies::ZIPL
+        elsif arch.ppc?
+          BootRequirementsStrategies::PReP
```
jreidinger: also wrong, as ppc64 `powernv` board do not need prep. IT is already supported in SP2, so it is regression.

src/lib/y2storage/autoinst_proposal.rb
imobachgs: That information is read from the AutoYaST profile, like this:

```xml
<drive>
  <device>/dev/sda</device>
  <use>all</use>
</drive>
```

Then, we use a `AutoinstDiskMap` to store the information from that profile and it offers a `#disk_names` method.

src/lib/y2storage/autoinst_proposal.rb
imobachgs: Ok, I'll add a note. The idea is to consider only the disks which are mentioned in the AutoYaST profile.

src/lib/y2storage/boot_requirements_checker.rb
```diff
@@ -61,26 +67,36 @@ def needed_partitions(target = :desired)
...
+        else
+          # Fallback to Legacy as default
+          BootRequirementsStrategies::Legacy
+        end
+      @strategy ||= klass.new(devicegraph, planned_devices, boot_disk_name)
```
jreidinger: why `||=`? it is already checked above.

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -0,0 +1,199 @@
...
+    # information (regarding calculation of boot requirements) about the
+    # expected final system.
+    class Analyzer
+      # Constructor
+      #
+      # @param devicegraph     [Devicegraph] starting situation.
+      # @param planned_devices [Array<Planned::Device>] devices that are already planned to be
+      #   added to the starting devicegraph.
+      # @param boot_disk_name  [String, nil] device name of the disk that the system will try to
+      #   boot first. Only useful in some scenarios like legacy boot.
```
jreidinger: also document nil here. I am fine if in previous method you just link this one via `see`

src/lib/y2storage/autoinst_proposal.rb
imobachgs: I just forgot the comment.

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -0,0 +1,199 @@
...
+        @boot_disk ||= boot_disk_from_devicegraph
+        @boot_disk ||= devicegraph.disks.first
+
+        @boot_disk
+      end
+
+      # Whether the root (/) filesystem is going to be in a LVM logical volume
+      #
+      # @return [Boolean] true if the root filesystem is going to be in a LVM
+      #   logical volume.
```
jreidinger: it should be mentioned here, that if root is not specified in graph or in planned devices, it returns false.

jreidinger: and same for other methods

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -0,0 +1,199 @@
...
+          false
+        end
+      end
+
+      # Whether the root (/) filesystem is going to be Btrfs
+      #
+      # @return [Boolean] true if the root filesystem is going to be Btrfs
+      def btrfs_root?
+        if root_planned_dev
+          root_planned_dev.respond_to?(:filesystem_type) && root_planned_dev.filesystem_type.is?(:btrfs)
```
jreidinger: can it happen that device do not respond to `filesystem_type` ? actually from my POV, problem can be that `filesystem_type` is nil and then `root_planned_dev.filesystem_type.is?` crashes with access nil.

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -0,0 +1,199 @@
...
+      # device or by any device in the devicegraph
+      #
+      # @param path [String] mount point to check for
+      # @return [Boolean]
+      def free_mountpoint?(path)
+        return false if planned_devices.any? do |dev|
+          dev.mount_point && File.identical?(dev.mount_point, path)
+        end
+        return false if devicegraph.filesystems.any? do |fs|
+          fs.mountpoint && File.identical?(fs.mountpoint, path)
```
jreidinger: :+1: for File usage.

src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -22,61 +22,61 @@
...
 require "y2storage/disk_size"
 require "y2storage/filesystems/type"
 require "y2storage/planned"
+require "y2storage/boot_requirements_strategies/analyzer"
+require "y2storage/exceptions"
 
 module Y2Storage
   module BootRequirementsStrategies
-    class Error < RuntimeError
+    class Error < Y2Storage::Error
```
jreidinger: where it is used? I failed to see raise with it.

imobachgs: For example, in `Y2Storage::Proposal::AutoinstProposal` and even on `yast2-autoinstallation` (the next PR).

src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -22,61 +22,61 @@
...
+      def boot_partition_missing?
+        free_mountpoint?("/boot")
+      end
+
       def boot_partition(target)
         vol = Planned::Partition.new("/boot", Filesystems::Type::EXT4)
-        vol.disk = settings.root_device
+        vol.disk = boot_disk.name
         vol.min_size = target == :min ? DiskSize.MiB(100) : DiskSize.MiB(200)
         vol.max_size = DiskSize.MiB(500)
```
jreidinger: please use constants. I see there three constant parameters for boot size proposal.
so I prefer code like
```
BOOT_PARTITION_SIZES = {
  min: DiskSize.MiB(100),
  ideal: DiskSize.MiB(200),
  max: DiskSize.MiB(500)
}
# and use it later.
```

btw why here is only /boot? Why it is so special? e.g. /boo/zipl is similar in importance, at least for me.

src/lib/y2storage/boot_requirements_strategies/zipl.rb
```diff
@@ -28,25 +28,29 @@ module BootRequirementsStrategies
     # Strategy to calculate boot requirements in systems using ZIPL
     class ZIPL < Base
       def needed_partitions(target)
-        raise Error unless supported_root_disk?
-        [zipl_partition(target)]
+        raise Error unless supported_boot_disk?
```
jreidinger: why exception do not have any message?

src/lib/y2storage/boot_requirements_strategies/zipl.rb
```diff
@@ -28,25 +28,29 @@ module BootRequirementsStrategies
...
-      def supported_root_disk?
-        return false unless root_disk
-        if root_disk.is?(:dasd)
-          return false if root_disk.dasd_type.is?(:fba)
-          return false if root_disk.dasd_format.is?(:ldl)
+      def supported_boot_disk?
+        return false unless boot_disk
+        if boot_disk.is?(:dasd)
+          return false if boot_disk.dasd_type.is?(:fba)
+          return false if boot_disk.dasd_format.is?(:ldl)
```
jreidinger: to be honest I expect here raising error with explanation like
```
def supported_boot_disk!
  raise Error, "Missing root disk" unless boot_disk
  if boot_disk.is?(:dasd)
    raise Error, "Boot disk uses Fixed Block Access" if boot_disk.dasd_type.is?(:fba)
    raise Error, "Boot disk uses Linux disk layout format" if boot_disk.dasd_format.is?(:ldl)
```

jreidinger: this way it is immediately obvious what is problem without checking code, checking accronyms and checking from log how looks boot disk configuration.

src/lib/y2storage/guided_proposal.rb
```diff
@@ -0,0 +1,181 @@
...
+  #
+  #   proposal.proposed? # => true
+  #   proposal.devices   # Proposed layout
+  #   proposal.settings.use_separate_home = false # raises RuntimeError
+  #
+  class GuidedProposal
+    include Yast::Logger
+
+    # Settings used to calculate the proposal. They cannot be altered after
+    # calculating the proposal
```
jreidinger: if so, then why do not freeze it?

jreidinger: ah, it is freezed, then 👍 :)

src/lib/y2storage/planned/can_be_encrypted.rb
```diff
@@ -46,6 +51,72 @@ def initialize_can_be_encrypted
...
+      #
+      # FIXME: with the current implementation (using the device kernel name
+      # instead of UUID or something similar), the DeviceMapper for an encrypted
+      # /dev/sda5 would be "cr_sda5", which implies a quite high risk of
+      # collision with existing DeviceMapper names.
+      #
+      # Revisit this after improving libstorage-ng capabilities about
+      # alternative names and DeviceMapper.
+      #
+      # FIXME: this should probably be moved to Encryption.
```
jreidinger: 👍

src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -45,9 +45,12 @@ module CanBeFormatted
       # @return [String] Parent for all Btrfs subvolumes (typically "@")
       attr_accessor :default_subvolume
 
+      attr_accessor :reformat
```
jreidinger: please document what i means

src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -58,14 +61,17 @@ def initialize_can_be_formatted
       #
       # Do nothing if #filesystem_type is not set.
       #
+      # FIXME: temporary API. It should be improved.
+      #
       # @param blk_dev [BlkDevice]
       #
       # @return [Filesystems::BlkFilesystem] filesystem
-      def create_filesystem(blk_dev)
+      def format!(blk_dev)
```
jreidinger: please explain in comment why it have `!` in name. ( like that it modify device graph attached to argument ).

src/lib/y2storage/planned/device.rb
```diff
@@ -60,8 +61,39 @@ def self.to_string_attrs
         [:reuse]
       end
 
+      # Returns the (possibly processed) device to be used for the planned
+      # device.
+      #
+      # FIXME: temporary API. It should be improved.
+      def final_device!(plain_device)
+        plain_device.encrypted? ? plain_device.encryption : plain_device
```
jreidinger: well, all this is quite highly tight to encryption. I worry what happen if there is something similar like encryption child. How much code have to be adapted?

jreidinger: maybe better method in plain_device can help. Not sure about proper name, but something like `end_device` which means final device to be mounted or operated on.

src/lib/y2storage/planned/device.rb
```diff
@@ -60,8 +61,39 @@ def self.to_string_attrs
...
+
+      # Whether this planned device is expected to reuse an existing one
+      def reuse?
+        !(reuse.nil? || reuse.empty?)
+      end
+
     protected
 
+      def reuse_device!(dev)
+        log.info "Reusing #{reuse} (#{dev}) for #{self}"
```
jreidinger: ??? I am a bit confused by this method. in general I prefer to have protected documented, as it is also part of API to be used.

jreidinger: later it is used, but what if some child forget to implement it? I am not sure if it is robust enough.

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -0,0 +1,85 @@
...
+require "y2storage/proposal/partitions_distribution_calculator"
+# TODO: fix distribution calculator to don't require this
+require "y2storage/proposal/lvm_helper"
+require "y2storage/proposal/partition_creator"
+
+module Y2Storage
+  module Proposal
+    # Class to create and reuse devices during the AutoYaST proposal, based
+    # on the information contained in the profile.
+    class AutoinstDevicesCreator
```
jreidinger: why it is `Autoinst` and mentioned that it is only for AutoYaST? I do not see any autoyast specific code. Why it cannot be reused for other purposes?

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -0,0 +1,85 @@
...
+      def initialize(original_graph)
+        @original_graph = original_graph
+      end
+
+      # Devicegraph including all the specified planned devices
+      #
+      # @param planned_devices [Array<Planned::Partition>] Devices to create/reuse
+      # @param disk_names [Array<String>] Disks to consider
+      # @return [Devicegraph] New devicegraph in which all the planned devices have been allocated
+      def devicegraph(planned_devices, disk_names)
```
jreidinger: I suggest to at least rename it to `devicegraph_for` otherwise it is quite confusing as I already mention above.
BTW related to discussion on IRC. here it is not that arguments are not modified, but that it `return copy of of original graph` so it is not that argument is not modified, but that devicegraph is related to original one. So it should be mentioned.

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -0,0 +1,85 @@
...
+        planned_partitions = planned_devices.select { |dev| dev.is_a?(Planned::Partition) }
+        reused, created = planned_partitions.partition(&:reuse?)
+
+        log.info "Partitions to reuse (#{reused.map(&:reuse)}): #{reused}"
+        log.info "Partitions to create: #{created}"
+
+        dist = best_distribution(created, disk_names)
+        raise Error if dist.nil?
```
jreidinger: hmm, again I do not like this Error without any message. When I see in log that error is throw, but no info, it makes debugging harder and not easier.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -0,0 +1,155 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage/disk"
+
+module Y2Storage
+  module Proposal
+    # Class to generate a list of Planned::Device objects that must be allocated
+    # during the AutoYaST proposal.
+    class AutoinstDevicesPlanner
```
jreidinger: ok, here it use drives_map which is autoinst specific format, so fine for me. But maybe that drives map can be more generic interface and in such case, again this do not have to be autoinst specific code.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -0,0 +1,155 @@
...
+      # Constructor
+      #
+      # @param devicegraph [Devicegraph] Devicegraph to be used as starting point
+      def initialize(devicegraph)
+        @devicegraph = devicegraph
+      end
+
+      # Returns an array of planned devices according to the drives map
+      #
+      # @raise [Error] if not partitions were specified
```
jreidinger: typo `not` instead of `no`

jreidinger: btw maybe own exception for this kind of error makes lives easier on other places, like I comment above.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -0,0 +1,155 @@
...
+      #
+      # @param drives_map [Proposal::AutoinstDrivesMap] Drives map from AutoYaST
+      # @return [Array<Planned::Partition>] List of planned partitions
+      def planned_devices(drives_map)
+        result = []
+
+        drives_map.each_pair do |disk_name, drive_spec|
```
jreidinger: hint
```
result = drives_map.each_pair.reduce([]) do |result, disk_name, drive_spec|
```

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -0,0 +1,155 @@
...
+      # @param disk         [Disk] Disk to place the partitions on
+      # @param partitioning [Hash] Partitioning specification from AutoYaST
+      # @return [Array<Planned::Partition>] List of planned partitions
+      def planned_for_disk(disk, spec)
+        result = []
+        spec.fetch("partitions", []).each do |part_description|
+          # TODO: fix Planned::Partition.initialize
+          part = Y2Storage::Planned::Partition.new(nil, nil)
+          part.disk = disk.name
+          # part.bootable no está en el perfil (¿existe lógica?)
```
jreidinger: I am looking forward to learn some pepper offensive words in spanish, but google translate dissatisfied me with result 😉

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -0,0 +1,155 @@
...
+          part.filesystem_type = filesystem_for(part_description["filesystem"])
+          # TODO: set the correct id based on the filesystem type (move to Partition class?)
+          part.partition_id = 131
+          if part_description["crypt_fs"]
+            part.encryption_password = part_description["crypt_key"]
+          end
+          part.mount_point = part_description["mount"]
+          part.label = part_description["label"]
+          part.uuid = part_description["uuid"]
+          if part_description["create"] == false
```
jreidinger: so if I get it, default value is nil and nil is same as true, right? I think it should be more obvious from code. like
```
part_description["create"] = true unless part_description.key?("create")
if !part_description["create"]
```

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -0,0 +1,155 @@
...
+            partition_to_reuse = find_partition_to_reuse(devicegraph, part_description)
+            if partition_to_reuse
+              part.reuse = partition_to_reuse.name
```
jreidinger: this mixture of partition and part is a bit confusing, especially if part have bigger life-time then partition, which is unusual. Usually objects that live longer have longer and more descriptive names.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -0,0 +1,155 @@
...
+      #
+      # @see SIZE_REGEXP
+      def sizes_for(part_spec, disk)
+        normalized_size = part_spec["size"].to_s.strip.downcase
+
+        if normalized_size == "max" || normalized_size.empty?
+          return [disk.min_grain, DiskSize.unlimited]
+        end
+
+        _all, number, unit = SIZE_REGEXP.match(normalized_size).to_a
```
jreidinger: another option is 
`number, unit = SIZE_REGEXP.match(normalized_size).values_at(1,2)`

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -0,0 +1,155 @@
...
+      end
+
+      # Regular expression to detect which kind of size is being used in an
+      # AutoYaST <size> element
+      SIZE_REGEXP = /([\d,.]+)?([a-zA-Z%]+)/
+
+      # Returns min and max sizes for a partition specification
+      #
+      # @param description [Hash] Partition specification from AutoYaST
+      # @return [[DiskSize,DiskSize]] min are max sizes for the given partition
```
jreidinger: typo "are" instead "and"

src/lib/y2storage/proposal/autoinst_drives_map.rb
```diff
@@ -0,0 +1,94 @@
...
+      # @!method each_pair
+      #   Calls block once por each disk that contains the AutoYaST specification
+      #   passing as arguments the disk name and the specification itself.
+      #
+      #   @example
+      #     drives_map.each_pair do |disk_name, drive_spec|
+      #       puts "Drive for #{disk_name}: #{drive_spec}"
+      #     end
+      #
+      # @!method each
```
jreidinger: if there is each, maybe include also Enumerable mixin make sense?

src/lib/y2storage/proposal/autoinst_drives_map.rb
```diff
@@ -0,0 +1,94 @@
...
+module Y2Storage
+  module Proposal
+    # Utility class to map disk names to the corresponding AutoYaST <drive>
+    # specification that will be applied to that disk.
+    class AutoinstDrivesMap
+      extend Forwardable
+
+      # @!method each_pair
+      #   Calls block once por each disk that contains the AutoYaST specification
+      #   passing as arguments the disk name and the specification itself.
```
jreidinger: link to specification would be great, especially if someone want to reuse it for something different then only autoyast.

src/lib/y2storage/proposal/autoinst_space_maker.rb
```diff
@@ -0,0 +1,96 @@
...
+# find current contact information at www.suse.com.
+
+require "y2storage/disk"
+
+module Y2Storage
+  module Proposal
+    # Class to provide free space during the AutoYaST proposal by deleting
+    # partitions and partition tables according to the information in the
+    # AutoYaST profile.
+    class AutoinstSpaceMaker
```
jreidinger: I am a bit suprised that AutinstSpaceMaker is not child of SpaceMaker, but it looks like it do not have shared functionality.

jreidinger: another idea. Can be common SpaceMaker used just with settings tuned according to autoyast profile? I would like to see sharing functionality :)

src/lib/y2storage/boot_requirements_checker.rb
ancorgs: And `BootRequirementsStrategies::PReP` will not require a PReP partition for powernv, as already documented in the RSpec tests of that class and here:
https://github.com/yast/yast-storage-ng/blob/master/doc/boot-requirements.md#needed-partitions-in-a-ppc64-system

src/lib/y2storage/boot_requirements_checker.rb
ancorgs: aarm64 is not taken into account in the original code and this PR is not changing that behavior. Let's focus.

src/lib/y2storage/proposal/planned_devices_generator.rb
```diff
@@ -68,14 +68,17 @@ def planned_devices(target)
...
       # @return [Array<Planned::Device>]
-      def boot_devices
-        checker = BootRequirementsChecker.new(settings, devicegraph)
+      def boot_devices(root_dev)
+        checker = BootRequirementsChecker.new(
+          devicegraph, planned_devices: [root_dev], boot_disk_name: settings.root_device
+        )
         checker.needed_partitions(@target)
       rescue BootRequirementsChecker::Error => error
         raise NotBootableError, error.message
```
jreidinger: to be honest I do not get logic here. If I want to get list of boot devices to "decide if bootable system can be installated", I get here exception if it cannot be proposed?

src/lib/y2storage/proposal/skip_list.rb
```diff
@@ -0,0 +1,81 @@
...
+    #         <skip_key>driver</skip_key>
+    #         <skip_value>usb-storage</skip_value>
+    #       </listentry>
+    #       <listentry>
+    #         <skip_key>size_k</skip_key>
+    #         <skip_value>1048576</skip_value>
+    #         <skip_if_less_than config:type="boolean">true</skip_if_less_than>
+    #       </listentry>
+    #     </skip_list>
+    #   </drive>
```
jreidinger: link to autoyast documentation will really help here.

src/lib/y2storage/proposal/skip_list_rule.rb
```diff
@@ -0,0 +1,167 @@
...
+    #   disk = Disk.new(8192)
+    #   rule = Proposal::SkipListRule.new(:size_k, :less_than, 16384)
+    #   rule.matches?(disk) #=> true
+    #
+    # @example Creating a rule from an AutoYaST profile hash
+    #   hash = { "skip_key" => "size_k", "skip_if_less_than" => 16384,
+    #     "skip_value" => 1024 }
+    #   Proposal::SkipListRule.from_profile_hash(hash)
+    #
+    class SkipListRule
```
jreidinger: why it is `SkipListRule` and not simple `SkipRule` I do not see here any reason why it cannot be used outside of SkipList

src/lib/y2storage/proposal/skip_list_rule.rb
```diff
@@ -0,0 +1,167 @@
...
+
+      # Cast the reference value in order to do the comparison
+      #
+      # @param raw [String]
+      # @return [String,Fixnum,Symbol] Converted reference value
+      def cast_reference(raw, klass)
+        if klass == Fixnum
+          raw.to_i
+        elsif klass == Symbol
+          raw.respond_to?(:to_sym) ? raw.to_sym : :nothing
```
jreidinger: according to documentation raw is always string and string always respond to_sym. maybe missing documentation that parameter can be nil?
Another question is why `:nothing` is it link to something? if so, it have to be documented that nil act this way.

src/lib/y2storage/proposal/skip_list_value.rb
```diff
@@ -0,0 +1,64 @@
...
+      # @return [String] Full device name
+      def device
+        disk.name
+      end
+
+      # Device name
+      #
+      # @return [String] Last part of device name (for instance, sdb)
+      def name
+        disk.name.split("/").last
```
jreidinger: note. I see this usage also in partitioner, maybe make sense to finally add it to blk_device? and maybe add name method to lvm_vg? 😉

src/lib/y2storage/boot_requirements_checker.rb
ancorgs: I don't remember the reason (it was not even my decision). But yes, there was a conscious decision on using the Storage class instead of the YaST one.

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
ancorgs: Yes, it's documented here https://github.com/yast/yast-storage-ng/pull/259/files#diff-8929af5ad49feaa141b33ec8c31010adR56

So a `@see` should be enough

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -0,0 +1,184 @@
...
+require "y2storage/disk_analyzer"
+require "y2storage/proposal"
+require "y2storage/exceptions"
+
+module Y2Storage
+  # Class to calculate a storage proposal for autoinstallation
+  #
+  # @example Example
+  #   profile = Yast::Profile.current["partitioning"]
+  #   proposal = Y2Storage::AutoinstProposal.new(partitioning)
```
jreidinger: btw partitioning not defined, maybe you mean profile?

src/lib/y2storage/autoinst_proposal.rb
jreidinger: maybe it can be called ProfileBasedProposal? As I hope profile proposal can be reused elsewhere.

src/lib/y2storage/autoinst_proposal.rb
jreidinger: so maybe better name is needed. Maybe something like `space_maker.cleaned_devicegraph(initial_devicegraph, drives)`

jreidinger: My problem is that from method `provide_space` I expect it return provided space and not device graph

src/lib/y2storage/autoinst_proposal.rb
jreidinger: ah, so so it is chaining of modified device graphs. So maybe it helps to better name it like
```
cleaned_devicegraph = space_maker.cleaned_devicegraph(initial_devicegraph, drives)

@proposed_devicegraph = propose_devicegraph(cleaned_devicegraph, drives)
```

src/lib/y2storage/autoinst_proposal.rb
ancorgs: Because the planner doesn't know whether is going to be possible to make the devices fit. In the regular (guided) proposal, that's a way more complex process that implies deleting, reusing and shrinking devices between subsequent attempts of making the planned devices fit.

So the planner just set the goal and subsequent steps, in separate classes, try to address it. If it was not possible, a less ambitious goal is set and the process is tried again.

That's the approach followed from the very beginning and [described here](https://github.com/yast/yast-storage-ng/blob/master/doc/proposal.md)... which reminds me another missing task for this pull request - update that document that is not 100% accurate now.

src/lib/y2storage/autoinst_proposal.rb
jreidinger: ok, so it can create different devicegraph using constructor argument as initial one. I think it really helps to mention that it is copy of initial one in constructor. For me this is same case as String and its method. To mention that it returns copy, as graph is in fact not! argument, but part of object.

About naming. If I try to follow "say, not ask" I prefer names like `devicegraph_for` or `create` as it say what action we want. Simple `devicegraph` is for me like asking for devicegraph.

src/lib/y2storage/autoinst_proposal.rb
jreidinger: what happens if none is defined? can it get nil or []? BTW to clarity it would be also good to mention what is supported format, as I think in autoyast you can specify udev names like `/dev/disk/by-id/blabla`

src/lib/y2storage/boot_requirements_checker.rb
jreidinger: well, then consider this code confusing :) having strategy for prep that do not require prep is confusing. What about calling this strategy PPC strategy? Because you actually do not require prep, but require booting strategy for ppc include powernv

src/lib/y2storage/boot_requirements_checker.rb
ancorgs: Yes, it's documented here https://github.com/yast/yast-storage-ng/pull/259/files#diff-8929af5ad49feaa141b33ec8c31010adR56

So a @see should be enough

src/lib/y2storage/boot_requirements_checker.rb
ancorgs: On one hand, yes, the name of the class could be better... but having not-perfect names is almost a pattern :wink: and I would expect developers to check what the class does. For example in this context (a class using several strategies to calculate the list of needed partition) assuming a class called `Strategies::PReP` blindly create a PReP partition is maybe too much to infer from the name without checking that class first.

On the other hand, that re-reviewing code (or class names) that has been there for months. Out of the scope of this PR, I would say.

src/lib/y2storage/boot_requirements_strategies/base.rb
ancorgs: Here there is only `/boot` because this is the base strategy class. `/boot/zipl` is implemented in its own strategy.

ancorgs: And about the usage of constants, this is again re-reviewing quite old code just because the line before was modified. This PR is already big and complex enough. :wink:

src/lib/y2storage/boot_requirements_strategies/zipl.rb
ancorgs: Once again, I consider this to be quite out of the scope of this pull request. Maybe we will take a look, but I cannot promise it. :smile:

src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -99,6 +105,10 @@ def remove_shadowed_subvolumes!(other_mount_points)
         self.subvolumes = subvolumes.reject { |subvol| subvol.shadowed?(other_mount_points) }
       end
 
+      def reformat?
```
ancorgs: Documention also missing here

src/lib/y2storage/planned/device.rb
ancorgs: Yes, that's why there is a `FIXME: temporary API. It should be improved.` comment there. We plan to do it while adding support for LVM to the AutoYaST proposal.

src/lib/y2storage/planned/device.rb
ancorgs: This is part of the ` FIXME: temporary API. It should be improved.` initiative to be addressed during LVM implementation.

src/lib/y2storage/proposal/autoinst_devices_creator.rb
ancorgs: It is indeed AutoYaST specific because it implements the AutoYaST expected behavior/algorithm, which is completely different from the guided proposal behavior, for example.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
ancorgs: But again, this algorithm to plan devices is the AutoYaST algorithm, not the only one. The guided proposal has a different planner (currently with the horrible name `PlannedDevicesGenerator`, hopefully to be changed).

src/lib/y2storage/proposal/autoinst_devices_planner.rb
ancorgs: A matter of taste, most people from the storage team prefer the existing alternative to the `reduce`-base one.

src/lib/y2storage/proposal/autoinst_drives_map.rb
ancorgs: No so far. Let's keep it simple until needed.

src/lib/y2storage/proposal/autoinst_space_maker.rb
ancorgs: Definitely not an option, two complete different beasts.

src/lib/y2storage/proposal/planned_devices_generator.rb
ancorgs: Yes. You get a list of the devices that are needed (and missing) in order the for system to boot. If it's impossible to propose such list (because you are trying to setup a system that will never be able to boot) you get an exception.

src/lib/y2storage/proposal/skip_list_value.rb
ancorgs: I'm open to suggestions about the name of the new method

ancorgs: What about `short_name`?

src/lib/y2storage/autoinst_proposal.rb
imobachgs: This code contains AutoYaST specific code, so I do not think it will be reusable at all.

src/lib/y2storage/autoinst_proposal.rb
imobachgs: No, it does not modify the passed one.

src/lib/y2storage/proposal/skip_list_rule.rb
imobachgs: The documentation looks wrong: it could receive also `nil`. I'll updated it.

Regarding the `:nothing` thing, it was just the fallback value used in the [original implementation](https://github.com/yast/yast-autoinstallation/blob/cdfab9e0503bb368d62eba49269c8ae82396d3f0/src/modules/AutoinstStorage.rb#L812).

src/lib/y2storage/autoinst_proposal.rb
imobachgs: I've added the @return documentation.

src/lib/y2storage/proposal/skip_list_rule.rb
imobachgs: Actually it is wrong, because according to the old version I should do the same with the value when it is nil. I will fix it.

src/lib/y2storage/boot_requirements_checker.rb
jreidinger: agreed, see is enough

src/lib/y2storage/planned/device.rb
jreidinger: ok, fine

src/lib/y2storage/proposal/autoinst_devices_creator.rb
jreidinger: then it would be nice to document this different behavior. In which part it is different? it will help if e.g. me deciding which one I should use when I want to propose something.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
jreidinger: so again if it is different algorithm it would be nice to  document such behavior

src/lib/y2storage/proposal/planned_devices_generator.rb
jreidinger: ok, so please comment it above rescue, as it is not obvious

src/lib/y2storage/proposal/skip_list_value.rb
jreidinger: well, it is tricky as original is already wrong, as it is `path` and name. I suggest `kernel_name` or `kernel_device` as it is what it is. Name how kernel see it.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
imobachgs: I would avoid modifying the original object. But sure, we could do it more readable with something like:

```ruby
if !part_description.fetch("create", true)
  # ...
```

src/lib/y2storage/autoinst_proposal.rb
imobachgs: It's about retrying, so we will retry with a different target on any Storage error.

src/lib/y2storage/autoinst_proposal.rb
imobachgs: `AutoinstDrivesMap` will take care of finding a suitable disk if none was specified. Unfortunatelly, that behaviour is [not implemented yet](https://github.com/yast/yast-storage-ng/pull/259/files#diff-6a95b1be0007f9b4629f2e653394ee4cR62).

src/lib/y2storage/planned/can_be_formatted.rb
ancorgs: As already mentioned in this method documentation: temporary API (to be improved while adding LVM support).

src/lib/y2storage/proposal/autoinst_drives_map.rb
ancorgs: Already added in the class description

src/lib/y2storage/proposal/skip_list.rb
ancorgs: Added

PR #260, opened 2017-06-14
provide probe function

aschnell:
For https://trello.com/c/nxgcriYy/612-2-storageng-fix-instdisksactivate-client.


src/lib/y2storage/storage_manager.rb
```diff
@@ -144,6 +145,15 @@ def update_staging_revision
       @staging_revision += 1
     end
 
+    # Probes all storage devices.
+    #
+    # Invalidates the probed and staging devicegraph.
+    def probe
+      storage.probe
```
ancorgs: Until we kill all the legacy code (to support direct access to the `Storage` classes), we will also need this here:
```ruby
@y2probed = nil
@y2staging = nil
```

src/lib/y2storage/storage_manager.rb
```diff
@@ -81,6 +81,7 @@ def initialize(storage_environment)
       @storage.activate(activate_callbacks)
       @storage.probe
```
ancorgs: What if we simply call the just added `#probe` so don't duplicate responsibilities (like calling `Storage.probe` and updating `@staging_revision_after_probing`)?

I mean
```ruby
  ...
  @storage.activate(activate_callbacks)
  @staging_revision = 0
  @proposal = nil
  probe
end
```

ancorgs:
In addition, we should add a changelog entry and increase the version number. So we can adjust the dependencies in yast2-installation

ancorgs:
LGTM

PR #261, opened 2017-06-15
Fake from xml

jreidinger:


src/lib/y2storage/storage_manager.rb
```diff
@@ -230,6 +230,21 @@ def fake_from_yaml(yaml_file = nil)
         @instance
       end
 
+      # Use this as an alternative to the instance method.
+      # Probing is skipped and the device tree is initialized from xml_file.
+      # For xml definition ask or search source code of libstorage-ng.
```
ancorgs: Not sure if telling the developers to ask for examples is very professional. ;-) It makes it sound like this is a random ad-hoc xml format just made up for this method. But in fact, this is the "official" libstorage-ng format. In other parts, like [here](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/devicegraph.rb#L49), we simply refer to it as "the libstorage xml format". Maybe a `@see Devicegraph#load` would be useful here.

PR #262, opened 2017-06-16
Make proposal work with DASD

joseivanlopez:
PBI: https://trello.com/c/m7N8jeE6/506-2-storageng-make-proposal-work-with-dasd

### DIFF
--- a/src/lib/y2storage/devicegraph.rb
+++ b/src/lib/y2storage/devicegraph.rb
@@ -131,7 +131,7 @@ module Y2Storage
 
     # All disk devices. This method is necessary because DASD devices
     #   are not returned when asking for disks.
-    # @return [Array<Dasd|Disk>]
+    # @return [Array<Dasd, Disk>]
     def disk_devices
       dasds + disks
     end
--- a/src/lib/y2storage/devicegraph.rb
+++ b/src/lib/y2storage/devicegraph.rb
@@ -119,9 +119,21 @@ module Y2Storage
       Actiongraph.new(graph)
     end
 
+    # @return [Array<Dasd>]
+    def dasds
+      Dasd.all(self)
+    end
+
     # @return [Array<Disk>]
     def disks
       Disk.all(self)
+    end
+
+    # All disk devices. This method is necessary because DASD devices
+    #   are not returned when asking for disks.
+    # @return [Array<Dasd|Disk>]
+    def disk_devices
+      dasds + disks
     end
 
     # @return [Array<Partition>]
--- a/src/lib/y2storage/devicegraph.rb
+++ b/src/lib/y2storage/devicegraph.rb
@@ -119,18 +119,34 @@ module Y2Storage
       Actiongraph.new(graph)
     end
 
+    # All the DASDs in the devicegraph
+    #
+    # @note Based on the libstorage classes hierarchy, DASDs are not considered to be disks.
+    # See #disk_devices for a method providing the whole list of both disks and DASDs.
+    # @see #disk_devices
+    #
     # @return [Array<Dasd>]
     def dasds
       Dasd.all(self)
     end
 
+    # All the disks in the devicegraph
+    #
+    # @note Based on the libstorage classes hierarchy, DASDs are not considered to be disks.
+    # See #disk_devices for a method providing the whole list of both disks and DASDs.
+    # @see #disk_devices
+    #
     # @return [Array<Disk>]
     def disks
       Disk.all(self)
     end
 
-    # All disk devices. This method is necessary because DASD devices
-    #   are not returned when asking for disks.
+    # All the devices that are usually treated like disks by YaST
+    #
+    # Currently this method returns an array including all the disks and all the DASDs.
+    # @see #disks
+    # @see #dasds
+    #
     # @return [Array<Dasd, Disk>]
     def disk_devices
       dasds + disks
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -36,7 +36,7 @@ module Y2Storage
   #
   # Class to analyze the disk devices (the storage setup) of the existing system:
   # Check the existing disk devices (Dasd or Disk) and their partitions what candidates
-  # there areto install on, typically eliminate the installation media from that list
+  # there are to install on, typically eliminate the installation media from that list
   # (unless there is no other disk), check if there already are any
   # partitions that look like there was a Linux system previously installed
   # on that machine, check if there is a Windows partition that could be
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -25,7 +25,7 @@ require "yast"
 require "storage"
 require "fileutils"
 require "y2storage/disk_size"
-require "y2storage/disk"
+require "y2storage/blk_device"
 require "y2storage/lvm_pv"
 require "y2storage/partition_id"
 require "y2storage/existing_filesystem"
@@ -34,17 +34,15 @@ Yast.import "Arch"
 
 module Y2Storage
   #
-  # Class to analyze the disks (the storage setup) of the existing system:
-  # Check the existing disks and their partitions what candidates there are
-  # to install on, typically eliminate the installation media from that list
+  # Class to analyze the disk devices (the storage setup) of the existing system:
+  # Check the existing disk devices (Dasd or Disk) and their partitions what candidates
+  # there areto install on, typically eliminate the installation media from that list
   # (unless there is no other disk), check if there already are any
   # partitions that look like there was a Linux system previously installed
   # on that machine, check if there is a Windows partition that could be
   # resized.
   #
-  # Some of those operations involve trying to mount the underlying
-  # filesystem.
-  #
+  # Some of those operations involve trying to mount the underlying filesystem.
   class DiskAnalyzer
     include Yast::Logger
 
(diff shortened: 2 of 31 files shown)


src/lib/y2storage/devicegraph.rb
```diff
@@ -119,11 +119,23 @@ def actiongraph(from: nil)
...
+      Dasd.all(self)
+    end
+
     # @return [Array<Disk>]
     def disks
       Disk.all(self)
     end
 
+    # All disk devices. This method is necessary because DASD devices
+    #   are not returned when asking for disks.
```
ancorgs: The documentation of this method seems to be written taking into account the reader will have the implementation in front, so it's more focused in "why" that in "what". I would go for something more descriptive, since this is a very important method. Let me try an example:
```ruby
# All the devices that are usually treated like disks by YaST
#
# Currently this method returns an array including all the disks and all the DASDs.
# @see #disks
# @see #dasds
# @return [Array<Dasd, Disk>]
  def disk_devices
```
And then, I would add additional notes to the documentation of `disks` and `dasds`. Something like...
```ruby
# All the disks in the devicegraph
#
# @note Take into account that, based on the libstorage classes hierarchy, DASDs are not considered to be disks
# @see #disks_devices for a method providing the whole list of both disks and DASDs
# @return [Array<Disk>]
def disks
```

src/lib/y2storage/devicegraph.rb
```diff
@@ -119,11 +119,23 @@ def actiongraph(from: nil)
...
+    # @return [Array<Dasd, Disk>]
+    def disk_devices
+      dasds + disks
```
ancorgs: NP: are we sure this is the best order? I don't know, just asking.

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -34,17 +34,15 @@
 
 module Y2Storage
   #
-  # Class to analyze the disks (the storage setup) of the existing system:
-  # Check the existing disks and their partitions what candidates there are
-  # to install on, typically eliminate the installation media from that list
+  # Class to analyze the disk devices (the storage setup) of the existing system:
+  # Check the existing disk devices (Dasd or Disk) and their partitions what candidates
+  # there areto install on, typically eliminate the installation media from that list
```
ancorgs: missing blank space

src/lib/y2storage/devicegraph.rb
joseivanlopez: I am not sure, but I would say so, or at least it is a reasonable option. When we have dasd disks, the system should be installed in them, but I do not know if this is a good criteria to be first in the list. Another option would be to order them by name, for example.

src/lib/y2storage/partition_tables/base.rb
```diff
@@ -114,6 +115,26 @@ def inspect
...
+      # The partition id depends on the partition table type. For example,
+      # MSDOS partition tables use SWAP id for swap partitions, but DASD
+      # partiton tables always need LINUX id.
+      #
+      # With this method, each partition table can define the partition id
+      # it expects for each case.
+      #
+      # TODO: right now only :swap is considered.
+      #
+      # @param id_name [Symbol] a partition id name (e.g., :swap, :linux, :ntfs)
```
ancorgs: I don't think this should have a direct correlation to partitions ids. If that's the case, why not simple accepting `<PartitionId>` as argument?

I believe the intention of this argument is more like a role. `:swap`, `lvm_pv`, `windows`...

src/lib/y2storage/partition_tables/base.rb
```diff
@@ -114,6 +115,26 @@ def inspect
...
+      # @return [PartitionId]
+      def partition_id_for(id_name)
+        case id_name
+        when :swap
+          PartitionId::SWAP
+        else
```
ancorgs: Hmmm, if only `:swap` is considered... why do we have an `else` here? Not sure if I like how the whole situation of inheritance and this method is resolved.

src/lib/y2storage/partitionable.rb
```diff
@@ -43,6 +43,8 @@ class Partitionable < BlkDevice
     #   @return [Array<PartitionTables::Type>]
     storage_forward :possible_partition_table_types, as: "PartitionTables::Type"
 
+    storage_forward :default_partition_table_type, as: "PartitionTables::Type"
+
```
ancorgs: This method and `possible_partition_table_types` need documentation. Specially when compared to `preferred_ptable_type` (that is, btw, inserted in a different point of the hierarchy)

ancorgs: What I would do is:
1) Not exposing `default_partition_table_type` publicly, but define it as private
2) Moving `preferred_ptable_type` to `Partitionable`, implementing it as a direct forward to the private method (maybe just an alias).
3) Re-implementing it in `Disk` to always offer GPT.

src/lib/y2storage/partitionable.rb
```diff
@@ -181,6 +183,58 @@ def mbr_gap
...
+    #
+    # @example With a completely empty device
+    #   empty_disk.as_not_empty do
+    #     empty_disk.partition_table # => a temporary PartitionTable
+    #   end
+    #   empty_disk.partition_table # Not longer there
+    def as_not_empty
+      fake_ptable = nil
+      if !has_children?
+        fake_ptable = create_partition_table(preferred_ptable_type)
```
ancorgs: This is really weird, because `preferred_ptable_type` is used in `Partitionable`, but only defined in `Disk` and `Dasd`. See my comment above.

joseivanlopez: Yes, I don't realized about that :(

src/lib/y2storage/proposal/partition_creator.rb
```diff
@@ -132,11 +130,16 @@ def free_space_within(initial_free_space)
...
       #
-      def create_partition(planned_partition, partition_id, free_space, primary)
+      def create_partition(planned_partition, free_space, primary)
         log.info "Creating partition for #{planned_partition.mount_point} with #{planned_partition.size}"
         disk = free_space.disk
         ptable = partition_table(disk)
 
+        if !planned_partition.partition_id
+          id_name = planned_partition.mount_point == "swap" ? :swap : :other
+          planned_partition.partition_id = ptable.partition_id_for(id_name)
```
ancorgs: Modifying the passed argument looks wrong to me.

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -208,14 +208,15 @@ def free_spaces(graph, disk = nil)
...
-      # List of candidate disks in the given devicegraph
+      # List of candidate disk devices in the given devicegraph
       #
       # @param devicegraph [Devicegraph]
-      # @param disk [String] optional disk name to restrict result to
-      # @return [DisksList]
-      def disks_for(devicegraph, disk = nil)
-        filter = disk || candidate_disk_names
-        devicegraph.disks.select { |d| filter.include?(d.name) }
+      # @param device_name [String] optional device_name device name to restrict
```
ancorgs: optional device_name device name... ?

test/dasd_test.rb
```diff
@@ -0,0 +1,47 @@
...
+  subject { Y2Storage::Dasd.find_by_name(fake_devicegraph, "/dev/sda") }
+
+  describe "#usb?" do
+    it "returns false" do
+      expect(subject.usb?).to be_falsey
+    end
+  end
+
+  describe "#preferred_ptable_type" do
+    it "is the default partition table type" do
```
ancorgs: As suggested, I would keep `default_partition_table_type` as private and would rephrase this as
```ruby
it "returns the DASD partition table type"
```
checking directly for `PartitionTables::Type::DASD`.

test/data/devicegraphs/output/s390_dasd_zipl.yml
```diff
@@ -1,30 +1,32 @@
...
-- disk:
+- dasd:
     name: "/dev/sda"
     size: 50 GiB
-    partition_table: gpt
+    partition_table: dasd
     partitions:
     - partition:
         size: 200 MiB
+        start: 3 MiB
```
ancorgs: I would void adding these kind of extra lines that add no information for a human being. The changes on this file should only be line 2, line 5 (hopefully removing the line instead of changing it) and the swap partition if.

test/data/devicegraphs/output/s390_dasd_zipl-lvm-sep-home.yml
```diff
@@ -0,0 +1,43 @@
+---
+- dasd:
```
ancorgs: NP: I can live with these files as they are. But they look too much like a machine dumb. Removing accessory information like the `partition_table` line, the `start` positions, the redundant partition types and so on would be welcome. :wink:

test/fake_device_factory_test.rb
```diff
@@ -21,11 +21,62 @@
...
+    staging = storage.staging
+
+    # rubocop:disable Style/StringLiterals
+
+    input = ['---',
+             '- dasd:',
+             '    name: "/dev/sda"',
+             '    size: 256 GiB',
+             '    block_size: 4 KiB',
+             '    partition_table: dasd',
```
ancorgs: is this line really needed? It should not be. A `disk` can be created without that (which defaults to a MSDOS partition table, if I'm not wrong). Creating a `dasd` without that line makes even more sense, since there is only a valid type.

ancorgs: BTW, what does the reader do if we try to format directly a DASD, or to use it as PV? Is not supported by the class, but is detected as a syntax error?

joseivanlopez: Yes, that line is needed. In fact, I think it is always needed. Why do you think it is not? Right now I think it is not possible to create a partition table without this line. A different thing is to create a Disk with a direct filesystem over it.

Regarding the syntax errors, the current implementation accepts a filesystem directly over the disk, or even to use the disk as PV or MD. I would not consider these options as a syntax error only because the library it is not prepared yet for that. In my opinion, the yaml reader only should be adapted if we are totally sure that these options are not possible.

PR #263, opened 2017-06-19
Changelog entry

ancorgs:
Changelog entry that was missing in #262 Since Iván is not here, I decided to merge the PR and add the entry afterwards.


package/yast2-storage-ng.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Jun 19 06:58:54 UTC 2017 - jilopez@localhost
+
+- Adjustments to DASD to ensure the proposal works as expected
+  in such devices.
+- Added Devicegraph#disk_devices to get all DASDs and disks.
+- DASD support added to Yaml reader and writer.
```
imobachgs: np: YAML

PR #265, opened 2017-06-19
Added Y2Storage::Region #size, #cover?.

mvidner:
This is useful in https://github.com/yast/yast-partitioner/pull/15

### DIFF
--- a/src/lib/y2storage/disk_size.rb
+++ b/src/lib/y2storage/disk_size.rb
@@ -609,36 +609,48 @@ module Y2Storage
     # @example
     #   x = DiskSize.KB(1)   #=> <DiskSize 0.98 KiB (1000)>
     #   x.to_human_string    #=> "0.98 KiB"
-    def to_human_string
+    def to_human_string(round_method = :round)
       return "unlimited" if unlimited?
-      size, unit = human_string_components
-      format("%.2f %s", size, unit)
+      float, unit_s = human_string_components
+      rounded = (float * 100).public_send(round_method) / 100.0
+      # A plain "#{rounded} #{unit_s}" would not keep trailing zeros
+      format("%.2f %s", rounded, unit_s)
     end
 
-    # A human readable representation that does not exceed the exact size.
+    # A human readable string that does not exceed the exact size.
     #
-    # If we have 4.999 GiB of space and prefill the "Size" widget
-    # with a "5.00 GiB" it will then fail validation. We must round down.
+    # If we have 4.999 GiB of space, and ask the user how much of that
+    # should be used, prefilling the "Size" widget with the maximum
+    # rounded up to "5.00 GiB" it will then fail validation.
+    # We must round down.
+    #
+    # @return [String]
+    #
+    # @example
+    #   x = DiskSize.new(4095)  #=> <DiskSize 4.00 KiB (4095)>
+    #   x.to_human_string       #=> "4.00 KiB"
+    #   x.human_floor           #=> "3.99 KiB"
     #
     # @see to_human_string
     # @see human_ceil
     def human_floor
-      return "unlimited" if unlimited?
-      float, unit_s = human_string_components
-      "#{(float * 100).floor / 100.0} #{unit_s}"
+      to_human_string(:floor)
     end
 
     # A human readable representation that is at least the exact size.
     #
     # (This seems unnecessary because actual minimum sizes
-    # have few significant digits, but we use it for symmetry)
+    # have few significant digits, but we provide it for symmetry)
+    #
+    # @example
+    #   x = DiskSize.new(4097)  #=> <DiskSize 4.00 KiB (4097)>
+    #   x.to_human_string       #=> "4.00 KiB"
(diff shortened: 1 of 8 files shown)


src/lib/y2storage/disk_size.rb
```diff
@@ -615,6 +615,32 @@ def to_human_string
...
+    #
+    # (This seems unnecessary because actual minimum sizes
+    # have few significant digits, but we use it for symmetry)
+    #
+    # @see to_human_string
+    # @see human_floor
+    def human_ceil
+      return "unlimited" if unlimited?
+      float, unit_s = human_string_components
+      "#{(float * 100).ceil / 100.0} #{unit_s}"
```
mvidner: Code climate does not like this duplication.

I think we could solve this by
```rb
def to_human_string(round_method = :round)
  return "unlimited" if unlimited?
  float, unit_s = human_string_components
  rounded = (float * 100).public_send(round_method) / 100.0
  format("%.2f %s", rounded, unit_s) # the #format is necessary to get "5.00 B" instead of "5.0 B"
end

def human_ceil
  to_human_string(:ceil)
end
```

ancorgs: If properly documented, this makes sense to me. So please, go ahead

src/lib/y2storage/disk_size.rb
```diff
@@ -609,10 +609,48 @@ def floor(unit_size)
     # @example
     #   x = DiskSize.KB(1)   #=> <DiskSize 0.98 KiB (1000)>
     #   x.to_human_string    #=> "0.98 KiB"
-    def to_human_string
+    def to_human_string(round_method = :round)
```
ancorgs: This new argument is not documented. Maybe is enough to point to `#human_floor` and `#human_ceil`, but there must be something

src/lib/y2storage/disk_size.rb
```diff
@@ -609,10 +609,48 @@ def floor(unit_size)
...
+      rounded = (float * 100).public_send(round_method) / 100.0
+      # A plain "#{rounded} #{unit_s}" would not keep trailing zeros
+      format("%.2f %s", rounded, unit_s)
+    end
+
+    # A human readable string that does not exceed the exact size.
+    #
+    # If we have 4.999 GiB of space, and ask the user how much of that
+    # should be used, prefilling the "Size" widget with the maximum
+    # rounded up to "5.00 GiB" it will then fail validation.
```
ancorgs: Which validation you mean here?

Reading this text sounds like the "human_floored" value will become the final one used to create the partition (or whatever device), not just a representation displayed to the user. In this case, we will be loosing precision on the real thing just to fulfill it representation needs.

src/lib/y2storage/disk_size.rb
```diff
@@ -609,10 +609,48 @@ def floor(unit_size)
...
+    #
+    # @return [String]
+    #
+    # @example
+    #   x = DiskSize.new(4095)  #=> <DiskSize 4.00 KiB (4095)>
+    #   x.to_human_string       #=> "4.00 KiB"
+    #   x.human_floor           #=> "3.99 KiB"
+    #
+    # @see to_human_string
+    # @see human_ceil
```
ancorgs: In my opinion, these should be examples of `x.to_human_string(:floor)`. This places the useful information in the method that is just an alias.

src/lib/y2storage/disk_size.rb
ancorgs: Moreover, for those optional methods that are kind of unexpected[*] and have a default value, I usually prefer keywords attributes, like
```ruby
def to_human_string(round_method: :round)
```
In this same class, there are similar examples of such arguments, like `rounding` or `legacy_units`. So consistency would be another argument here.

But feel free to disagree, of course.

[*] Unexpected because I would usually not expect a `to_xx` to receive an argument and also because they are not really needed to perform the integral part of the functionality, they are there mostly to influence the functionality with some fancy hue.

src/lib/y2storage/disk_size.rb
mvidner: The use case is the Partition Size dialog, where you can choose
- Use the maximum size available (4.99GiB)
- Use a specific size: `4.99GiB`
- Use a specific region, from `12345` to `19999`



and yes, if you select the specific `4.99GiB` size, you will waste up to 1% of space. I don't know of a way to avoid this without confusing the user with subtle details.

Oh well, we could say: "Use a specific size: `5368601746` " which is **exact**, but slightly less than 5GiB. Would you know the difference?

Regarding "validation", is this better?
> then fail validation (checking that the entered value fits in the available space)

ancorgs: Ok. If the user has an option to not waste space by selecting "maximum size", then is fair enough (I assume there is no rounding at all in that case).

As a completely unrelated note, libstorage-ng does not use cylinders at all (that are an obsolete unit). I guess we can still "simulate" them in the "custom region" part. Or maybe we should switch to sectors or anything that makes sense nowadays.

PR #268, opened 2017-06-20
redefine '+', '-', and '%' operators to accept Strings as 2nd argument

wfeldt:


ancorgs:
The code looks good to me. Since the corresponding tests are in #265, I guess it makes sense to wait until that pull request is merged and then rebase this one to include removing the 'x' from the several 'xit'.

test/disk_size_test.rb
```diff
@@ -285,7 +305,15 @@
       disk_size = Y2Storage::DiskSize.MiB(12) / 3
       expect(disk_size.to_i).to be == 12 / 3 * 1024**2
     end
-    it "should refuse another type" do
+    it "should accept division by a float" do
```
mvidner: I think we are overspecifying. What should this operation mean anyway?

Because then we could also say that strings make sense, just like for `+` and `-`
It's complicated. So I think we should not commit to more than we need.

wfeldt: I'm testing what's implemented in the class. If you want to discuss class features, this is not the place.

mvidner: I mean, sometimes, things are implemented because it is easier to keep them than to leave them out. But that does not mean they make good sense.
But it is not such a big deal.

wfeldt: Well, it does make a big difference if you can increase a disk size by 20% by multiplying it by 1.2 or not. Likewise by division.

It's similar with DiskSize/DiskSize as you commented in the test - it's a big difference if the result
is expected to be an integral number of not. So IMO the tests should cover the status quo here.

PR #269, opened 2017-06-21
Extract some logic from LvmHelper into new classes

imobachgs:
Preparation for adding LVM support to AutoYaST.


src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -0,0 +1,223 @@
...
+      end
+
+      # Returns a copy of the original devicegraph in which the volume
+      # group (and its logical volumes, if needed) have been created.
+      #
+      # @param planned_vg [Planned::LvmVg]   Volume group to create
+      # @param pv_partitions [Array<String>] names of the newly created
+      #     partitions that should be added as PVs to the volume group
+      # @return [Devicegraph] New devicegraph containing the planned volume group
+      def create_volumes(planned_vg, pv_partitions = [])
```
imobachgs: We should use `planned_vg.pvs` instead of an additional `pv_partitions`  argument.

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -0,0 +1,223 @@
...
+      end
+
+      # Create a volume group in a devicegraph
+      #
+      # The volume group will be DEFAULT_VG_NAME.
+      #
+      # @param devicegraph [Devicegraph] Starting point
+      # @return [Devicegraph] New devicegraph containing the new volume group
+      def create_volume_group(devicegraph)
+        name = available_name(DEFAULT_VG_NAME, devicegraph)
```
imobachgs: Support for using custom volume groups will come in the next PR.

src/lib/y2storage/planned/lvm_vg.rb
```diff
@@ -0,0 +1,113 @@
...
+      # @return [Array<Planned::LvmLv>] List of logical volumes
+      attr_accessor :lvs
+
+      # @return [Array<Planned::Device>] List of physical volumes
+      attr_accessor :pvs
+
+      # @return [DiskSize] Size of one extent
+      attr_accessor :extent_size
+
+      # @return [DiskSize] Total size of the volume group
```
ancorgs: It's strange to have a writer for this. I would expect that value to be automagically set for reused VGs (that is, as a consequence of `.from_real_vg`), but not to have a writer to set it by hand.

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -0,0 +1,223 @@
...
+      # Returns a copy of the original devicegraph in which the volume
+      # group (and its logical volumes, if needed) have been created.
+      #
+      # @param planned_vg [Planned::LvmVg]   Volume group to create
+      # @param pv_partitions [Array<String>] names of the newly created
+      #     partitions that should be added as PVs to the volume group
+      # @return [Devicegraph] New devicegraph containing the planned volume group
+      def create_volumes(planned_vg, pv_partitions = [])
+        new_graph = original_devicegraph.duplicate
+        return new_graph if planned_vg.lvs.empty?
```
ancorgs: So if there are no LVs the VG is not created at all (or the PVs not added if the VG already existed)?

ancorgs: That's highly unexpected and should be documented.

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -0,0 +1,223 @@
...
+
+        vg = planned_vg.reuse? ? find_vg(planned_vg, new_graph) : create_volume_group(new_graph)
+
+        assign_physical_volumes!(vg, pv_partitions, new_graph)
```
ancorgs: I know the original methods were written by me. But I have recently learn that using `!` without a non-`!` counterpart is an anti-pattern. Is not enough for a method to be destructive to deserve a `!`. If there is no non-destructive version, then the `!` should not be used.

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -0,0 +1,223 @@
...
+      def assign_physical_volumes!(volume_group, part_names, devicegraph)
+        partitions = devicegraph.partitions.select { |p| part_names.include?(p.name) }
+        partitions.each do |partition|
+          device = partition.encryption || partition
+          volume_group.add_lvm_pv(device)
+        end
+      end
+
+      # Makes sure the given volume group has enough free extends to allocate
+      # all the planned volumes, by deleting the existing volume groups.
```
ancorgs: deleting the existing **logical volumes**

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -0,0 +1,223 @@
...
+        @original_devicegraph = original_devicegraph
+      end
+
+      # Returns a copy of the original devicegraph in which the volume
+      # group (and its logical volumes, if needed) have been created.
+      #
+      # @param planned_vg [Planned::LvmVg]   Volume group to create
+      # @param pv_partitions [Array<String>] names of the newly created
+      #     partitions that should be added as PVs to the volume group
+      # @return [Devicegraph] New devicegraph containing the planned volume group
```
ancorgs: If it's reusing a VG, it will kill the old LVs in order to allocate the new ones if needed. That should be very clearly documented. Not sure if that's the desired behavior in the AutoYaST case.

src/lib/y2storage/proposal/lvm_helper.rb
```diff
@@ -69,16 +61,9 @@ def initialize(planned_lvs, encryption_password: nil)
...
-
-        vg = reused_volume_group ? find_reused_vg(new_graph) : create_volume_group(new_graph)
-
-        assign_physical_volumes!(vg, pv_partitions, new_graph)
-        make_space!(vg)
-        create_logical_volumes!(vg)
-
-        new_graph
+        lvm_creator = LvmCreator.new(original_graph)
+        # FIXME: it won't be needed in the future
```
ancorgs: I guess this FIXME does not apply to this line, but to the whole `LvmHelper#create_volumes` method.

src/lib/y2storage/proposal/lvm_creator.rb
imobachgs: That's the old behaviour and this PR is only about extracting some logic out of LvmHelper. I'll add a document and we'll fix it when implementing changes for AutoYaST.

src/lib/y2storage/proposal/lvm_creator.rb
imobachgs: Old behaviour again: this code was not modified, just moved. I'll add a note to not overlook it when implemeting AutoYaST support.

src/lib/y2storage/proposal/lvm_creator.rb
imobachgs: I agree.

src/lib/y2storage/proposal/lvm_creator.rb
imobachgs: old comment :)

src/lib/y2storage/planned/lvm_vg.rb
```diff
@@ -77,11 +73,22 @@ def self.from_real_vg(real_vg)
...
         @pvs = pvs
       end
 
+      # Initializes the object taking the values from a real volume group
+      #
+      # @param real_vg [Y2Storage::LvmVg] Real volume group
+      def initialize_from_real_vg(real_vg)
+        @extent_size = real_vg.extent_size
+        @total_size = real_vg.total_size
+        @pvs = real_vg.lvm_pvs.map { |v| v.blk_device.name }
```
ancorgs: I wonder if this could be potentially dangerous at some point since names can change as we create and destroy partitions. We will find out. :wink:

PR #271, opened 2017-06-23
run new partitioner

wfeldt:


src/lib/y2storage/clients/inst_disk_proposal.rb
```diff
@@ -97,7 +97,7 @@ def guided_setup(settings)
       end
 
       def expert_partitioner
-        ExpertPartitioner::MainDialog.new.run
+        Y2Partitioner::Clients::Main.run
```
ancorgs: Would it be too picky to say that, since we don't have yast2-partitioner as a dependency, we should detect here if Y2Partitioner is not available and show an error instead of trying to run it?

Dockerfile
```diff
@@ -1,3 +1,4 @@
 FROM yastdevel/storage-ng
+RUN zypper --gpg-auto-import-keys --non-interactive in yast2-partitioner
```
lslezak: Alternatively you can add the package to https://github.com/yast/docker-storage-ng/blob/master/Dockerfile, then it will be available for all packages at Travis.

src/lib/y2storage/clients/inst_disk_proposal.rb
ancorgs: Of course, with the corresponding extension to `inst_disk_proposal_test.rb`

lslezak:
Do we need to adapt the RPM deps?

jreidinger:
rpm deps probably should not be changed, but dependencies here are a bit confusing, showing whole architecture, where storage is backend, partitioner is frontend, but storage itself is also partially frontend.

lslezak:
Then the `require "y2partitioner/..."` call might fail. Shouldn't we handle it better than displaying the generic crash dialog?

lslezak:
Sorry, misclecked, reopening...

Dockerfile
wfeldt: Actually that's a good thing to do when we go mainline with storage-ng. For now I'd just wait and see.

src/lib/y2storage/clients/inst_disk_proposal.rb
wfeldt: That's a good question and IMO shows that the package split logic is kind of broken here.

The 'lib' package requires the package that uses it.

I'd leave it as is for now. More would be out of scope, I think.

jreidinger:
@lslezak well, it is installation only, so it is responsibility of installer to have all required code in place. At least that is how it is done also for other modules, where installation depends on them.

lslezak:
OK, then I'm fine with that.

lslezak:
Actually this is `yast2-storage-ng` package, that would be OK for `yast2-installation`, but this is indirect dependency, out of sight of `yast2-installation` or `istallation-images`... So maybe a better error handling would be nice.

src/lib/y2storage/clients/inst_disk_proposal.rb
lslezak: (BTW that check would have to be placed at the require above, that would fail earlier...)

wfeldt:
We can either assume all packages are correctly added to the inst-sys (as with the other yast packages)
or we make storage-ng require yast2-partitioner (which looks weird).

For now I'd leave it as is.

PR #272, opened 2017-06-23
require yast2-partitioner

wfeldt:
I've just realized it's by far the easiest way to get yast2-partitioner into instsys.


wfeldt:
won't work - yast2-partitioner has a BuildRequires on yast2-storage-ng

PR #273, opened 2017-06-23
disable partitioner test

wfeldt:
We can't require yast2-partitioner atm as yast2-partitioner BuildRequires
yast2-storage-ng. So, disable this test to be able to build yast2-storage-ng.

### DIFF
--- a/test/clients/inst_disk_proposal_test.rb
+++ b/test/clients/inst_disk_proposal_test.rb
@@ -21,9 +21,19 @@
 # find current contact information at www.suse.com.
 
 require_relative "../spec_helper"
-require "y2storage/clients/inst_disk_proposal"
 
-describe Y2Storage::Clients::InstDiskProposal do
+# FIXME
+#
+# We can't run this test due to depedency hell.
+#
+# But we need the test suit to succeed without yast2-partitioner installed -
+# so disable for now.
+#
+
+# require "y2storage/clients/inst_disk_proposal"
+
+xdescribe "Y2Storage::Clients::InstDiskProposal" do
+
   subject(:client) { described_class.new }
 
   describe "#run" do


test/clients/inst_disk_proposal_test.rb
```diff
@@ -21,9 +21,22 @@
...
+# But we need the test suit to succeed without yast2-partitioner installed -
+# so disable for now.
+#
+
+# require "y2storage/clients/inst_disk_proposal"
+# fake the class for xdescribe below
+class Y2Storage::Clients::InstDiskProposal
+end
+
+xdescribe Y2Storage::Clients::InstDiskProposal do
```
lslezak: You can use `xdescribe "Y2Storage::Clients::InstDiskProposal"` (as string) without the fake class.

ancorgs:
As already commented in the previous PR. I believe our code should check if the expert partitioner is there or not before trying to run it. So we don't need an strict dependency.

And I also believe we can test that behavior in RSpec (checking if the partitioner is there and running it in that case),  without needing to disable the whole test.

wfeldt:
The current package split is plain wrong. We'll never be able to Require yast2-partitioner. We need to change this.

test/clients/inst_disk_proposal_test.rb
jreidinger: another option is to create here something like stub_dir, where lives partitioner client and just modify ruby load_path before require "y2storage/clients/inst_disk_proposal". This way, it can be even tested.

wfeldt: cool, didn't know that :-)

ancorgs:
Of course the package split is debatable. But while the split is reconsidered, I believe we can move forward with a proper, tested and robust enough solution that is doable.

Is not the only place in which we have a "weak" dependency, so we try to require a file and we act in consequence depending on whether the file is there.

Is not that much work.

wfeldt:
I don't think optional tests are a good idea.

wfeldt:
replaced by https://github.com/yast/yast-storage-ng/pull/274

PR #274, opened 2017-06-23
allow easy mocking clients from partitioner

jreidinger:
alternative approach to https://github.com/yast/yast-storage-ng/pull/273


wfeldt:
lgtm

PR #276, opened 2017-06-26
AutoYaST LVM support

imobachgs:


src/lib/y2storage/planned/partition.rb
```diff
@@ -60,6 +60,9 @@ class Partition < Device
       #   flag but is not needed in our grub2 setup.
       attr_accessor :bootable
 
+      # @return [String] name of the LVM volume group where the partition belongs
+      attr_accessor :lvm_volume_group_name
```
ancorgs: I guess this should, somehow, replace the existing `lvm_pv?` and its usage during the proposal.

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -63,8 +63,10 @@ def populated_devicegraph(planned_devices, disk_names)
         part_creator = Proposal::PartitionCreator.new(original_graph)
         result = part_creator.create_partitions(dist)
 
-        reused.each { |r| r.reuse!(result) }
-        result
+        reused.each { |r| r.reuse!(result.devicegraph) }
```
ancorgs: Now the variable name `result` turns out to be confusing. Is not exactly the result of this method anymore.

ancorgs: We should maybe rethink a little bit the messages displayed before. I mean those "Partitions to reuse" and "partitions to create".

src/lib/y2storage/proposal/autoinst_drives_map.rb
```diff
@@ -46,13 +46,16 @@ class AutoinstDrivesMap
...
+      SUPPORTED_TYPES = [:CT_DISK, :CT_LVM].freeze
+
       # Constructor
       #
       # @param devicegraph  [Devicegraph] Devicegraph where the disks are contained
       # @param partitioning [Array<Hash>] Partitioning layout from an AutoYaST profile
       def initialize(devicegraph, partitioning)
         # By now, consider only regular disks
-        disks = partitioning.select { |i| i.fetch("type", :CT_DISK) == :CT_DISK }
+        disks = partitioning.select { |i| SUPPORTED_TYPES.include?(i.fetch("type", :CT_DISK)) }
```
ancorgs: The name of the variable and even the comment ("consider only regular disks") are confusing now.

ancorgs: Hmmm, is it possible to have a LVM_CT without `<device>`? If so, it will be matched to nothing (only disks are considered in `#first_usable_disk`). If not... that's impossible to spot from this code with apparently makes no difference between disks and VGs.

src/lib/y2storage/proposal/autoinst_space_maker.rb
```diff
@@ -65,6 +65,7 @@ def cleaned_devicegraph(original_devicegraph, drives_map)
       # @option drive_spec [Boolean] "initialize" Initialize the device
       # @option drive_spec [String]  "use"        Partitions to remove ("all" or "linux")
       def delete_stuff(devicegraph, disk, drive_spec)
+        return unless disk.respond_to?(:partition_table)
```
ancorgs: I guess this implies there is no way to specify in the profile what to do with the existing LVs in a VG. If that's the case, is not already wrong the fact that `#delete_stuff` is sometimes called on something that is not "partitionable"? (I assume that's the reason to for this `return`.

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -51,23 +49,21 @@ def initialize(original_devicegraph)
...
+      #   partitions that should be added as PVs to the volume group
       # @return [Devicegraph] New devicegraph containing the planned volume group
       def create_volumes(planned_vg, pv_partitions = [])
         new_graph = original_devicegraph.duplicate
-        return new_graph if planned_vg.lvs.empty?
-
-        vg = planned_vg.reuse? ? find_vg(planned_vg, new_graph) : create_volume_group(new_graph)
+        vg =
+          if planned_vg.reuse?
+            find_vg(planned_vg, new_graph)
```
ancorgs: Maybe I'm being nitpicking here. But this `if` reveals quite some inconsistencies in names and arguments order of the methods. `find_vg(planned_vg, graph)` vs `create_volume_group(graph, planned_vg)`.

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -149,18 +176,32 @@ def create_logical_volumes(volume_group, planned_lvs)
...
+      # This method modifies the volume group received as first argument.
       #
       # @param volume_group [LvmVg] Volume group
       # @param planned_lv   [Planned::LvmLv] Planned logical volume to be used as reference
       #   for the new one
       def create_logical_volume(volume_group, planned_lv)
         name = planned_lv.logical_volume_name || DEFAULT_LV_NAME
         name = available_name(name, volume_group)
-        lv = volume_group.create_lvm_lv(name, planned_lv.size.to_i)
+        lv = volume_group.create_lvm_lv(name, size_for(volume_group, planned_lv).to_i)
```
ancorgs: I would say the  `to_i` is not needed.

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -149,18 +176,32 @@ def create_logical_volumes(volume_group, planned_lvs)
...
+      # Returns the size for the logical volume
+      #
+      # It returns the planned size (Planned::LogicalVolume#size) unless a
+      # percentage has been specified. In that case, it will use the volume
+      # group size and Planned::LogicalVolume#percent_size to calculate the
+      # desired size.
+      #
+      # @param [LvmVg]          Volume group where the logical volume will be placed
+      # @param [Planned::LvmLv] Planned logical volume
+      def size_for(volume_group, planned_lv)
```
ancorgs: This looks like a method that should be in `PlannedLv`.

src/lib/y2storage/proposal/partition_creator.rb
```diff
@@ -27,6 +27,16 @@
 
 module Y2Storage
   module Proposal
+    # Partition creation result
+    #
+    # @!method devicegraph
+    # @return[Devicegraph] Devicegraph containing the new devices
```
ancorgs: I think this yardoc needs nesting to work properly. Same immediately below.

src/lib/y2storage/proposal/partition_creator.rb
```diff
@@ -93,17 +106,20 @@ def process_free_space(free_space, partitions, usable_size, num_logical)
       # @param initial_free_space [FreeDiskSpace]
       # @param num_logical [Symbol] logical partitions. See {#process_space}
       def create_planned_partitions(planned_partitions, initial_free_space, num_logical)
```
ancorgs: The new returned value is not documented

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -86,6 +88,66 @@ def best_distribution(planned_partitions, disk_names)
 
         calculator.best_distribution(planned_partitions, spaces)
       end
+
+    private
+
+      # Creates planned partitions in the given devicegraph
+      #
+      # @param planned_partitions [Array<Planned::Partition>] Devices to create/reuse
```
ancorgs: strictly speaking, only partitions to create. Isn't it?

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -86,6 +88,66 @@ def best_distribution(planned_partitions, disk_names)
...
+        part_creator.create_partitions(dist)
+      end
+
+      # Creates volume groups in the given devicegraph
+      #
+      # @param vgs         [Array<Planned::LvmVg>]           List of planned volume groups to add
+      # @param devicegraph [Devicegraph]                     Starting point
+      # @param devices_map [Hash<String,Planned::Partition>] Map of device names and partitions
+      # @return [Devicegraph] Devicegraph containing the specified volume groups
+      def create_volume_groups(vgs, devicegraph, devices_map)
```
ancorgs: Nitpicking: strictly speaking, this is not necessarily creating VGs. I would say the more important part here (that may not be obvious from the name) is that it's creating LVs. At least, not all the VGs included in the first argument are finally created, some of them are already there (reuse) and are simply "completed" with more LVs...

Nevermind, I'm probably thinking too much ;-)

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -86,6 +88,66 @@ def best_distribution(planned_partitions, disk_names)
...
+      end
+
+      # Reuses volume groups for the given devicegraph
+      #
+      # @param reused_vgs  [Array<Planned::LvmVg>] Volume groups to reuse
+      # @param devicegraph [Devicegraph]           Devicegraph to reuse partitions
+      def reuse_vgs(reused_vgs, devicegraph)
+        reused_vgs.each do |vg|
+          vg.reuse!(devicegraph)
+          vg.lvs.each { |v| v.reuse!(devicegraph) }
```
ancorgs: Maybe a comment here clarifying that the fact that we call `#reuse!` for all the planned LVs doesn't imply all of them are going to be reused (that's decided by each PlannedDevice). That's actually fine, but in the case of partitions we only call `#reuse!` for those for which `#reuse?` is true. That can be confusing. I don't care if we fix it with a comment or with a `if v.reuse?`.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -52,7 +52,12 @@ def planned_devices(drives_map)
 
         drives_map.each_pair do |disk_name, drive_spec|
           disk = Disk.find_by_name(devicegraph, disk_name)
```
ancorgs: We should use either `BlkDevice.find_by_name` or `devicegraph.disk_devices` here, since DADs are not disks.

src/lib/y2storage/proposal/autoinst_space_maker.rb
```diff
@@ -40,14 +40,15 @@ def initialize(disk_analyzer)
...
-      # @param planned_partitions [Array<Planned::Partition>] set of partitions
-      #     to make space for.
+      # @param original_devicegraph [Devicegraph] initial devicegraph
+      # @param drives_map           [Array<Planned::Partition>] set of partitions
+      #   to make space for.
       def cleaned_devicegraph(original_devicegraph, drives_map)
         devicegraph = original_devicegraph.dup
 
         drives_map.each_pair do |disk_name, drive_spec|
           disk = Disk.find_by_name(devicegraph, disk_name)
```
ancorgs: Same as above, we could be overlooking DASDs here

PR #279, opened 2017-06-26
Md enum to human

jreidinger:


src/lib/y2storage/md_level.rb
```diff
@@ -27,7 +27,25 @@ module Y2Storage
...
+    TRANSLATIONS = {
+      unknown: N_("Unknown"),
+      raid0:   N_("RAID0"),
+      raid1:   N_("RAID1"),
+      raid5:   N_("RAID5"),
+      raid6:   N_("RAID6"),
+      raid10:  N_("RAID10")
+    }
+
+    def to_human_string
```
ancorgs: It can raise an exception. That should be documented.

src/lib/y2storage/md_parity.rb
```diff
@@ -29,5 +29,39 @@ class MdParity
...
+
+      near_2:             N_("Two copies near"),
+      near_3:             N_("Three copies near"),
+      offset_2:           N_("Two copies offset"),
+      offset_3:           N_("Three copies offset"),
+      far_2:              N_("Two copies far"),
+      far_3:              N_("Three copies far")
+    }
+
+    def to_human_string
```
ancorgs: Same than above.

src/lib/y2storage/md_level.rb
```diff
@@ -27,7 +27,25 @@ module Y2Storage
...
     wrap_enum "MdLevel"
+
+    TRANSLATIONS = {
+      unknown: N_("Unknown"),
+      raid0:   N_("RAID0"),
+      raid1:   N_("RAID1"),
+      raid5:   N_("RAID5"),
+      raid6:   N_("RAID6"),
+      raid10:  N_("RAID10")
+    }
```
ancorgs: `:private_constant` is probably a good idea here.

src/lib/y2storage/md_level.rb
```diff
@@ -27,7 +27,25 @@ module Y2Storage
...
+
+    def to_human_string
+      textdomain "storage"
```
ancorgs: I'm not against having this here, but we usually place it in the initializer. So it's unexpected.

src/lib/y2storage/md_parity.rb
```diff
@@ -29,5 +29,39 @@ class MdParity
     include StorageEnumWrapper
 
     wrap_enum "MdParity"
+
+    TRANSLATIONS = {
+      # TODO: should be tested by some real OPS guy, I (JR) pick human wording from `man mdadm`
```
ancorgs: Thanks god we have git blame. Otherwise, it would be hard to find who JR is. :) My first though was:

JR

src/lib/y2storage/md_level.rb
jreidinger: well, I would like to have it simple. If we need translations on more places, we should move it to initialize

src/lib/y2storage/md_level.rb
```diff
@@ -39,11 +40,15 @@ class MdLevel
       raid6:   N_("RAID6"),
       raid10:  N_("RAID10")
     }
+    private_constant :TRANSLATIONS
 
+    # Returns human readable representation of enum.
+    # @return [String]
+    # @raise [RuntimeError] when called on unknown ennum value.
```
ancorgs: Typo: ennum

src/lib/y2storage/md_level.rb
```diff
@@ -27,7 +27,30 @@ module Y2Storage
...
+      raid1:   N_("RAID1"),
+      raid5:   N_("RAID5"),
+      raid6:   N_("RAID6"),
+      raid10:  N_("RAID10")
+    }
+    private_constant :TRANSLATIONS
+
+    # Returns human readable representation of enum.
+    # @return [String]
+    # @raise [RuntimeError] when called on unknown ennum value.
```
ancorgs: Typo: ennum.

Since you have to change the documentation anyway to fix the typo, it would be cool if we clarify (is never too much clarification) that the resulting string would already be translated.

And talking about clarification... maybe we should also mention than an "unknown ennum value" means a value for which we don't have a i18n-enabled representation, which doesn't imply the value is invalid or unknown for Y2Storage/libstorage in general.

PR #286, opened 2017-06-29
Move examples from src/lib to src/examples

imobachgs:
IMHO, that example code belongs to `src/examples`.


src/examples/used_storage_features_demo.rb
```diff
@@ -0,0 +1,32 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+$LOAD_PATH.unshift(File.expand_path("../../lib", __FILE__))
+
+require "y2storage"
+
+devicegraph = Y2Storage::StorageManager.instance.probed
```
ancorgs: We should probably show a warning if not executed as root

shundhammer: Oh yes, right...

shundhammer: Now really LGTM.

imobachgs:
Thanks!

ancorgs:
LGTM also :)

PR #287, opened 2017-06-29
Fall back to less swap (min 512 MB) if not enough disk space for 2 GB

shundhammer:
If there is not enough disk space to satisfy all desired partition / volume sizes, now falling back to a minimum of 512 MB for swap, but keeping the maximum (by default - unless "enlarge for swap" is set) of 2 GB. It now uses a weight to get some disk space beyond the minimum: Any disk space above the cumulated minimum sizes of all planned partitions / volumes is distributed among those that have a weight set.

Right now, swap gets a weight of 100, i.e. it gets as much extra disk space as root and home together. The value is hard-coded right now. We might make it configurable some day, but right now this would just add to the complexity.

In practice all this means that while swap no longer gets its full 2 GB if disk space is low, it at least typically gets more than the minimum 512 MB; in the unit test added in this PR it usually gets ~1.7 GB which seems to be a reasonable compromise.

### DIFF
--- a/src/lib/y2storage/proposal/planned_devices_generator.rb
+++ b/src/lib/y2storage/proposal/planned_devices_generator.rb
@@ -37,8 +37,9 @@ module Y2Storage
 
       attr_accessor :settings
 
-      DEFAULT_SWAP_SIZE = DiskSize.GiB(2)
       MIN_SWAP_SIZE = DiskSize.MiB(512)
+      MAX_SWAP_SIZE = DiskSize.GiB(2) # This is also DEFAULT_SWAP_SIZE
+      SWAP_WEIGHT = 100 # Importance of swap if low free disk space
 
       def initialize(settings, devicegraph)
         @settings = settings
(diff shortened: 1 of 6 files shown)


test/proposal_low_space_test.rb
```diff
@@ -0,0 +1,43 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require_relative "spec_helper"
+require "storage"
```
ancorgs: NP: this is not needed

test/proposal_low_space_test.rb
```diff
@@ -0,0 +1,43 @@
...
+describe Y2Storage::GuidedProposal do
+  describe "#propose" do
+    include_context "proposal"
+
+    subject(:proposal) { described_class.new(settings: settings) }
+    let(:architecture) { :x86 }
+
+    context "in a PC with a small (25GiB) disk" do
+      let(:scenario) { "empty_hard_disk_gpt_25GiB" }
+      let(:expected_scenario) { "empty_hard_disk_gpt_25GiB" }
```
ancorgs: NP: as far as I remember, you don't need :expected_scenario if it matches the scenario

src/lib/y2storage/proposal/planned_devices_generator.rb
```diff
@@ -98,38 +100,44 @@ def additional_devices
...
-      def swap_partition(size)
+      def swap_partition(min_size, max_size)
         part = Planned::Partition.new("swap", Filesystems::Type::SWAP)
         part.encryption_password = settings.encryption_password
         # NOTE: Enforcing the re-use of an existing partition limits the options
         # to propose a valid distribution of the planned partitions. For swap we
         # already have mechanisms to reuse UUIDs and labels, so maybe is smarter
         # to never reuse partitions as-is.
-        reuse = reusable_swap(size)
+        reuse = reusable_swap(max_size)
```
ancorgs: NP: is debatable if we should use `max_size` or `min_size` here. But I don't care much. As the comment states, I hope we will remove this reusing anyway in the future.

ancorgs:
Merging directly to allow rebase of the other pending pull requests.

src/lib/y2storage/proposal/planned_devices_generator.rb
shundhammer: Yes, I was thinking about that, too. One could argue that if we have a minimum size for swap, this should also apply to reusing an existing swap partition, but OTOH if disk space is NOT low, this might easily reuse swap partitions of a size that made sense many years ago, but maybe today's demands are higher. But this is probably mostly academic. Reusing existing swap space really makes sense only in very limited scenarios such as a user having a multi-boot setup with several kinds of Linux where he wants to share the swap space between them. We only try to reuse it to keep an existing UUID of the swap space so that doesn't need to be changed in possibly other Linux installations on that same machine; otherwise the best approach is almost always to remove it and recreate it in the size we really want.

test/proposal_low_space_test.rb
shundhammer: OK. I copied that from existing tests, so it's probably not needed there, either.

test/proposal_low_space_test.rb
shundhammer: OK. Same as above: Copy & paste from an existing test. But anyway, this is explicit and not relying on implicit behaviour, making people who are not that familiar with the mechanics of this test suite wonder what's going on here.

PR #289, opened 2017-07-03
Adapt proposal to different situations

joseivanlopez:
PBI: https://trello.com/c/B2Qv7y6d/639-5-storageng-proposal-adapt-to-the-situation.

Tasks:

* Increase root size based on snapshots and not on filesystem type.
* Add a mechanism to adjust the initial ProposalSettings (no home, no snapshots, etc.) and retry.


src/lib/y2storage/guided_proposal.rb
```diff
@@ -57,11 +57,66 @@ class GuidedProposal
...
+    class << self
+      # Calculates the initial proposal
+      #
+      # If a proposal is not possible by honoring current settings, other settings
+      # are tried. For example, a proposal without separate home or without snapshots
+      # will be calculated.
+      #
+      # @see GuidedProposal#initialize
+      #
+      # @param settings [ProposalSettings]
```
ancorgs: And thinking about that (although I know it's not 100% related to this change)... It would be nice to improve the initialization to default values, so when `probed` is used we don't create a new analyzer but we also use `probed_disk_analyzer`. So just calling this with no arguments (so falling back to defaults) is more efficient. Ask me offline if my explanation is not understandable.

test/guided_proposal_test.rb
```diff
@@ -27,13 +27,89 @@
...
 
 describe Y2Storage::GuidedProposal do
-  describe "#propose" do
-    using Y2Storage::Refinements::SizeCasts
+  using Y2Storage::Refinements::SizeCasts
+  let(:architecture) { :x86 }
+
+  include_context "proposal"
+
+  describe ".initial" do
```
ancorgs: Good tests! What about adding also another to verify that it uses `ProposalSettings.new_for_current_product` by default?

src/lib/y2storage/guided_proposal.rb
```diff
@@ -57,16 +57,81 @@ class GuidedProposal
...
     #   devicegraph will be used
     # @param disk_analyzer [DiskAnalyzer] if nil, a new one will be created
-    #   based in the initial devicegraph
+    #   based on the initial devicegraph
     def initialize(settings: nil, devicegraph: nil, disk_analyzer: nil)
       @settings = settings || ProposalSettings.new_for_current_product
       @proposed = false
-      @initial_graph = devicegraph
-      @disk_analyzer = disk_analyzer
+      if devicegraph.nil?
```
ancorgs: I like this change. But not `#initial_graph` should become a straightforward `attr_reader`. Otherwise it's confusing.

PR #290, opened 2017-07-03
Partitioner additions

jreidinger:


test/partition_tables_test.rb
```diff
@@ -47,4 +47,11 @@
       expect(ptable.inspect).to include "Partition /dev/sdb7 10237 MiB"
     end
   end
+
+  describe "#delete_all_partitions" do
+    it "deletes all partitions in table" do
```
ancorgs: We should specify more explicitly that this test implies logical partitions, just in case somebody feels tempted to change the subject in the future and points it to a too simplistic partition table (lowering the quality of the test).

PR #291, opened 2017-07-04
Use root_base_size for home_min_size

shundhammer:
Use root_base_size for home_min_size until we have a separate entry for this in control.xml

https://trello.com/c/B2Qv7y6d/639-5-storageng-proposal-adapt-to-the-situation

### DIFF
--- a/src/lib/y2storage/proposal_settings.rb
+++ b/src/lib/y2storage/proposal_settings.rb
@@ -209,8 +209,10 @@ module Y2Storage
       @subvolumes                    = SubvolSpecification.fallback_list
 
       # Not yet in control.xml
-      @home_min_size            = DiskSize.GiB(10)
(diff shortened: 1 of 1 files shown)


shundhammer:
**REVOKED.**

Reasoning:

I tried to use `root_base_size` for `home_min_size`, but not only does this break 30 test cases, it also turned out to be not desirable and did not contribute anything to the purpose of this PBI:

This would reduce `home_min_size` to 3 GB (which is the common value for `root_base`). This is turn meant that /home would start out with just 3 GB if there was not enough space to satisfy the _desired_ sizes of all partitions / volumes, so they all start with the _min_ sizes, and the rest is distributed according to the relative _weight_ of each partition / volume.

Of course that means that the sizes are all different from what they were before, thus breaking all test scenarios where we had to fall back to _min_ sizes and distributing the remaining free space.

PR #296, opened 2017-07-05
Base proposal class

joseivanlopez:
Create a new class `Proposal::Base`.

PBI https://trello.com/c/B2Qv7y6d/639-5-storageng-proposal-adapt-to-the-situation.

### DIFF
--- a/src/lib/y2storage/autoinst_proposal.rb
+++ b/src/lib/y2storage/autoinst_proposal.rb
@@ -19,6 +19,7 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "yast"
 require "y2storage/proposal_settings"
 require "y2storage/exceptions"
 require "y2storage/proposal"
@@ -55,23 +56,19 @@ module Y2Storage
       @partitioning = AutoinstProfile::PartitioningSection.new_from_hashes(partitioning)
     end
 
+  private
+
     # Calculates the proposal
     #
-    # @raise [UnexpectedCallError] if called more than once
     # @raise [NoDiskSpaceError] if there is no enough space to perform the installation
-    def propose
-      raise UnexpectedCallError if proposed?
-
+    def calculate_proposal
       drives = Proposal::AutoinstDrivesMap.new(initial_devicegraph, partitioning)
 
       space_maker = Proposal::AutoinstSpaceMaker.new(disk_analyzer)
       devicegraph = space_maker.cleaned_devicegraph(initial_devicegraph, drives)
 
       @devices = propose_devicegraph(devicegraph, drives)
-      @proposed = true
     end
-
-  private
 
     # Proposes a devicegraph based on given drives map
     #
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -19,42 +19,29 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "yast"
```
ancorgs: Strictly speaking, the `require "yast"` is needed for the subsequent requires to work (yes, I know we usually skip it, but we probably shouldn't). I don't see the point in removing this particular line.

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -19,42 +19,29 @@
...
-require "y2storage/storage_manager"
-require "y2storage/disk_analyzer"
-require "y2storage/proposal"
+require "y2storage/proposal_settings"
```
ancorgs: Why? The `ProposalSettings` are only useful for `GuidedProposal`. Why was this require added?

src/lib/y2storage/proposal/base.rb
```diff
@@ -0,0 +1,97 @@
...
+      end
+
+      # Calculates the proposal
+      # @note It should be defined by derived classes
+      #
+      # @raise [NotImplementedError]
+      def propose
+        raise NotImplementedError,
+          "Subclasses of #{Module.nesting.first} must override #{__method__}"
+      end
```
ancorgs: Although I admit I'm not a big fan of the dependency it introduces, we use `abstract_method` for this.
https://github.com/search?q=user%3Ayast+abstract_method&type=Code

src/lib/y2storage/guided_proposal.rb
```diff
@@ -19,43 +19,31 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "yast"
```
ancorgs: Same comment than above.

src/lib/y2storage/proposal/base.rb
ancorgs: On the other hand, I'm not sure if this method should be completely abstract here. I believe these are important lines that should be in the base class.
```ruby
raise UnexpectedCallError if proposed?
@proposed = true
maybe_a_call_here_to_the_really_abstract_method_to_be_implemented_down_there
```

src/lib/y2storage/autoinst_proposal.rb
joseivanlopez: It is used here: https://github.com/yast/yast-storage-ng/pull/296/files#diff-493889b4c9485e86ca3002d10d8112ecL187

src/lib/y2storage/proposal/base.rb
```diff
@@ -0,0 +1,104 @@
...
+          @disk_analyzer ||= StorageManager.instance.probed_disk_analyzer
+        end
+        # Create new disk analyzer when devicegraph is passed but not disk analyzer
+        @disk_analyzer ||= DiskAnalyzer.new(@initial_devicegraph)
+      end
+
+      # Calculates the proposal
+      # @see #calculate_proposal
+      #
+      # @raise [UnexpectedCallError] if called more than once
```
ancorgs: We should probably also document that it can raise other types of exception based on the `#calculate_proposal` implementation of the subclass

PR #297, opened 2017-07-06
Adding check for disk in network

gilsonsouza:


src/lib/y2storage/devicegraph.rb
```diff
@@ -166,6 +166,12 @@ def filesystems
       Filesystems::Base.all(self)
     end
 
+    # @return [Boolean]
+    def filesystem_in_network?(mountpoint)
+      filesystem = filesystems.find { |i| i.mountpoint == mountpoint }
```
ancorgs: What if there is no such mountpoint? `filesystem` would be nil and this would explode.... which makes me think we miss good unit tests here.

src/lib/y2storage/disk.rb
```diff
@@ -68,6 +68,13 @@ def usb?
       transport.to_sym == :usb
     end
 
+    # Checks if it's in network
+    #
+    # @return [Boolean]
+    def network_trasporter?
+      transport.to_sym == :iscsi or transport.to_sym == :fcoe
```
ancorgs: Please, use `||` instead of `or`. The second one if not an operator, is a control flow sentence.

src/lib/y2storage/disk.rb
```diff
@@ -68,6 +68,13 @@ def usb?
       transport.to_sym == :usb
     end
 
+    # Checks if it's in network
+    #
+    # @return [Boolean]
+    def network_trasporter?
```
ancorgs: Why transporter instead of transport?

src/lib/y2storage/disk.rb
ancorgs: I would expect this comparison to be in the `DataTransport` class. So here you would do:
```ruby
transport.network?
```
Instead of comparing here for all the possible values.

src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -132,6 +132,13 @@ def btrfs_subvolumes
         storage_subvols.to_a.map { |vol| BtrfsSubvolume.new(vol) }
       end
 
+      def in_network?
+        disks = ancestors.find_all { |d| d.is?(:disk) }
+        disks.append(self) if is?(:disk)
```
ancorgs: How can this be a disk if it's a `BlkFilesystem`?

src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -132,6 +132,13 @@ def btrfs_subvolumes
...
+        return true if disks.any? { |d| d.network_trasporter? }
+        false
```
ancorgs: Instead of
```ruby
return true if disks.any? { |d| d.network_trasporter? }
false
```

You could simply have this, which is way more straightforward
```ruby
disks.any?(&:network_transporter?)
```

test/devicegraph_test.rb
```diff
@@ -142,4 +142,30 @@ def with_sda2_deleted(initial_graph)
       expect(device_names).to include("/dev/mapper/cr_sdd")
     end
   end
+
+  describe "#filesystem_in_network?" do
+    context "when filesystem is in network" do
```
ancorgs: Missing scenario: "where there is no such mountpoint"

test/filesystems/blk_filesystem_test.rb
```diff
@@ -88,4 +88,28 @@
...
     end
   end
+
+  describe "#in_network?" do
+    context "for a disk in network" do
+      let(:dev_name) { "/dev/sda1" }
+
+      it "returns true" do
+        allow_any_instance_of(Y2Storage::Disk)
+          .to receive(:network_trasporter?).and_return(true)
```
ancorgs: `allow_any_instance_of` is discouraged. Why don't you do?
```ruby
expect(blk_device).to receive #blah blah
```

test/filesystems/blk_filesystem_test.rb
```diff
@@ -88,4 +88,28 @@
       end
     end
   end
+
+  describe "#in_network?" do
```
ancorgs: Can I suggest to move this test to `test/filesystems/base` and then have scenarios like these altogether?
```
for a NFS filesystem
  it returns true
for a filesystem in an LVM logical volume
  if any of the PVs is a network device
    it returns true
  if all the PVs are local
    it returns false
for a filesystem in a partition
  if the partition belongs to a local disk
     it returns false
  if the partition belongs to a DASD
     it returns false
  if the partition belongs to a network disk
     it returns true
for a filesystem directly in a network disk
  it returns true
for a filesystem directly in a local disk
  it returns false
```
Maybe it's a little bit too much, but the idea is that anybody reading the output knows exactly what to expect from the method (that's where the "spec" of RSpec comes from :wink:) and with your output I wouldn't have a clear picture of the method being also available for NFS or about what happens with LVM.

test/filesystems/blk_filesystem_test.rb
```diff
@@ -88,4 +88,28 @@
...
+
+        expect(filesystem.in_network?).to eq true
+      end
+    end
+
+    context "for a local disk" do
+      let(:dev_name) { "/dev/sda1" }
+
+      it "returns false" do
+        allow_any_instance_of(Y2Storage::Disk)
```
ancorgs: Same than above.

src/lib/y2storage/filesystems/base.rb
```diff
@@ -37,6 +37,10 @@ class Base < Mountable
       #   @return [Array<Filesystems::Base>] all the filesystems in the given devicegraph
       storage_class_forward :all, as: "Filesystems::Base"
 
+      def in_network?
```
ancorgs: Missing yardoc

test/filesystems/blk_filesystem_test.rb
ancorgs: I meant `allow(blk_device)`. Sorry

test/filesystems/blk_filesystem_test.rb
gilsonsouza: If I create a file  `test/filesystems/base`, it would look like I'm testing the `Base` class, when in fact, these tests that you wrote would test subclasses of the `Base` class (`BlkFilesystem` and `NFS`). In the future, if I wanna see how I would expect the behavior of `BlkFilesystem`, I would have to go to the tests of the `Base` class.
I don't know if that's how we do it in YaST, but it seems strange to me...

ancorgs: You are right. My point was that I wanted to see everything together (a better name for that would be `test/filesystem_test.rb` or `test/filesystems_test.rb`, I guess). Specially because of the NFS vs others thingie. But feel free to leave it where it is.

src/lib/y2storage/data_transport.rb
```diff
@@ -29,5 +29,9 @@ class DataTransport
     include StorageEnumWrapper
 
     wrap_enum "Transport"
+
+    def network?
```
ancorgs: yardoc missing

src/lib/y2storage/devicegraph.rb
```diff
@@ -166,6 +166,13 @@ def filesystems
       Filesystems::Base.all(self)
     end
 
+    # @return [Boolean]
```
ancorgs: yardoc

src/lib/y2storage/disk.rb
```diff
@@ -68,6 +68,13 @@ def usb?
       transport.to_sym == :usb
     end
 
+    # Checks if it's in network
+    #
+    # @return [Boolean]
+    def network_transport?
+      transport.network?
+    end
```
ancorgs: Having a method `network_transport?` that is implemented as `transport.network?` is weird. I would say the problem is that the name of the method reveals too much about the logic behind it. Don't you think a better name would be simply `in_network?`. Most of the users of the `Disk` class don't even know the "transport" concept.

As a plus, `Disk#in_network?` is more consistent with `Filesystems::Base#in_network?`

src/lib/y2storage/filesystems/base.rb
```diff
@@ -37,6 +37,12 @@ class Base < Mountable
       #   @return [Array<Filesystems::Base>] all the filesystems in the given devicegraph
       storage_class_forward :all, as: "Filesystems::Base"
 
+      # @!method self.in_network?
```
ancorgs: What does this `@!method` do here?

test/filesystems/blk_filesystem_test.rb
```diff
@@ -88,4 +88,60 @@
       end
     end
   end
+
+  describe "#in_network?" do
+    let(:disk) { Y2Storage::BlkDevice.find_by_name(fake_devicegraph, "/dev/sda") }
+
+    context "for a single disk in network" do
```
ancorgs: In my opinion, this new version of the test is too coupled to the implementation (basically a 1:1 repetition of the code in the method). But let's not block the PR because of that.

src/lib/y2storage/filesystems/base.rb
gilsonsouza: This was the first method that I created I just copied the lines above... My mistake...

src/lib/y2storage/devicegraph.rb
```diff
@@ -166,6 +166,14 @@ def filesystems
       Filesystems::Base.all(self)
     end
 
+    # @param from [String] mountpoint of the filesystem (e.g. "/").
```
ancorgs: I'm pretty sure yardoc would complain about the mismatch in the name of the argument. Have you checked?

PR #298, opened 2017-07-06
Use expert partitioner

joseivanlopez:
PBI: https://trello.com/c/pol3pLNV/608-1-storageng-enter-the-expert-partitioner-from-installation.

### DIFF
--- a/src/lib/y2storage/clients/inst_disk_proposal.rb
+++ b/src/lib/y2storage/clients/inst_disk_proposal.rb
@@ -101,7 +101,7 @@ module Y2Storage
         dialog = Y2Partitioner::Dialogs::Main.new
         result = dialog.run(storage_manager.y2storage_probed, @devicegraph)
         case result
-        # TODO: method Y2Partitioner::Dialogs::Main#run does not return :abort when aborting.
+        # NOTE: method Y2Partitioner::Dialogs::Main#run does not return :abort when aborting.
         when :abort
           @result = :abort
         when :next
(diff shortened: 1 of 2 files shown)


src/lib/y2storage/clients/inst_disk_proposal.rb
```diff
@@ -99,7 +98,9 @@ def guided_setup(settings)
       end
 
       def expert_partitioner
-        Y2Partitioner::Clients::Main.run
+        dialog = Y2Partitioner::Dialogs::Main.new
+        result = dialog.run(storage_manager.y2storage_probed, @devicegraph)
+        @devicegraph = dialog.device_graph if result == :next
```
ancorgs: Shouldn't `@proposal` be set to nil?

src/lib/y2storage/clients/inst_disk_proposal.rb
```diff
@@ -99,7 +98,16 @@ def guided_setup(settings)
       end
 
       def expert_partitioner
-        Y2Partitioner::Clients::Main.run
+        dialog = Y2Partitioner::Dialogs::Main.new
+        result = dialog.run(storage_manager.y2storage_probed, @devicegraph)
+        case result
+        # TODO: method Y2Partitioner::Dialogs::Main#run does not return :abort when aborting.
```
ancorgs: Well, there is nothing TODO here. We are just prepared to the moment in which the dialog is fixed. This is, at most, a `NOTE`, but not a `TODO`.

PR #300, opened 2017-07-06
Summary expert partitioner

joseivanlopez:
PBI: https://trello.com/c/pol3pLNV/608-1-storageng-enter-the-expert-partitioner-from-installation.

Task: (optional, to be replaced with workflow later) Possible to enter partitioner from summary (https://trello.com/c/Bf7k0xKs/601-storageng-showing-proposal-in-the-general-proposal-overview).

### DIFF
--- a/src/lib/y2storage/clients/partitions_proposal.rb
+++ b/src/lib/y2storage/clients/partitions_proposal.rb
@@ -25,6 +25,7 @@ require "yast"
 require "y2storage"
 require "y2storage/actions_presenter"
 require "installation/proposal_client"
+require "y2partitioner/dialogs/main"
 
 Yast.import "Popup"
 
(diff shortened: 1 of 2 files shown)


ancorgs:
I tested this manually and, to be honest, it doesn't work very will. Looks like it tries to refresh the proposal UI without actually having closed the partitioner before. Weird stuff on pressing `next`, `back`...

ancorgs:
@joseivanlopez Fortunately I have the receipt to make everything work as expected (verified to work):

1.- Rebase with master
2.- Include `InstDialogMixin` in `PartitionsProposal`
3.- Change your version of the `expert_partitioner` method into this:
```ruby
def expert_partitioner
  dialog = Y2Partitioner::Dialogs::Main.new
  result = without_title_on_left do
    probed = storage_manager.y2storage_probed
    staging = storage_manager.y2storage_staging
    dialog.run(probed, staging)
  end
  storage_manager.staging = dialog.device_graph if result == :next
  result
end 
```

As a bonus, it will also be rendered properly (full size).

ancorgs:
Now it should work, according to my previous tests. LGTM

PR #301, opened 2017-07-07
MD RAID support for AutoYaST

ancorgs:
It also includes the first steps for cloning (exporting) MD arrays to the AutoYaST profile.


src/lib/y2storage/proposal/md_creator.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        md.md_level = planned_md.md_level if planned_md.md_level
+        md.chunk_size = planned_md.chunk_size if planned_md.chunk_size
+        md.md_parity = planned_md.md_parity if planned_md.md_parity
+
+        devices = device_names.map do |dev_name|
+          device = Y2Storage::BlkDevice.find_by_name(new_graph, dev_name)
+          device.encryption || device
+        end
+        planned_md.add_devices(devices, md)
```
joseivanlopez: These methods are a bit weird. They receive the real `md` as parameter and modify it. Would it not be better to have the real device as an attribute of the planned device?

test/autoinst_profile/partitioning_section_test.rb
```diff
@@ -103,15 +103,48 @@
...
+    end
 
-  describe "#md_drives" do
-    it "returns drives which type :CT_MD"
+    describe "#md_drives" do
+      it "returns drives which type is :CT_MD, even if they look invalid" do
+        expect(section.md_drives).to contain_exactly(drive5, drive6, wrongdrv2, wrongdrv5)
+      end
+
+      it "doesn't include drives of other types with device='/dev/md'" do
```
joseivanlopez: Nitpicking: we are trying to use "does not" in texts for the user.

test/autoinst_profile/raid_options_section_test.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    it "initializes parity_algorithm" do
+      expect(raid_options.parity_algorithm).to eq("left_asymmetric")
+    end
+
+    it "initializes raid_type" do
+      expect(raid_options.raid_type).to eq("raid0")
+    end
+
+    it "initializes device_order"
```
ancorgs: No. `.new_from_storage` is actually used for cloning the system (export), which will be a future PBI. Imo just created a first unfinished draft for it and left the pending test as some kind of "bonus track". :)

test/autoinst_profile/raid_options_section_test.rb
```diff
@@ -0,0 +1,131 @@
...
+      end
+    end
+
+    context "when device order is not specified" do
+      let(:spec) { {} }
+
+      it "sets device order to an empty array" do
+        expect(raid_options.device_order).to eq([])
+      end
+    end
```
ancorgs: No, when any other attribute is not specified, it gets initialized to `nil` (as anyone would have expected from a struct-like class).  This is the only one that is special because it's not an scalar but an array, so we have to make sure it gets an empty array by default instead of `nil`.

test/autoinst_proposal_test.rb
```diff
@@ -274,6 +270,130 @@
...
+      end
+
+      it "creates a RAID" do
+        proposal.propose
+        devicegraph = proposal.devices
+        expect(devicegraph.md_raids).to contain_exactly(
+          an_object_having_attributes(
+            "number" => 1
+          )
+        )
```
ancorgs: Yes, I will add such check.

src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -162,6 +167,15 @@ def init_from_disk(disk)
         true
       end
 
+      # Device name to be used for the real MD device
+      #
+      # @see PartitionSection#name_for_md for details
+      #
+      # @return [String] MD RAID device name
+      def name_for_md
+        partitions.first.name_for_md
```
ancorgs: No if the profile is correct. We have a separate PBI to add proper error reporting for cases in which we receive non-sense profiles. But I will add a comment to make sure we don't skip this particular point when adding such error reporting.

src/lib/y2storage/autoinst_profile/raid_options_section.rb
```diff
@@ -0,0 +1,92 @@
...
+        result = new
+        result.init_from_raid(device)
+        result
+      end
+
+      # Method used by {.new_from_storage} to populate the attributes when
+      # cloning RAID options
+      #
+      # @param md [Md] RAID device
+      def init_from_raid(md)
```
ancorgs: Is actually undocumented and a complete non-sense, since all RAIDs use persistent superblock by default since the '90s. Even Arvin was not aware of that option of the possibility of having a non-persistent superblock (something not supported even by the old libstorage). I will document all that.

src/lib/y2storage/planned/md.rb
```diff
@@ -0,0 +1,99 @@
...
+      # @see Y2Storage::Md#devices
+      #
+      # The block devices will be added to the array in the order specified by
+      # {#devices_order}. If #{devices_order} is nil or empty, the devices will
+      # be added in alphabetical order (AutoYaST documented behavior).
+      #
+      # @param devices [Array<Y2Storage::BlkDevice>] block devices to add
+      # @param md_device [Y2Storage::Md] MD device created to represent this
+      #   planned device. It will be modified.
+      def add_devices(devices, md_device)
```
ancorgs: I will switch the order. And about the bang, I recently read a post from Matz (Ruby father) saying that bang methods only make sense if there is a bang-free alternative. So I stopped abusing bang names. The rest will be corrected little by little.

src/lib/y2storage/proposal/md_creator.rb
ancorgs: Maybe. It's done in this way for consistency with the rest of the `Planned` and `WhateverCreator` API. But I will add a comment about reconsidering it in the future.

PR #303, opened 2017-07-07
Run the partitioner in full size during installation

ancorgs:
Added as a separate method because it will also be needed here https://github.com/yast/yast-storage-ng/pull/300


src/lib/y2storage/inst_dialog_mixin.rb
```diff
@@ -0,0 +1,47 @@
...
+    #
+    # @example
+    #   dialog = Y2Partitioner::Dialogs.main.new
+    #   dialog_result = without_title_on_left do
+    #     dialog.run
+    #   end
+    def without_title_on_left(&block)
+      Yast::Wizard.OpenNextBackDialog
+      result = block.call
+      Yast::Wizard.CloseDialog
```
jreidinger: what about using ensure to close wizard even if block raise exception?

ancorgs: good point

src/lib/y2storage/inst_dialog_mixin.rb
```diff
@@ -0,0 +1,48 @@
...
+    #   dialog = Y2Partitioner::Dialogs.main.new
+    #   dialog_result = without_title_on_left do
+    #     dialog.run
+    #   end
+    def without_title_on_left(&block)
+      Yast::Wizard.OpenNextBackDialog
+      result = block.call
+    ensure
+      Yast::Wizard.CloseDialog
+      result
```
jreidinger: this result is skipped.
try it on your own in irb:
```
irb(main):001:0> def a
irb(main):002:1> return 5
irb(main):003:1> ensure
irb(main):004:1* 6
irb(main):005:1> end
=> :a
irb(main):006:0> a
=> 5
```

so it can looks like
```
def without_title_on_left(&block)
￼   Yast::Wizard.OpenNextBackDialog
￼   result = block.call
ensure
  Yast::Wizard.CloseDialog
end
```

PR #304, opened 2017-07-10
Update installer-hacks

gilsonsouza:


ancorgs:
Well, if the hack is gone now I would simply remove the whole entry instead of saying it uses storage-ng now. I guess the goal is to get an empty document (no hacks) in the end. :-)

PR #305, opened 2017-07-10
Add PartitionId#to_human_string and PartitionId::formatable?

jreidinger:


test/partition_id_test.rb
```diff
@@ -95,4 +95,24 @@
...
+    it "returns true if partition can be formatted" do
+      expect(Y2Storage::PartitionId::NTFS.formattable?).to eq true
+    end
+
+    it "returns false otherwise" do
+      expect(Y2Storage::PartitionId::PREP.formattable?).to eq true
+    end
+  end
+
+  describe "#to_human_string" do
```
ancorgs: Good tests to detect new constants added in libstorage.

jreidinger: yes, that is intention. To catch it early enough.

PR #307, opened 2017-07-11
Legacy boot flag in proposal

ancorgs:
Depends on the discussion at https://trello.com/c/RAPZuM0j/640-3-storageng-bootloader-bootloader-does-not-detect-prep-boot-partition

If we conclude that the legacy boot must be set for PReP, this PR should do it. Otherwise, this can be closed right away.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jul 11 06:21:36 UTC 2017 - ancor@suse.com
+
+- In the partitioning proposal, manage the legacy_boot flag in GPT
+  partition tables in the same way than the boot flag in MBR ones.
+
+-------------------------------------------------------------------
 Thu Jul  6 08:11:59 UTC 2017 - gsouza@suse.com
 
 - Added verifications for disk in network.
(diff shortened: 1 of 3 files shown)


ancorgs:
The conclusion is that we solve it in three places:

 - Bootloader should not "overcheck". That's already addressed in https://github.com/yast/yast-bootloader/pull/448) which should be enough to make the installation succeed.
- `PartitionCreator` should be adapted to:
  - never touch the legacy_boot flag in PPC
  - to treat it as the MBR boot flag (that is, honoring `Planned::Partition#bootable` like suggested in this PR) in other architectures
- To make things more clear in `BootRequirementsChecker` (specially in the RSpec tests we are using to discuss stuff), we should adapt it to never propose `#bootable == true` for a PReP planned partition within a GPT disk.

So I will modify this PR to cover the last two points.

wfeldt:
To clarify this `legacy_boot` flag in GPT which `grub2` sets, here is the relevant part of the UEFI spec:

> UEFI Spec 2.6
> 
> 5.3.3 GPT Partition Entry Array
> 
> Attribute, Bit 2 (Legacy BIOS Bootable)
> 
> This bit is set aside by this specification to let systems with traditional PC-AT BIOS firmware implementations inform certain limited, special-purpose software running on these systems that a GPT partition may be bootable.

GPT just doesn't have a 'bootable' flag.. Instead it uses dedicated partition (gui)ids to indicate partitions to use for booting. Like for `prep` or `efi-bios`.

teclator:
@ancorgs this PR is probably out of date, and taking in account how fat storage-ng has been changing maybe it does not make sense anymore.

Could you check it and close if it is not needed.

ancorgs:
Too old.

PR #308, opened 2017-07-11
version + changelog

mvidner:
documenting 3ac7b91afe2f4ba13a4295303a0c3e2c029d0f67 from #305 so that
yast2-partitioner can depend on it in https://github.com/yast/yast-partitioner/pull/26

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jul 11 11:18:06 UTC 2017 - jreidinger@suse.com
+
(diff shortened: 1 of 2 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jul 11 11:18:06 UTC 2017 - jreidinger@suse.com
+
+- Added PartitionId#to_human_string and PartitionId::formatable?
```
gilsonsouza: PartitionId::formattable? with double t

mvidner: right; fixed; thanks!

PR #309, opened 2017-07-11
Filesystem type extend

jreidinger:


package/yast2-storage-ng.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jul 11 15:08:42 UTC 2017 - jreidinger@suse.com
+
+- Added Filesystem::Type#supported_fstab_options and
```
ancorgs: It's actually `Filesystems::Type` (plural)

src/lib/y2storage/filesystems/type.rb
```diff
@@ -148,6 +171,24 @@ def to_human_string
...
+      # @return [Array<String>]
+      def supported_fstab_options
+        properties = PROPERTIES[to_sym]
+        default = []
+        return default unless properties
+        properties[:fstab_options] || default
+      end
+
+      # return the best fitting partition id for given filesystem type
+      def default_partition_id
```
ancorgs: Maybe add a note here like:
```ruby
#
# @note: Take into account that the default partition id can be inappropriate for some
#   partition tables. Consider using {PartitionTable::Base#partition_id_for} to translate
#   the result to a supported id before assigning it to a partition.
```

Other than that, I miss the `@return` tag here

jreidinger: will add it, thanks for note

jreidinger:
failing due to monkey patch in partitioner, so do not merge without partitioner pr ( WIP)

ancorgs:
Do not create the corresponding partitioner PR. I will handle everything as part of https://trello.com/c/Be6kThDA/647-2-merge-yast2-partitioner-and-yast2-storage-ng

jreidinger:
@ancorgs thanks for doing it. This sprint we have no partitioner PBI to hide it into it. So I have to focus at other parts of code.

ancorgs:
Replaced by #310 (still in progress)

PR #310, opened 2017-07-12
Repatriate the expert partitioner into yast2-storage-ng

ancorgs:
See https://trello.com/c/Be6kThDA/647-2-merge-yast2-partitioner-and-yast2-storage-ng

Reviewing commit by commit should be quite easy


jreidinger:
Please do not forget to merge also documentation from partitioner README especially hints about testing.

ancorgs:
@jreidinger Yes, merging READMEs was in the tasks list of the PBI... and now is done. So this PR is ready for final review.

src/lib/y2partitioner/refinements/filesystem_type.rb
```diff
@@ -0,0 +1,42 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage"
+
+module Y2Partitioner
+  module Refinements
+    # Temporary refinement for Y2Storage::Filesystems::Type. It should be
+    # replaced by a proper class or any other mechanism at some point.
```
jreidinger: in FACT I prefer to have it as constant in format and mount or method, that decide if format UI should be offered

package/yast2-storage-ng.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Wed Jul 12 15:37:16 UTC 2017 - ancor@suse.com
+
+- Replaced the original prototype of the expert partitioner with
+  the one coming from yast2-partitioner.
+- Removed Y2Storage::DevicesLists and not longer needed modules
+  from Y2Storage::Refinements
```
jreidinger: I am missing here mention for new methods in Filesystem::Type

.rubocop.yml
```diff
@@ -18,30 +18,6 @@ Metrics/LineLength:
...
-    - 'src/lib/expert_partitioner/tab_views/*/*'
-    - 'src/lib/expert_partitioner/tree_views/*'
-    - 'src/lib/expert_partitioner/tree.rb'
-
-# These methods use CamelCase to mimic the libyui API
-Style/MethodName:
-  Exclude:
-    - 'src/lib/expert_partitioner/ui_extensions.rb'
-
 # This two are not equivalent if using refinements (and we use them to add
```
jreidinger: it this one still needed? how much Refinements we have there?

ancorgs: This is not part of this PBI, but of another one for cleaning up legacy stuff. See task in https://trello.com/c/BWtX9ccg/548-2-remove-refinements-patches-extensions-etc-from-yast-storage-ng

The point is that, even if we stop using refinements, some team members hate the SymbolProc syntax. So this exception will probably stay for different reasons. Something to negotiate... but not now.

package/yast2-storage-ng.changes
ancorgs: I can add it but... are we planning to document in the changelog every single method added to the Y2Storage classes? It made sense (together with the version increase) when a separate y2-partitioner package needed them. But...

jreidinger: well, it makes sense for me, as it is public API and y2-partitioner is not the only user of API.
another option for me is to not mention it in changelog and use instead `@since` tag so it is easy to see which version add it. And same we should do for method removal, so lifetime of API is clear

src/lib/y2partitioner/refinements/filesystem_type.rb
ancorgs: Well, I don't care much about the final solution. The scope of this PBI was just removing the monkey patch so nothing explodes. I assume this part of the partitioner is still under heavy development.

PR #312, opened 2017-07-13
Dasd alignment

joseivanlopez:
WIP

PBI: https://trello.com/c/UBCLA6Qo/651-2-storageng-and-s-390-partitions-not-aligned.

This PR depends on https://github.com/openSUSE/libstorage-ng/pull/303.

### DIFF
--- a/test/y2storage/partition_tables_test.rb
+++ b/test/y2storage/partition_tables_test.rb
@@ -61,6 +61,9 @@ describe Y2Storage::PartitionTables::Base do
     end
   end
 
+  # This test ensures the behavior of libstorage is stable in the future, because, in order
+  # to generate aligned partitions, the proposal code heavily relies on the fact that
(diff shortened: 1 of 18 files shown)


test/y2storage/partition_tables_test.rb
```diff
@@ -49,11 +51,38 @@
...
+
     it "deletes all partitions in table" do
       # NOTE: it is important here to have partition table with logical partitions
       # as it need special handling, so test will cover it
       ptable.delete_all_partitions
       expect(ptable.partitions).to be_empty
     end
   end
+
+  describe "#unused_partition_slots" do
```
ancorgs: I would maybe add some comment here explaining why we added this test. Something like:
```ruby
# This tests ensures the behavior of libstorage is stable in the future, because, in order
# to generate aligned partitions, the proposal code heavily relies on the fact that
# Y2Storage::PartitionTable#unused_partition_slots returns aligned regions.
```

PR #316, opened 2017-07-18
Dasd end-alignment

joseivanlopez:
PBI: https://trello.com/c/UBCLA6Qo/651-2-storageng-and-s-390-partitions-not-aligned.

Enforced end-alignment for last partition in a DASD partition table.

### DIFF
--- a/src/lib/y2storage/planned/has_size.rb
+++ b/src/lib/y2storage/planned/has_size.rb
@@ -88,7 +88,7 @@ module Y2Storage
         #     partitions.
         # @return [Array] list containing devices with an adjusted value
         #     for Planned::HasSize#size
-        def distribute_space(devices, space_size, rounding: nil, align_grain: nil)
+        def distribute_space(devices, space_size, rounding: nil, align_grain: nil, end_alignment: false)
           raise RuntimeError if space_size < DiskSize.sum(devices.map(&:min))
 
           rounding ||= align_grain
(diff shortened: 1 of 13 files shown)


src/lib/y2storage/planned/has_size.rb
```diff
@@ -169,7 +173,8 @@ def distributable?(size, rounding)
           size >= rounding
         end
 
-        def adjust_size_to_last_slot!(device, space_size, align_grain)
+        def adjust_size_to_last_slot!(device, space_size, align_grain, end_alignment)
+          return if align_grain.nil? || end_alignment
```
ancorgs: NP: I'm not sure whether it's a good idea to move the condition here. It implies the method receives an extra argument that is only used to force a quick exit of the method... so not really an argument for the calculations. That's weird.

And it also makes harder to see the relation between calling (or not) `adjust_size_to_last_slot!` (by the way, nice occasion to remove the bang) and assigning the unused space... which makes me realize we should have something like
```ruby
# The last space is extended until the end if we are working with partitions (align_grain is
# not nil) and the partition table allows that (end_alignment is false)
adjust_to_end = !align_grain.nil? && !end_alignment

adjust_size_to_last_slot(new_list.last, space_size, align_grain) if adjust_to_end
extra_size = space_size - DiskSize.sum(new_list.map(&:size))
unused = distribute_extra_space!(new_list, extra_size, rounding)
new_list.last.size += unused if adjust_to_end && unused < align_grain
```

PR #320, opened 2017-07-20
Popup to ask for multipath activation

ancorgs:
Manually tested

### DIFF
--- a/test/y2storage/callbacks/activate_test.rb
+++ b/test/y2storage/callbacks/activate_test.rb
@@ -69,4 +69,32 @@ describe Y2Storage::Callbacks::Activate do
       end
     end
   end
+
+  describe "#multipath" do
+    before do
+      allow(Yast::Popup).to receive(:YesNo).and_return answer
+    end
+    let(:answer) { true }
+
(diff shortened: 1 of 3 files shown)


ancorgs:
@aschnell The tests run in my system but fails in travis. Could it be that libstorage-ng is detecting multipath in Travis and trying to activate it. If that's the case, shouldn't multipath activation been disabled when running with `Storage::ProbeMode_NONE`?

aschnell:
activate() is a separate function call independent of probe(). If you don't need it you should not call it. And in unit tests it should not be called as if modifies the system and needs root privileges. FMPOV activate() should only be called during installation and update, likely not even during installation from a Live DVD.

test/y2storage/callbacks/activate_test.rb
```diff
@@ -69,4 +69,32 @@
...
     end
   end
+
+  describe "#multipath" do
+    before do
+      allow(Yast::Popup).to receive(:YesNo).and_return answer
+    end
+    let(:answer) { true }
+
+    it "asks the user wheter to activate multipath" do
```
joseivanlopez: Typo: whether.

PR #321, opened 2017-07-21
Correct partition names

joseivanlopez:
PBI: https://trello.com/c/yh9J6T5F/656-5-storageng-multipath-and-dm-raid-in-the-proposal.

Generate partitions with correct name.

### DIFF
--- a/test/spec_helper.rb
+++ b/test/spec_helper.rb
@@ -52,4 +52,10 @@ require_relative "support/storage_helpers"
 
 RSpec.configure do |c|
   c.include Yast::RSpec::StorageHelpers
+
+  c.before do
+    # Storage::Storage#activate should not be called during testing because we are not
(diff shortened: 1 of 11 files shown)


test/spec_helper.rb
```diff
@@ -52,4 +52,10 @@
 
 RSpec.configure do |c|
   c.include Yast::RSpec::StorageHelpers
+
+  c.before do
+    # Storage::Storage#activate should not be called during testing because we are not
+    # doing real probing.
+    allow_any_instance_of(Storage::Storage).to receive(:activate).and_return(true)
```
ancorgs: An acceptable workaround hopefully to be removed when completing https://trello.com/c/BWtX9ccg/548-2-remove-refinements-patches-extensions-etc-from-yast-storage-ng

PR #322, opened 2017-07-24
Remove old API from StorageManager

joseivanlopez:
PBI: https://trello.com/c/BWtX9ccg/548-2-remove-refinements-patches-extensions-etc-from-yast-storage-ng.

### DIFF
--- a/src/lib/y2storage/storage_manager.rb
+++ b/src/lib/y2storage/storage_manager.rb
@@ -29,10 +29,8 @@ require "y2storage/disk_analyzer"
 require "y2storage/callbacks/activate.rb"
 
 module Y2Storage
-  #
   # Singleton class to provide access to the libstorage Storage object and
   # to store related state information.
-  #
   class StorageManager
     include Yast::Logger
     extend Forwardable
@@ -50,6 +48,7 @@ module Y2Storage
     # Zero means no modification (still not probed). Incremented every
     # time the staging devicegraph is re-assigned.
     # @see #copy_to_staging
+    # @see #staging_changed
     #
     # @return [Integer]
     attr_reader :staging_revision
@@ -79,6 +78,8 @@ module Y2Storage
     def initialize(storage_environment)
       @storage = Storage::Storage.new(storage_environment)
       @probed = false
+      reset_probed
+      reset_staging
       reset_staging_revision
     end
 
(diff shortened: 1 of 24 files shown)


src/lib/y2storage/storage_manager.rb
```diff
@@ -164,25 +190,6 @@ def commit
       storage.commit
```
ancorgs: Out of scope, but maybe good to add a FIXME since we are touching all this: we need to add proper CommitCallbacks to yast2-storage-ng

src/lib/y2storage/storage_manager.rb
```diff
@@ -257,14 +276,17 @@ def create_instance(storage_environment = nil)
...
+        @instance.reset_staging_revision
 
         fake_graph = Devicegraph.new(@instance.storage.create_devicegraph("fake"))
         Y2Storage::FakeDeviceFactory.load_yaml_file(fake_graph, yaml_file) if yaml_file
-        fake_graph.copy(@instance.y2storage_probed)
-        fake_graph.copy(@instance.y2storage_staging)
+
+        fake_graph.to_storage_value.copy(@instance.storage.probed)
+        fake_graph.to_storage_value.copy(@instance.storage.staging)
         @instance.storage.remove_devicegraph("fake")
```
ancorgs: Shouldn't the removal of `fake` be within an `ensure`?

src/lib/y2storage/storage_manager.rb
```diff
@@ -257,14 +276,17 @@ def create_instance(storage_environment = nil)
       #
```
ancorgs: Leftover `@see #refresh!`

src/lib/y2storage/storage_manager.rb
```diff
@@ -281,20 +303,14 @@ def fake_from_yaml(yaml_file = nil)
...
```
ancorgs: Leftover `@see #refresh!`

src/lib/y2storage/storage_manager.rb
```diff
@@ -281,20 +303,14 @@ def fake_from_yaml(yaml_file = nil)
...
       # @return [StorageManager] singleton instance
       def fake_from_xml(xml_file)
```
ancorgs: We maybe should rethink these methods a little bit. The were creating (literally from documentation) as "an alternative to the instance method"... when the instance method implied probed. Now that probing and creating the instance are different things, maybe we should reconsider how they work.

test/y2partitioner/test_helper.rb
```diff
@@ -5,7 +5,8 @@
 
 def devicegraph_stub(name)
   path = File.join(TEST_PATH, "data", "devicegraphs", name)
-  storage = Y2Storage::StorageManager.fake_from_yaml(path)
+  Y2Storage::StorageManager.instance.probe_from_yaml(path)
+  storage = Y2Storage::StorageManager.instance
```
ancorgs: NP: doesn't it more sense swapping the lines like this?
```ruby
storage = Y2Storage::StorageManager.instance
storage.probe_from_yaml(path)
```

PR #326, opened 2017-07-28
Add btrfs edition to expert partitioner

joseivanlopez:
PBI: https://trello.com/c/eCoRKS7f/667-3-storageng-expert-partitioner-btrfs.


src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,94 @@
+require "yast"
+require "y2partitioner/dialogs/popup"
+
+module Y2Partitioner
+  module Dialogs
+    # Popup dialog to create a btrfs subvolume
+    class BtrfsSubvolume < Popup
+      attr_accessor :path
+      attr_accessor :nocow
```
jreidinger: where is this used? I do not see any read or write of it.

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,94 @@
...
+        )
+      end
+    end
+  end
+
+  # Input field to set the subvolume path
+  class SubvolumePath < CWM::InputField
+    attr_reader :dialog
+
+    def initialize(dialog)
```
jreidinger: I suggest to not call it dialog. Widgets should be independent on dialog that is using it. I suggest to call it something like generic data holder. So maybe something like `data` and document that requirement is to have methods `path/path=` . This way it is much easier in future to reuse this widget in different dialog and its dependency is much cleaner.

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,94 @@
...
+    def focus
+      Yast::UI.SetFocus(Id(widget_id))
+    end
+  end
+
+  # Input field to set the subvolume nocow attribute
+  class SubvolumeNocow < CWM::CheckBox
+    attr_reader :dialog
+
+    def initialize(dialog)
```
jreidinger: same as above.

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,94 @@
...
+      _("noCoW")
+    end
+
+    def store
+      dialog.nocow = value
+    end
+
+    def init
+      self.value = dialog.nocow || false
+    end
```
jreidinger: dialog does not have help text? otherwise I think thiat noCoW should have help method with explanation what it stands for.

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,94 @@
...
+    def label
+      _("noCoW")
+    end
+
+    def store
+      dialog.nocow = value
+    end
+
+    def init
+      self.value = dialog.nocow || false
```
jreidinger: for boolean value this is discouraged, as in future accidental change of default to true breaks it.
I suggest to have code like
```
self.value = dialog.nocow.nil? ? dialog.nocow : false
```
and maybe use constant for default value so it looks like

```
self.value = dialog.nocow.nil? ? dialog.nocow : DEFAULT_VALUE
```

src/lib/y2partitioner/dialogs/popup.rb
```diff
@@ -0,0 +1,37 @@
...
+          MinSize(50, 18, ReplacePoint(Id(:contents), Empty())),
+          VSpacing(1),
+          VStretch(),
+          ButtonBox(
+            PushButton(Id(:help), Opt(:helpButton), Yast::Label.HelpButton),
+            PushButton(Id(:ok), Opt(:default), Yast::Label.OKButton),
+            PushButton(Id(:cancel), Yast::Label.CancelButton)
+          )
+        )
+      end
```
jreidinger: This looks quite usefull, maybe we can in future add it to CWM itself. Only questionable part is that MinSize.

src/lib/y2partitioner/widgets/btrfs_subvolumes.rb
```diff
@@ -0,0 +1,64 @@
...
+require "y2partitioner/widgets/btrfs_subvolumes_delete_button"
+require "y2partitioner/device_graphs"
+
+Yast.import "Mode"
+
+module Y2Partitioner
+  module Widgets
+    # Widget to manage btrfs subvolumes of a specific filesystem
+    #
+    # FIXME: How to handle events directly from a CWM dialog?
```
jreidinger: well, ideally it should be handled by CWM, but maybe something is wrong here. It need some investigation.

src/lib/y2partitioner/widgets/btrfs_subvolumes_add_button.rb
```diff
@@ -0,0 +1,111 @@
...
+        result = subvolume_dialog.run
+
+        path = subvolume_dialog.path
+        nocow = subvolume_dialog.nocow
+
+        if result == :ok
+          ensure_default_subvolume
+
+          if path.empty?
+            Yast::Popup.Message(_("Empty subvolume path not allowed."))
```
jreidinger: this should be in SubvolumePath in validation method.

src/lib/y2partitioner/widgets/btrfs_subvolumes_add_button.rb
```diff
@@ -0,0 +1,111 @@
...
+          else
+            path = fix_path(path)
+
+            if exist_path?(path)
+              Yast::Popup.Message(format(_("Subvolume name %s already exists."), path))
```
jreidinger: and same here. Validation of content is widget job.

src/lib/y2partitioner/widgets/btrfs_subvolumes_add_button.rb
```diff
@@ -0,0 +1,111 @@
...
+      end
+
+      def add_subvolume(path, nocow)
+        parent = filesystem.default_btrfs_subvolume
+        subvol = parent.create_btrfs_subvolume(path)
+        subvol.nocow = nocow
+        subvol.mountpoint = path
+      end
+
+      def ensure_default_subvolume
```
jreidinger: this smells like it should be method of Btrfs Filesystem class and not widget or dialog.

src/lib/y2partitioner/widgets/btrfs_table.rb
```diff
@@ -0,0 +1,158 @@
...
+      end
+
+      def items
+        filesystems.map { |f| values_for(f) }
+      end
+
+      def selected_filesystem
+        return nil if items.empty? || !value
+
+        device_name = value[/table:partition:(.*)/, 1]
```
jreidinger: maybe can be also added to BlkDeviceTable helper.

src/lib/y2storage/devicegraph.rb
```diff
@@ -90,6 +90,12 @@ class Devicegraph
     #   @return [Device]
     storage_forward :find_device, as: "Device"
 
+    # @!method remove_device(device)
+    #   Removes the device from the devicegraph
+    #
+    #   @param device [Device]
```
jreidinger: does it return something? and can it raise exception?

src/lib/y2storage/fake_device_factory.rb
```diff
@@ -689,15 +690,15 @@ def find_default_subvolume(btrfs)
     def create_subvolume(parent, args)
       log.info("#{__method__}( #{parent}, #{args} )")
       path  = args["path"]
-      nocow = args["nocow"] || "false"
+      nocow = args["nocow"] || false
```
jreidinger: same as above, having `|| false` looks wrong.

jreidinger: ideally here `args.fetch("nocow", false)`

src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -121,6 +121,35 @@ def top_level_btrfs_subvolume
...
+      #
+      # @param path [String] subvolume path
+      # @return [BtrfsSubvolume, nil] nil if it does not find a subvolume with this path
+      def find_btrfs_subvolume_by_path(path)
+        return nil unless supports_btrfs_subvolumes?
+
+        btrfs = Storage.to_btrfs(to_storage_value)
+        storage_subvol = btrfs.find_btrfs_subvolume_by_path(path)
+        BtrfsSubvolume.new(storage_subvol)
+      rescue Storage::BtrfsSubvolumeNotFoundByPath
```
jreidinger: there is no way to check it in advance? It looks wrong for me to have Exception part of common workflow.

jreidinger: btw why not use simplier
```
btrfs_subvolumes.find { |s| s.path == path }
```

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
joseivanlopez: It is used by the widgets, but I agree it is a bit weird. I will change it.

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
joseivanlopez: This is a very is small dialog to enter a new btrfs subvolume. For now, the help for **path** and **noCoW** attributes is in the table from you launch this small dialog.

src/lib/y2storage/filesystems/blk_filesystem.rb
joseivanlopez: Well, here different things happen. First, and in my opinion this is questionable, we do not have a specific class for each type of filesystem in storage-ng as there is in libstorage-ng. I think the most correct it would be to have one wrapper class for each filesystem. Right now, we only have `BlkFilesystem` class, and some specific btrfs methods are implemented in that class calling directly to libstorage-ng methods, see for example

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,192 @@
...
+            PushButton(Id(:ok), Opt(:default), Yast::Label.OKButton),
+            PushButton(Id(:cancel), Yast::Label.CancelButton)
+          )
+        )
+      end
+
+      # Form object for the dialog
+      #
+      # Widgets use this object to storage data
+      class Form
```
jreidinger: I suggest to use slightly longer data, that is more obvious. Like `SubvolumeForm`. Form standalone is too ambitious name

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,192 @@
...
+
+        # Validates the path
+        #
+        # Path cannot be empty
+        # Path must start by default subvolume path
+        # Path must be uniq
+        def validate
+          valid = true
+
+          focus
```
jreidinger: this is strange, it should not set focus when it is valid, as it make usage harder when there are more widgets. So set it only when invalid.

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
joseivanlopez: Actually it is a subclass: `Dialogs::Subvolume::Form`.

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,192 @@
...
+          self.value = form.path
+        end
+
+        # Validates the path
+        #
+        # Path cannot be empty
+        # Path must start by default subvolume path
+        # Path must be uniq
+        def validate
+          valid = true
```
jreidinger: this is not needed if you use other form `invalid` as it is if not set `nil`.

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,192 @@
...
+            valid = false
+          elsif !filesystem.nil?
+            fix_path
+            if exist_path?
+              Yast::Popup.Message(format(_("Subvolume name %s already exists."), value))
+              valid = false
+            end
+          end
+
+          valid
```
jreidinger: so with my suggestions how it will look like:
```
if value.empty?
  Yast::Popup.Message(_("Empty subvolume path not allowed."))
  invalid = true
elseif !filesystem.nil?
  fix_path
  if exist_path?
    Yast::Popup.Message(format(_("Subvolume name %s already exists."), value))
     invalid = true
  end
end

if invalid
  focus
  return false
else
  true
end
```
```

File: src/lib/y2partitioner/dialogs/btrfs_subvolume.rb [hunk 19]
```diff
@@ -0,0 +1,192 @@
...
+        def fix_path
+          default_subvolume_path = default_path
+          prefix = default_subvolume_path + "/"
+
+          return value if value.start_with?(prefix)
+
+          message = format(
+            _("Only subvolume names starting with \"%s\" currently allowed!\n" \
+              "Automatically prepending \"%s\" to name of subvolume."), prefix, prefix
+          )
```
jreidinger: format support named parameters so you can use - https://ruby-doc.org/core-2.2.0/Kernel.html#method-i-format
```
          format(
            _("Only subvolume names starting with \"%{prefix}\" currently allowed!\n" \
              "Automatically prepending \"%{prefix}\" to name of subvolume."), prefix: prefix
          )
```

jreidinger: in general I suggest to use named name always when there is at least two arguments, as it allows translators to even switch placeholders, but it it not possible if plain `%s` is used

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -0,0 +1,192 @@
...
+
+        # @param form [Dialogs::BtrfsSubvolume::Form]
+        # @param filesystem [Y2Storage::Filesystems::BlkFilesystem] a btrfs filesystem
+        def initialize(form, filesystem: nil)
+          @form = form
+          @filesystem = filesystem
+        end
+
+        def label
+          _("noCoW")
```
jreidinger: I really suggest to write translator hint here. Consider that you are not IT expert and you have see such string, what it is? I suggest something like
```
# TRANSLATORS: noCoW is accronym to "not use Copy on Write" feature for Btrfs. If no suitable expression exists in your language, then better keep it as it is as it is expert value.
```

src/lib/y2partitioner/dialogs/fstab_options.rb
```diff
@@ -18,42 +19,13 @@ def title
...
-          ButtonBox(
-            PushButton(Id(:help), Opt(:helpButton), Yast::Label.HelpButton),
-            PushButton(Id(:ok), Opt(:default), Yast::Label.OKButton),
-            PushButton(Id(:cancel), Yast::Label.CancelButton)
-          )
+        HBox(
+          HStretch(),
+          HSpacing(1),
+          HVSquash(Widgets::FstabOptions.new(@options)),
+          HStretch(),
```
jreidinger: I am not sure if it make sense to have in HBox two Stretch, as first stretch should already eat all space, if I am not wrong.

src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -27,6 +29,16 @@ def handle
         @pager.handle("ID" => id)
       end
 
+      # Device object selected in the table
+      #
+      # @return [Y2Storage::BlkDevice, nil] nil if anything is selected
+      def selected_device
+        return nil if items.empty? || !value
+
+        device_name = value[/table:(.*):(.*)/, 2]
```
jreidinger: first matching in regexp is useless. So I suggest to use `value[/table:.*:(.*)/, 1]`

src/lib/y2partitioner/widgets/btrfs_edit_button.rb
```diff
@@ -0,0 +1,39 @@
...
+        _("Edit...")
+      end
+
+      # Opens a dialog to manage the list of subvolumes. In case of there is no
+      # selected table row, it shows an error.
+      def handle
+        filesystem = table.selected_filesystem
+
+        if filesystem.nil?
+          Yast::Popup.Error(_("There are no filesystems to edit."))
```
jreidinger: I am not good at English, but maybe here should be `There is no filesystem selected to edit`. @mvidner how do you see it.

src/lib/y2partitioner/widgets/btrfs_subvolumes.rb
```diff
@@ -0,0 +1,82 @@
...
+          VSpacing(1),
+          table,
+          HBox(
+            Widgets::BtrfsSubvolumesAddButton.new(table),
+            Widgets::BtrfsSubvolumesDeleteButton.new(table)
+          )
+        )
+      end
+
+      # FIXME: this should belongs to the dialog
```
jreidinger: well, in fact this help is fine to have here. PRoblem I see is `handle_help` which is dialog job to collect help. ( so move FIXME to that handle method ).

src/lib/y2partitioner/widgets/btrfs_subvolumes_delete_button.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      # Deletes a subvolume
+      #
+      # A confirmation popup is showed before deleting the subvolume.
+      # An error message is presented when there is no selected table row.
+      def handle
+        subvolume = table.selected_subvolume
+
+        if subvolume.nil?
+          Yast::Popup.Error(_("There are any subvolume selected"))
```
jreidinger: here, sentence does not make sense. Maybe missing "not"? in general I suggest simple "Nothing selected". It should be easy enough and maybe can be used also in "Edit"

src/lib/y2partitioner/widgets/btrfs_subvolumes_table.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      # Column titles
+
+      def path_title
+        # TRANSLATORS: table header, subvolume path e.g. "@/home"
+        _("Path")
+      end
+
+      def nocow_title
+        # TRANSLATORS: table header, nocow subvolume attribute
```
jreidinger: 👍 for translators help. I just suggest to write what nocow accronym strands for ( do not use copy on write feature )

src/lib/y2partitioner/widgets/btrfs_subvolumes_delete_button.rb
```diff
@@ -0,0 +1,58 @@
...
+        else
+          result = Yast::Popup.YesNo(
+            # TRANSLATORS: %s is the path of the subvolume to be deleted
+            format(_("Really delete subvolume %s?"), subvolume.path)
```
joseivanlopez: Actually, the mountpoint and path could be different, but we are using the same at this moment. As that could change, I will not use "subvolume at".

src/lib/y2partitioner/widgets/btrfs_subvolumes_table.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      # Help
+
+      def path_help
+        _("<b>Path</b> shows the subvolume path.")
+      end
+
+      def nocow_help
+        _("<b>noCoW</b> shows the subvolume noCoW attribute. " \
+          "If set, the subvolume is not copy on write.")
```
jreidinger: what about short explanation about CoW? something like
```
_("<b>noCoW</b> shows the subvolume noCoW attribute. " \
     "If set, the subvolume explicitelly does not use btrfs copy on write feature. This means, that when something is copied, it always create new data instead of postponed write when data is modified. Useful when runtime performance is required, but initialization can be slower, so slow copy is done during intialization and there is no risk for postponed copy during run of application.")
```

src/lib/y2partitioner/widgets/btrfs_table.rb
```diff
@@ -0,0 +1,164 @@
...
+
+      # Table help
+      #
+      # Shows help of all columns
+      def help
+        header = _(
+          "<p>This view shows all Btrfs filesystems.</p>" \
+          "<p>The table contains:</p>" \
+        )
+        fields = HELP_FIELDS.map { |f| helptext_for(f) }.join("\n")
```
jreidinger: Do you test it? As help use richtext, maybe it is better to join it with `<br>` instead of `\n`.

src/lib/y2storage/devicegraph.rb
```diff
@@ -90,6 +90,13 @@ class Devicegraph
...
+    #   @raise [Exception] if the device is not found
```
jreidinger: 👍

src/lib/y2storage/fake_device_factory.rb
```diff
@@ -689,15 +690,15 @@ def find_default_subvolume(btrfs)
     def create_subvolume(parent, args)
       log.info("#{__method__}( #{parent}, #{args} )")
       path  = args["path"]
-      nocow = args["nocow"] || "false"
+      nocow = args.fetch("nocow", false)
```
jreidinger: so `args["nocow"]` is now boolean and not string?

src/lib/y2storage/filesystems/blk_filesystem.rb
jreidinger: @ancorgs how do you see it? Does it make sense to have specific Btrfs class like libstorage-ng have?

src/lib/y2storage/fake_device_factory.rb
joseivanlopez: Yes, the correct it should be `false`.

src/lib/y2partitioner/dialogs/fstab_options.rb
joseivanlopez: It seems to it is necessary the second one if we want the content centered.

jreidinger: For centering, do you try HCenter? https://yast-ui-bindings.surge.sh/classYAlignment.html

src/lib/y2partitioner/widgets/btrfs_edit_button.rb
```diff
@@ -27,7 +27,7 @@ def handle
         filesystem = table.selected_filesystem
 
         if filesystem.nil?
-          Yast::Popup.Error(_("There are no filesystems to edit."))
+          Yast::Popup.Error(_("There is no filesystem selected to edit."))
```
jreidinger: It should be consistent with Error at `btrfs_subvolumes_delete_button.rb`

src/lib/y2partitioner/widgets/btrfs_subvolumes_table.rb
```diff
@@ -115,7 +116,12 @@ def path_help
 
       def nocow_help
         _("<b>noCoW</b> shows the subvolume noCoW attribute. " \
-          "If set, the subvolume is not copy on write.")
+          "If set, the subvolume explicitly does not use Btrfs copy on write feature. " \
+          "Copy on write means that when something is copied, the resource is shared without " \
+          "doing a real copy. The shared resource is actually copied when first write operation " \
+          "is performed. With noCoW, the resource is always copied during initialization. " \
+          "This is useful when runtime performace is required, and there is no risk for delaying " \
```
jreidinger: I think it need someone with better english skill, but at least I replace ", and" with ", so "

src/lib/y2partitioner/dialogs/fstab_options.rb
ancorgs: I have seen (and used) @joseivanlopez's solution in several places of YaST. It works better when resizing and all that

src/lib/y2storage/devicegraph.rb
```diff
@@ -90,6 +90,13 @@ class Devicegraph
     #   @return [Device]
     storage_forward :find_device, as: "Device"
 
+    # @!method remove_device(device)
+    #   Removes the device from the devicegraph
+    #
+    #   @param device [Device]
```
ancorgs: Strictly speaking, I believe this would also work receiving a sid as argument, not only when receiving the Device object.

src/lib/y2storage/devicegraph.rb
```diff
@@ -90,6 +90,13 @@ class Devicegraph
     #   @return [Device]
     storage_forward :find_device, as: "Device"
 
+    # @!method remove_device(device)
+    #   Removes the device from the devicegraph
```
ancorgs: This documentation should be improved to alert against its usage. Right now, the documentation in `libstorage-ng` is better and more explicit than this forwarded method. That should not happen in general.

src/lib/y2storage/filesystems/blk_filesystem.rb
ancorgs: Whatever we do, I think this PBI is the right time to address it. Note that "this PBI" doesn't imply this PR. I'm fine with merging as it is and fixing the subclassing issue in a separate PR inside the same PBI.

Back to the subject, IMO we need to subclass because there is too much Btrfs specific code here now. The question is whether that should imply that we need to create a wrapper for each possible `BlkFilesystem` subclass (that is, we must create all the sibling classes for `Btrfs`). As I see it, we have two options:

- Create only the `Btrfs < BlkFilesystem` class, but no other subclasses. That's simpler and makes sense since `Btrfs` is the only block filesystem with an extended API. All the other `BlkFilesystem` subclasses in libstorage simply adhere to the parent class API, without offering any new method (so they don't deserve to be separate classes on their own). This approach have a problem, that we have a class that have one subclass (`Btrfs < BlkFilesystem`) but it's not abstract. That could be considered an "API smell".

- Another solution would be to create the following 9 classes as subclasses of block filesystem - `Btrfs`, `Ext`, `Iso9660`, `Ntfs`, `Reiserfs`, `Swap`, `Udf`, `Vfat` and `Xfs`. That's closer to libstorage-ng (which is good and bad). The point is that 8 of those 9 classes offer absolutely nothing else than the parent `BlkFilesystem` class, so they are irrelevant from the API point of view and they just need maintenance job to yast-storage-ng (we need to add just another empty subclass for each newly supported filesystem).

So we have to choose one API smell here: having a subclass with no siblings (solution 1) or having many subclasses that are only different in the implementation but not from the API point of view (solution 2).

src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -121,15 +132,67 @@ def top_level_btrfs_subvolume
...
+        nil
+      end
+
+      # Creates a new default btrfs subvolume if it does not exist or the top level
+      # subvolume is set as default
```
ancorgs: It would be good to succinctly explain why do we need to create an extra subvolume to become the default when the top level subvolume is already the default. Sure it makes sense from the Btrfs expert point of view... but most developers will forget the implications and will need to re-check the rationale often.

src/lib/y2storage/filesystems/blk_filesystem.rb
jreidinger: @ancorgs what about to use second option, but generate all non-existing subclasses automatic? So maintenance is zero and we can even test it.

ancorgs: I'm not only worried about the maintenance, that's just one aspect. I'm even more worried about the API smell (making public classes out of something that should only be private strategies).

OTOH, generating classes via metaprogramming will probably not help to make the thing more understandable.

src/lib/y2storage/filesystems/blk_filesystem.rb
ancorgs: On the other hand, from the description it looks like it creates a new subvolume or does nothing, but that's not true. In some cases, it takes an existing subvolume and sets it as default. That's unexpected from the description.

src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -121,15 +132,67 @@ def top_level_btrfs_subvolume
...
+        return nil unless supports_btrfs_subvolumes?
+
+        subvolume = default_btrfs_subvolume
+
+        if subvolume.nil? || subvolume.top_level?
+          subvolume = btrfs_subvolumes.detect { |s| s.path == default_btrfs_subvolume_path }
+
+          if subvolume.nil?
+            subvolume = create_default_subvolume
+          elsif !subvolume.default_btrfs_subvolume?
```
ancorgs: So if `default_btrfs_subvolume_path` is the empty string and the top level subvolume is set as default it actually does nothing. Isn't it? That's again contradictory with the description.

Sorry if I'm asking stupid questions, but this top_level vs default subvolume thingies are confusing.

src/lib/y2storage/filesystems/blk_filesystem.rb
joseivanlopez: Ok, maybe it is something to discuss. My idea was that if it is necessary to create a default subvolume and the default path is `@foo` and it already exists (but not as default), then it is set as default. I agree the doc should be better.

test/y2partitioner/dialogs/btrfs_subvolume_test.rb
```diff
@@ -0,0 +1,130 @@
...
+            expect(Yast::Popup).to receive(:Message)
+            subject.validate
+          end
+
+          it "removes extra slashes and prepend the default subvolume path" do
+            expect(subject).to receive(:value=).with("foo").ordered
+            expect(subject).to receive(:value=).with(/^@\/.*/).ordered
+            subject.validate
+          end
+        end
```
ancorgs: I miss a test for what happens if:
- The default subvolume is the top one (that is, no "@")
- The user tries to introduce an absolute path (starting with "/")

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -0,0 +1,131 @@
...
+        return if subvolume.nil?
+
+        subvolume.remove_descendants
+        devicegraph.to_storage_value.remove_device(subvolume.to_storage_value)
+      end
+
+      # The path that a new default btrfs subvolume should have
+      #
+      # @return [String]
+      def self.default_btrfs_subvolume_path
```
ancorgs: I believe that using this to create the default subvolume when creating a new btrfs filesystem (or when setting the type of the filesystem to btrfs) should be part of this pull request.

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -0,0 +1,131 @@
...
+      #
+      # If a specific default subvolume path is requested, create it; if not,
+      # use the toplevel subvolume that is implicitly created by mkfs.btrfs.
+      #
+      # This default subvolume is the parent for all others on Btrfs 'filesystem'.
+      #
+      # @param path [String, nil] path for the default subvolume
+      #
+      # @return [BtrfsSubvolume]
+      def get_or_create_default_btrfs_subvolume(path: nil)
```
joseivanlopez: Only one detail. With that version, if default subvolume is "@":

```
filesystem.get_or_create_default_btrfs_subvolume(path: "")  # => subvolume for "@"
```
I guess it should be:

```
filesystem.get_or_create_default_btrfs_subvolume(path: "")  # => subvolume for ""
filesystem.get_or_create_default_btrfs_subvolume(path: nil)  # => subvolume for "@"
```

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -0,0 +1,126 @@
...
+      # If a specific default subvolume path is requested, returns a subvolume with
+      # that path; if not, use the toplevel subvolume that is implicitly created by mkfs.btrfs.
+      #
+      # This default subvolume is the parent for all others on Btrfs 'filesystem'.
+      #
+      # @param path [String, nil] path for the default subvolume
+      #
+      # @return [BtrfsSubvolume]
+      def ensure_default_btrfs_subvolume(path: nil)
+        # If no path is specified, we are fine with whatever the current default subvolume is
```
ancorgs: This documentation about the meaning for `path: nil` should be in the Yardoc, not here.

src/lib/y2storage/filesystems/btrfs.rb
ancorgs: This returns "" if the value is indeed "" (meaning the top level subvolume) and if the value is not there. Just asking - wouldn't it be better to return nil if the value is not specified?

I don't think is relevant in general, but it would make more sense when compared to the `#ensure_default_btrfs_subvolume` (which really makes a difference between nil and "").

As said, not a big deal in reality. Check `#ProposalSettings#product_feature` to see how to differentiate nil and "".

src/lib/y2partitioner/format_mount/base.rb
```diff
@@ -50,7 +50,14 @@ def apply_format_options!
           @partition.password = @options.password
         end
 
-        @partition.create_filesystem(@options.filesystem_type) if @options.format
+        if @options.format
+          filesystem = @partition.create_filesystem(@options.filesystem_type)
+
+          if filesystem.is?(:btrfs)
```
ancorgs: It would be better to use `#supports_btrfs_subvolumes?`

PR #327, opened 2017-08-02
Fix alignment bug

joseivanlopez:
Ensure end-alignment when distributing extra space.

This should solve alignment problem detected in s380 openQA test: https://yast-openqa.suse.cz/tests/12374#step/start_install/5.

Part of PBI: https://trello.com/c/UBCLA6Qo/651-2-storageng-and-s-390-partitions-not-aligned.

### DIFF
--- a/src/lib/y2storage/planned/has_size.rb
+++ b/src/lib/y2storage/planned/has_size.rb
@@ -160,10 +160,10 @@ module Y2Storage
           extra_size = extra_size.ceil(rounding)
           extra_size = available_size.floor(rounding) if extra_size > available_size
 
-          new_size = extra_size + device.size
+          new_size = device.size + extra_size
           if new_size > device.max_size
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/planned/has_size.rb
```diff
@@ -160,10 +160,10 @@ def device_extra_size(device, total_size, total_weight, assigned_size, rounding)
           extra_size = extra_size.ceil(rounding)
           extra_size = available_size.floor(rounding) if extra_size > available_size
 
-          new_size = extra_size + device.size
+          new_size = device.size + extra_size
```
jreidinger: what is reason for such change? I hope you do not overwrite plus to not be commutative.

joseivanlopez: Only because I think is easier to understand the line, but it has no effect.

PR #328, opened 2017-08-02
- adapted to libstorage-ng API change

aschnell:
For https://trello.com/c/2tJ3LlpM/672-5-storageng-bios-raid-via-md-devices.


aschnell:
Tests failed because Travis still has the old libstorage-ng. OBS already has the new and fails with the old yast2-storage-ng.

PR #330, opened 2017-08-08
adjust examples to new y2storage interface

wfeldt:


src/examples/fake_probing_demo.rb
```diff
@@ -24,11 +24,11 @@
 
 require "y2storage"
 
-devicegraph = Y2Storage::StorageManager.fake_from_yaml.y2storage_probed
+Y2Storage::StorageManager.instance.probe_from_yaml
```
ancorgs: Probably, a more clear way to achieve this would be `Y2Storage::StorageManager.create_test_instance` instead of faking with an empty yaml.

PR #331, opened 2017-08-09
submit to default YaST project (YaST:Head)

wfeldt:


wfeldt:
Also replace changes file with a complete log?

PR #332, opened 2017-08-11
[WIP] Subvolumes shadowing

joseivanlopez:
Some refac, doc and unit tests are pending.

PBI: https://trello.com/c/2XZ1k97t/681-5-storageng-expert-partitioner-manage-shadowing-of-btrfs-subvolumes.

### DIFF
--- a/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
+++ b/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
@@ -1,6 +1,7 @@
 require "yast"
 require "cwm"
 require "y2partitioner/dialogs/popup"
+require "y2storage/filesystems/btrfs"
 
 Yast.import "Popup"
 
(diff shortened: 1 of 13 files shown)


src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -150,6 +155,23 @@ def fix_path
         def exist_path?
           filesystem.btrfs_subvolumes.any? { |s| s.path == value }
         end
+
+        # Checks if the mount point already exists in the system
+        # @see Mountable#shadowed?
+        #
+        # @return [Boolean]
+        def exist_mount_point?
```
ancorgs: Being shadowed and being duplicated is not the same.

src/lib/y2partitioner/format_mount/root_subvolumes_builder.rb
```diff
@@ -0,0 +1,248 @@
...
+
+require "yast"
+require "y2storage"
+require "y2storage/subvol_specification"
+require "y2partitioner/device_graphs"
+
+module Y2Partitioner
+  module FormatMount
+    # Class to manage root subvolumes in expert partitioner
+    class RootSubvolumesBuilder
```
ancorgs: I find this approach to singleton quite confusing because the public instance API is not really public (since there is no reasonable way to get an instance of `RootSubvolumesBuilder`). The public API of an instance is in fact the "internal" API used by the class, than then offer the really public API of the whole thing.

So, for example, `reload` or `remove_shadowed_subvolumes` are never called by the users of the class although they are public and need to remain public for the class methods to manipulate the encapsulated instance.

Too much brain twisting for my taste.

ancorgs:
This is obsolete. Superseded by https://github.com/yast/yast-storage-ng/pull/337

PR #338, opened 2017-08-21
PartitionId#to_human_string more similar to old storage

ancorgs:


wfeldt:
Personally, I'd like 'Linux' more. But since we've had the 'native'  before, lgtm.

ancorgs:
Yes, same for me. The "native" part is a little bit weird, but...

PR #341, opened 2017-08-25
First proposal of control.xml structure (to be refined)

ancorgs:
For easier review, take a look to the [rendered version of the document](https://github.com/ancorgs/yast-storage-ng/blob/3303b8b8f8e379c447a52681c70de901fa382071/doc/old_and_new_proposal.md)

### DIFF
--- a/doc/old_and_new_proposal.md
+++ b/doc/old_and_new_proposal.md
@@ -368,14 +368,16 @@ fashion.
 
 ## Some examples of control file with the new format
 
-**WIP.**
-
 This section shows how it would be possible to reproduce the current behavior of
 several products via the new proposal and the new suggested format for
 `control.xml`. Hopefully, it also illustrates how easy would be to accommodate
 changes that are currently very hard or impossible to achieve.
 
 ### Current behavior of SLES
+
+Some values are a direct translation of the current `control.xml` and others
+(like the relationship with Windows partitions) are inferred from the typical SLES
+use case.
 
 ```xml
 <partitioning>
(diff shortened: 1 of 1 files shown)


doc/old_and_new_proposal.md
```diff
@@ -108,18 +178,304 @@ addition, every volume gets a "weight" (so far, based on the
...
+have read and understood the above section titled "How the new proposal
+distributes the space".
+
+### General structure
+
+The new `partitioning` section will still contain the
+`root_subvolume_read_only`, `proposal_settings_editable`,
+`expert_partitioner_warning` and `use_separate_multipath_module` options
+described in this document's subsection "Other general options" at "Settings in
+yast-storage". These options will retain its exact form and meaning.
```
joseivanlopez: their exact form ...

doc/old_and_new_proposal.md
```diff
@@ -108,18 +178,304 @@ addition, every volume gets a "weight" (so far, based on the
...
+The `proposal` subsection will be used to configure some general aspects of the
+storage proposal (referenced as "guided setup" in the UI) and will contain the
+following options.
+
+  * `default_lvm`. Whether LVM should be used by default.
+  * `default_encrypt`. Whether encryption should be used by default.
+  * `default_windows_delete`. Default value for the automatic delete mode for
+    Windows partitions. It can be `none`, `all` or `ondemand`. For more
+    information, see the description of the new proposal above.
+  * `default_linux_delete`. Default value for the automatic delete mode for
```
joseivanlopez: Only to clarify, I would say explicitly that the options are `none`, `all` or `ondemand` again.

doc/old_and_new_proposal.md
```diff
@@ -108,18 +178,304 @@ addition, every volume gets a "weight" (so far, based on the
...
+    Linux partitions. See `default_windows_delete` above.
+  * `default_other_delete`. Default value for the automatic delete mode for
```
joseivanlopez: And the same here.

doc/old_and_new_proposal.md
```diff
@@ -108,18 +178,304 @@ addition, every volume gets a "weight" (so far, based on the
...
+    other partitions. See `default_windows_delete` above.
+  * `default_resize_windows`. Default value for the user setting deciding
+    whether to resize Windows systems if needed.
+  * `lvm_vg_strategy`. If the user decides to use LVM, strategy to decide the
+    size of the volume group (and, thus, the number and size of created physical
+    volumes). It can be `use_available`, `use_needed` or `use_vg_size`. See the
```
joseivanlopez: For the possible `lvm_vg_strategy` options, I understood the following:

* `use_available`: the VG will be created with all available free space in the system, then the VG size can be greater than the sum of LVs size.
* `use_needed`: the VG size is equal to LVs size.
* `use_vg_size`: fix a size for the VG, so VG size is greater or equal than LVs size.

Am I right? Maybe this explanation should be included.

doc/old_and_new_proposal.md
```diff
@@ -108,18 +178,304 @@ addition, every volume gets a "weight" (so far, based on the
...
+    given, YaST will use a fallback one based on the mount point.
+  * `default_fstype`. Default file system type to format the volume.
+  * `desired_size`. Initial size to use in the first proposal attempt.
+  * `min_size`. Initial size to use in the second proposal attempt.
+  * `max_size`. Maximum size to assign to the volume. It can also contain the
+    value "unlimited" (meaning as big as possible). This will be considered the
+    default value if the option is not present.
+  * `max_size_lvm`. When LVM is used, this option can be used to override the
+    value at `max_size`.
+  * `weight`. Value used to distribute the extra space (after assigning the
```
joseivanlopez: I would say: "after assigning the initial ones" instead of "the minimal ones".

doc/old_and_new_proposal.md
```diff
@@ -108,18 +178,304 @@ addition, every volume gets a "weight" (so far, based on the
...
+    minimal ones) among the volumes.
+  * `adjust_by_ram`. Whether the initial and max sizes of each attempt should
+    be adjusted based in the RAM size. Possible values are "true", "false",
+    "optional", with "false" being the default if not present and "optional"
+    meaning the user can decide via UI. So far the adaptation consist in
```
joseivanlopez: Typo, consists.

doc/old_and_new_proposal.md
```diff
@@ -108,18 +178,304 @@ addition, every volume gets a "weight" (so far, based on the
...
+    ensuring all the sizes are, at least, as big as the RAM. In the future, an
+    extra `adjust_by_ram_mode` option could be added to allow other approaches.
+  * `default_adjust_by_ram`. If the previous value is "optional", whether to
```
joseivanlopez: This option looks a little bit weird to me. I would expect something like `adjust_by_ram` with values `true` or `false`, and then an option `optional_adjust_by_ram` with values `true` or `false` meaning that the user can change the value.

doc/old_and_new_proposal.md
```diff
@@ -108,18 +178,304 @@ addition, every volume gets a "weight" (so far, based on the
...
+
+Some options will only apply if the chosen filesystem type for the volume is
+Btrfs, in some cases with the same name and meaning that in the old
+`control.xml` format. The main difference is that now the setting will apply to
+the volume in which it's included, not necessarily to the root ("/") one. In the
+expert partitioner, if a Btrfs filesystem is created and assigned to the mount
+point of the volume, these settings will also be used to suggest the filesystem
+options.
+
+  * `snapshots`. Whether snapshots should be activated. Possible values are
```
joseivanlopez: Same than before. Maybe it would be clearer to have `snaphsots` and `optional_snapshots`.

doc/old_and_new_proposal.md
ancorgs: Yes, you are (right). I will include that.

doc/old_and_new_proposal.md
```diff
@@ -108,18 +178,453 @@ addition, every volume gets a "weight" (so far, based on the
...
+In the first iteration, it will look for the lowest number there. If
+`adjust_by_swap` is optional in that volume and enabled, it will disable it. If
+that is not enough and snapshots are optional but enabled, it will disable them
+and try again (assuming Btrfs is being used). If that's still not enough, it
+will disable the whole volume if it's optional.
+
+If that's not enough, it will keep those settings and look for the next volume
+with some value in `disable_order` to perform the same operation in a cumulative
+fashion.
+
```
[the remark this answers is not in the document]
ancorgs: The number and order of attempts will depend on the control file. But in the SLES example provided in this document it would be still reasonable:

* First disabling /home
* Second also not trying to enlarge swap (if it makes sense)
* Third without snapshots.

PR #348, opened 2017-09-04
Snapshots on proposal

ancorgs:
Last commit to cover the Guided proposal part of https://trello.com/c/ytK5Yoe4/554-5-storageng-installation-to-root-subvolume (the others are already merged)

### DIFF
--- a/src/lib/y2storage/planned/can_be_formatted.rb
+++ b/src/lib/y2storage/planned/can_be_formatted.rb
@@ -127,7 +127,7 @@ module Y2Storage
       # @return [Boolean] true if snapshots are requested (see {#snapshots=}) and
       #   possible (so far, only for Btrfs root filesystems).
       def snapshots?
-        @snapshots && btrfs? && mount_point == "/"
+        @snapshots && btrfs? && root?
       end
 
     protected
--- a/src/lib/y2storage/planned/can_be_formatted.rb
+++ b/src/lib/y2storage/planned/can_be_formatted.rb
@@ -54,6 +54,7 @@ module Y2Storage
       def initialize_can_be_formatted
         @subvolumes = []
         @reformat = false
+        @snapshots = false
       end
 
       # Creates a filesystem for the planned device on the specified real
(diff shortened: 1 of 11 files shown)


src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -115,15 +116,32 @@ def reformat?
...
       # Creates subvolumes in the previously created filesystem that is placed
       # in the final device.
       #
-      # This also sets other Btrfs attributes, like the default subvolume.
+      # This also sets other Btrfs attributes, like the default subvolume or
+      # Filesystems::Btrfs#configure_snapper
       #
       # @param filesystem [Filesystems::BlkFilesystem]
       def btrfs_setup(filesystem)
+        filesystem.configure_snapper = snapshots? if filesystem.respond_to?(:configure_snapper=)
```
joseivanlopez: Does it have sense to "configure snapper" if the fs does not support btrfs subvolumes? Maybe to move this line after `return unless filesystem.supports_btrfs_subvolumes?` ?

src/lib/y2storage/planned/can_be_mounted.rb
```diff
@@ -37,6 +37,12 @@ module CanBeMounted
       # Initializations of the mixin, to be called from the class constructor.
       def initialize_can_be_mounted
       end
+
+      # Whether the device will be mounted as root
+      # @return [Boolean]
+      def root?
+        mount_point == "/"
```
joseivanlopez: We already have `Y2Storage::Mountable::ROOT_PATH`. Should we start to use that constant instead of literal "/"?

src/lib/y2storage/proposal/devices_planner.rb
```diff
@@ -173,23 +173,32 @@ def root_device
...
         multiplicator = 1.0 + settings.btrfs_increase_percentage / 100.0
+        log.info "Increasing root filesystem size for snapshots (#{multiplicator})"
         planned_device.min_size *= multiplicator
         planned_device.max_size *= multiplicator
       end
 
-      def init_btrfs_subvolumes!(planned_device)
-        return unless planned_device.btrfs? && settings.subvolumes
+      def init_btrfs_subvolumes(planned_device)
+        return unless settings.subvolumes
```
joseivanlopez: I have doubts about we should set the default subvolume path even when the list of subvolumes is empty.

src/lib/y2storage/storage_manager.rb
```diff
@@ -187,6 +191,7 @@ def staging_changed?
     #
     # Beware: this method can cause data loss
     def commit
```
joseivanlopez: It would be nice to explain the implications of activating `Yast2::FsSnapshot.configure_on_install`, or simply to refer to some documentation (if exists).

src/lib/y2storage/planned/can_be_formatted.rb
ancorgs: In theory, Snapper can work with other filesystem types, although so far we don't consider it. So supporting snapper and supporting btrfs subvolumes are different questions.

src/lib/y2storage/proposal/devices_planner.rb
ancorgs: Good point... but not for this PR :wink:

src/lib/y2storage/storage_manager.rb
ancorgs: Well, the method is documented via yardoc in yast-yast2. But I can add a sentence, of course.

PR #350, opened 2017-09-06
Disk size parse case insensitive

imobachgs:


package/yast2-storage-ng.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Sep  6 14:59:46 UTC 2017 - igonzalezsosa@suse.com
+
+- Ignore case when parsing sizes (bsc#1056715)
```
ancorgs: This should also fix https://bugzilla.suse.com/show_bug.cgi?id=1055913. Either include that number also here or mark as duplicate.

PR #353, opened 2017-09-07
Snapshots in AutoYaST

ancorgs:


package/yast2-storage-ng.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Sep  7 08:34:59 UTC 2017 - ancor@suse.com
+
+- AutoYaST: support for Btrfs snapshots (part of #fate#318196).
```
imobachgs: I would remove the `#` at the beginning.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -188,6 +199,22 @@ def add_common_device_attrs(device, section)
...
+      #
+      # This method modifies the first argument
+      #
+      # @param device  [Planned::Device] Planned device
+      # @param drive_section [AutoinstProfile::DriveSection] AutoYaST specification
+      def add_snapshots(device, drive_section)
+        return unless device.respond_to?(:root?) && device.root?
+
+        # Always try to enable snapshots if possible
+        snapshots = true
```
imobachgs: np: I would prefer a ternary operation, but I do not have an strong opinion.

ancorgs: I found it to be less readable

ancorgs: Or at less, less obvious

PR #354, opened 2017-09-07
[Review] Request from 'schubi2' @ 'yast/yast-storage-ng/review_170907_ay_size_auto'

schubi2:
Please review the following changes:
  * f7bdd4c AY: size is -auto-

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,5 +1,5 @@
 -------------------------------------------------------------------
-Thu Sep  7 13:28:28 CEST 2017 - schubi@suse.de
+Thu Sep  7 14:28:28 CEST 2017 - schubi@suse.de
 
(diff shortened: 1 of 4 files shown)


ancorgs:
The AutoYaST documentation says that "auto" may only be used with swap and boot and that it results in "a reasonable size". https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#ay.partition_configuration

So I would say assuming "auto" is equivalent to "max" is far from handling the value correctly.

ancorgs:
@jsrain This actually is a good reminder that we need to address this sooner than later now than we are in SLE15 https://trello.com/c/vAcAibrC/638-autoyast-ng-level-3-error-handling

Otherwise, we will keep adding workarounds for the crash/exceptions... and we will end up forgetting some.

PR #358, opened 2017-09-13
Extend udev

jreidinger:


src/lib/y2storage/blk_device.rb
```diff
@@ -180,13 +194,35 @@ def filesystem_label
...
     end
 
+    # full path of the udev by-label link or `nil` if it does not exist.
+    # e.g. "/dev/disk/by-label/DATA"
+    # @see #udev_paths
+    # @return [String]
+    def udev_full_label
+      label = filesystem_label
+
+      return nil unless label
```
ancorgs: Well, you should check for empty label here. Take into account that attributes coming from libstorage-ng (like `filesystem#label`) are never nil. At most, they are empty.

src/lib/y2storage/blk_device.rb
```diff
@@ -180,13 +194,35 @@ def filesystem_label
...
       return nil unless blk_filesystem
       blk_filesystem.uuid
     end
 
+    # full path of the udev by-uuid link or `nil` if it does not exist.
+    # e.g. "/dev/disk/by-uuid/a1dc747af-6ef7-44b9-b4f8-d200a5f933ec"
+    # @see #udev_paths
+    # @return [String]
+    def udev_full_uuid
+      uuid = filesystem_uuid
```
ancorgs: Same here. `nil != ""`

src/lib/y2storage/blk_device.rb
jreidinger: so it can be nil if there is no filesystem and can be also empty, which also means not defined, right?

test/y2storage/blk_device_test.rb
```diff
@@ -345,4 +345,86 @@
       end
     end
   end
+
+  describe "#udev_full_label" do
+    let(:device_name) { "/dev/sda1" }
+    before { allow(device).to receive(:filesystem_label).and_return(label) }
+
+    context "for devices without label" do
```
ancorgs: The tests still check only for `filesystem_label.nil?` which I'm not sure covers all the possibilities. What about some tests less coupled with the implementation? I mean, really using a device:

* Without filesystem
* With a filesystem that has no label (an empty label, nil is not possible here)
* With a labeled filesystem

Instead of just mocking the call to `filesystem_label`, which is an implementation detail. The very same methods could be implemented not relying on `filesystem_label`. I prefer mocking the whole situation when possible instead of coupling the current implementation and the tests so much.

src/lib/y2storage/blk_device.rb
```diff
@@ -180,13 +194,35 @@ def filesystem_label
       blk_filesystem.label
     end
 
+    # full path of the udev by-label link or `nil` if it does not exist.
+    # e.g. "/dev/disk/by-label/DATA"
+    # @see #udev_paths
+    # @return [String]
```
mvidner: @jreidinger To be precise, we should say `@return [String,nil]`

PR #361, opened 2017-09-19
WIP: MD raid creation

ancorgs:
This is still half baked, but it already works to a big extend. It allows to create a MD device, defining its level (aka RAID type) and the list of devices.


ancorgs:
Obsoleted by #366

PR #363, opened 2017-09-21
Ported simple_mode proposal to storage-ng

shundhammer:
https://trello.com/c/s8sAvZrA/723-1-storageng-provide-the-simple-proposal

https://bugzilla.suse.com/show_bug.cgi?id=1058736

Test result: 
casp-4 0-overview

### DIFF
--- a/src/lib/y2storage/clients/partitions_proposal.rb
+++ b/src/lib/y2storage/clients/partitions_proposal.rb
@@ -53,11 +53,13 @@ module Y2Storage
       def initialize
         textdomain "storage"
         @failed = false
+        @simple_mode = false
         ensure_proposed unless storage_manager.staging_changed?
         self.class.update_state
       end
 
-      def make_proposal(_attrs)
+      def make_proposal(param)
+        @simple_mode = param["simple_mode"] || false
         failed ? failed_proposal : successful_proposal
       end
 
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/clients/partitions_proposal.rb
```diff
@@ -137,12 +139,26 @@ def failed_proposal
...
         {
-          "preformatted_proposal" => actions_presenter.to_html,
+          "preformatted_proposal" => preformatted_proposal,
           "links"                 => actions_presenter.events,
           "language_changed"      => false
         }
       end
 
+      def preformatted_proposal
+        return nil if @failed
```
ancorgs: It is safe/expected to return `nil` here?

shundhammer: yes, this is just for consistence; it is not called from failed_proposal, but this is the value failed_proposal returns for this field.

shundhammer: See
https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/clients/partitions_proposal.rb#L130

test/y2storage/clients/partitions_proposal_test.rb
```diff
@@ -189,6 +189,29 @@
...
+      it "returns a label_proposal with 'Default'" do
+        proposal = subject.make_proposal("simple_mode" => true)
+        expect(proposal).to be_a Hash
+        expect(proposal).to include("label_proposal")
+        expect(proposal["label_proposal"]).to be_a Array
+        expect(proposal["label_proposal"]).to eq ["Default"]
+      end
+
+      it "returns a label_proposal with 'Custom' after using the partitioner" do
+        subject.make_proposal("simple_mode" => true)
```
joseivanlopez: I guess this first call to `#make_proposal` is not necessary, isn't it?

shundhammer: Since it does that normally before a user can enter the expert partitioner, this makes sure that everything is properly initialized. It might work without this, but it's more realistic to have it IMHO.

PR #366, opened 2017-09-25
Partitioner: workflow to add an MD array

ancorgs:
This allows to create a MD device, defining its level (aka RAID type), name and the list of devices. It also can be formatted, encrypted, etc.

https://trello.com/c/YDRIZHHh/700-8-storageng-expert-partitioner-create-md-array


src/lib/y2partitioner/sequences/md_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+      #
+      # It removes any previous children (like filesystems) from the devices and
+      # adapts the partition id if possible.
+      #
+      # @param devices [Array<Y2Storage::BlkDevice>]
+      def devices_in_md=(devices)
+        md.devices.each { |dev| md.remove_device(dev) }
+        devices.each do |dev|
+          dev.adapted_id = Y2Storage::PartitionId::RAID if dev.is?(:partition)
+          dev = dev.encryption if dev.encrypted?
```
joseivanlopez: @aschnell  Does it make sense to reuse the encryption when adding a partition to a MD?

ancorgs: In other words, if a partition is already encrypted and the user chooses to add that partition to the RAID, should we add the encrypted device to the RAID or should we delete the encryption device and add only the plain partition?

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -667,8 +672,53 @@ def items
...
+      end
+
+      # @macro seeAbstractWidget
+      def enable
+        @selector.enable if @selector
+      end
+
+      # @macro seeAbstractWidget
+      def disable
+        @selector.enable if @selector
```
joseivanlopez: Should it be `@selector.disable` ?

ancorgs: Indeed. Checked and I introduced a small bug. Fixing

src/lib/y2partitioner/sequences/md_controller.rb
aschnell: In theory it should work to use encrypted devices for MD RAID. But I think that was so far not supported and may not work in practice.

src/lib/y2storage/callbacks/commit.rb
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env ruby
```
joseivanlopez: Shebang necessary?

PR #367, opened 2017-09-28
allow extra volumes to be proposed (wip)

wfeldt:


ancorgs:
As commented here[1], I wouldn't abuse the current `DevicesPlanner` to support the new format IN ADDITION to all the custom logic that is already there to try to mimic the old format and behavior.

I would have the legacy (current) `DevicesPlanner` that works with the old format and a separate one that simply process the new `<volumes>` section, adding whatever comes from `BootRequirementsChecker` afterwards.

[1] https://trello.com/c/I9K56rmD/566-5-allow-to-propose-a-data-partition-changes-in-the-proposal-logic#comment-59cccf21aef6f9df8e37b7cf

wfeldt:
obsolete

PR #368, opened 2017-09-29
Handling use, mkfs_options, fstopt and mountby elements

imobachgs:
It should fix [bsc#1061042](https://bugzilla.suse.com/show_bug.cgi?id=1061042).

### DIFF
--- a/src/lib/y2storage/proposal/autoinst_space_maker.rb
+++ b/src/lib/y2storage/proposal/autoinst_space_maker.rb
@@ -72,16 +72,27 @@ module Y2Storage
 
         # TODO: resizing of partitions
 
-        case drive_spec.use
+        delete_by_use(devicegraph, disk, drive_spec.use)
+      end
+
+      # Deletes unwanted partition according to the "use" element
+      #
+      # @param devicegraph [Devicegraph]
+      # @param disk        [Disk]
+      # @param use         [String,Array<Integer>] Use value ("all", "linux", "free"
+      #   or a list of partition numbers)
+      def delete_by_use(devicegraph, disk, use)
+        return if use == "free"
+        case use
         when "all"
           disk.partition_table.remove_descendants if disk.partition_table
         when "linux"
           delete_linux_partitions(devicegraph, disk)
         when Array
-          delete_partitions_by_number(devicegraph, disk, drive_spec.use)
+          delete_partitions_by_number(devicegraph, disk, use)
         else
           # TODO: better error handling
-          log.warn "Unknown value '#{drive_spec.use}' for 'use' element in driver specification"
+          log.warn "Unknown value '#{use}' for 'use' element in driver specification"
         end
       end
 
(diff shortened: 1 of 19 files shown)


src/lib/y2storage/proposal/autoinst_space_maker.rb
```diff
@@ -72,11 +72,27 @@ def delete_stuff(devicegraph, disk, drive_spec)
...
+      #
+      # @param devicegraph [Devicegraph]
+      # @param disk        [Disk]
+      # @param use         [String,Array<Integer>] Use value ("all", "linux", "free"
+      #   or a list of partition numbers)
+      def delete_by_use(devicegraph, disk, use)
+        return if use == "free"
+        case use
         when "all"
           disk.partition_table.remove_descendants if disk.partition_table
```
joseivanlopez: Any reason to not delete the partition table?

test/y2storage/proposal/autoinst_devices_planner_test.rb
```diff
@@ -187,6 +191,26 @@
         expect(root.filesystem_type).to eq(Y2Storage::Filesystems::Type::EXT4)
         expect(home.filesystem_type).to eq(Y2Storage::Filesystems::Type::XFS)
       end
+
+      it "sets the mountby properties" do
```
joseivanlopez: Here the filesystem is checked, but not the mountby, right?

src/lib/y2storage/proposal/autoinst_space_maker.rb
imobachgs: That's a fair question. From the AutoYaST point of view, these are different use cases. If you want to wipe the partition table, you set the `initialize` element in the profile; if you only want to use the whole disk (but keeping the partition table if it exists), you set `use=all` (this case).

test/y2storage/proposal/autoinst_devices_planner_test.rb
imobachgs: Yes, you are absolutely right. I will fix it.

src/lib/y2storage/proposal/autoinst_space_maker.rb
joseivanlopez: Ok, then your code does what it is expected :). Thanks.

src/lib/y2storage/storage_manager.rb
```diff
@@ -190,12 +190,13 @@ def staging_changed?
...
     #
     # Beware: this method can cause data loss
-    def commit
+    def commit(force_rw: false)
       # Tell FsSnapshot whether Snapper should be configured later
       Yast2::FsSnapshot.configure_on_install = configure_snapper?
       callbacks = Callbacks::Commit.new
       storage.calculate_actiongraph
-      storage.commit(callbacks)
+      commit_options = ::Storage::CommitOptions.new(force_rw)
```
joseivanlopez: Maybe `CommitOptions` should have its own wrapper class.

imobachgs: As we do not need any special behavior (it is just some kind of 'struct' to pass options), I would keep it simple by now.

PR #369, opened 2017-09-29
Partitioner: fix custom size validation

joseivanlopez:
Bug: https://bugzilla.suse.com/show_bug.cgi?id=1060864

https://trello.com/c/t5RCEg1L/729-sles15-p5-1060864-alpha-5-add-partition-with-empty-size-results-in-error-client-call-failed-with-not-a-number-typeerror


imobachgs:
I general, it looks good. I am just missing the update to changes file (and version number).

PR #370, opened 2017-09-29
separate planning strategies from DevicesPlanner into DevicesPlannerStrategies module

wfeldt:
There are currently `:legacy` and a stub for `:ng`.

The `:legacy` code is basically a 1:1 copy of the old `devices_planner.rb`.

### DIFF
--- a/src/lib/y2storage/proposal/devices_planner.rb
+++ b/src/lib/y2storage/proposal/devices_planner.rb
@@ -21,11 +21,7 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "fileutils"
-require "y2storage/planned"
-require "y2storage/disk_size"
-require "y2storage/boot_requirements_checker"
-require "y2storage/exceptions"
+require "y2storage/proposal/devices_planner_strategies"
 
 module Y2Storage
   module Proposal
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/proposal/devices_planner_strategies/legacy.rb
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/env ruby
```
joseivanlopez: The class should not be considered a script itself, so we have not to include a shebang.

src/lib/y2storage/proposal/devices_planner_strategies/ng.rb
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env ruby
```
joseivanlopez: Same here.

src/lib/y2storage/proposal/devices_planner_strategies/ng.rb
```diff
@@ -0,0 +1,57 @@
...
+#
+# encoding: utf-8
+
+# Copyright (c) [2016-2017] SUSE LLC
```
joseivanlopez: Strictly speaking, this should be only 2017.

src/lib/y2storage/proposal/devices_planner.rb
```diff
@@ -35,215 +31,33 @@ module Proposal
...
-      MAX_SWAP_SIZE = DiskSize.GiB(2) # This is also DEFAULT_SWAP_SIZE
-      SWAP_WEIGHT = 100 # Importance of swap if low free disk space
+      attr_accessor :settings
 
       def initialize(settings, devicegraph)
         @settings = settings
         @devicegraph = devicegraph
+        strategy = @settings.format
+        @strategy = STRATEGIES[strategy].new(settings, devicegraph)
+      rescue
```
joseivanlopez: Here, an exception could be raised for different reasons (unknown strategy, failure in new, etc). Maybe, we should check specifically the correctness of the strategy:

```
def initialize(settings, devicegraph)
  @settings = settings
  @devicegraph = devicegraph
  @strategy = strategy
end

def strategy
  return @strategy unless @strategy.nil?

  strategy_class = 
    case @settings.format
    when :legacy
      DevicesPlannerStrategies::Legacy
    when :ng
      DevicesPlannerStrategies::Ng
    else
      err_msg = "Unsupported device planner strategy :#{@settings.format}"
      log.error err_msg
      raise DevicesPlannedStrategyError, err_msg
    end

  @strategy = strategy_class.new(@settings, @devicegraph)
end
```

joseivanlopez: In this case, the constant `DevicesPlanner::STRATEGIES` would not be necessary.

src/lib/y2storage/proposal/devices_planner_strategies/legacy.rb
wfeldt: Fine with me. But there's no consistency in the storage-ng code here.

src/lib/y2storage/proposal/devices_planner.rb
wfeldt: I'm not convinced. Having different errors for all kind of things provides a useless level of detail. In the end you'll just check for an error and not distinguish several cases while setting a strategy.

Is there anything bad with having a `DevicesPlanner::STRATEGIES` constant?

joseivanlopez: Well, I am not against about using `STRATEGIES` or raise `ArgumentError`, but what I do not see correct is to raise always the same message error independently of the exception. For example, an exception could happen in `@settings.format` or in `STRATEGIES[strategy].new` even when the `STRATEGIES[strategy]` is known.

So, I agree with you propose, but maybe raising the exception only when the strategy is not known:

```
def initialize(settings, devicegraph)
  @settings = settings
  @devicegraph = devicegraph
  strategy = @settings.format
  if STRATEGIES[strategy]
    @strategy = STRATEGIES[strategy].new(settings, devicegraph)
  else
    err_msg = "Unsupported device planner strategy :#{strategy}"
    log.error err_msg
    raise ArgumentError, err_msg 
  end
end

```

joseivanlopez:
Test coverage decreases a little bit, but it should be increased after next PRs related with this PBI.

PR #375, opened 2017-10-04
New format for proposal settings

joseivanlopez:
PBI: https://trello.com/c/I9K56rmD/566-5-allow-to-propose-a-data-partition-changes-in-the-proposal-logic

### DIFF
--- a/doc/old_and_new_proposal.md
+++ b/doc/old_and_new_proposal.md
@@ -39,7 +39,7 @@ The whole list of proposal-specific attributes is: `try_separate_home`,
 `vm_home_max_size`, `vm_keep_unpartitioned_region` and
 `btrfs_increase_percentage`.
 
-### Options to influence proposal and partitioner 
+### Options to influence proposal and partitioner
 
 In addition to the list above, there are two options, related to the usage of
 Btrfs for the root filesystem, that are used by the proposal and also by the
@@ -153,7 +153,7 @@ the default).
 
 ## How the new proposal distributes the space
 
-The new proposal uses the following two-steps approach when 
+The new proposal uses the following two-steps approach when
 deciding which partitions should be created/deleted/resized and how much
 space to assign to every new partition or logical volume.
 
@@ -264,12 +264,12 @@ use case.
 ```xml
 <partitioning>
   <proposal>
-    <lvm config:type="boolean">false</default_lvm>
-    <encrypt config:type="boolean">false</default_encrypt>
+    <lvm config:type="boolean">false</lvm>
+    <encrypt config:type="boolean">false</encrypt>
     <windows_delete_mode>all</windows_delete_mode>
-    <linux_delete>ondemand</linux_delete_mode>
+    <linux_delete_mode>ondemand</linux_delete_mode>
     <other_delete_mode>ondemand</other_delete_mode>
-    <lvm_vg_strategy_mode>use_available</lvm_vg_strategy>
+    <lvm_vg_strategy_mode>use_available</lvm_vg_strategy_mode>
   </proposal>
 
   <volumes config:type="list">
--- a/doc/old_and_new_proposal.md
+++ b/doc/old_and_new_proposal.md
@@ -269,7 +269,7 @@ use case.
     <windows_delete_mode>all</windows_delete_mode>
     <linux_delete_mode>ondemand</linux_delete_mode>
     <other_delete_mode>ondemand</other_delete_mode>
-    <lvm_vg_strategy_mode>use_available</lvm_vg_strategy_mode>
+    <lvm_vg_strategy>use_available</lvm_vg_strategy>
   </proposal>
 
   <volumes config:type="list">
--- a/src/lib/y2storage/partitioning_features.rb
+++ b/src/lib/y2storage/partitioning_features.rb
@@ -24,15 +24,45 @@ require "yast"
 Yast.import "ProductFeatures"
 
 module Y2Storage
-  # Mixin that enables a class to define attributes that are never exposed via
-  #   #inspect, #to_s or similar methods, with the goal of preventing
-  #   unintentional leaks of sensitive information in the application logs.
+  # Mixin that allows to load features from partitioning section of control file
   module PartitioningFeatures
+    # Assign the value of a specific feature into an object attribute.
+    # @param feature [List<Symbol, String>] feature path (e.g., [:proposal, :lvm])
+    # @param to [Symbol, String] attribute name where to store the feature value
+    # @param source [Hash] from where to read the features
+    #
+    # @example
+    #   settings.use_lvm #=> nil
+    #   settings.load_feature(:proposal_lvm, to: :use_lvm)
+    #   settings.use_lvm #=> false
+    #
+    # When :to parameter is not indicated, the last value of feature path is considered
+    # the attribute name:
+    #
+    # @example
+    #   settings.lvm #=> nil
+    #   settings.load_feature(:proposal, :lvm)
+    #   settings.lvm #=> false
+    #
+    # When the feature needs to be loaded from a specific subsection, the subsection can
+    # be passed as :source parameter:
+    #
+    # @example
+    #   settings.lvm #=> nil
+    #   settings.load_feature(:lvm, source: {lvm: true})
+    #   settings.lvm #=> true
+    #
     def load_feature(*feature, to: nil, source: partitioning_section)
       value = feature(*feature, source: source)
       set_feature(*feature, to, value)
     end
 
+    # Assign an integer feature (see {#load_feature})
+    # For backwards compatibility reasons, features with a value of zero are ignored.
+    #
+    # @param feature [List<Symbol, String>] feature path (e.g., [:proposal, :lvm])
+    # @param to [Symbol, String] attribute name where to store the feature value
+    # @param source [Hash] from where to read the features
     def load_integer_feature(*feature, to: nil, source: partitioning_section)
       value = feature(*feature, source: source)
       return nil if value.nil?
@@ -41,6 +71,12 @@ module Y2Storage
       set_feature(*feature, to, value) if value >= 0
     end
 
+    # Assing a size feature (see {#load_feature})
+    # The feature value is converted to a {DiskSize}.
+    #
+    # @param feature [List<Symbol, String>] feature path (e.g., [:proposal, :lvm])
+    # @param to [Symbol, String] attribute name where to store the feature value
+    # @param source [Hash] from where to read the features
     def load_size_feature(*feature, to: nil, source: partitioning_section)
       value = feature(*feature, source: source)
       return nil if value.nil?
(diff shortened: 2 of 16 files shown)


doc/old_and_new_proposal.md
```diff
@@ -264,12 +264,12 @@ use case.
...
-    <lvm config:type="boolean">false</default_lvm>
-    <encrypt config:type="boolean">false</default_encrypt>
+    <lvm config:type="boolean">false</lvm>
+    <encrypt config:type="boolean">false</encrypt>
     <windows_delete_mode>all</windows_delete_mode>
-    <linux_delete>ondemand</linux_delete_mode>
+    <linux_delete_mode>ondemand</linux_delete_mode>
     <other_delete_mode>ondemand</other_delete_mode>
-    <lvm_vg_strategy_mode>use_available</lvm_vg_strategy>
+    <lvm_vg_strategy_mode>use_available</lvm_vg_strategy_mode>
```
ancorgs: Actually the fix would be the other way around. The correct one is `lvm_vg_strategy` (without the "mode" part)

joseivanlopez: Ohh, you are right :)

src/lib/y2storage/partitioning_features.rb
```diff
@@ -0,0 +1,138 @@
...
+    # When :to parameter is not indicated, the last value of feature path is considered
+    # the attribute name:
+    #
+    # @example
+    #   settings.lvm #=> nil
+    #   settings.load_feature(:proposal, :lvm)
+    #   settings.lvm #=> false
+    #
+    # When the feature needs to be loaded from a specific subsection, the subsection can
+    # be passed as :source parameter:
```
ancorgs: I got confused by the combination of this text and the example. From the text I was expecting this argument to be the name of the subsection, not its whole content.

Maybe rephrase it like:

By default, the feature is loaded from the content of the `<partitioning>` section of the product features. The argument `source` can be used to provide the content of any other section or subsection.

joseivanlopez: Yes, much better.

src/lib/y2storage/proposal_settings.rb
```diff
@@ -24,58 +22,113 @@
...
+    attr_accessor :resize_windows
+
+    # @note :legacy format
+    # @return [DiskSize] root size used when calculating the :min size for
+    #   the proposal.
+    attr_accessor :root_base_size
+
+    # @note :legacy format
+    # @return [DiskSize] root size used when calculating the :desired size for
+    #   the proposal.
```
ancorgs: Although it's true that this is sometimes currently used as initial size for `:desired` (because we don't have a better alternative). That's not the real meaning of the attribute. This is the maximum size.

src/lib/y2storage/proposal_settings.rb
```diff
@@ -24,58 +22,113 @@
...
+    # @return [Numeric] used to adjust size when distributing extra space
+    attr_accessor :root_space_percent
+
+    # @note :legacy format
+    # @return [Numeric] used to adjust size when using snapshots
+    attr_accessor :btrfs_increase_percentage
+
+    # @note :legacy format
+    # @return [DiskSize] min disk size to allow a separate home
+    attr_accessor :min_size_to_use_separate_home
```
ancorgs: Wow! Sure we can't come with a sorter name?

src/lib/y2storage/proposal_settings.rb
```diff
@@ -24,58 +22,113 @@
...
+
+    # @note :legacy format
+    # @return [String] default btrfs subvolume path
+    attr_accessor :btrfs_default_subvolume
+
+    # @note :legacy format
```
ancorgs: In fact, this is not a proposal setting. This setting is used by the installer at the end of installation, but it doesn't affect the proposal.

src/lib/y2storage/proposal_settings.rb
ancorgs: Nervermind, I just realized it was already called like that. You only documented it

joseivanlopez: Yes, I only reused it. In fact, right now this setting is not used at all. Do you have any alternative name in mind?

src/lib/y2storage/proposal_settings.rb
ancorgs: So it's not used at all in the current `Y2Storage` and we should take the opportunity to remove it.

src/lib/y2storage/proposal_settings.rb
```diff
@@ -24,58 +22,113 @@
...
+    # @return [DiskSize] home size used when calculating the :min size for
+    #   the proposal. If space is tight, {root_base_size} is used instead.
+    #   See also {Y2Storage::DevicesPlannerStrategies::Legacy#home_device}.
+    attr_accessor :home_min_size
+
+    # @note :legacy format
+    # @return [DiskSize] home size used when calculating the :desired size for
+    #   the proposal
```
ancorgs: Again, this is not accurate. By definition this is the maximum, not the initial for `:desired`

src/lib/y2storage/proposal_settings.rb
```diff
@@ -24,58 +22,113 @@
...
     # @return [Boolean] whether to enlarge swap based on the RAM size, to ensure
     #   the classic suspend-to-ram works
     attr_accessor :enlarge_swap_for_suspend
 
+    # @note :legacy format
     # @return [String] device name of the disk in which / must be placed. If set
     #   to nil, the proposal will try to find a good candidate
     attr_accessor :root_device
 
+    # @note :legacy format
```
ancorgs: This is also used in the `:ng` format

src/lib/y2storage/proposal_settings.rb
```diff
@@ -84,32 +137,74 @@ class ProposalUserSettings
...
-      @windows_delete_mode      = :ondemand
-      @linux_delete_mode        = :ondemand
-      @other_delete_mode        = :ondemand
+    attr_reader :other_delete_mode
+
+    # @note :ng format
+    # @return [Symbol] if the user decides to use LVM, strategy to decide the size
+    #   of the volume group (and, thus, the number and size of created physical volumes).
+    #
+    #   * :use_available The VG will be created to use all the available space, thus the
```
ancorgs: NP: Details in the yardoc formatting:

1)  `:use_available` vs `use_needed:` inconsistent placing of the colon.
2) For long descriptions that are not a one-liner we usually do

```ruby
# Description
#
# @note bla
#
# @return [Symbol]
attr_reader :whatever
```

Instead of that syntax with the `@return` on top.

src/lib/y2storage/proposal_settings.rb
```diff
@@ -84,32 +137,74 @@ class ProposalUserSettings
...
+    end
+
+    # Current format used in control file
+    #
+    # Format is considered :ng only if subsections 'proposal' and 'volumes' are
+    # present in the 'partitioning' section.
+    #
+    # @return [Symbol] :ng or :legacy
+    def format
+      ng_format? ? NG_FORMAT : LEGACY_FORMAT
```
ancorgs: I see a problem with the design of `#format`, `#ng_format?` and so on. `ProductFeatures` is not immutable. Quite the opposite, nowadays looks like we are introducing more and more ways of changing it. So the mode should not be determined dynamically based on its current content. It should be determined during `#load_features` and stored in the `ProductFeatures` object alongside the read features.

src/lib/y2storage/proposal_settings.rb
```diff
@@ -142,204 +237,185 @@ def delete_forced(type)
...
-        "  root_max_size: #{root_max_size}\n" \
-        "  root_space_percent: #{root_space_percent}\n" \
-        "  btrfs_increase_percentage: #{btrfs_increase_percentage}\n" \
-        "  min_size_to_use_separate_home: #{min_size_to_use_separate_home}\n" \
-        "  btrfs_default_subvolume: #{btrfs_default_subvolume}\n" \
-        "  root_subvolume_read_only: #{root_subvolume_read_only}\n" \
-        "  home_min_size: #{home_min_size}\n" \
-        "  home_max_size: #{home_max_size}\n" \
-        "  subvolumes: \n#{subvolumes}\n"
+    def apply_ng_defaults
```
ancorgs: NP: why are the right sides of the assignation to far away from the left sides?

src/lib/y2storage/proposal_settings.rb
joseivanlopez: Stored in the `ProductFeatures`? Do you mean `ProposalSettings`?

src/lib/y2storage/proposal_settings.rb
```diff
@@ -142,204 +237,185 @@ def delete_forced(type)
...
+      self.lvm                           ||= false
+      self.resize_windows                ||= true
+      self.windows_delete_mode           ||= :ondemand
+      self.linux_delete_mode             ||= :ondemand
+      self.other_delete_mode             ||= :ondemand
+      self.lvm_vg_strategy               ||= :use_needed
```
ancorgs: So far this is the only implemented strategy. But in the future I believe it should not be the default, because I'm pretty sure it's going to be by far the slowest one and because is not what we currently do in SLE or openSUSE. `:use_available` would be a more sensible default, I would say.

joseivanlopez: Maybe a #NOTE or #FIXME comment?

src/lib/y2storage/volume_specification.rb
```diff
@@ -0,0 +1,202 @@
...
+
+    alias_method :proposed?, :proposed
+    alias_method :proposed_configurable?, :proposed_configurable
+    alias_method :adjust_by_ram?, :adjust_by_ram
+    alias_method :adjust_by_ram_configurable?, :adjust_by_ram_configurable
+    alias_method :snapshots?, :snapshots
+    alias_method :snapshots_configurable?, :snapshots_configurable
+
+    def initialize(volume_features)
+      @volume_features = volume_features
```
ancorgs: Does this mean we store in memory everything twice (processed and raw)? Is there a practical reason or do we do it "just in case"?

src/lib/y2storage/volume_specification.rb
```diff
@@ -0,0 +1,202 @@
...
+    def root?
+      mount_point && mount_point == "/"
+    end
+
+    def apply_fallbacks
+      apply_subvolumes_fallback
+      apply_fs_types_fallback
+    end
+
+    def apply_subvolumes_fallback
```
ancorgs: Mmm, this makes impossible to specify that you want an empty list of subvolumes for root. Is that what we want?

src/lib/y2storage/volume_specification.rb
```diff
@@ -0,0 +1,202 @@
...
+    def apply_fs_types_fallback
+      return unless fs_types.empty?
+
+      types = case mount_point
+      when "/"
+        Filesystems::Type.root_filesystems
+      when "/home"
+        Filesystems::Type.home_filesystems
+      else
+        []
```
ancorgs: Maybe `[fs_type]` is a more sensible default here.

src/lib/y2storage/proposal_settings.rb
```diff
@@ -142,204 +237,216 @@ def delete_forced(type)
...
+      ng_format? ? load_ng_features : load_legacy_features
     end
 
-  protected
+    # FIXME: Improve implementation. Use composition to encapsulate logic for
+    # ng and legacy formats
+    def apply_ng_defaults
+      self.lvm                 ||= false
+      self.resize_windows      ||= true
+      # TODO: once other strategies are available, change default value (:use_available)
```
ancorgs: This comment ended up in the wrong line. It refers to lvm_vg_strategy, not to windows_delete_mode.

The same happens again in the legacy version of the method

PR #376, opened 2017-10-04
Added .desktop file

shundhammer:
https://trello.com/c/85gke1CE/745-0-storageng-missing-desktop-file-for-expert-partitioner

Directory structure and Rakefile changes taken from yast-packager, a package that does not also use the autotools. desktop files from src/desktop are automatically picked up and installed to /usr/share/applications/YaST by our default rake rules.

The .desktop file is copied verbatim from yast-storage-old.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Oct  4 17:28:24 CEST 2017 - shundhammer@suse.de
+
+- Added disk.desktop file for partitioner (bsc#1059528)
+- 3.3.25
+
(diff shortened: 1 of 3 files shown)


lslezak:
> The .desktop file is copied verbatim from yast-storage-old.

And that's probably the problem, I cannot see any `disk` client in https://github.com/yast/yast-storage-ng/tree/master/src/clients. Maybe it should be `partitioner` now?

shundhammer:
Argh. You are right, of course. Fixed the name.

lslezak:
There's one more place in the file to update:

    X-SuSE-YaST-Call=disk

I'm not sure if that tag is still used, but should be updated as well...

PR #378, opened 2017-10-06
Refactor Sequences namespace. More shared and organized code

ancorgs:
Previous steps to ease implementation of https://trello.com/c/1MiaL1lF/733-5-expert-partitioner-create-a-lv

This reorganizes a little bit the `Sequences` namespace, that was a little bit hairy.

- Common functionality extracted into classes and mixins
- Controllers grouped into its own namespace

It also cleanups the requires mess for the  `Dialogs` namespace.


src/lib/y2partitioner/sequences/transaction_wizard.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+
+    protected
+
+      # Specification of the steps of the sequence.
+      #
+      # To be defined by each subclass.
+      #
+      # See UI::Sequence in yast2.
+      def sequence_hash
```
joseivanlopez: Maybe to define this as an abstract method?

ancorgs: Hmm, maybe. I was also unsure. But as an exception, I will merge as it is... because it's my last day before a long absence and I don't want to go through all the testing cycle again. :wink:

PR #380, opened 2017-10-09
Ng strategy for devices planner

joseivanlopez:
PBI: https://trello.com/c/I9K56rmD/566-5-allow-to-propose-a-data-partition-changes-in-the-proposal-logic

Implement logic for devices planner ng strategy.

### DIFF
--- a/src/lib/y2storage/proposal/devices_planner_strategies/base.rb
+++ b/src/lib/y2storage/proposal/devices_planner_strategies/base.rb
@@ -40,7 +40,7 @@ module Y2Storage
           @devicegraph = devicegraph
         end
 
-        # @!method planned_devices
+        # @!method planned_devices(target)
         #   List of devices (read: partitions or volumes) that need to be
         #   created to satisfy the settings.
         #
--- a/src/lib/y2storage/proposal/devices_planner_strategies/base.rb
+++ b/src/lib/y2storage/proposal/devices_planner_strategies/base.rb
@@ -1,0 +1,116 @@
+# encoding: utf-8
+
+# Copyright (c) [2016-2017] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage/planned"
+require "y2storage/disk_size"
+require "y2storage/boot_requirements_checker"
+require "y2storage/exceptions"
+require "abstract_method"
+
+module Y2Storage
+  module Proposal
+    module DevicesPlannerStrategies
+      # Abstract base class with common functionalty for different devices
+      # planner strategies.
+      class Base
+        include Yast::Logger
+
+        attr_accessor :settings
+
+        def initialize(settings, devicegraph)
+          @settings = settings
+          @devicegraph = devicegraph
+        end
+
+        # @!method planned_devices
+        #   List of devices (read: partitions or volumes) that need to be
+        #   created to satisfy the settings.
+        #
+        #   @note This method must be reimplemented by derived classes.
+        #
+        #   @see DevicesPlanner#planned_devices
+        #
+        #   @param target [Symbol] :desired, :min
+        #   @return [Array<Planned::Device>]
+        abstract_method :planned_devices
+
+      protected
+
+        # @return [Devicegraph]
+        attr_reader :devicegraph
+
+        # @return [Symbol] :desired or :min
+        attr_reader :target
+
+        # Planned devices needed by the bootloader
+        #
+        # @return [Array<Planned::Device>]
+        def planned_boot_devices(planned_devices)
+          checker = BootRequirementsChecker.new(
+            devicegraph, planned_devices: planned_devices, boot_disk_name: settings.root_device
+          )
+          checker.needed_partitions(target)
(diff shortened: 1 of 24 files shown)


src/lib/y2storage/proposal/devices_planner_strategies/base.rb
```diff
@@ -0,0 +1,116 @@
...
+  module Proposal
+    module DevicesPlannerStrategies
+      # Abstract base class with common functionalty for different devices
+      # planner strategies.
+      class Base
+        include Yast::Logger
+
+        attr_accessor :settings
+
+        def initialize(settings, devicegraph)
```
imobachgs: Please, document arguments.

src/lib/y2storage/proposal/devices_planner_strategies/base.rb
```diff
@@ -0,0 +1,116 @@
...
+
+        # @return [Devicegraph]
+        attr_reader :devicegraph
+
+        # @return [Symbol] :desired or :min
+        attr_reader :target
+
+        # Planned devices needed by the bootloader
+        #
+        # @return [Array<Planned::Device>]
```
imobachgs: Missing documentation for `planned_devices`.

src/lib/y2storage/proposal/devices_planner_strategies/base.rb
```diff
@@ -0,0 +1,116 @@
...
+          # exception if it's impossible to get a bootable system, even adding
+          # more partitions.
+          raise NotBootableError, error.message
+        end
+
+        # Swap partition that can be reused.
+        #
+        # It returns the smaller partition that is big enough for our purposes.
+        #
+        # @return [Partition]
```
imobachgs: Missing documentation for `required_size`.

src/lib/y2storage/proposal/devices_planner_strategies/base.rb
```diff
@@ -0,0 +1,116 @@
...
+        def reusable_swap(required_size)
+          return nil if settings.use_lvm || settings.use_encryption
+
+          partitions = devicegraph.disk_devices.map(&:swap_partitions).flatten
+          partitions.select! { |part| part.size >= required_size }
+          # Use #name in case of #size tie to provide stable sorting
+          partitions.sort_by { |part| [part.size, part.name] }.first
+        end
+
+        # Delete shadowed subvolumes from each planned device
```
imobachgs: Missing documentation for `planned_devices`.

src/lib/y2storage/proposal/devices_planner_strategies/base.rb
```diff
@@ -0,0 +1,116 @@
...
+              device.subvolumes.delete(subvolume)
+            end
+          end
+        end
+
+        # Return the total amount of RAM as DiskSize
+        #
+        # @return [DiskSize] current RAM size
+        def ram_size
+          # FIXME: use the .proc.meminfo agent and its MemTotal field
```
imobachgs: Is this something to be solved in this PR?

src/lib/y2storage/proposal/devices_planner_strategies/ng.rb
```diff
@@ -19,35 +19,136 @@
...
+        def adjust_sizes(planned_device, volume)
+          max_size = volume.max_size
+          max_size = volume.max_size_lvm if settings.lvm && volume.max_size_lvm
+          planned_device.max_size = max_size
+
+          min_size = target == :min ? volume.min_size : volume.desired_size
+          planned_device.min_size = min_size
+
+          if volume.adjust_by_ram?
+            min_size = [planned_device.min_size, ram_size].max
```
imobachgs: np: You could merge both lines.

src/lib/y2storage/proposal/devices_planner_strategies/ng.rb
```diff
@@ -19,35 +19,136 @@
...
+          planned_type = settings.lvm ? Planned::LvmLv : Planned::Partition
+          planned_device = planned_type.new(volume.mount_point, volume.fs_type)
+
+          adjust_to_settings(planned_device, volume)
+
+          planned_device
+        end
+
+        # Adjusts planned device values according to settings
+        #
```
imobachgs: I would document that planned_device is modified.

test/support/devices_planner_context.rb
```diff
@@ -0,0 +1,59 @@
...
+    allow(boot_checker).to receive(:needed_partitions).and_return(
+      [
+        Y2Storage::Planned::Partition.new("/one_boot", xfs),
+        Y2Storage::Planned::Partition.new("/other_boot", vfat)
+      ]
+    )
+    allow(devicegraph).to receive(:disk_devices).and_return [disk]
+    allow(disk).to receive(:swap_partitions).and_return(swap_partitions)
+
+    allow(Yast::Arch).to receive(:x86_64).and_return(arch == :x86_64)
```
imobachgs: I would do:

```ruby
allow(Yast::Arch).to receive(:architecture).and_return(arch)
```

And then you can define:

```ruby
let(:arch) { "x86_64" }
```

and it should work.

test/support/proposal_context.rb
```diff
@@ -67,6 +72,15 @@
     settings
   end
 
+  let(:ng_settings) do
+    file = File.join(DATA_PATH, "control_files", control_file)
+    settings = Yast::XML.XMLToYCPFile(file)
+    Yast::ProductFeatures.Import(settings)
```
imobachgs: Why do you need to import them? Is not enough with returning and doing something like:

```ruby
allow(Y2Storage::ProposalSettings).to receive(:new_for_current_product)
  .and_return(settings)
```

src/lib/y2storage/proposal/devices_planner_strategies/base.rb
joseivanlopez: Not, it is out of scope.

test/support/proposal_context.rb
joseivanlopez: I don't think so because that `settings` is only a hash, and we need a `ProposalSettings` object with all parsers done over `settings` value.

imobachgs: What looks wrong to me is that the `let` block has side-effects. IMHO the import should take care on a different place (in a before block, for instance).

On the other hand, I can live with that.

imobachgs:
It still LGTM :)

PR #381, opened 2017-10-11
Fix documentation of new settings

joseivanlopez:


wfeldt:
So far ok, but the XML examples need bigger changes: there are many more `config:type` attributes needed now. Also, the document links to an `old_and_new_control.md` that does not exist.

That said, I would in fact re-write/re-organize the entire document to make it document the old and new settings. ATM it's written in the style of a WIP doc describing some research and our plans.

Also, we'd need to update the partitioning section in the control file doc in yast-installation.

PR #389, opened 2017-10-18
Partitioner add lv

joseivanlopez:
PBI: https://trello.com/c/1MiaL1lF/733-2-left-5-expert-partitioner-create-a-lv


src/lib/y2partitioner/widgets/lvm_validations.rb
```diff
@@ -0,0 +1,65 @@
...
+      #
+      # @param vg [LvmVg]
+      # @return [Boolean]
+      def validate_add_lv(vg)
+        if vg.nil?
+          Yast::Popup.Error(_("No device selected"))
+          return false
+        end
+
+        if vg.number_of_free_extents == 0
```
ancorgs: I'm not going to do a full review, but I'm curious what's the point/reason to introduce this redundancy with `Sequences::AddLvmLv#preconditions`.

joseivanlopez: Well, the reason is because showing the popup error once the wizard has started results in a weird UI behaviour (I will do a screenshots). To avoid that, it is necessary to validate and show the popup before launching the wizard.

ancorgs:
Ok. But at least when I asked, the check was in both places. That is,
duplicated

src/lib/y2partitioner/widgets/blk_device_button.rb
```diff
@@ -0,0 +1,71 @@
...
+          Yast::Popup.Error(_("No device selected"))
+          return nil
+        end
+
+        actions
+      end
+
+    protected
+
+      attr_reader :pager
```
imobachgs: If you are not going to add documentation for each `attr_reader`, I would merge both of them.

src/lib/y2partitioner/widgets/controller_radio_buttons.rb
```diff
@@ -0,0 +1,123 @@
...
+        end
+        nil
+      end
+
+      # Get the currently selected radio button from the UI
+      def value
+        Yast::UI.QueryWidget(Id(widget_id), :CurrentButton)
+      end
+
+      # Tell the UI to change the currently selected radio button
```
imobachgs: Document for param `val` is missing.

src/lib/y2partitioner/widgets/controller_radio_buttons.rb
```diff
@@ -0,0 +1,123 @@
...
+
+    private
+
+      # @return [Array<id>]
+      def ids
+        @ids ||= items.map(&:first)
+      end
+
+      def buttons_with_widgets
+        items = self.items
```
imobachgs: why are you assigning these values? If needed, I would choose another name.

src/lib/y2partitioner/widgets/lvm_validations.rb
```diff
@@ -0,0 +1,67 @@
...
+          Yast::Popup.Error(
+            # TRANSLATORS: %s is a volume group name (e.g. "system")
+            _("No free space left in the volume group \"%s\".") % vg.vg_name
+          )
+          return false
+        end
+
+        return true
```
imobachgs: np: return is not needed here.

src/lib/y2partitioner/widgets/pages/lvm_vg.rb
```diff
@@ -95,13 +119,23 @@ def label
           _("Log&ical Volumes")
         end
 
+        def initial
```
imobachgs: just curious: is it needed?

test/y2partitioner/dialogs/lvm_lv_info_test.rb
```diff
@@ -0,0 +1,179 @@
...
+          expect(widget.validate).to be(true)
+        end
+      end
+    end
+
+    describe "#store" do
+      let(:value) { "lv3" }
+
+      it "sets #lv_name in the controller" do
+        expect(controller.lv_name).to be_nil
```
imobachgs: You can use the `expect { }.to change` form ,as IMHO is more expressive:

```ruby
expect { widget.store }.to change { controller.lv_name }.from(nil).to(value)
```

There are more places where you could use the same approach, but I am not going to write a comment for every one (maybe you are not interesting in applying this change).

test/y2partitioner/dialogs/lvm_lv_info_test.rb
```diff
@@ -0,0 +1,179 @@
...
+          expect(controller.thin_pool).to be_nil
+          widget.store
+          expect(controller.thin_pool).to eq(used_pool)
+        end
+      end
+
+      context "when the selected type is not :thin" do
+        let(:value) { :normal }
+
+        it "does not set #thin_pool in the controller" do
```
imobachgs: What if `controller.thin_pool` already contained a value? Is also set to `nil`?

test/y2partitioner/dialogs/lvm_lv_size_test.rb
```diff
@@ -0,0 +1,222 @@
...
+    before do
+      allow(widget).to receive(:value).and_return(value)
+    end
+
+    let(:value) { nil }
+
+    include_examples "CWM::AbstractWidget"
+
+    describe "#validate" do
+      context "when indicated value in not a valid size" do
```
imobachgs: really-np: I would change `indicated` by `given` (it is a common term when writing tests).

test/y2partitioner/dialogs/lvm_lv_size_test.rb
```diff
@@ -0,0 +1,222 @@
...
+          expect(widget.validate).to eq(false)
+        end
+
+        it "shows an error popup" do
+          expect(Yast::Popup).to receive(:Error)
+          widget.validate
+        end
+      end
+
+      context "when indicated value in less than min size" do
```
imobachgs: typo: *in* -> *is* ?

test/y2partitioner/dialogs/lvm_lv_size_test.rb
```diff
@@ -0,0 +1,222 @@
...
+    describe "#store" do
+      before do
+        allow(Y2Partitioner::Dialogs::LvmLvSize::StripesNumberSelector)
+          .to receive(:new).and_return(stripes_number_widget)
+
+        allow(Y2Partitioner::Dialogs::LvmLvSize::StripesSizeSelector)
+          .to receive(:new).and_return(stripes_size_widget)
+      end
+
+      let(:stripes_number_widget) { double("StripesNumberSelector", value: stripes_number) }
```
imobachgs: I tend to use `instance_double` when possible because it checks if stubbed methods exist in the original class and I think it worth it.

test/y2partitioner/sequences/controllers/lvm_lv_test.rb
```diff
@@ -0,0 +1,285 @@
...
+      expect(lv.stripes).to eq(stripes_number)
+      expect(lv.stripe_size).to eq(stripes_size)
+    end
+
+    it "stores the new created lv" do
+      expect(controller.lv).to be_nil
+
+      controller.create_lv
+
+      expect(controller.lv).to_not be_nil
```
imobachgs: np: I think the test below makes this one unneeded, but it is true that error message could be slightly better.

test/y2partitioner/sequences/controllers/lvm_lv_test.rb
```diff
@@ -0,0 +1,285 @@
...
+      end
+
+      controller.create_lv
+    end
+  end
+
+  describe "#delete_lv" do
+    context "when a lv has not been created" do
+      it "does not modify the editing vg" do
+        lvs = controller.vg.lvm_lvs
```
imobachgs: You also have the `expect {}.to_not change {}` syntax :)

Anyway, if you prefer to avoid that syntax, I would use a more obvious name like `old_lvs` for that variable.

test/y2partitioner/widgets/add_lvm_button_test.rb
```diff
@@ -0,0 +1,146 @@
...
+
+    let(:event) { { "ID" => selected_option } }
+
+    context "when a vg is selected in the table" do
+      let(:selected_device) { vg }
+
+      context "and option for adding vg is selected" do
+        let(:selected_option) { :add_volume_group }
+
+        xit "opens the workflow for adding a new vg" do
```
imobachgs: Maybe you already know: if you do not specify the `do...end` block, it will work as `xit`. The advantage of `xit` is that it could be easier to find it in your editor.

src/lib/y2partitioner/dialogs/lvm_lv_info.rb
```diff
@@ -0,0 +1,188 @@
...
+
+        # @macro seeAbstractWidget
+        def store
+          @controller.type_choice = value
+          @controller.thin_pool = current_widget.value if value == :thin
+        end
+      end
+
+      # Selector for used thin pool
+      # TODO: thin provisioning should be implemented in libstorage-ng
```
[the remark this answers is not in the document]
joseivanlopez: Well, probably libstorage-ng is already ready to work with thin provisioning, so the comment should be only "TODO" because the implementation of thin pools is out of scope in this PBI.

src/lib/y2partitioner/widgets/controller_radio_buttons.rb
joseivanlopez: Well, to be honest I did not implement this part. I guess they are assigned because these methods do some work and it is not necessary to do it again and again. I will use other local variable names.

src/lib/y2partitioner/widgets/pages/lvm_vg.rb
joseivanlopez: Yes, this sets the tab as selected by default.

test/y2partitioner/dialogs/lvm_lv_info_test.rb
joseivanlopez: Nice, I did not know that, but at least in this case, I prefer the "classic" way. Yes, maybe I am too much familiar with this way.

test/y2partitioner/dialogs/lvm_lv_info_test.rb
joseivanlopez: I isn't, and that is wrong. Fixing.

test/y2partitioner/widgets/add_lvm_button_test.rb
joseivanlopez: I didn't know this trick either. As you mention, I will leave `xit` syntax to find it easier.

imobachgs:
If Travis is happy, I am :) Thanks!

PR #391, opened 2017-10-23
have volumes proposed by default

wfeldt:
... and let users change their settings.

It's just irritating having the volumes defined in control.xml and seeing
they are not created.

### DIFF
--- a/src/lib/y2storage/volume_specification.rb
+++ b/src/lib/y2storage/volume_specification.rb
@@ -155,6 +155,8 @@ module Y2Storage
     def apply_defaults
       @max_size ||= DiskSize.unlimited
       @max_size_lvm ||= DiskSize.zero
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/volume_specification.rb
```diff
@@ -155,6 +155,8 @@ def initialize(volume_features)
     def apply_defaults
       @max_size ||= DiskSize.unlimited
       @max_size_lvm ||= DiskSize.zero
+      @proposed ||= true
+      @proposed_configurable ||= true
```
joseivanlopez: I think we should update the documentation to indicate the default values for these settings.

wfeldt: Yes, not only for these.

PR #393, opened 2017-10-24
update control.xml doc

wfeldt:


doc/old_and_new_proposal.md
```diff
@@ -188,53 +188,59 @@ This chapter presents a brief summary of the new elements in the
...
+Most settings are grouped into two subsections of the `partitioning` section:
   * `proposal`  
     Holds general settings for the proposal.
   * `volumes`  
     A list of `volume` elements holding specific settings for each volume that should be created. Note
     that you really must add one section for each volume. There are no defaults.
 
+Besides these, there are two further elements:
+  * (**FIXME - not done**) `expert_partitioner_warning` *(boolean, default: `false`)*
+  * (**FIXME - not done**) `use_separate_multipath_module` *(boolean, default: `false`)*
```
joseivanlopez: NP: I would use (** TODO **) instead of that FIXME.

doc/old_and_new_proposal.md
```diff
@@ -662,6 +676,8 @@ following options.
     * `use_needed`  
       The created VG will match the requirements 1:1, so its size
       will be exactly the sum of all the LVs sizes.
+    * `use_vg_size`  
+      The VG will have exactly the size specified in `lvm_vg_size`.
```
joseivanlopez: Right now, only `use_needed` is implemented. Maybe we should add `TODO` to the other two options?

doc/old_and_new_proposal.md
```diff
@@ -662,6 +676,8 @@ following options.
...
   * `lvm_vg_size`  
     Specifies the predefined size of the LVM volume group.
```
joseivanlopez: Probably it would be good to clarify that this setting only applies when `lvm_vg_strategy == :use_vg_size`.

doc/old_and_new_proposal.md
joseivanlopez: Well, after a second reading, now I prefer the FIXME way :)

PR #394, opened 2017-10-24
Clone subvolumes properly

imobachgs:
Now subvolumes are cloned in a proper way.

### DIFF
--- a/src/lib/y2storage/autoinst_profile/partition_section.rb
+++ b/src/lib/y2storage/autoinst_profile/partition_section.rb
@@ -297,8 +297,10 @@ module Y2Storage
 
         @default_btrfs_subvolume_name = fs.default_subvolume_name
 
+        snapshots_root = File.join(@default_btrfs_subvolume_name.to_s, ".snapshots")
         valid_subvolumes = fs.btrfs_subvolumes.reject do |subvol|
-          subvol.path.empty? || subvol.path == @default_btrfs_subvolume_name
+          subvol.path.empty? || subvol.path == @default_btrfs_subvolume_name ||
+            subvol.path.start_with?(snapshots_root)
         end
 
         @subvolumes = valid_subvolumes.map do |subvol|
--- a/src/lib/y2storage/autoinst_profile/partition_section.rb
+++ b/src/lib/y2storage/autoinst_profile/partition_section.rb
@@ -70,7 +70,8 @@ module Y2Storage
           { name: :raid_name },
           { name: :raid_options },
           { name: :mkfs_options },
-          { name: :fstab_options, xml_name: :fstopt }
+          { name: :fstab_options, xml_name: :fstopt },
+          { name: :default_btrfs_subvolume_name }
         ]
       end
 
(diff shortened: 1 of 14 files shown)


src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -301,6 +301,18 @@ def restore_unshadowed_subvolumes(devicegraph)
...
       end
 
+      # Determines the default subvolume name
+      #
+      # When a default subvolume name have been used, a subvolume named after
+      # it lives under the #top_level_btrfs_subvolume. Otherwise, an empty
+      # string will be taken as the default subvolume name.
+      #
+      # @return [String] Default subvolume name
+      def default_subvolume_name
```
joseivanlopez: I am not sure if this is the best name for the method. We have a `Btrfs#default_btrfs_subvolume` method, and `#default_subvolume_name` looks like the name (path) of that subvolume, but it is not. Probably we need a new subvolume concept, something like `top_parent_subvolume`. Something to discuss in future, I guess.

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -288,6 +291,23 @@ def init_mount_options(fs)
         @fstab_options = fs.fstab_options unless fs.fstab_options.empty?
       end
 
+      # @param fs [Filesystem::BlkFilesystem]
```
joseivanlopez: It is obvious, but maybe to add a comment saying that it populates `@subvolumes` ¿?

test/y2storage/filesystems/btrfs_test.rb
```diff
@@ -639,4 +639,22 @@
...
+    end
+
+    context "when default subvolume name is not set" do
+      let(:dev_name) { "/dev/sdd1" }
+
+      it "returns an empty string" do
+        expect(subject.default_subvolume_name).to eq("")
+      end
+    end
+  end
```
joseivanlopez: I would add one more tests where the `default_btrfs_subvolume.path` and `default_subvolume_name` are different, to clarify that they are not related. For example, with snapshots, the `default_btrfs_subvolume.path` would be `@/.snapshots`, isn't it?

src/lib/y2storage/filesystems/btrfs.rb
imobachgs: Yes, the name can be tricky. For instance, this concept is named `btrfs_default_subvolume` in the control file, which is even worst: https://github.com/yast/yast-storage-ng/blob/master/test/data/control_files/caasp.xml#L31

test/y2storage/filesystems/btrfs_test.rb
imobachgs: It would be `@/.snapshots/1/snapshot` (or whatever snapshot it is the current one). And the `default_subvolume_name` could be `@`.

For instance, the second test checks that `default_subvolume_name` is not set and, although subvolumes are named like `@/blah`, the `default_subvolume_name` is still `""`.

src/lib/y2storage/filesystems/btrfs.rb
imobachgs: The problem when using `top` is that the `top_level` is a different thing and it could be confusing too.

imobachgs:
As discussed with @joseivanlopez on IRC, I will hold the PR until we discuss about `default_subvolume_name` name.

imobachgs:
After the discussion, I've changed that confusing name to `subvolumes_prefix`.

imobachgs:
Thanks!

PR #397, opened 2017-10-26
WIP: UI for configuring storage proposal parameters

shundhammer:
https://trello.com/c/xzXwFeek/703-3-allow-to-propose-a-data-partition-ui-changes

### DIFF
--- a/src/examples/proposal_settings_dialog.rb
+++ b/src/examples/proposal_settings_dialog.rb
@@ -30,7 +30,7 @@ require "y2storage/dialogs/guided_setup"
 require "pp"
 
 DATA_DIR = "../../test/data"
-FALLBACK_CONTROL_FILE = DATA_DIR + "/control_files/volumes_ng/control.SLE-like.xml"
+FALLBACK_CONTROL_FILE = DATA_DIR + "/control_files/volumes_ng/control.SLE-with-data.xml"
 FALLBACK_DEVICEGRAPH_FILE = DATA_DIR + "/devicegraphs/empty_hard_disk_gpt_50GiB.yml"
 
 control_file = Yast::WFM.Args[0] || FALLBACK_CONTROL_FILE
--- a/src/examples/proposal_settings_dialog.rb
+++ b/src/examples/proposal_settings_dialog.rb
@@ -1,0 +1,51 @@
+#!/usr/bin/env ruby
+# encoding: utf-8
+
+# Copyright (c) [2016] SUSE LLC
+#
+# All Rights Reserved.
+#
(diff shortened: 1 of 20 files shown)


ancorgs:
In general, I don't like the fact that this new dialog has hard-coded specific behaviors for root, home and swap that don't respect the specification.

For example, according to the specification I can use `proposed_configurable` in any volume and the effect should be that a checkbox is shown (or not) to enable/disable that partition. Even more, the default value for that attribute is true, meaning that all volumes can be deactivated by the user unless specified otherwise in the control file. But now that checkbox is never shown for root and is never shown for swap because the UI decided it makes not sense in those two cases (although it works as expected for other volumes). Something similar happens with snapshots, that are ignored for any volume that is not root, and for many other fields.

In my opinion, all that is not UI business and is highly unexpected that YaST offers different options for two `<volume>` that are identical in the control.xml in everything but in the mountpoint. The control.xml specification is designed to be very flexible and this UI is removing that flexibility by creating specific handling for specific volumes. As somebody touching control.xml I would be very surprised if the effect of my changes is different for different partitions, while the specification says that all volumes follow the same specification.

Maybe some cases look absurd, like the release manager specifying `proposed_configurable` to true for the root partition, but other situations that are limited by the current UI could eventually make sense. For example, what if the release manager wants to offer the possibility to disabling swap? Why should the UI ignore that?

So let's treat all volumes in a generic way (with specific handling only for the wording) and let the release manager shoot their own feet.

jsrain:
I agree that there is no need to limit the options that the control file format gives. Something that currently looks like behaving sane can in the future be an obstacle (and ignored control file parameter turn into unnecessary debug session). Taking as an example the possibility to disable swap creation - there are know cases where swap is not wanted; as a multi-purpose OS installer, we should allow the installer to propose system without swap (just one example).

I don't insist on full flexibility now, OTOH it does not make sense to limit it while it does not simplify the code itself.

And, one more note: The current implementation enforces order of the partition. Yes, root first makes sense but why should there be home before everything else?

jsrain:
I agree that there is no need to limit the options that the control file format gives. Something that currently looks like behaving sane can in the future be an obstacle (and ignored control file parameter turn into unnecessary debug session). Taking as an example the possibility to disable swap creation - there are know cases where swap is not wanted; as a multi-purpose OS installer, we should allow the installer to propose system without swap (just one example).

I don't insist on full flexibility now, OTOH it does not make sense to limit it while it does not simplify the code itself.

And, one more note: The current implementation enforces order of the partition. Yes, root first makes sense but why should there be home before everything else?

ancorgs:
Closed in favor of #412

PR #399, opened 2017-10-27
Limit maximal proposed size of EFI partition (bsc#1062775)

jsrain:


ancorgs:
It would be good to also update the tests and the autogenerated documentation to better reflect the new behavior. That is, in `test/support/proposed_partitions_examples.rb` change these lines...

```ruby
context "when aiming for the recommended size" do
  let(:target) { :desired }

  it "requires /boot/efi to be at least 500 MiB large" do
    expect(efi_part.min_size).to eq 500.MiB
  end
end

context "when aiming for the minimal size" do
  let(:target) { :min }

  it "requires /boot/efi to be at least 33 MiB large" do
    expect(efi_part.min_size).to eq 33.MiB
  end
end
```

...into these ones...

```ruby
context "when aiming for the recommended size" do
  let(:target) { :desired }

  it "requires /boot/efi to be exactly 500 MiB large" do
    expect(efi_part.min_size).to eq 500.MiB
    expect(efi_part.max_size).to eq 500.MiB
  end
end

context "when aiming for the minimal size" do
  let(:target) { :min }

  it "requires /boot/efi to be between 33 MiB and 500 MiB large" do
    expect(efi_part.min_size).to eq 33.MiB
    expect(efi_part.max_size).to eq 500.MiB
  end
end
```

I have not tested it, but it should work like that.

And, of course, run `rake doc:bootspecs` to refresh `doc/boot-requirements.md`.

Am I asking too much? ;-)

doc/boot-requirements.md
```diff
@@ -1,183 +1,2 @@
 
 [//]: # (document was automatically created using 'rake doc:bootspecs')
```
ancorgs: Erm... Why does this empties the documentation?

PR #400, opened 2017-10-27
Add context to Autoinst issues

imobachgs:
This PR adds some context to Autoinst issues adding the section where they were detected.

* `PartitioningSection`, `SectionsWithAttributes` (and derived classes) and `DriveSection` have a reference to its parent.
* Issues hold a reference to the section where the problem was found.
* Some minor refactoring regarding issues lists tests was performed.

With those changes in place, AutoYaST will benefit when showing errors information to users.

What is missing?

* Add the `parent` method to `SkipListSection`.
* Improve sections tests, including one (possible shared) to check the parent method.


src/lib/y2storage/autoinst_issues/invalid_value.rb
```diff
@@ -26,28 +26,32 @@ module AutoinstIssues
...
         @attr = attr
-        @value = value
         @new_value = new_value
       end
 
+      # Return the invalid value
+      #
+      # @return [Integer,String,Symbol] Invalid value
+      def value
+        section.send(attr)
```
jreidinger: well, this allow to also call private methods, which is not so nice from my POV. `public_send` is not enough?

src/lib/y2storage/autoinst_issues/invalid_value.rb
```diff
@@ -60,11 +64,10 @@ def severity
...
       def message
-        # TRANSLATORS: 'value' is a generic value (number or string)
-        # 'attr' is an AutoYaST element name; 'device' is a device name (eg. /dev/sda1);
-        # 'new_value_message' is a short explanation about what should be done with the value.
-        _("Invalid value '%{value}' for attribute '%{attr}' on '%{device}' (%{new_value_message}).") %
-          { value: value, attr: attr, device: device, new_value_message: new_value_message }
+        # TRANSLATORS: 'value' is a generic value (number or string) 'attr' is an AutoYaST element
+        # name; 'new_value_message' is a short explanation about what should be done with the value.
+        _("Invalid value '%{value}' for attribute '%{attr}' (%{new_value_message}).") %
+          { value: value, attr: attr, new_value_message: new_value_message }
```
jreidinger: rubocop is fine with this? I think it usually suggest to use format, but maybe it is here more benevolent.

src/lib/y2storage/autoinst_issues/issue.rb
```diff
@@ -7,6 +7,9 @@ module AutoinstIssues
     class Issue
       include Yast::I18n
 
+      # @return [Object] section where the problem was detected (see {AutoinstProfile})
```
jreidinger: well, to be honest documenting only Object as parent is not much helpful. is there any methods on top of pure Object that it have in common? I think for documentation purpose it would be helpful to have class that define common methods for all sections.

jreidinger: currently it looks like at least parent and section_name is must have, not?

src/lib/y2storage/autoinst_profile/section_with_attributes.rb
```diff
@@ -95,6 +108,23 @@ def to_hashes
...
+      # @example
+      #   section = PartitioningSection.new
+      #   section.section_name #=> "partitioning"
+      #
+      # @return [String] Section name
+      def section_name
+        klass_name = self.class.name.split("::").last
+        klass_name
+          .gsub(/([a-z])([A-Z])/, "\\1_\\2").downcase
+          .gsub(/_section\z/, "")
```
jreidinger: gsub here is not needed, it can be simple sub or even better `.chomp("_section")`

test/support/autoinst_profile_examples.rb
```diff
@@ -0,0 +1,35 @@
+# encoding: utf-8
```
jreidinger: why file is called profile_examples, when it is section-examples?

test/support/autoinst_profile_examples.rb
```diff
@@ -0,0 +1,35 @@
...
+# find current contact information at www.suse.com.
+
+require "y2storage/autoinst_profile/section_with_attributes"
+
+RSpec.shared_examples "autoinst section" do
+  let(:parent) do
+    Y2Storage::AutoinstProfile::SectionWithAttributes.new
+  end
+
+  describe ".new_from_hashes" do
```
jreidinger: so common method for all sections?

jreidinger: in general I see it as problem when you want to create new section.

test/support/autoinst_profile_examples.rb
imobachgs: Because the module is called AutoinstProfile.

test/support/autoinst_profile_examples.rb
imobachgs: What I want to test is that the `parent` is set always when using `new_from_hashes`. About the method itself, we have a `new_from_hashes` and `new_from_storage` to build sections from the profile or from the storage layer.

What is exactly the problem?

src/lib/y2storage/autoinst_issues/invalid_value.rb
imobachgs: Yes, it is fine.

src/lib/y2storage/autoinst_issues/issue.rb
imobachgs: Well, that's partially true: these classes will only hold a reference to those objects (to be used by others later). But it's true that maybe something like `#parent,#section_name` would be better than `Object`.

test/support/autoinst_profile_examples.rb
jreidinger: so it is common behavior for everything in given namespace, right?

imobachgs: BTW, I have renamed to a hopefully a better name :)

test/support/autoinst_profile_examples.rb
jreidinger: my problem is documentation to see what is important and I have to do when I want to create new section.

imobachgs:
Thanks!

PR #402, opened 2017-10-30
Partitioner delete lv

joseivanlopez:
PBI: https://trello.com/c/k8ntmZMZ/735-2-expert-partitioner-delete-a-lv


src/lib/y2partitioner/actions.rb
```diff
@@ -21,22 +19,10 @@
...
-          formattable?
-        end
-      end
-    end
+  # Namespace to group all the actions of the expert partitioner
+  module Actions
   end
 end
+
+require "y2partitioner/actions/delete_blk_device"
```
ancorgs: Looks like the comment I added here before disappeared 🤔  I will repeat it:

This `y2partitioner/actions.rb` seems to be there to make possible to require all the actions at once, but it only requires one of the existing actions. It's a little bit confusing to me. I would either kill the file or make sure it actually requires all the actions. Having a file there that just requires another random file looks weird.

BTW, it's fun how git detects this as a renaming of the old refinement. :-)

joseivanlopez: I forgot to include the rest of actions. I like having a separated file for the module definition.

ancorgs:
I have another comment about a piece of code that is not shown in the diff (in general, the interpretation of the changes git did for this PR is interesting to say the least), but is very relevant here:

```ruby
confirm_recursive_delete(
  names,
  _("Confirm Deleting of All Partitions"),
  # TRANSLATORS: type stands for type of device and name is its identifier
  format(_("The %{type} \"%{name}\" contains at least one another device.\n" \
    "If you proceed, the following devices will be deleted:"),
    name: device.name,
    type: device.is?(:disk) ? _("disk") : _("partition")),
  format(_("Really delete all devices on \"%s\"?"), device.name)
)
```

The "All Partitions" seems to refer to the very particular case in which the action is called on a disk. But I believe it will also be displayed when deleting a LV or a particular partition.

On the other hand, strings like `"The %{type} \"%{name}\" contains at least bla"` can be very tricky for the translations. Just take Spanish as an example. If type is partition, the correct translation would be "**La** partición /dev/sda1 contiene..." but if it's a LV it would be "**El** volume /dev/system/foo contiene...". Having a placeholder makes sense often, but strings that are generated programatically are a PITA for the translators.

And last but not least, `%{type}' is partition for all devices that are not a disk... which is not precise.

src/lib/y2partitioner/actions/delete_disk.rb
```diff
@@ -0,0 +1,73 @@
...
+      #
+      # @see DeleteDevice#dependent_devices
+      # @see DeleteDevice#confirm_recursive_delete
+      def confirm
+        confirm_recursive_delete(
+          dependent_devices,
+          _("Confirm Deleting of All Partitions"),
+          # TRANSLATORS: name is the name of the disk to be deleted (e.g., /dev/sda)
+          format(_("The disk \"%{name}\" contains at least one partition.\n" \
+            "If you proceed, the following partitions will be deleted:"), name: device.name),
```
ancorgs: NP: it says "the following partitions", but the list can show more than partitions. It may also include VGs or RAIDs based on those partitions. Isn't it?

joseivanlopez: Yes, only mimic the old partitioner.

src/lib/y2partitioner/actions/delete_partition.rb
```diff
@@ -29,13 +29,94 @@ module Actions
...
         disk_device.partition_table.delete_partition(device)
         UIState.instance.select_row(disk_device)
       end
+
+    private
+
+      # Checks whether the partition is used as physical volume
+      #
+      # @return [Boolean] true if partition belongs to a volume group; false otherwise
+      def used_by_lvm?
```
ancorgs: There is a Proposal::PartitionKiller#lvm_vg? that is too similar to this. Time to have a method `Partition#lvm_vg` (or `BlkDevice#lvm_vg`)?

ancorgs: Or maybe a more explicit name to make obvious that the VG is a descendant of the block device and not the other way around.

src/lib/y2partitioner/actions/delete_partition.rb
```diff
@@ -29,13 +29,94 @@ module Actions
...
+      #
+      # @return [Boolean] true if partition belongs to a md raid; false otherwise
+      def used_by_md_raid?
+        !md_raid.nil?
+      end
+
+      # Md Raid that the partition belongs to
+      #
+      # @return [Y2Storage::Md, nil] nil if the partition does not belong to a md raid
+      def md_raid
```
ancorgs: Relatively similar to the former, we maybe should move this to `BlkDevice` with a good name.

PR #403, opened 2017-10-30
 Add button to run partitioner from probed 

ancorgs:
This implements https://trello.com/c/LvwStMCI/704-1-partitioner-start-from-scratch

Easier to review in a per-commit fashion.

Manually tested in the inst-sys, see screenshots.

partitioner_from_scratch_qt

partitioner_from_scratch_ncurses

### DIFF
--- a/src/lib/y2storage/dialogs/proposal.rb
+++ b/src/lib/y2storage/dialogs/proposal.rb
@@ -124,10 +124,10 @@ module Y2Storage
       def expert_partitioner
         items = []
         if devicegraph
-          items << Item(Id(:expert_from_proposal), _("Start with Current Proposal"))
+          items << Item(Id(:expert_from_proposal), _("Start with &Current Proposal"))
         end
-        items << Item(Id(:expert_from_probed), _("Start with Existing Partitions"))
-        MenuButton(_("Expert Partitioner"), items)
+        items << Item(Id(:expert_from_probed), _("Start with Existing &Partitions"))
+        MenuButton(_("&Expert Partitioner"), items)
       end
 
       def dialog_content
@@ -135,7 +135,7 @@ module Y2Storage
           2, 1,
           VBox(
             MinHeight(8, summary),
-            PushButton(Id(:guided), _("Guided Setup")),
+            PushButton(Id(:guided), _("&Guided Setup")),
             expert_partitioner
           )
         )
--- a/src/lib/y2storage/dialogs/proposal.rb
+++ b/src/lib/y2storage/dialogs/proposal.rb
@@ -60,8 +60,12 @@ module Y2Storage
         finish_dialog(:guided)
       end
 
-      def expert_handler
-        finish_dialog(:expert)
+      def expert_from_proposal_handler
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/dialogs/proposal.rb
```diff
@@ -117,13 +121,22 @@ def dialog_title
         _("Suggested Partitioning")
       end
 
+      def expert_partitioner
+        items = []
+        if devicegraph
+          items << Item(Id(:expert_from_proposal), _("Start with Current Proposal"))
+        end
+        items << Item(Id(:expert_from_probed), _("Start with Existing Partitions"))
+        MenuButton(_("Expert Partitioner"), items)
```
jreidinger: btw when you touch it, you can try to make openQA life easier with suggested keyboard shortcuts with "&" in name. Also I do not see it, but if final state is 2-3 options, maybe 3 push buttons instead of small menu is better? but I expect you already discuss it.
Otherwise it looks good.

ancorgs: Yes, creating a menu-button was actually requested in the PBI. Looks like it's the UI suggested by the UI team.

ancorgs:
Hmm, the coverage report looks bogus to me. The `Y2Storage::Dialogs::Proposal#expert_from_probed_handler` and `Y2Storage::Dialogs::Proposal#expert_from_proposal_handler` methods are reported as not covered. I'm pretty sure they are (I added a commit just to cover them).

ancorgs:
And indeed, those two methods are reported as covered if I run simplecov locally. Very strange.

jreidinger:
@ancorgs maybe some mocking that skips its execution?

ancorgs:
But if it's due to some mocking. Why are the lines reported as covered when I run the tests locally and why tests stop there if I insert a byebug call?

I will open a parallel PR to do some experiments.

ancorgs:
As probed by this debugging PR, which includes this very same code plus a couple of `puts` (https://github.com/yast/yast-storage-ng/pull/404), the code is being executed. See the Travis output (which includes the messages) and the coverage increase.

But I don't know whether is Travis, Docker or Coveralls fault.

src/lib/y2storage/dialogs/proposal.rb
ancorgs: And whether the shortcuts should be explicitly added by us, that's relatively controversial. I understand it makes the whole thing more stable for openQA to test an English system, but I have seen some discussions about how good/bad is to enforce them at developer level instead of letting the automatic mechanisms in LibUI do their job.

jreidinger: well, in past it was common practice to have UI shortcuts stable. I also personally find very annoying to have different shortcuts depending on options like it is for users ( when old system is detected, then shortcuts are different ). Automatic mechanism just ensure that there is sortcut, but there is no guarantie that it will be stable. Which we can do on other hand with giving prefered keyword there. ( as it have precedence over automatic ones, so at first it tries to assign all suggested and then find remaining ones, that is not already set ).

PR #406, opened 2017-10-31
Fix devices reuse

imobachgs:
Fix several issues when trying to reuse devices in AutoYaST and improves error handling.


package/yast2-storage-ng.changes
```diff
@@ -1,4 +1,16 @@
 -------------------------------------------------------------------
+Tue Oct 31 12:14:18 UTC 2017 - igonzalezsosa@suse.com
+
+- AutoYaST: fix reuse of partitions (bsc#1060637).
+- AutoYaST: improve error handling when trying to reuse a partition
+  fails.
+- AutoYaST: when a problem is detected while creating a proposal,
+  register in which section it was found.
+- AutoYaST: register an issue when a proposel does not contain
```
joseivanlopez: Typo: proposel --> proposal

test/y2storage/proposal/autoinst_devices_planner_test.rb
```diff
@@ -98,6 +98,42 @@
...
+
+        it "adds a new partition" do
+          devices = planner.planned_devices(drives_map)
+          root = devices.find { |d| d.mount_point == "/" }
+          expect(root.reuse).to be_nil
+        end
+
+        it "registers an issue" do
+          planner.planned_devices(drives_map)
+          issue = issues_list.find { |i| i.is_a?(Y2Storage::AutoinstIssues::MissingReusableDevice) }
```
joseivanlopez: I would check that the issues list is empty or does not contain that issue before planning the devices.

test/y2storage/proposal/autoinst_devices_planner_test.rb
```diff
@@ -98,6 +98,42 @@
...
+
+        it "adds a new partition" do
+          devices = planner.planned_devices(drives_map)
+          root = devices.find { |d| d.mount_point == "/" }
+          expect(root.reuse).to be_nil
+        end
+
+        it "registers an issue" do
+          planner.planned_devices(drives_map)
+          issue = issues_list.find { |i| i.is_a?(Y2Storage::AutoinstIssues::MissingReuseInfo) }
```
joseivanlopez: Same comment than before. In general, when finding some issue, first I would check that the issue does not exist.

imobachgs: Issues are supposed to be added when running the planner (`planned_services`), so I think it is safe to assume that there are no issues at the beginning of the test. But I've added them to be on the safe side :)

PR #408, opened 2017-11-02
Get hardware information for skip lists

imobachgs:
This PR adds support to get hardware information through `hwinfo`.

* Adds a `Y2Storage::BlkDevice#hwinfo` method to get such information. That information is cached (but refreshed when probing a second time).
* Extends skip lists to query that information, partially fixing [bsc#1065668](https://bugzilla.suse.com/show_bug.cgi?id=1065668). The following keys are now supported: unique_id, parent_id, sysfs_id, sysfs_busid, sysfs_device_link, hardware_class, model, vendor, device, revision, serial_id, *driver*, *driver_modules,* device_file, device_files, device_number, geometry_logical, size, capacity, config_status, attached_to, bus. Some of those values might need some refinement, but it is a nice starting point.
* Support for multi-valued keys (for instance, `driver` or `driver_modules`).
* Do not crash if no suitable disk for installation was found.


src/lib/y2storage/devicegraph.rb
```diff
@@ -27,6 +27,7 @@
 require "y2storage/filesystems/base"
 require "y2storage/filesystems/blk_filesystem"
 require "y2storage/filesystems/nfs"
+require "y2storage/hwinfo_reader"
```
joseivanlopez: Why is it required here?

src/lib/y2storage/hwinfo_reader.rb
```diff
@@ -0,0 +1,120 @@
...
+    # @param name [String] Device name (eg. "/dev/sda")
+    # @return [OpenStruct] Hardware information
+    def for_device(name)
+      data[name]
+    end
+
+    # Reset the cache
+    #
+    # The values will be loaded when #for_device is called again.
+    def reset!
```
joseivanlopez: Lately, we are only using bang methods when we have two versions, with and without exclamation mark.

src/lib/y2storage/hwinfo_reader.rb
```diff
@@ -0,0 +1,120 @@
...
+      @data = nil
+    end
+
+  private
+
+    def data
+      return @data if @data
```
joseivanlopez: NP: Do you not prefer `@data ||= data_from_hwinfo` ?

test/y2storage/hwinfo_reader_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    it "caches hardware information" do
+      expect(Yast::Execute).to receive(:on_target!)
+        .with("/usr/sbin/hwinfo", "--disk", "--listmd", stdout: :capture)
+        .and_return(hwinfo_output)
+        .once
+      reader.for_device("/dev/sda")
+      reader.for_device("/dev/sda")
+    end
```
joseivanlopez: I know that it is implicit tested, by maybe it would be nice to have an explicit test for `#reset!`.

test/y2storage/autoinst_issues/no_disk.rb
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env rspec
```
joseivanlopez: The name of the test should end with **_test.rb**.

joseivanlopez:
Now I have finished. Nothing important, only the name of the test file.

PR #411, opened 2017-11-03
Fix use

imobachgs:
This fixes an important bug: [bsc#1066398](https://bugzilla.suse.com/show_bug.cgi?id=1066398).

Basically, AutoinstSpaceMaker kicks in when partitions have been planned and does not remove any partition that could be reused.

The tricky part is mapping the `sid` with the `device name` in order to reassign `reuse` properties after making space (as, for instance, removing a logical partition might cause other partitions to be renamed).

### DIFF
--- a/src/lib/y2storage/proposal/autoinst_space_maker.rb
+++ b/src/lib/y2storage/proposal/autoinst_space_maker.rb
@@ -187,7 +187,7 @@ module Y2Storage
             all << devicegraph.partitions.find { |p| device.reuse == p.name }
           when Y2Storage::Planned::LvmVg
             vg = devicegraph.lvm_vgs.find { |v| File.join("/dev", device.reuse) == v.name }
-            all.concat(vg.lvm_pvs.map(&:blk_device))
+            all.concat(vg.lvm_pvs)
           when Y2Storage::Planned::Md
             all << devicegraph.md_raids.find { |r| device.reuse == r.name }
           end
--- a/src/lib/y2storage/proposal/autoinst_space_maker.rb
+++ b/src/lib/y2storage/proposal/autoinst_space_maker.rb
@@ -22,6 +22,8 @@
 # find current contact information at www.suse.com.
 
 require "y2storage/disk"
+require "y2storage/planned/lvm_vg"
+require "y2storage/planned/lvm_lv"
 
 module Y2Storage
   module Proposal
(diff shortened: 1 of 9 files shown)


src/lib/y2storage/proposal/autoinst_space_maker.rb
```diff
@@ -131,6 +158,76 @@ def register_invalid_use_value(drive_spec)
...
+      # @param planned_devices [Array<Planned::Partition>] set of partitions
+      # @return [Hash<String,Array<String>>] disk name to list of reused partitions map
+      def find_reused_partitions(devicegraph, planned_devices)
+        reused_devices = planned_devices.select(&:reuse).each_with_object([]) do |device, all|
+          case device
+          when Y2Storage::Planned::Partition
+            all << devicegraph.partitions.find { |p| device.reuse == p.name }
+          when Y2Storage::Planned::LvmVg
+            vg = devicegraph.lvm_vgs.find { |v| File.join("/dev", device.reuse) == v.name }
+            all.concat(vg.lvm_pvs.map(&:blk_device))
```
joseivanlopez: Method `LvmPv#blk_device` could return the encryption device instead of the blk device. There is no problem if the device hosting the PV is a partition due to you are searching then by partitions in the ancestors list. But there is a possible corner case: a PV over an encrypted disk. Is it necessary to take into account that scenario? Probably not because you are only working with partitions here, isn't it?

test/y2storage/proposal/autoinst_space_maker_test.rb
```diff
@@ -111,23 +219,39 @@
...
       end
+
+      context "and some device will be used" do
+        let(:scenario) { "lvm-two-vgs" }
+        let(:planned_devices) { [planned_partition] }
+        let(:planned_partition) do
+          Y2Storage::Planned::Partition.new("/").tap { |p| p.reuse = "/dev/sda1" }
+        end
+
+        it "does not removes the partition table" do
```
joseivanlopez: Typo: removes --> remove

src/lib/y2storage/proposal/autoinst_space_maker.rb
imobachgs: Thanks for the info. Yes, I am working with partitions. If I can reach the partition through the "#ancestors" method, it will be ok. If not, I guess I should handle it.

imobachgs:
Thanks!

PR #414, opened 2017-11-06
Fix new settings

joseivanlopez:
Part of PBI https://trello.com/c/I9K56rmD/566-5-allow-to-propose-a-data-partition-changes-in-the-proposal-logic

### DIFF
--- a/doc/old_and_new_proposal.md
+++ b/doc/old_and_new_proposal.md
@@ -54,12 +54,12 @@ Btrfs for the root filesystem, that are used by the proposal and also by the
 expert partitioner when suggesting the default configuration for such
 filesystem.
 
-  * `subvolumes`  
+  * `subvolumes`
     Optional list of Btrfs subvolumes. If it is missing, a
     hard-coded list is used. If the section is there but empty, no subvolumes
     are created. Each subvolume section has a mandatory `path` and optional
     `copy_on_write` and `archs` elements.
-  * `btrfs_default_subvolume`  
+  * `btrfs_default_subvolume`
     The default subvolume is not represented by an
     entry in the `subvolumes` list, but with a separate option that specifies only
     its path. The path of the default subvolume is prepended to the path of all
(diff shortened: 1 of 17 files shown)


test/y2storage/planned/lvm_lv_test.rb
```diff
@@ -27,12 +27,42 @@
...
+        it "sets lv name based on mount point" do
+          expect(subject.logical_volume_name).to eq("var_lib_docker")
+        end
+      end
+    end
+
+    context "when creating a planned lv without mount point" do
+      let(:mount_point) { nil }
+
+      it "does not set lv name" do
```
ancorgs: This makes me think. Proposing volumes without a mount point (to be used raw by some database or something like that) is a real use case that is coming just around the corner. With our current implementation it will work if only one of such volumes is specified (it will receive the "lv" name thanks to the `LvmCreator` fallback). A more robust solution is needed (either introduce support for the LV name in control.xml or improve the `LvmCreator` mechanism). We need a PBI to track this.

@jsrain Do you think it makes sense to introduce a `lv_name` attribute in control.xml (I'm all for it)? The alternative is to use things like "var_lib_docker" for LVs with a mount point (already implemented) and "lv0", "lv1", etc. for volumes without mount point (right now, only "lv" without the number, so the second unmounted volume would explode).

ancorgs: Out of the scope of this PR, anyways :-)

test/y2storage/proposal/autoinst_size_parser_test.rb
```diff
@@ -69,7 +69,11 @@
           expect(size_info.min).to eq(128.MiB)
           expect(size_info.max).to eq(1.GiB)
           expect(size_info.percentage).to be_nil
-          expect(size_info.weight).to be_nil
+        end
+
+        it "sets weight according to control file value" do
```
ancorgs: So we stopped testing the case with `"weight" => nil` in order to start testing `weight != nil`? Why are we replacing the test instead of adding one?

test/y2storage/planned/lvm_lv_test.rb
ancorgs: Hmm, I just remembered we have a mechanism to avoid duplicated LV names, even if we didn't add the conflicting LV. So having several non-mounted LVs will not fail with the current implementation. The non-mounted LVs will be successfully created with names like "lv", "lv0", "lv1"...

So this is not as bad/urgent as I originally feared. But still one question stands - does it make sense to add a `lv_name` tag to the volume specification in control.xml? With that, release managers could get nicer names than `var_lib_docker` or `lv`.

doc/old_and_new_proposal.md
```diff
@@ -482,8 +482,8 @@ enable or disable in the "Guided Setup" (the former "Proposal Settings").
     <!-- The '/' filesystem -->
     <volume>
       <mount_point>/</mount_point>
-      <!-- Enforce Btrfs for root by not offering any other option -->
-      <fs_types>btrfs</fs_types>
+      <!-- Default == final, since the user can't change it -->
+      <fs_type>btrfs</fs_type>
```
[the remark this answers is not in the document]
joseivanlopez: The second one will also fallback.

PR #416, opened 2017-11-07
Add some missing skip list keys

imobachgs:
Here is the second part of #408, adding some missing keys that were available in SLE12. Check https://github.com/yast/yast-autoinstallation/pull/364 to see them in action.

### DIFF
--- a/src/lib/y2storage/autoinst_profile/skip_list_value.rb
+++ b/src/lib/y2storage/autoinst_profile/skip_list_value.rb
@@ -19,8 +19,47 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "y2storage/dasd"
+
 module Y2Storage
   module AutoinstProfile
+    # @return [Array<Symbol>] hwinfo keys that can be used as a skip list key
+    HWINFO_KEYS = [
+      :bios_id,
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/autoinst_profile/skip_list_value.rb
```diff
@@ -68,11 +107,73 @@ def name
         disk.basename
       end
 
+      # @return [String,nil] DASD format or nil if not a DASD device
+      #   backward compatibility
+      def dasd_format
+        return nil unless disk.is_a?(Y2Storage::Dasd)
```
ancorgs: It's better to use `unless disk.is?(:dasd)`. It's more duck-typing oriented and doesn't depend on the concrete hierarchy enforced by libstorage-ng at the moment.

src/lib/y2storage/autoinst_profile/skip_list_value.rb
```diff
@@ -68,11 +107,73 @@ def name
...
+        disk.format.to_s
+      end
+
+      # @return [String,nil] DASD type or nil if not a DASD device
+      def dasd_type
+        return nil unless disk.is_a?(Y2Storage::Dasd)
```
ancorgs: Same than above.

PR #417, opened 2017-11-08
Fix disklabel

imobachgs:
This PR fixes partition table creation in AutoYaST (now is set to the right type). But, even more important, it also fixes space distribution in the following situation:

* no partition table exists (or profile sets `<initialize>` element to `true`)
* any partition size is set to `max` (or not set at all)

In such a scenario, the storage proposal runs out of space. There is a possible workaround: setting all partition sizes, but the user would need to take the partitions table size into account.

For the records, the problem was that the partition table was created too late (not explicitly created at all), so calculations about space distribution were just wrong.

### DIFF
--- a/test/y2storage/autoinst_proposal_test.rb
+++ b/test/y2storage/autoinst_proposal_test.rb
@@ -643,7 +643,6 @@ describe Y2Storage::AutoinstProposal do
         allow(Y2Storage::BootRequirementsChecker).to receive(:new).and_return(boot_checker)
       end
 
-
       let(:partitioning) do
         [{ "use" => "all", "partitions" => [swap, root], "disklabel" => "gpt" }]
       end
@@ -662,7 +661,6 @@ describe Y2Storage::AutoinstProposal do
         it "creates the boot partition" do
           proposal.propose
           devicegraph = proposal.devices
-          disk = devicegraph.disk_devices.first
           boot = devicegraph.partitions.find { |p| p.id == Y2Storage::PartitionId::BIOS_BOOT }
           expect(boot).to_not be_nil
         end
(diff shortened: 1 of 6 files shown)


test/y2storage/autoinst_proposal_test.rb
```diff
@@ -629,6 +629,101 @@
...
+
+      context "when partition table exist" do
+        let(:scenario) { "windows-linux-free-pc" }
+
+        context "and a different type is requested" do
+          let(:partitioning) do
+            [
+              {
+                "use" => "all", "partitions" => [swap, root], "disklabel" => "gpt",
+                "initialize" => true
```
ancorgs: Shouldn't we also have a test with `initialize == false`? In that case the existing partition table would be respected and `disklabel` would be ignored. Isn't it?

It's the reasonable behavior, but maybe is worth an explicit RSpec test.

test/y2storage/autoinst_proposal_test.rb
```diff
@@ -629,6 +629,101 @@
...
+      it "creates the boot partition" do
+        proposal.propose
+        devicegraph = proposal.devices
+        disk = devicegraph.disk_devices.first
+        expect(disk.name).to eq("/dev/sda")
+        expect(disk.partition_table.type).to eq(Y2Storage::PartitionTables::Type::GPT)
+      end
+    end
+
+    describe "partition table" do
```
ancorgs: And we maybe should have some kind of test to ensure the partition table and the other partitions are already there when calling BootRequirementsChecker. That was part of the problem, anyway, that BootRequirementsChecker was called before creating the new partition tables.

imobachgs:
Thanks for the review. Actually, two of those tests were just wrong. I've addded the suggested tests and fixed the current ones.

PR #418, opened 2017-11-08
handle partition ids for different partition types (bsc #1060993)

wfeldt:
- step 1: don't break if we messed up the ids - instead, log an error and continue with a reasonable default

- step 2: map partition ids to allowed values


ancorgs:
Approved, but I miss a `test/y2storage/partition_tables/gpt_test.rb` now that GPT has its own partition_id mapping. :-) Let's see what coveralls thinks about it

PR #419, opened 2017-11-08
Filter out '@' subvolumes when importing

imobachgs:
It should fix https://bugzilla.suse.com/show_bug.cgi?id=1061253 and supersedes https://github.com/yast/yast-autoinstallation/pull/365.


imobachgs:
Thanks! I will wait for Travis before merging it.

PR #420, opened 2017-11-08
Partitioner add vg

joseivanlopez:
PBI: https://trello.com/c/CGjosZ0p/734-3-expert-partitioner-create-a-vg

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/lvm_vg.rb
+++ b/src/lib/y2partitioner/actions/controllers/lvm_vg.rb
@@ -255,7 +255,7 @@ module Y2Partitioner
           !(extent_size &&
             extent_size > Y2Storage::DiskSize.KiB(1) &&
             extent_size % Y2Storage::DiskSize.KiB(128) == Y2Storage::DiskSize.zero &&
-            extent_size.power_of_two?)
+            extent_size.power_of?(2))
         end
 
         # Error message to show when the given extent size is not valid
--- a/src/lib/y2partitioner/actions/controllers/lvm_vg.rb
+++ b/src/lib/y2partitioner/actions/controllers/lvm_vg.rb
@@ -82,43 +82,28 @@ module Y2Partitioner
           vg.extent_size = extent_size
         end
 
-        # Checks whether the given volume group name is empty
+        # Error messages for the volume group name
         #
-        # @return [Boolean] true if volume group name is "" or nil; false otherwise
-        def empty_vg_name?
-          vg_name.nil? || vg_name.empty?
+        # @note When there is no error, an empty list is returned.
+        #
+        # @return [Array<String>] list of errors
+        def vg_name_errors
+          errors = []
+          errors << empty_vg_name_message if empty_vg_name?
+          errors << illegal_vg_name_message if vg_name && illegal_vg_name?
+          errors << duplicated_vg_name_message if vg_name && duplicated_vg_name?
+          errors
         end
 
-        # Checks whether the givem volume group name has illegal characters
+        # Error messages for the extent size
         #
-        # @see ALLOWED_NAME_CHARS
+        # @note When there is no error, an empty list is returned.
         #
-        # @return [Boolean] true if volume group name has illegal characters;
-        #   false otherwise
-        def illegal_vg_name?
-          vg_name.split(//).any? { |c| !ALLOWED_NAME_CHARS.include?(c) }
-        end
-
-        # Checks whether it exists another device with the given volume group name
-        #
-        # @return [Boolean] true if there is another device with the same name;
-        #   false otherwise
-        def duplicated_vg_name?
-          name = File.join("/dev", vg_name)
-          Y2Storage::BlkDevice.all(working_graph).map(&:name).include?(name)
-        end
-
-        # Checks whether the given extent size is not valid
-        #
-        # @note Extent size is valid if it is bigger than 1 KiB, multiple of 128 KiB
-        #   and power of 2.
-        #
-        # @return [Boolean] true if the extent size is not valid; false otherwise
-        def invalid_extent_size?
-          !(extent_size &&
-            extent_size > Y2Storage::DiskSize.KiB(1) &&
-            extent_size % Y2Storage::DiskSize.KiB(128) == Y2Storage::DiskSize.zero &&
-            extent_size.power_of_two?)
+        # @return [Array<String>] list of errors
+        def extent_size_errors
+          errors = []
+          errors << invalid_extent_size_message if invalid_extent_size?
+          errors
         end
 
         # Devices that can be selected to become physical volume of a volume group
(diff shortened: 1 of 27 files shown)


src/lib/y2storage/disk_size.rb
```diff
@@ -549,6 +549,25 @@ def zero?
...
+    #
+    # @example
+    #   x = DiskSize.KiB(4) #=> <DiskSize 4.00 KiB (4096)>
+    #   x.power_of_two?     #=> true
+    #
+    #   x = DiskSize.KB(4) #=> <DiskSize 3.91 KiB (4000)>
+    #   x.power_of_two?    #=> false
+    def power_of_two?
+      return false if zero? || unlimited?
+      size.to_s(2).count("1") == 1
```
ancorgs: I still haven't decided whether this is very neat or just hacky :wink: (the line between those two is always thin).

Why not something more generic and less surprising like this?
```ruby
def power_of?(exp)
  return false if unlimited?
  (Math.log(size, exp) % 1).zero?
end
```

src/lib/y2partitioner/widgets/md_add_button.rb
```diff
@@ -0,0 +1,49 @@
...
+# find current contact information at www.suse.com.
+
+require "yast"
+require "cwm"
+require "y2partitioner/actions/add_md"
+
+module Y2Partitioner
+  module Widgets
+    # Button for openng a wizard to add a new MD array
+    class MdAddButton < CWM::PushButton
```
ancorgs: Why wasn't this turned into a `DeviceButton`  like `LvmLvAddButton` and `PartitionAddButton`?

src/lib/y2partitioner/dialogs/lvm_vg.rb
```diff
@@ -0,0 +1,267 @@
...
+        end
+
+        # @macro seeAbstractWidget
+        # Checks whether the given value is valid
+        #
+        # @note An error popup is shown when necessary.
+        #
+        # @return [Boolean]
+        def validate
+          focus
```
ancorgs: Why?

src/lib/y2partitioner/dialogs/lvm_vg.rb
```diff
@@ -0,0 +1,267 @@
...
+          Yast::UI.SetFocus(Id(widget_id))
+        end
+
+        # Checks whether some value was entered
+        #
+        # @return [Boolean]
+        def presence_validation
+          return true unless controller.empty_vg_name?
+
+          Yast::Popup.Error(
```
ancorgs: NP: In general, having the detection of errors in the controller and the strings explaining them in the widgets feels wrong. We are forced to keep them in sync. We can live like that for the time being (deadlines are around the corner), but maybe some FIXMEs wouldn't hurt.

src/lib/y2partitioner/actions/controllers/lvm_vg.rb
```diff
@@ -0,0 +1,298 @@
...
+        # @return [Y2Storage::DiskSize]
+        def vg_size
+          vg.size
+        end
+
+        # Applies given values (i.e., volume group name and extent size) to the
+        # volume group
+        def apply_values
+          vg.vg_name = vg_name
+          vg.extent_size = extent_size
```
ancorgs: Assigning nil here to `extend_size` is correctly handled? Doesn't the library go nuts? Even more important, is that situation unit-tested?

test/y2partitioner/dialogs/lvm_vg_test.rb
```diff
@@ -0,0 +1,542 @@
...
+      end
+
+      it "sets its value with the extent size stored in the controller" do
+        expect(subject).to receive(:value=).with("128 KiB")
+        subject.init
+      end
+    end
+
+    describe "#handle" do
+      before do
```
ancorgs: I would like to see a unit test for empty string and/or an invalid string here. Just to ensure it doesn't break today or in the future.

src/lib/y2partitioner/dialogs/lvm_vg.rb
```diff
@@ -0,0 +1,267 @@
...
+        end
+
+        # Checks whether there is not another device with the entered name
+        #
+        # @return [Boolean]
+        def uniqueness_validation
+          return true unless controller.duplicated_vg_name?
+
+          Yast::Popup.Error(
+            format(
```
ancorgs: Maybe a comment for translators here. My feeling is that they are used to the `%s` and `%1` notation, but not so much to the Ruby keyword placeholders one.

src/lib/y2partitioner/actions/controllers/lvm_vg.rb
```diff
@@ -0,0 +1,298 @@
...
+        def illegal_vg_name?
+          vg_name.split(//).any? { |c| !ALLOWED_NAME_CHARS.include?(c) }
+        end
+
+        # Checks whether it exists another device with the given volume group name
+        #
+        # @return [Boolean] true if there is another device with the same name;
+        #   false otherwise
+        def duplicated_vg_name?
+          name = File.join("/dev", vg_name)
```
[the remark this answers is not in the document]
joseivanlopez: Added comment to explain it.

src/lib/y2partitioner/dialogs/lvm_vg.rb
joseivanlopez: It is necessary to focus the field when an error happens. Commented in the code.

src/lib/y2partitioner/dialogs/lvm_vg.rb
joseivanlopez: I agree, so the messages generation has been moved to the controller.

src/lib/y2partitioner/dialogs/lvm_vg.rb
joseivanlopez: Added.

src/lib/y2partitioner/widgets/md_add_button.rb
joseivanlopez: As commented, `LvmVgAddButton` and `MdAddButton` do not work over a specific device. The class `DeviceButton` is meant to act over a device.

src/lib/y2storage/disk_size.rb
joseivanlopez: Totally agree (but you cannot say that the hack was quite cute)

test/y2partitioner/dialogs/lvm_vg_test.rb
joseivanlopez: Added.

src/lib/y2partitioner/actions/controllers/lvm_vg.rb
joseivanlopez: This method is only called when all validations are passed, so in practice it is not possible to assign a nil value.

src/lib/y2storage/disk_size.rb
ancorgs: Yes. It was cool. Unnecessary but cool :-)

src/lib/y2partitioner/actions/controllers/lvm_vg.rb
```diff
@@ -0,0 +1,364 @@
...
+          # generate the volume group name has been duplicated here due to its simplicity.
+          name = File.join("/dev", vg_name)
+          Y2Storage::BlkDevice.all(working_graph).map(&:name).include?(name)
+        end
+
+        # Error message to show when the given volume group name is duplicated
+        #
+        # @return [String]
+        def duplicated_vg_name_message
+          # TRANSLATORS: vg_name is replaced by the name of the volume group
```
ancorgs: NP: I would say: `%{vg_name} is replaced...` The whole point is that translators may not be familiar with the `%{whatever}` syntax.

PR #424, opened 2017-11-15
Enhance criteria to choose the best partitions distribution 

ancorgs:
See https://hackweek.suse.com/projects/yast2-storage-ng-ensure-the-best-layout-is-proposed

Better review in a per-commit basis

### DIFF
--- a/src/lib/y2storage/planned/assigned_space.rb
+++ b/src/lib/y2storage/planned/assigned_space.rb
@@ -67,6 +67,13 @@ module Y2Storage
         end
       end
 
+      # Sum of the weights of all the planned partitions assigned to this space
+      #
+      # @return [Integer]
+      def total_weight
+        partitions.map { |p| p.weight || 0 }.reduce(:+)
+      end
+
       # Checks if the volumes really fit into the assigned space
       #
       # TODO: We do not check for start_offset. Anyways,
--- a/src/lib/y2storage/planned/partitions_distribution.rb
+++ b/src/lib/y2storage/planned/partitions_distribution.rb
@@ -117,6 +117,10 @@ module Y2Storage
       #
       # @return [Integer] -1, 0, 1 like <=>
       def better_than(other)
+        # Sorted list of criteria (starting with the most important) to use in
+        # order to select which distribution is best. Basically names of methods
+        # that return something comparable using #<=>, followed by the direction
+        # of the comparison (whether a bigger value is better or worse).
         criteria = [
           # The smaller gaps the better
           [:gaps_total_size, :smaller],
--- a/src/lib/y2storage/planned/partitions_distribution.rb
+++ b/src/lib/y2storage/planned/partitions_distribution.rb
@@ -164,11 +164,8 @@ module Y2Storage
       # @see #better_than
       # @return [Float] a number between 0.0 and 1.0
       def weight_space_deviation
-        extra_sizes = spaces.map { |i| i.usable_extra_size.to_i }
-        total_extra_size = extra_sizes.reduce(:+)
-
-        weights = spaces.map(&:total_weight)
-        total_weight = weights.reduce(:+)
+        total_extra_size = spaces.map(&:usable_extra_size).reduce(:+)
+        total_weight = spaces.map(&:total_weight).reduce(:+)
 
         # Edge case #1:
         # No partition looks interested in growing, so we are fine whatever
@@ -179,11 +176,10 @@ module Y2Storage
         # any partition to grow
         return 1.0 if total_extra_size.zero?
 
-        normalized_sizes = extra_sizes.map! { |i| i.to_f / total_extra_size }
-        normalized_weights = weights.map! { |i| i.to_f / total_weight }
-
-        diffs = normalized_sizes.each_with_index.map do |size, idx|
-          (size - normalized_weights[idx]).abs
+        diffs = spaces.map do |space|
+          normalized_size = space.usable_extra_size.to_i.to_f / total_extra_size.to_i
+          normalized_weight = space.total_weight.to_f / total_weight
+          (normalized_size - normalized_weight).abs
         end
         diffs.reduce(:+)
       end
--- a/src/lib/y2storage/planned/partitions_distribution.rb
+++ b/src/lib/y2storage/planned/partitions_distribution.rb
@@ -100,6 +100,12 @@ module Y2Storage
         DiskSize.sum(spaces.map(&:disk_size))
       end
 
+      # Number of assigned spaces in the distribution
+      # @return [Integer]
+      def spaces_count
+        spaces.count
+      end
+
       # Total number of planned partitions included in the distribution
       # @return [Integer]
       def partitions_count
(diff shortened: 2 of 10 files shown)


src/lib/y2storage/planned/assigned_space.rb
```diff
@@ -67,6 +67,13 @@ def partition_type
         end
       end
 
+      # Sum of the weights of all the planned partitions assigned to this space
+      #
+      # @return [Integer]
+      def total_weight
+        partitions.map { |p| p.weight || 0 }.reduce(:+)
```
jreidinger: does it make sense to have default in that weight instead? so nil is 0 defined in one place. I cannot see semantics in weight definition, so I am not so sure.

src/lib/y2storage/planned/partitions_distribution.rb
```diff
@@ -96,7 +96,7 @@ def gaps_count
 
       # Total space available for the planned partitions
       # @return [DiskSize]
-      def spaces_total_disk_size
+      def spaces_total_size
```
jreidinger: well, when you renaming it, why not remove `_total_`? for me `gaps_size` and `spaces_size`.

src/lib/y2storage/planned/partitions_distribution.rb
```diff
@@ -100,6 +100,12 @@ def spaces_total_size
         DiskSize.sum(spaces.map(&:disk_size))
       end
 
+      # Number of assigned spaces in the distribution
+      # @return [Integer]
+      def spaces_count
```
jreidinger: is it really worth? I do not see benefit of writting `spaces_count` over `spaces.count`

jreidinger: ah, reading it below I see the benefit

src/lib/y2storage/planned/partitions_distribution.rb
```diff
@@ -111,31 +117,77 @@ def partitions_count
...
-        # The biggest installation the better
-        res = other.spaces_total_size <=> spaces_total_size
-        return res unless res.zero?
+          # We want the partitions with biggest weight to be placed in the spaces
+          # in which there is more extra space to be distributed
+          [:weight_space_deviation, :smaller],
+
+          # The less fragmentation the better
+          [:spaces_count, :smaller]
+        ]
```
jreidinger: btw this start smelling for `Comparator` class, which gets two object and compare it, so you can have even more complex criteria. and does not need to have that helpers.

src/lib/y2storage/planned/partitions_distribution.rb
```diff
@@ -111,31 +117,77 @@ def partitions_count
...
+      #
+      # In general, for two distributions that are considered equivalent in
+      # any other aspect, the one in which the partitions with biggest weights
+      # are placed in the spaces with more "growth potential" is considered the
+      # best. This coefficient is zero if the proportion of weights (of every
+      # assigned space compared with the total) is equal to the proportion of
+      # extra spaces (same calculation).
+      #
+      # @see #better_than
+      # @return [Float] a number between 0.0 and 1.0
```
jreidinger: why it return Float when better then need Integer?

src/lib/y2storage/planned/partitions_distribution.rb
```diff
@@ -111,31 +117,77 @@ def partitions_count
...
+        # any partition to grow
+        return 1.0 if total_extra_size.zero?
+
+        normalized_sizes = extra_sizes.map! { |i| i.to_f / total_extra_size }
+        normalized_weights = weights.map! { |i| i.to_f / total_weight }
+
+        diffs = normalized_sizes.each_with_index.map do |size, idx|
+          (size - normalized_weights[idx]).abs
+        end
+        diffs.reduce(:+)
```
jreidinger: ugh, very complex method. As I said, here object that do such comparison makes some sense for me.

src/lib/y2storage/planned/assigned_space.rb
ancorgs: Well, this is a sum and 0 is the [identity element](https://en.wikipedia.org/wiki/Identity_element) for addition operations. Since there is nothing like the "canonical default weight" (in fact, the weights only make full sense in concrete contexts, when compared to others), using the identity element of the operation is the most reasonable approach.

src/lib/y2storage/planned/partitions_distribution.rb
ancorgs: Because `disk_size` has proven to be confusing in the past. And `total` has never confused anyone and I personally find it useful for clarification purposes.

src/lib/y2storage/planned/partitions_distribution.rb
ancorgs: `better_than` doesn't need Integer. It needs two values that are comparable via `<=>`. For example. `#gaps_total_size` or `#spaces_total_size` are not integer either.

src/lib/y2storage/planned/partitions_distribution.rb
```diff
@@ -111,31 +117,77 @@ def partitions_count
...
+        # All the spaces are fully packet since the beginning, so no chance for
+        # any partition to grow
+        return 1.0 if total_extra_size.zero?
+
+        normalized_sizes = extra_sizes.map! { |i| i.to_f / total_extra_size }
+        normalized_weights = weights.map! { |i| i.to_f / total_weight }
+
+        diffs = normalized_sizes.each_with_index.map do |size, idx|
+          (size - normalized_weights[idx]).abs
+        end
```
jreidinger: this was quite confusing to have it at first as two separated values and now connect it via index. What about different data structure like. array of spaces and then do operation on top of it? something like

```
total_extra_size = spaces.map { |i| i.usable_extra_size.to_i }.reduce(:+)
total_weight = spaces.map(&:total_weight).reduce(:+)

# edge cases here
...

diffs = spaces.map do |space|
  normalized_size = space.usable_extra_size.to_i / total_extra_size
  normalized_weigh = space.total_weight / total_weight
  (normalized_size - normalized_weight).abs
end

diffs.reduce(:+)
```

this is at least for me more readable and easier to recognize what algorithm did.

src/lib/y2storage/planned/partitions_distribution.rb
ancorgs: How would a separate object help in making this calculation less complex? I'm not sure if this can be written in a more simple way (codeclimate and rubocop didn't complain and they tend to be quite picky).

But I'm open to any suggested alternative that allows to evaluate the same. That is: "_given a draw in all the previous criteria, which distribution places partitions in the spaces in which they can grow in a way that is as consistent as possible with the partitions' weights_"... or something like that.

src/lib/y2storage/planned/partitions_distribution.rb
jreidinger: ah, maybe document it on top of criteria, that it require method that return something where `<=>` can be applied

src/lib/y2storage/planned/partitions_distribution.rb
```diff
@@ -111,31 +117,73 @@ def partitions_count
...
 
-        # The biggest installation the better
-        res = other.spaces_total_disk_size <=> spaces_total_disk_size
-        return res unless res.zero?
+          # We want the partitions with biggest weight to be placed in the spaces
+          # in which there is more extra space to be distributed
+          [:weight_space_deviation, :smaller],
+
+          # The less fragmentation the better
+          [:spaces_count, :smaller]
```
jreidinger: BTW, why not adding here also `[:comparable_string, :smaller]` ? so it is in one list.
and also it smell like a constant for me.

ancorgs: Well, the previous loop contains a `return res unless res.zero?` as the last line. I wanted to empathize two things:

- That `unless res.zero?` does not affect `#comparable_string`. No matter if the result is zero (which in fact can NEVER happen for two distributions that are in fact different), the result will be returned.
- That `#comparable_string` is not a real criteria. It have no meaning about the value of the distribution for our purposes. It's just a last-resort fallback.

src/lib/y2storage/planned/partitions_distribution.rb
ancorgs: Comment added

PR #426, opened 2017-11-22
Avoid RAID chunk size below 64 kiB

shundhammer:
https://trello.com/c/8mxdS9qy/765-1-sles15-p2-1065381-build-3215ppc64-openqa-test-fails-in-to-create-raid10-stderrmdadm-runarray-failed-invalid-argument

https://bugzilla.suse.com/show_bug.cgi?id=1065381

For most RAID levels now use a chunk size of 64 kiB except for some where the page size is irrelevant. 

Now no longer using a fixed list of chunk sizes, but calculating it on the fly, depending on the default chunk size. If that is less than 64 kiB, use it as the starting point.

Also, after discussing this with Arvin extended the list of chunk sizes to up to 64 MiB since this might be more efficient and thus desired by many customers.

Caveat: It turned out that DiskSize.to_s likes to use fractional numbers in some cases, so the list contains `0.5 MiB` instead of `512 MiB`. This is just another way of presenting the same information, yet it might look a bit odd to some users. Let's see if users actually dislike this.

## RAID 0
raid0-chunks

-------------------------

## RAID 1

raid1-chunks

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/md.rb
+++ b/src/lib/y2partitioner/actions/controllers/md.rb
@@ -157,6 +157,20 @@ module Y2Partitioner
           md.minimal_number_of_devices
         end
 
+        # Possible chunk sizes for the Md object depending on its md_level.
+        #
+        # @return [Array<Y2Storage::DiskSize>]
+        def chunk_sizes
+          sizes = []
+          size = min_chunk_size
+
+          while size <= max_chunk_size
+            sizes << Y2Storage::DiskSize.new(size)
+            size *= 2
+          end
+          sizes
+        end
+
       private
 
         def working_graph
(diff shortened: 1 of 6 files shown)


src/lib/y2partitioner/actions/controllers/md.rb
```diff
@@ -173,18 +187,26 @@ def available?(partition)
...
           when :raid1
             Y2Storage::DiskSize.KiB(4)
           when :raid5, :raid6
             Y2Storage::DiskSize.KiB(128)
           when :raid10
-            Y2Storage::DiskSize.KiB(32)
+            Y2Storage::DiskSize.KiB(64)
           else
-            Y2Storage::DiskSize.KiB(4)
+            Y2Storage::DiskSize.KiB(64)
```
jreidinger: I see four times in code 64. What about using constant with reference to this bug explaining why now(!) 64kb is reasonable minimum, so we know it for future?

shundhammer: Those values seem to fall from the sky. There is not exactly any reference we could point to. Just see the PBI.

PR #427, opened 2017-11-22
Partitioner: resize md

joseivanlopez:
PBI: https://trello.com/c/2xffkeJR/738-3-expert-partitioner-resize-md-raid

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/md.rb
+++ b/src/lib/y2partitioner/actions/controllers/md.rb
@@ -27,9 +27,9 @@ require "y2partitioner/ui_state"
 module Y2Partitioner
   module Actions
     module Controllers
-      # This class stores information about an MD RAID being created and takes
-      # care of updating the devicegraph when needed, so the different dialogs
-      # can always work directly on a real Md object in the devicegraph.
+      # This class stores information about an MD RAID being created or modified
+      # and takes care of updating the devicegraph when needed, so the different
+      # dialogs can always work directly on a real Md object in the devicegraph.
       class Md
         include Yast::I18n
         extend Forwardable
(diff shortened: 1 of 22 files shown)


src/lib/y2partitioner/actions/controllers/md.rb
```diff
@@ -123,10 +131,20 @@ def md_name=(name)
         end
 
         # Title to display in the dialogs during the process
+        #
+        # @param action [Symbol] action currently being performed (:add, :resize)
         # @return [String]
-        def wizard_title
-          # TRANSLATORS: dialog title. %s is a device name like /dev/md0
-          _("Add RAID %s") % md.name
+        def wizard_title(action: nil)
```
ancorgs: Couldn't the action been set in the constructor? If an md is passed to `initialize`, then it's `:resize`. So we can avoid this argument here and in the dialogs. Just an idea.

src/lib/y2partitioner/actions/resize_md.rb
```diff
@@ -0,0 +1,120 @@
...
+            name: controller.md.name
+          )
+        )
+
+        false
+      end
+
+      # @return [Boolean] true if the MD RAID is already created on disk; false otherwise
+      def committed?
+        controller.md.exists_in_probed?
```
ancorgs: This should use `DevicesGraph.system`, not probed.

src/lib/y2partitioner/dialogs/md_options.rb
```diff
@@ -19,7 +19,7 @@ def initialize(controller)
 
       # @macro seeDialog
       def title
-        @controller.wizard_title
+        @controller.wizard_title(action: :add)
```
ancorgs: See comment above about the action being decided in the controller.

PR #428, opened 2017-11-22
Name-based stable sorting for (auto)installation

ancorgs:
Main pull request for https://trello.com/c/6szyiBWI/763-2-sle15-storage-ng-select-the-disk-to-use-consistently

### DIFF
--- a/test/y2storage/devicegraph_test.rb
+++ b/test/y2storage/devicegraph_test.rb
@@ -200,7 +200,7 @@ describe Y2Storage::Devicegraph do
       it "includes all disks and DASDs" do
         expect(graph.disk_devices.map(&:name)).to contain_exactly(
           "/dev/dasda", "/dev/dasdb", "/dev/sda", "/dev/sdb", "/dev/sdc",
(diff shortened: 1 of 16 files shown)


test/y2storage/devicegraph_test.rb
```diff
@@ -197,20 +202,41 @@ def with_sda2_deleted(initial_graph)
...
+          "/dev/dasda", "/dev/dasdb", "/dev/nvme0n1", "/dev/sda", "/dev/sdb",
+          "/dev/sdc", "/dev/sdd", "/dev/sdf", "/dev/sdh", "/dev/sdaa"
+        ]
+      end
+
+      context "even if Disk.all and Dasd.all return unsorted arrays" do
+        before do
+          allow(Y2Storage::Disk).to receive(:all) do |devicegraph|
+            # Let's shuffle things a bit
+            Y2Storage::Partitionable.all(devicegraph).select { |i| i.is?(:disk) }.shuffle
```
joseivanlopez: NP: Maybe it would be better to avoid `#shuffle` usage, because this makes the test runs with "different" input each time.

PR #430, opened 2017-11-24
Use Partition#adapted_id

imobachgs:
Use the safer `Partition#adapted_id` method. It fixes [bsc#1067207](https://bugzilla.suse.com/show_bug.cgi?id=1067207).


imobachgs:
Thanks!

PR #431, opened 2017-11-24
Never try to copy staging into itself. Bug#1069671

ancorgs:
Verified manually that it fixes https://bugzilla.suse.com/show_bug.cgi?id=1069671

### DIFF
--- a/test/y2storage/storage_manager_test.rb
+++ b/test/y2storage/storage_manager_test.rb
@@ -104,8 +104,8 @@ describe Y2Storage::StorageManager do
     end
 
     context "when trying to assign staging to itself" do
-      # It used to be in bug basically running staging.copy(staging), which
-      # resulted in staging becoming empty
+      # In the past, copying staging into itself, i.e. staging.copy(staging),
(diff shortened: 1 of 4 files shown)


test/y2storage/storage_manager_test.rb
```diff
@@ -102,6 +102,33 @@ def devicegraph_from(file_name)
       manager.staging = new_graph
       expect(manager.proposal).to be_nil
     end
+
+    context "when trying to assign staging to itself" do
+      # It used to be in bug basically running staging.copy(staging), which
```
wfeldt: > It used to be in bug basically running staging.copy(staging)

Could you rephrase this? It's not very clear what is meant.

ancorgs: What about: "In the past this tried to copy staging into itself, i.e. staging.copy(staging), which caused it to become a completely empty devicegraph"

wfeldt: Why not just:

In the past, copying staging into itself, i.e. staging.copy(staging), caused ...

PR #432, opened 2017-11-24
WIP: sorting of blk devices by name

ancorgs:
@aschnell, @joseivanlopez please check.

This is a proof of concept of an alternative algorithm to the `compare_by_name` offered by libstorage-ng. It follows the same principle, that is, alphabetical-like order but dealing well with stuff like `sdb < sdaa` or `vda2 < vda10`.

There is no need to keep it in Ruby, I'm fine with moving it to the library if we agree this supports more cases than the current one in the library and introduces no regression (I just was more comfortable with Ruby for writing a POC).

Advantages:

 * It supports sorting all kind of blk devices, even those that include several variable parts in the name (like "/dev/nvmeXvYpZ")
 * It works correctly when sorting sets of devices with a common name structure but that belongs to different types. That is, is not affected by the problem demonstrated at http://paste.opensuse.org/64738254

Several unit tests included to probe it works in all situations (so far). The regular expression for several blk device types are still missing, but this should be enough to get the idea.

### DIFF
--- a/test/y2storage/blk_device_test.rb
+++ b/test/y2storage/blk_device_test.rb
@@ -527,4 +527,34 @@ describe Y2Storage::BlkDevice do
       end
     end
   end
+
+  describe ".sorted_by_name" do
+    before { fake_scenario("sorting/disks_and_dasds1") }
+
+    it "returns all the blk devices sorted by name" do
+      devices = Y2Storage::BlkDevice.sorted_by_name(fake_devicegraph)
+      expect(devices.map(&:basename)).to eq %w(
+        dasda dasda1 dasda2 dasda10 dasdb dasdab dasdb1 dasdb2 dasdb3
+        nvme0n1 nvme0n1p1 nvme0n1p2 nvme0n1p3 nvme0n1p4 nvme0n1p10 nvme0n1p11 nvme0n1p40
+        nvme0n2 nvme0n2p1 nvme0n2p2 nvme1n1 nvme1n1p1 nvme1n1p2
+        sda sdb sdaa sdb1 sdb2 sdc sdc1 sdc2 sdc3 sdc4 sdc10 sdc21 sdaa1 sdaa2 sdaa3
+      )
+    end
+  end
+
+  describe "#compare_by_name" do
+    context "when devices of different types share a common name structure" do
+      before { fake_scenario("sorting/disks_and_dasds2") }
+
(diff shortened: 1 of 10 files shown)


test/y2storage/blk_device_test.rb
```diff
@@ -527,4 +527,34 @@
...
     end
   end
+
+  describe ".sorted_by_name" do
+    before { fake_scenario("sorting/disks_and_dasds1") }
+
+    it "returns all the blk devices sorted by name" do
+      devices = Y2Storage::BlkDevice.sorted_by_name(fake_devicegraph)
+      expect(devices.map(&:basename)).to eq %w(
+        dasda dasda1 dasda2 dasda10 dasdb dasdab dasdb1 dasdb2 dasdb3
```
aschnell: I would expect dasdb1 before dasdab.

test/y2storage/blk_device_test.rb
```diff
@@ -527,4 +527,34 @@
...
+        nvme0n1 nvme0n1p1 nvme0n1p2 nvme0n1p3 nvme0n1p4 nvme0n1p10 nvme0n1p11 nvme0n1p40
+        nvme0n2 nvme0n2p1 nvme0n2p2 nvme1n1 nvme1n1p1 nvme1n1p2
+        sda sdb sdaa sdb1 sdb2 sdc sdc1 sdc2 sdc3 sdc4 sdc10 sdc21 sdaa1 sdaa2 sdaa3
```
aschnell: I would expect sdb1 before sdaa.

test/y2storage/blk_device_test.rb
ancorgs: @aschnell That's doable. See https://github.com/yast/yast-storage-ng/pull/434 which works as you expect.

Here is the diff between both https://github.com/yast/yast-storage-ng/pull/434/commits/e3e87907c063df95f3107fe62abfc4f1eb95311f

For the first POC I just wanted to keep it simple to make the code easier to follow (for the sake of the discussion) and, eventually, to translate to C++. But I'm fine with any of the alternatives.

I can try to merge any of them into yast-storage-ng (whichever you guys prefer), with the translation to C++ as a separate PBI or extra task for this PBI.

ancorgs:
Closed in favor of https://github.com/yast/yast-storage-ng/pull/454, which uses the new sort-key-based implementation in https://github.com/openSUSE/libstorage-ng/pull/400

PR #433, opened 2017-11-24
Allow to work with BIOS RAIDs

joseivanlopez:
PBI: https://trello.com/c/wuFHiJ9X/781-betasles15-p3-1067349-no-disk-appear-on-yast-partitioner-guided-setup-with-intel-rste-swraid

Bug report: https://bugzilla.suse.com/show_bug.cgi?id=1067349


src/lib/y2partitioner/actions/add_partition.rb
```diff
@@ -57,29 +63,77 @@ def size
...
+      # @return [Controllers::Partition]
       attr_reader :part_controller
 
       # @see TransactionWizard
       def sequence_hash
         {
-          "ws_start" => "type",
-          "type"     => { next: "size" },
-          "size"     => { next: new_blk_device_step1, finish: :finish }
+          "ws_start"          => "delete_filesystem",
```
ancorgs: And what about disks directly used as PV or as part of a RAID or similar situations? Don't we want to handle them in a similar way?

src/lib/y2storage/md_container.rb
```diff
@@ -0,0 +1,65 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage/storage_class_wrapper"
+require "y2storage/md"
+
+module Y2Storage
+  # MD Container
+  #
+  # This is a wrapper for Storage::MdContainer
```
ancorgs: I would appreciate more documentation here about what an MdContainer is. The inheritance-based solution to distinguish software MD RAIDs from BIOS ones is counterintuitive to say the least. So in general I would add a couple of sentences to Md, another couple to MdContainer, same for MdMember and maybe a deeper explanation of `#software_defined?` and its implications.

src/lib/y2partitioner/actions/add_partition.rb
joseivanlopez: Good point. We should prevent modification of used devices.

test/y2storage/md_test.rb
```diff
@@ -217,4 +217,9 @@
 
   end
 
+  describe "#software_defined?" do
```
ancorgs: I know the corresponding tests are in `md_member_test.rb` and `md_container_test.rb`, but I think it would be better to understand the method if we would have here a more rich example with some objects returning true and others returning false.

src/lib/y2storage/blk_device.rb
```diff
@@ -315,5 +328,15 @@ def basename
...
       Y2Storage::HWInfoReader.instance.for_device(name)
     end
+
+    # Checks whether the device is in use
+    #
+    # @note A device is in use when it is used as physical volume, belongs to a
+    #   multipath or belongs to a raid.
+    #
+    # @return [Boolean]
+    def used?
```
ancorgs: I don't like the name of the method. A disk with formatted partitions is also "used" from some point of view.

Even worst, this method lives in `BlkDevice` but it internally calls `#partittion_table` which is defined in `Partitionable`. I would expect fireworks if this method is called on a `Encryption` or `LvmLv`, that are also block devices. And I would expect it to always return false for objects of type `Partition` (`#partition_table` cannot be nil) 

So this method should be in `Partitionable`. About the name... that's tricky. It checks whether the partitionable as a whole (not through its partitions) is used as a component of some other bigger device (like a RAID, a VG or a multipath).

ancorgs: I would even give up on providing such a specific check directly in `Y2Storage::Partitionable`. It's a one-liner, so it can be checked directly in the controller of the action in the partitioner. Specially taking into account that BIOS RAIDs and Multipath are kind of irrelevant there, since those disks will not be shown anyways.

test/y2partitioner/actions/controllers/partition_test.rb
```diff
@@ -148,6 +148,72 @@
     end
   end
 
+  describe "#used?" do
```
ancorgs: The method is called `#disk_used?`

PR #434, opened 2017-11-26
WIP: sorting of blk devices by name + feedback

ancorgs:


aschnell:
AFAIS you do sorting only based on the name. I think it is not possible to get
a good and consistent sorting this way. E.g. logical volumes should be sorted
lexicographical, e.g. /dev/system/a, /dev/system/aa, /dev/system/b,
/dev/system/c0, /dev/system/c10, /dev/system/c2. And named MD RAIDs should
have the same ordering, so e.g. /dev/md/a1, /dev/md/a10, /dev/md/a2,
/dev/md/b1. Partitions should be sorted by number, e.g. /dev/sda1, /dev/sda2,
/dev/sda10. Unfortunately it is unclear whether /dev/md/a10 is a named MD RAID
or a partition on a named MD RAID but that influences the ordering.

joseivanlopez:
So, according to commented by @aschnell, comparison should take into account the string you are comparing to and also the class of the compared object:

```
<Md name="/dev/md/a10"> <=> <Md name="/dev/md/2"> #=> 1

<Partition name="/dev/md/a10"> <=> <Md name="/dev/md/2"> #=> 1

<Partition name="/dev/md/a10"> <=> <Partition name="/dev/md/2"> #=> -1
```

ancorgs:
So the objects don't only have to tell us their naming pattern, but also whether their prefer alphabetical or natural order. Got it. Let's try a new refinement having that into account.

aschnell:
I thought a bit about the sorting problem and I think the approach is broken
since the compare function does not fulfill the strict weak ordering
requirement (in C++, Ruby documentation is silent about any such details) - at
least when extended.

E.g. there are three devices, /dev/sdaa (Disk), /dev/sdkit/src (LvmLv) and
/dev/sdz (Disk). The fallback comparison yields /dev/sdaa < /dev/sdkit/src and
/dev/sdkit/src < /dev/sdz. It is thus required that /dev/sdaa < /dev/sdz. But
we want /dev/sdz < /dev/sdaa.

Maybe introducing a sort-key is better and safer. Simply replace /dev/sda by
"/dev/sd  a" and "/dev/sdaa" by "/dev/sd aa". That will produce the desired
sorting for disks. Surely /dev/sdkit/src will be behind all disks starting
with /dev/sd.

aschnell:
Compare using a sort-key is implemented in https://github.com/openSUSE/libstorage-ng/pull/400.

ancorgs:
Closed in favor of https://github.com/yast/yast-storage-ng/pull/454, which uses the new sort-key-based implementation in https://github.com/openSUSE/libstorage-ng/pull/400

PR #435, opened 2017-11-27
PartitionTables::partition_id_supported? is now exported

wfeldt:


ancorgs:
We need to wait a bit for the method to be available in a libstorage-ng binary and then restart travis.

PR #436, opened 2017-11-28
WIP: Added can_resize? checks

shundhammer:


src/lib/y2storage/device.rb
```diff
@@ -111,6 +111,16 @@ class Device
...
 
+    # @!method detect_resize_info
+    #   Information about the possibility of resizing a given device.
+    #   If the device has any children, they are also taken into account;
+    #   the result of this method is the combined information about this device
+    #   and all its children.
+    #   @see can_resize?
+    #
+    #   @return [ResizeInfo]
+    storage_forward :detect_resize_info
```
ancorgs: I believe this should be:
```ruby
storage_forward :detect_resize_info, as: "ResizeInfo"
```
So the method retains the old behavior of returning a `Y2Storage::ResizeInfo` instead of a `Storage::ResizeInfo`

src/lib/y2storage/device.rb
```diff
@@ -170,19 +180,15 @@ def siblings
...
-      return nil unless exists_in_probed?
-      probed_device = StorageManager.instance.probed.find_device(sid)
-      probed_device.storage_detect_resize_info
+    # If the device has any children, they are also taken into account;
+    # the result of this method is the combined information about this device
+    # and all its children.
+    #
+    # @return [Boolean] true if the device can be resized, false if not.
+    def can_resize?
+      detect_resize_info.resize_ok
```
ancorgs: In combination with the change suggested above, this would become `detect_resize_info.resize_ok?`

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/resize_info.rb#L34

src/lib/y2storage/device.rb
shundhammer: ok

src/lib/y2storage/device.rb
shundhammer: ok

src/lib/y2storage/device.rb
```diff
@@ -170,19 +174,15 @@ def siblings
...
-    def detect_resize_info
-      return nil unless exists_in_probed?
-      probed_device = StorageManager.instance.probed.find_device(sid)
-      probed_device.storage_detect_resize_info
+    # If the device has any children, they are also taken into account;
+    # the result of this method is the combined information about this device
+    # and all its children.
+    #
+    # @return [Boolean] true if the device can be resized, false if not.
+    def can_resize?
```
ancorgs: Just an afterthought on this. If `detect_resize_info` is an expensive operation (I don't know, It's just what the name seems to suggest), maybe "hiding" the call on this method is not such a good idea. Just looking at the resulting API I would be tempted to call `#can_resize?` over and over instead of operating with the cached `ResizeInfo` object.

shundhammer: I'll add a note to the documentation.

PR #437, opened 2017-11-28
Fix size=max for logical volumes

imobachgs:
Set weight properly so the logical volume will grow when distributing space.

### DIFF
--- a/src/lib/y2storage/proposal/autoinst_devices_planner.rb
+++ b/src/lib/y2storage/proposal/autoinst_devices_planner.rb
@@ -395,6 +395,7 @@ module Y2Storage
           lv.min_size = size_info.min
           lv.max_size = size_info.max
         end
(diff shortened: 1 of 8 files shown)


src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -395,6 +395,7 @@ def assign_size_to_lv(vg, lv, lv_section)
           lv.min_size = size_info.min
           lv.max_size = size_info.max
         end
+        lv.weight = 1 if size_info.max == DiskSize.unlimited
```
ancorgs: NP: if `size_info.max` is always set (I hope so) you can do `size_info.max.unlimited?` Just for your information.

imobachgs: Nice. I've extended AutoinstSize (which is the class which `size_info` belongs to) with an `unlimited?` method.

PR #438, opened 2017-11-28
Fixed unit tests

shundhammer:
- Removed now obsolete unit test for detect_resize_info
- Added new unit test for can_resize?

### DIFF
--- a/test/y2storage/device_test.rb
+++ b/test/y2storage/device_test.rb
@@ -92,6 +92,7 @@ describe Y2Storage::Device do
     subject(:device) { Y2Storage::Partition.find_by_name(fake_devicegraph, "/dev/sda2") }
 
     it "reports that it can resize an ext4 partition" do
+      allow(resize_info).to receive(:resize_ok).and_return true
       allow(device.to_storage_value).to receive(:detect_resize_info).and_return resize_info
       expect(device.can_resize?).to eq true
     end
--- a/test/y2storage/device_test.rb
+++ b/test/y2storage/device_test.rb
@@ -87,36 +87,13 @@ describe Y2Storage::Device do
     end
   end
 
-  describe "#detect_resize_info" do
-    let(:probed) { double(Y2Storage::Devicegraph) }
-    let(:probed_partition) { double(Y2Storage::Partition, storage_detect_resize_info: resize_info) }
-    let(:resize_info) { double(Y2Storage::ResizeInfo) }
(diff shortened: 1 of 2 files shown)


test/y2storage/device_test.rb
```diff
@@ -87,36 +87,14 @@
...
-
-    context "if the device exists in probed" do
-      let(:in_probed) { true }
-
-      it "returns the resize info from the equivalent partition in probed" do
-        expect(probed).to receive(:find_device).with(444).and_return probed_partition
-        expect(staging_partition.detect_resize_info).to eq resize_info
-      end
+  describe "#can_resize?" do
+    let(:resize_info) { Y2Storage::ResizeInfo.new(true) }
```
ancorgs: This is not the correct way of creating a `Y2Storage::ResizeInfo` object. `Y2Storage::ResizeInfo` is a wrapper around `Storage::ResizeInfo`, so the initializer receives a `Storage::ResizeInfo` object.

You cannot wrap what is not there. That's why a double like this was used before:
```ruby
let(:resize_info) { double(Y2Storage::ResizeInfo, resize_ok?: resize_ok) }
```

test/y2storage/device_test.rb
```diff
@@ -87,36 +87,14 @@
...
-    end
-
-    context "if the device exists in probed" do
-      let(:in_probed) { true }
-
-      it "returns the resize info from the equivalent partition in probed" do
-        expect(probed).to receive(:find_device).with(444).and_return probed_partition
-        expect(staging_partition.detect_resize_info).to eq resize_info
-      end
+  describe "#can_resize?" do
```
ancorgs: I know writing a meaningful test for such a small method is hard. But at least we should test the two possible scenarios: something can be resized vs something cannot.

These tests would still pass if the implementation would be `def can_resize?; true; end` 😉 

This will cover at least the two possible results:

```ruby
describe "#can_resize?" do
    subject(:device) { Y2Storage::Partition.find_by_name(fake_devicegraph, "/dev/sda1") }
    let(:resize_info) { double(Y2Storage::ResizeInfo, resize_ok?: resize_ok) }
 
    before { allow(device).to receive(:detect_resize_info).and_return resize_info }
 
    context "if libstorage-nd reports that resizing is possible" do
      let(:resize_ok) { true }
 
      it "returns true" do
        expect(device.can_resize?).to eq true
      end
    end
 
    context "if libstorage-ng reports that resizing is not possible" do
      let(:resize_ok) { false }
 
      it "returns true" do
        expect(device.can_resize?).to eq false
      end
    end
  end
```

PR #440, opened 2017-11-29
show only partition ids suitable for a partition type

wfeldt:


src/lib/y2storage/partition_id.rb
```diff
@@ -158,5 +158,14 @@ def is?(*names)
         end
       end
     end
+
+    # List of partition ids suitable for a specific partition table.
+    #
+    # @param ptable [Y2Storage::PartitionTables::Base]
+    #
+    # @return [Array<PartitionId>]
+    def self.all_for_partition_table(ptable)
```
ancorgs: Wouldn't it be way more natural to have this method/responsibility in the partition table instead of a class method here?

I mean, doesn't `a_device.partition_table.supported_partition_ids` look more natural than `Y2Storage::PartitionId.all_for_partition_table(a_device.partition_table)`?

wfeldt: Oh, whatever. To me it looks the same. :-)

ancorgs: It's not only more natural, also sorter in general. :-)

Class methods are a resource to be used with care in API design. If it's possible to have an instance method, the resulting API will almost always be more object oriented. 

This is specially true when you have `PartitionTable#partition_id_supported?`. If the instance of that class has the responsibility of knowing if a partition id is supported, it makes sense that such instance also has the responsibility of knowing all the supported ids.

test/y2storage/partition_tables/dasd_test.rb
```diff
@@ -76,4 +76,18 @@
       expect(subject.partition_id_supported?(0)).to eq false
     end
   end
+
+  describe "#:supported_partition_ids" do
```
ancorgs: NP: surplus colon

test/y2storage/partition_tables/dasd_test.rb
```diff
@@ -76,4 +76,18 @@
...
+    it "for dasd, list includes the LINUX id" do
```
ancorgs: NP: Not sure if it makes much sense to repeat "for dasd" in every description of the tests of `PartitionTables::Dasd`

test/y2storage/partition_tables/dasd_test.rb
```diff
@@ -76,4 +76,18 @@
...
+      expect(subject.supported_partition_ids.include?(Y2Storage::PartitionId::LINUX)).to eq true
```
ancorgs: Slightly sorter and way more readable IMHO:
```ruby
expect(subject.supported_partition_ids).to include Y2Storage::PartitionId::LINUX
```

test/y2storage/partition_tables/dasd_test.rb
wfeldt: rspec looks more readable for me

wfeldt: I'd also have to change it in `partition_id_supported?`, consequently

PR #450, opened 2017-12-01
Resize partitions

joseivanlopez:
PBI: https://trello.com/c/qDcdjjBL/697-5-partitioner-plain-partitioning-resize


src/lib/y2partitioner/actions/resize_partition.rb
```diff
@@ -0,0 +1,135 @@
...
+      # new size meets all alignment requirements, but is still between
+      # min_size and max_size. This may change the partition's size (and
+      # region).
+      #
+      # @param resize_info [ResizeInfo]
+      def fix_end_alignment(resize_info)
+        return if partition.nil?
+
+        ptable = partition.partition_table
+        return if !ptable.require_end_alignment? || partition.end_aligned?
```
ancorgs: So we are aligning only if absolutely needed (DASD). Don't we want to align basically always than is possible?

src/lib/y2partitioner/actions/resize_partition.rb
```diff
@@ -0,0 +1,135 @@
...
+        grain_blocks = (ptable.align_grain.to_i / region.block_size.to_i)
+
+        partition.region = fix_region_end(region, min_blocks, max_blocks, grain_blocks)
+      end
+
+      # Make sure a region's end is between min_blocks and max_blocks. If it
+      # is not, add or subtract blocks in grain_blocks increments. All sizes
+      # are specified in that region's block size.
+      #
+      # @param region [Y2Storage::Region]
```
ancorgs: please document that this can be modified by the method

src/lib/y2partitioner/actions/resize_partition.rb
ancorgs: I'm asking because the partitioner already forces alignments of the partition starts. So if a user creates N partitions in a row using a big space and we align all the starts but none of the endings we are introducing gaps that make very little sense.

src/lib/y2partitioner/actions/resize_partition.rb
```diff
@@ -0,0 +1,135 @@
...
+      # @param min [Fixnum]
+      # @param max [Fixnum]
+      # @param grain [Fixnum]
```
ancorgs: Use Integer. Fixnum is deprecated.

src/lib/y2partitioner/actions/resize_partition.rb
```diff
@@ -0,0 +1,135 @@
...
+      #
+      # @return [Y2Storage::Region] adjusted region
+      def fix_region_end(region, min, max, grain)
+        if region.length < min
+          region.adjust_length(grain * ((min.to_f - region.length) / grain).ceil)
+        elsif region.length > max
+          region.adjust_length(grain * ((max.to_f - region.length) / grain).floor)
```
ancorgs: Sure this works?

Let's say the length of the region is 9, the max is 8 and the grain is 3. I would expect region to be adjusted by -3, so it's still aligned to a grain of 3 and is smaller than the max.

But `(8-9 / 3).floor` is 0. So the region would not be adjusted at all. 

It's Friday afternoon, so I could be completely wrong even in the most simple calculation, but...

joseivanlopez: I think it is correct:

```
(8.0 - 9) / 3 #=> -0.33333
-0.33333.floor #=> -1
```

ancorgs: Sure it's correct. As I said, Friday afternoon. :wink:

src/lib/y2partitioner/dialogs/partition_resize.rb
```diff
@@ -0,0 +1,321 @@
...
+        widgets = []
+        widgets << current_size_info
+        widgets << used_size_info unless space_info.nil?
+        VBox(*widgets)
+      end
+
+      # Widget for current size
+      def current_size_info
+        size = partition.size.to_human_string
+        Left(Label(format("Current size: %{size}", size: size)))
```
ancorgs: Maybe the usual comments for translators just in case they feel tempted to translate the `%{size}` part? The translators are more used to the classic '%s' than to named placeholders.

src/lib/y2partitioner/dialogs/partition_resize.rb
```diff
@@ -0,0 +1,321 @@
...
+      end
+
+      # Widget for used size
+      def used_size_info
+        size = used_size.to_human_string
+        Left(Label(format("Currently used: %{size}", size: size)))
```
ancorgs: Same than above about advising translators (and/or directly using `%s`)

src/lib/y2partitioner/widgets/device_resize_button.rb
```diff
@@ -35,13 +38,39 @@ def label
...
+        super
+      end
+
+      # Checks whether the device supports resizing
+      #
+      # @note An error popup is shown when the device does not support resizing.
+      #
+      # @return [Boolean] true if the device supports resizing; false otherwise.
+      def device_validation
+        return true if device.is?(:partition, :md)
```
ancorgs: Shouldn't we be checking here for software defined MDs instead MDs in general?

src/lib/y2partitioner/widgets/pages/disks.rb
```diff
@@ -63,9 +64,12 @@ def contents
               )
             ),
             table,
-            HBox(
-              # TODO: add and edit need to be also added
-              DeviceDeleteButton.new(pager: pager, table: table)
+            Left(
+              HBox(
+                # TODO: add, edit and move buttons need to be also added
```
ancorgs: Why not take the opportunity to add the missing buttons, as we are doing in most PBIs to avoid openQA headache.

Even more, add and edit should even work already. So the only dummy button we would be adding would be "move" (...which makes me think that maybe we should not add such button... let's see if nobody misses it).

src/lib/y2storage/partition_tables/base.rb
```diff
@@ -116,6 +116,16 @@ class Base < Device
       #   @return [Boolean] whether a partition can have this partition id.
       storage_forward :partition_id_supported?
 
+      # @!method align(region, align_policy = AlignPolicy::ALIGN_END, align_type = AlignType::OPTIMAL)
+      #   Align the region according to align policy and align type.
```
ancorgs: Does this return an new region that is aligned without modifying the passed one (ruby style) or does it modifies the passed argument (libstorage-ng style)? I would like to know from the documentation.

src/lib/y2partitioner/widgets/device_resize_button.rb
ancorgs: Never mind, I just realize it's not necessary since that button will never be displayed for BIOS defined RAIDs.

src/lib/y2partitioner/actions/resize_partition.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      # @return [Y2Storage::ResizeInfo]
+      attr_reader :resize_info
+
+      # Runs the dialog to resize the partition
+      #
+      # @return [Symbol] :finish if the dialog returns :next; dialog result otherwise.
+      def resize
+        result = Dialogs::PartitionResize.run(partition, resize_info)
+        fix_end_alignment(resize_info)
```
ancorgs: Why is `resize_info` passed as argument if we have an attr_reader for it. It looks inconsistent to what we do with `partition`.

src/lib/y2partitioner/widgets/device_resize_button.rb
ancorgs: Which makes me think. If resizing a partition (actually defining a new size for it) is a completely different action from resizing an MD or an VG (in fact, we don't define a new size but we add/remove devices to it). Why do we share the button class? Just because they have the same label?

src/lib/y2storage/partition.rb
```diff
@@ -126,6 +126,30 @@ def disk
...
+    # Whether the final sector of the partition is aligned according to
+    # the partition table grain.
+    #
+    # @return [Boolean]
+    def end_aligned?
+      grain = partition_table.align_grain
+      block_size = region.block_size
+      sector = region.end
+      overhead = (block_size * sector + block_size) % grain
+      overhead.zero?
```
ancorgs: I'm surprised that none of our cops complains about code duplication in `start_aligned?` end `end_aligned?`. Just saying :wink:

BTW, is `start_aligned?` really used of it's just defined for symmetry reasons?

test/y2partitioner/actions/resize_partition_test.rb
```diff
@@ -0,0 +1,186 @@
...
+  using Y2Storage::Refinements::SizeCasts
+
+  before do
+    allow(Yast::Wizard).to receive(:OpenNextBackDialog)
+    allow(Yast::Wizard).to receive(:CloseDialog)
+  end
+
+  subject(:action) { described_class.new(partition) }
+
+  context "With a mixed partition setup" do
```
ancorgs: Is this context really needed? It looks to me like it adds not useful information and its only effect is to indent everything one level.

test/y2partitioner/actions/resize_partition_test.rb
```diff
@@ -0,0 +1,186 @@
...
+          end
+
+          it "returns :abort" do
+            expect(action.run).to eq(:abort)
+          end
+        end
+      end
+    end
+
+    describe "#fix_region_end" do
```
ancorgs: A purist would tell you that you should test this using the public API instead of using `send` to call private methods. But I will not be that purist... today :wink:

test/y2partitioner/dialogs/partition_resize_test.rb
```diff
@@ -0,0 +1,256 @@
...
+  describe "#contents" do
+    def find_label(contents, text)
+      contents.nested_find do |widget|
+        widget.respond_to?(:value) &&
+          widget.value == :Label &&
+          widget.params.any? { |i| i.include?(text) }
+      end
+    end
+
+    it "contents a widget for selecting the new size" do
```
ancorgs: I guess you mean "it contains a widget". Content as a verb means to satisfy someone/something, so "it contents a widget" would mean it makes the widget happy.

src/lib/y2partitioner/widgets/device_resize_button.rb
joseivanlopez: Well, as commented face to face, the only one reason is to avoid to have 5 identical classes with only 1 different line.

test/y2partitioner/actions/resize_partition_test.rb
joseivanlopez: Yes, I agree.

src/lib/y2storage/partition.rb
joseivanlopez: `start_aligned?` is only used in tests.

src/lib/y2partitioner/widgets/pages/disks.rb
joseivanlopez: Well, we need to add some specific logic for those buttons. For example, when a partition is selected, the "add" button is meant to add a new partition over the device that the partition belongs to. And when a disk is selected, the "edit" button should jump to the disk page.

It is more or less easy to implement, but I would do it as a separate PR.

src/lib/y2storage/partition_tables/base.rb
joseivanlopez: Yes, it always returns a new object. Commented.

src/lib/y2partitioner/widgets/pages/disks.rb
ancorgs: Fine with me

src/lib/y2storage/partition.rb
```diff
@@ -126,27 +126,51 @@ def disk
...
+    def end
+      region.end
+    end
+
+    # Size of a single block
+    #
+    # @return [DiskSize]
+    def block_size
+      region.block_size
+    end
```
ancorgs: Shouldn't `#start`, `#end` and `#block_size` be defined in `BlkDevice` instead of this subclass? They apply to any region-based device.

joseivanlopez: I was sure that I was adding it to `BlkDevice` :S

PR #453, opened 2017-12-01
Fix boot partition detection

imobachgs:
It should fix [bsc#1070621](https://bugzilla.suse.com/show_bug.cgi?id=1070621)


imobachgs:
Thanks!

PR #454, opened 2017-12-01
Use the improved Device.compare_by_name from libstorage-ng 

ancorgs:
Based on https://github.com/openSUSE/libstorage-ng/pull/400

Replaces https://github.com/yast/yast-storage-ng/pull/434 and #432

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Dec  1 16:58:14 UTC 2017 - ancor@suse.com
+
+- Rely on the new improved mechanism of libstorage-ng to sort
+  devices by name. Preparation for bsc#1049901 and part of
+  fate#31896.
+- 4.0.47
+
+-------------------------------------------------------------------
 Fri Dec  1 15:34:11 UTC 2017 - igonzalezsosa@suse.com
 
 - Fix boot partition detection (bsc#1070621).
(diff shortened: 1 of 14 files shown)


ancorgs:
@aschnell yes

aschnell:
@ancorgs I tried now, but no.

ancorgs:
@aschnell Do you have a pull request or at least a diff with the changes in libstorage-ng, so I can check what breaks if yast-storage-ng?

ancorgs:
@aschnell Just to clarify, does "_the remaining sorting of result_" mean all the `WhateverDevice.all` methods that were previously sorted using `compare_by_name`? In that case I don't think there is anything left and I believe everything works already.

Or does it means any sorting from any result, including more "local" stuff like `PartitionTable#partitions` (used to be sorted by partition number), `LvmVg#lvm_lvs` (used to be sorted by lv name) and so on? I didn't understand we wanted to go as far as "libstorage-ng will never sort any result" as part of this PBI, but adapting yast2-storage-ng for that should be a matter of 5 minutes (already tested locally) if that's the intention.

aschnell:
The intension is indeed to not sort any return list/vectors anymore. To test this just remove the four sort calls in DevicegraphImpl.h (in the functions marked as to be deleted).

ancorgs:
@aschnell Got it. I didn't understand we wanted to go that far as part of the current PBI, but here it is https://github.com/yast/yast-storage-ng/pull/456. I tested it as you suggested and everything looks sane to me.

ancorgs:
@aschnell #456 is merged. Feel free to retry the sorting removal in the library.

PR #455, opened 2017-12-01
AutoYaST can disable snapshot even if partitions are not defined

imobachgs:
See [bsc#1070790](https://bugzilla.suse.com/show_bug.cgi?id=1070790).

### DIFF
--- a/src/lib/y2storage/autoinst_profile/drive_section.rb
+++ b/src/lib/y2storage/autoinst_profile/drive_section.rb
@@ -117,6 +117,7 @@ module Y2Storage
         super
         @type ||= default_type_for(hash)
         @use = use_value_from_string(hash["use"]) if hash["use"]
+        @enable_snapshots = true unless hash.key?("enable_snapshots")
         @partitions = partitions_from_hash(hash)
         @skip_list = SkipListSection.new_from_hashes(hash.fetch("skip_list", []), self)
       end
--- a/src/lib/y2storage/autoinst_profile/drive_section.rb
+++ b/src/lib/y2storage/autoinst_profile/drive_section.rb
@@ -117,7 +117,6 @@ module Y2Storage
         super
         @type ||= default_type_for(hash)
         @use = use_value_from_string(hash["use"]) if hash["use"]
-        @enable_snapshots = true unless hash.key?("enable_snapshots")
         @partitions = partitions_from_hash(hash)
         @skip_list = SkipListSection.new_from_hashes(hash.fetch("skip_list", []), self)
       end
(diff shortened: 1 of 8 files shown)


test/y2storage/autoinst_proposal_test.rb
```diff
@@ -519,6 +519,28 @@
...
+          proposal.propose
+        end
+      end
+
+      context "when snapshots are enabled in the profile" do
+        let(:partitioning) do
+          [{ "device" => "/dev/sda", "use" => use, "enable_snapshots" => true }]
+        end
+
+        it "enables use_snapshot setting" do
```
teclator: use_snapshots

test/y2storage/autoinst_proposal_test.rb
```diff
@@ -519,6 +519,28 @@
           expect(sda.partitions.size).to eq(3)
         end
       end
+
+      context "when snapshots are disabled in the profile" do
+        let(:partitioning) do
+          [{ "device" => "/dev/sda", "use" => use, "enable_snapshots" => false }]
+        end
+
+        it "disables use_snapshot setting" do
```
teclator: use_snapshots

src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -117,6 +117,7 @@ def init_from_hashes(hash)
         super
         @type ||= default_type_for(hash)
         @use = use_value_from_string(hash["use"]) if hash["use"]
+        @enable_snapshots = true unless hash.key?("enable_snapshots")
```
ancorgs: The documentation of this method says "_It only enforces default values for #type (:CT_DISK) and #use ("all") since the {AutoinstProposal} algorithm relies on them_". And indeed, IIRC the philosophy of these `init_from_hashes` methods of the `XXXSection` classes was not to set the default values in general. The `DriveSection` class is supposed to be a quite dumb class translating the hash-based format to something more useful, but turning nils into default values should not be part of this. Isn't it?

imobachgs: Ok, fine with me. I will change it.

src/lib/y2storage/proposal/autoinst_drives_map.rb
```diff
@@ -97,6 +97,13 @@ def partitions?
         @drives.values.any? { |i| !i.partitions.empty? }
       end
 
+      # Determine whether any of the drives sets snapshots enabled
+      #
+      # @return [Boolean]
+      def use_snapshots?
+        @drives.empty? || @drives.values.any?(&:enable_snapshots)
```
ancorgs: I believe that, to be consistent with the current philosophy, the decision about what to do if the field is nil in the profile should be taken here.

ancorgs: The `enable_snapshots` field, I meant.

PR #456, opened 2017-12-05
 Not longer trust any sorting coming from libstorage-ng 

ancorgs:
Last steps to implement https://trello.com/c/T8B7oRJs/769-yast2-storage-ng-do-not-rely-on-libstorage-ng-sorting-anymore

@aschnell announced its intention to drop any sorting from libstorage here https://lists.opensuse.org/yast-devel/2017-11/msg00009.html

As discussed in #454, that affects not only the collection of devices that used to be sorted by name (`Disks.all`, `Dasd.all`, etc.) but also collections within an object that used to be sorted by other criteria (partitions within a partition table will not be longer sorted by number, LVs within an VG will not be longer sorted by name, etc.)

This should prepare yast-storage-ng for such change. Manually tested by removing all the calls to `sort` in `DevicegraphImpl.h` in libstorage-ng.

### DIFF
--- a/src/lib/y2storage/partitionable.rb
+++ b/src/lib/y2storage/partitionable.rb
@@ -263,7 +263,7 @@ module Y2Storage
     def partitions_with_id(*ids)
       # Sorting is not mandatory, but keeping the output stable looks like a
       # sane practice.
-      partitions.sort_by(&:number).reject { |p| p.type.is?(:extended) }.select { |p| p.id.is?(*ids) }
+      partitions.reject { |p| p.type.is?(:extended) }.select { |p| p.id.is?(*ids) }.sort_by(&:number)
     end
   end
 end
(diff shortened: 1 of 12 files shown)


src/lib/y2storage/partitionable.rb
```diff
@@ -259,7 +261,9 @@ def ensure_partition_table
     #
     # @return [Array<Partition>}]
     def partitions_with_id(*ids)
-      partitions.reject { |p| p.type.is?(:extended) }.select { |p| p.id.is?(*ids) }
+      # Sorting is not mandatory, but keeping the output stable looks like a
+      # sane practice.
+      partitions.sort_by(&:number).reject { |p| p.type.is?(:extended) }.select { |p| p.id.is?(*ids) }
```
imobachgs: I would place the call to `sort_by` at the end for two reasons: 1) be consistent with L169 and 2) sort a potentially smaller collection (after reject and select).

ancorgs:
BTW, this can be considered like a follow up on https://github.com/yast/yast-storage-ng/pull/428 in which `sorted_by_name` was introduced and used for all the `Partitionable` subclasses (thanks to https://github.com/openSUSE/libstorage-ng/pull/400 now it's also available for other `BlkDevice` classes and for `LvmVg`).

PR #457, opened 2017-12-05
Fix ESP and other partition_id conversions

imobachgs:
DIAG is, AFAIS, in the same situation.

### DIFF
--- a/src/lib/y2storage/partition_id.rb
+++ b/src/lib/y2storage/partition_id.rb
@@ -76,7 +76,7 @@ module Y2Storage
     # Partition ids for which the internal numeric id is the same than the
     # corresponding fsid in the old libstorage.
     # See {.new_from_legacy} and {#to_i_legacy}.
-    LEGACY_KEPT = [DOS12, DOS16, DOS32, NTFS, EXTENDED, PREP, LINUX, SWAP, LVM, RAID]
+    LEGACY_KEPT = [DOS12, DOS16, DOS32, NTFS, EXTENDED, PREP, LINUX, SWAP, LVM, RAID, ESP]
 
     # Matching between fsids in the old libstorage and the corresponding
     # partition id.
--- a/src/lib/y2storage/partition_id.rb
+++ b/src/lib/y2storage/partition_id.rb
@@ -76,16 +76,20 @@ module Y2Storage
     # Partition ids for which the internal numeric id is the same than the
     # corresponding fsid in the old libstorage.
     # See {.new_from_legacy} and {#to_i_legacy}.
-    LEGACY_KEPT = [DOS12, DOS16, DOS32, NTFS, EXTENDED, PREP, LINUX, SWAP, LVM, RAID, ESP]
+    LEGACY_KEPT = [DOS12, DOS16, DOS32, NTFS, EXTENDED, PREP, LINUX, SWAP, LVM, RAID, DIAG, ESP]
 
     # Matching between fsids in the old libstorage and the corresponding
     # partition id.
     # See {.new_from_legacy} and {#to_i_legacy}.
     LEGACY_TO_CURRENT = {
+      4   => DOS16, # Known as "FAT16 <32M"
       5   => EXTENDED, # In the past both 5 and 15 were recognized as extended
+      11  => DOS32, # Known as "Win95 FAT32" as an alternative to 0x0c (Win95 FAT32 LBA)
+      14  => DOS16, # Known as "Win95 FAT16" as an alternative to 0x06 (FAT16)
       257 => UNKNOWN, # 257 used to mean mac_hidden, but is BIOS_BOOT now
       258 => UNKNOWN, # 258 used to mean mac_hfs, but is WINDOWS_BASIC_DATA now
-      259 => BIOS_BOOT, # 259 is MICROSOFT_RESERVED now
+      259 => ESP, # 259 is MICROSOFT_RESERVED now
+      261 => MICROSOFT_RESERVED,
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/partition_id.rb
```diff
@@ -76,7 +76,7 @@ def formattable?
     # Partition ids for which the internal numeric id is the same than the
     # corresponding fsid in the old libstorage.
     # See {.new_from_legacy} and {#to_i_legacy}.
-    LEGACY_KEPT = [DOS12, DOS16, DOS32, NTFS, EXTENDED, PREP, LINUX, SWAP, LVM, RAID]
+    LEGACY_KEPT = [DOS12, DOS16, DOS32, NTFS, EXTENDED, PREP, LINUX, SWAP, LVM, RAID, ESP]
```
ancorgs: Hmm, originally I was aware of all the fsids defined at https://github.com/yast/yast-storage/blob/master/src/modules/Partitions.rb#L50 which doesn't include ESP. But I just found there were other supported ids that were not declared but, apparently, were handled in some situations. Here https://github.com/yast/yast-storage/blob/master/src/modules/Partitions.rb#L689

So give me 10 minutes to check what else is missing.

ancorgs: After a closer look (way more than 10 minutes), I believe we should change this:

```diff
-    LEGACY_KEPT = [DOS12, DOS16, DOS32, NTFS, EXTENDED, PREP, LINUX, SWAP, LVM, RAID]
+    LEGACY_KEPT = [DOS12, DOS16, DOS32, NTFS, EXTENDED, PREP, LINUX, SWAP, LVM, RAID, DIAG, ESP]

     LEGACY_TO_CURRENT = { 
+      4   => DOS16, # Known as "FAT16 <32M"
       5   => EXTENDED, # In the past both 5 and 15 were recognized as extended
+      11  => DOS32, # Known as "Win95 FAT32" as an alternative to 0x0c (Win95 FAT32 LBA)
+      14  => DOS16, # Known as "Win95 FAT16 as an alternative to 0x06 (FAT16)
       257 => UNKNOWN, # 257 used to mean mac_hidden, but is BIOS_BOOT now 
       258 => UNKNOWN, # 258 used to mean mac_hfs, but is WINDOWS_BASIC_DATA now 
-      259 => BIOS_BOOT, # 259 is MICROSOFT_RESERVED now
+      259 => ESP, # 259 is MICROSOFT_RESERVED now
+      261 => MICROSOFT_RESERVED,
       263 => BIOS_BOOT,
       264 => PREP
     }   

     CURRENT_TO_LEGACY = { 
-      BIOS_BOOT.to_i          => 259, # BIOS_BOOT.to_i is 257, that used to mean mac_hidden
+      BIOS_BOOT.to_i          => 263, # BIOS_BOOT.to_i is 257, that used to mean mac_hidden
+      ESP.to_i                => 259, # ESP.to_i is 239, that used to have no special meaning
       WINDOWS_BASIC_DATA.to_i => 0, # WINDOWS_BASIC_DATA.to_i is 258, that used to mean mac_hfs
-      MICROSOFT_RESERVED.to_i => 0, # MICROSOFT_RESERVED.to_i is 259, that used to mean BIOS_BOOT
+      MICROSOFT_RESERVED.to_i => 261, # MICROSOFT_RESERVED.to_i is 259, that used to mean BIOS_BOOT
     }
```

ancorgs: An additional question comes to mind. If we receive an id that cannot be handled by libstorage-ng we currently return `UNKNOWN`. It has the advantage of failing in a very obvious way (libstorage-ng raises an exception)... but it has the disadvantage of failing in a very obvious way. :-)

So the usual question, should we fallback to `LINUX` (optionally reporting the error) and continue (the show must go on) or it's fine to use `UNKNOWN` causing the libstorage-ng exception?

Anyway, in the first case this class wouldn't be the place for the fallback, so I'm talking about unrelated stuff here. :-)

src/lib/y2storage/partition_id.rb
```diff
@@ -94,9 +98,10 @@ def formattable?
     # them in the old libstorage.
     # See {.new_from_legacy} and {#to_i_legacy}.
     CURRENT_TO_LEGACY = {
-      BIOS_BOOT.to_i          => 259, # BIOS_BOOT.to_i is 257, that used to mean mac_hidden
+      BIOS_BOOT.to_i          => 263, # BIOS_BOOT.to_i is 257, that used to mean mac_hidden
+      ESP.to_i                => 259, # ESP.to_i is 239, that used to have no special meaning
       WINDOWS_BASIC_DATA.to_i => 0, # WINDOWS_BASIC_DATA.to_i is 258, that used to mean mac_hfs
```
ancorgs: I'm actually kind of worried about `WINDOWS_BASIC_DATA`. This is the type used by Windows in GPT. Exporting it as Unknown instead of 258 may be a way to avoid confusions with the old mac_hfs fsid. But it's a recipe for disaster for the future. Let me think it a little bit better. :-)

ancorgs: @imobachgs Let me do some experiments with a Windows installation in GPT and the old yast-storage

ancorgs: So far, I can't export such partition with AutoYaST. But the old expert partitioner displays it as 0x07. So I guess we need more complex logic in the new AutoYaST to make the `NTFS <=> WINDOWS_BASIC_DATA` conversion based on the type of partition table. fsid really mixed a little bit of everything :-)

PR #458, opened 2017-12-05
Add buttons to Hard Disks view

joseivanlopez:
PBI: https://trello.com/c/kE0SSbi5/698-0-partitioner-plain-partitioning-hard-disks-overview-add-edit-buttons

### DIFF
--- a/src/lib/y2partitioner/widgets/blk_device_edit_button.rb
+++ b/src/lib/y2partitioner/widgets/blk_device_edit_button.rb
@@ -29,21 +29,52 @@ Yast.import "Popup"
 
 module Y2Partitioner
   module Widgets
-    # Button for opening the editing workflow (basically mount and format
-    # options) on a block device.
+    # Button for editing a block device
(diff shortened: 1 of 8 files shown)


src/lib/y2partitioner/widgets/blk_device_edit_button.rb
```diff
@@ -29,21 +29,52 @@
...
+      # @return [Booelan]
+      def partition?
+        device.is?(:partition)
+      end
+
+      # Whether the device is a software raid
+      #
+      # @return [Booelan]
+      def software_raid?
+        device.is?(:md) && device.software_defined?
```
ancorgs: Probably not 100% related, but this makes me think that `Md` and `Dm` objects should return true to `device.is?(:raid)`. Just thinking aloud.

joseivanlopez: Yes, it makes sense.

PR #459, opened 2017-12-07
Mount /mnt/run during installation

imobachgs:
Fixes installation part of [bsc#1071745](https://bugzilla.suse.com/show_bug.cgi?id=1071745).


imobachgs:
I do not think using memory is a good idea after all. Maybe it would a better idea to just clean-up the files after installation/upgrade.

Issue #463, opened 2017-12-13
Nasty typo in translated string "in invalid" must be "is invalid"


sriemer:
https://github.com/yast/yast-storage-ng/blob/27656b8f1c75654d9be54b08ed9b0597692690c3/src/lib/y2partitioner/actions/controllers/lvm_vg.rb#L267

Here is the typo. Please change this to "is invalid".

Found while checking:
https://l10n.opensuse.org/translate/yast-storage/master/de/?type=untranslated

Please fix this. TIA

sriemer:
The typo has been introduced by commit 02c83d1a829f8bd77f7c3bd93dc1b0e1a9ddd4cf.

aschnell:
Thanks, fixed in https://github.com/yast/yast-storage-ng/pull/464.

PR #464, opened 2017-12-13
- fixed typo

aschnell:
For https://github.com/yast/yast-storage-ng/issues/463.


ancorgs:
I guess we will get tons of these, now that the translators are looking into it.

ancorgs:
Wow! We got a unit test checking for an error containing the substring "data entered in invalid". So a test ensuring the bug was not fixed :(

PR #466, opened 2017-12-14
Don't rely on partitions order in the tests

ancorgs:


joseivanlopez:
Umm, travis is complaining here https://travis-ci.org/yast/yast-storage-ng/builds/316390559#L8012

PR #468, opened 2017-12-15
Add storage checks

joseivanlopez:
PBI: https://trello.com/c/PxAc8OlA/766-5-expert-partitioner-add-sanity-checks


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Jan 15 08:22:32 UTC 2018 - jlopez@suse.com
+
+- Added sanity checks for partitioning setup.
```
ancorgs: NP: I would mention the expert partitioner here (and maybe also the new control.xml format somehow).

src/lib/y2storage/planned/md.rb
```diff
@@ -94,6 +96,20 @@ def add_devices(md_device, devices)
...
+
+    protected
+
+      # Values for volume specification matching
+      #
+      # @see MatchVolumeSpec
+      def volume_match_values
+        {
+          mount_point:  mount_point,
+          size:         DiskSize.zero,
```
ancorgs: Why zero?

src/lib/y2storage/md.rb
```diff
@@ -288,7 +288,11 @@ def md_users
     end
 
     def types_for_is
-      super << :md
+      types = super
+      types << :md
+      types << :raid if software_defined?
```
ancorgs: I guess this `if software_defined?` is here to exclude MD containers (in combination with the `types << :raid` on `MdMember` below). It definitely needs some comment because it's a really tricky implementation.

src/lib/y2storage/boot_requirements_strategies/zipl.rb
```diff
@@ -48,12 +64,34 @@ def zipl_partition_missing?
         free_mountpoint?("/boot/zipl")
       end
 
+      # @return [VolumeSpecification]
+      def zipl_volume
+        return @zipl_volume unless @zipl_volume.nil?
+
+        @zipl_volume = VolumeSpecification.new({})
+        @zipl_volume.mount_point = "/boot/zipl"
+        @zipl_volume.fs_types = [Filesystems::Type::EXT2]
```
ancorgs: I guess any EXT{2..4} is supported here. Somebody should know :-)

src/lib/y2storage/boot_requirements_strategies/zipl.rb
```diff
@@ -48,12 +64,34 @@ def zipl_partition_missing?
...
+        planned_partition.disk = boot_disk.name
+        planned_partition
+      end
+
+      # Specific error when the boot disk is not valid for booting
+      #
+      # @return [SetupError]
+      def unsupported_boot_disk_error
+        # TRANSLATORS: error message
+        error_message = _("Current boot disk cannot be used for booting")
```
ancorgs: We need a better message here, specially taking into account that this (as all boot-related errors) could be a false positive.

Something like "Looks like the system is going to be installed on a FBA or LDL device. Booting from such device is not supported". And yes, I know the message is not agnostic enough if we change the conditions in `supported_boot_disk?` but sometimes you need to duplicate that to have human-readable messages.

src/lib/y2storage/setup_checker.rb
```diff
@@ -0,0 +1,138 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2storage"
```
ancorgs: Is not this some kind of circular requirement?

src/lib/y2storage/setup_error.rb
```diff
@@ -0,0 +1,228 @@
...
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "yast/i18n"
+require "y2storage"
```
ancorgs: Same here. Looks a bit circular to me. Really needed?

src/lib/y2storage/setup_error.rb
```diff
@@ -0,0 +1,228 @@
...
+      elsif fs_types_info
+        message_with_fs
+      else
+        message_without_mount_point_default
+      end
+    end
+
+    # @return [String]
+    def message_with_mount_point_and_partition_id_and_fs
+      # TRANSLATORS: error message, where %{mount_point} is repalaced by a mount point
```
ancorgs: typo: repalaced... unless it means we build a new palace for it.

joseivanlopez: :-)

src/lib/y2partitioner/setup_errors_presenter.rb
```diff
@@ -0,0 +1,93 @@
...
+    # @return [SetupChecker] checker for the current setup
+    attr_reader :setup_checker
+
+    # HTML representation for boot errors
+    #
+    # @return [String]
+    def boot_errors_html
+      errors = setup_checker.boot_errors
+      # TRANSLATORS
+      header = _("The system could not load because the following errors were found:\n")
```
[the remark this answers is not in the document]
joseivanlopez: Any suggestion? Probably something like "It is possible the system does not boot properly for the following reason:"

src/lib/y2partitioner/setup_errors_presenter.rb
```diff
@@ -0,0 +1,93 @@
...
+    # HTML representation for a set of errors
+    #
+    # @note The representation is composed by a header message and the list error messages.
+    #
+    # @param header [String] header text
+    # @param errors [Array<SetupError>] list of errors
+    #
+    # @return [String]
+    def errors_html(header, errors)
+      return nil if errors.empty?
```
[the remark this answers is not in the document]
joseivanlopez: Yes, probably it is better to always return a valid html representation, even when there is no error.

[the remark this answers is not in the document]
joseivanlopez: I mean for the `#to_html` public method. For those private ones probably it is easier when them return nil.

src/lib/y2storage/boot_requirements_strategies/zipl.rb
joseivanlopez: I seems there is no guarantee about that.

src/lib/y2storage/planned/md.rb
joseivanlopez: Probably it does not make sense. The problem is that Planned::Md has not a size and I set it to zero as the equivalent nil value for a DiskSize. I will change it to actually use nil and it will be taken into account when matching size.

package/yast2-storage-ng.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:        4.0.68
+Version:        4.0.70
```
ancorgs: Why are you jumping two numbers here?

package/yast2-storage-ng.changes
```diff
@@ -6,6 +14,7 @@ Mon Jan 15 12:51:01 UTC 2018 - ancor@suse.com
   alignment types, in addition to the optimal one.
 - Added a (temporary) workaround to a possible bug in libstorage-ng
   regarding alignment.
+- 4.0.69
```
ancorgs: This was not intended to introduce a version bump, no need to "correct" that

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Mon Jan 15 14:22:32 UTC 2018 - jlopez@suse.com
+
+- Added sanity checks for partitioning setup.
+- Partitioner: setup issues are shown to the user before continue.
+  Mandatory product volumes are required according to control file.
```
ancorgs: I miss the storage-ng fate entry and also mentioning the several bug reports we have about the missing checks.

test/support/proposed_partitions_examples.rb
```diff
@@ -69,7 +69,6 @@
     it "requires it to be between 1 and 8MiB, despite the alignment" do
       expect(grub_part.min).to eq 1.MiB
       expect(grub_part.max).to eq 8.MiB
-      expect(grub_part.align).to eq :keep_size
```
ancorgs: Let's remove also the "despite the alignment" part from the textual description

test/support/proposed_partitions_examples.rb
```diff
@@ -79,7 +78,6 @@
     it "requires it to be between 256KiB and 8MiB, despite the alignment" do
       expect(grub_part.min).to eq 256.KiB
       expect(grub_part.max).to eq 8.MiB
-      expect(grub_part.align).to eq :keep_size
```
ancorgs: Same here for "despite the alignment"

test/support/proposed_partitions_examples.rb
```diff
@@ -138,7 +136,6 @@
     it "requires it to be between 1MiB and 8MiB, despite the alignment" do
       expect(prep_part.min).to eq 1.MiB
       expect(prep_part.max).to eq 8.MiB
-      expect(prep_part.align).to eq :keep_size
```
ancorgs: Same here for "despite the alignment"

test/support/proposed_partitions_examples.rb
```diff
@@ -148,7 +145,6 @@
     it "requires it to be between 256KiB and 8MiB, despite the alignment" do
       expect(prep_part.min).to eq 256.KiB
       expect(prep_part.max).to eq 8.MiB
-      expect(prep_part.align).to eq :keep_size
```
ancorgs: Same here for "despite the alignment"

src/lib/y2storage/md.rb
```diff
@@ -287,6 +287,11 @@ def md_users
       to_storage_value.in_holders.to_a.map { |h| Storage.to_md_user(h) }
     end
 
+    # All Software RAIDs are Md devices, but MdMember derived class represents
+    # BIOS RAIDs. Class MdContainer is also derived from Md, but objects of this
+    # class are not be considered neither Software nor BIOS RAIDs.
+    # To properly exclude MdContainer as :raid device, here only Software RAIDs
+    # are considered to be raid type.
```
ancorgs: NP: Wouldn't be too much to ask to have this explanation right on top of the `types << :raid if software_defined?` line instead of the heading of the method?

package/yast2-storage-ng.spec
joseivanlopez: Arfff

src/lib/y2storage/planned/md.rb
```diff
@@ -105,7 +105,7 @@ def self.to_string_attrs
       def volume_match_values
         {
           mount_point:  mount_point,
-          size:         DiskSize.zero,
+          size:         nil,
```
ancorgs: Why don't we want to check the size of the Md devices? Doesn't `Md#size` return a sensible value just like e.g. `Partition#size`?

src/lib/y2storage/filesystems/type.rb
```diff
@@ -152,6 +154,17 @@ def self.legacy_home_filesystems
...
       end
 
+      # Allowed filesystems for zipl boot partition
+      #
+      # @note See page 13 in following link
+      # https://share.confex.com/share/123/webprogram/Handout/\
+      #   Session15694/SHARE_Bootloader_Ihno_PittsPPT_0.09.pdf
+      #
+      # @return [Array<Filesystems::Type>]
+      def self.zipl_filesystems
```
ancorgs: It should be mentioned that the first one is the preferred one (i.e. used by the guided setup)

src/lib/y2storage/boot_requirements_strategies/zipl.rb
```diff
@@ -90,7 +90,10 @@ def zipl_partition(target)
       # @return [SetupError]
       def unsupported_boot_disk_error
         # TRANSLATORS: error message
-        error_message = _("Current boot disk cannot be used for booting")
+        error_message = _(
+          "Looks like the system is going to be installed on a FBA " \
+          "or LDL device. Booting from such device is not supported"
```
ancorgs: Missing full period at the end?

src/lib/y2partitioner/widgets/partition_description.rb
```diff
@@ -60,7 +60,7 @@ def device_attributes_list
           device_udev_by_path.join(Yast::HTML.Newline),
           device_udev_by_id.join(Yast::HTML.Newline),
           # TRANSLATORS: acronym for Filesystem Identifier
-          format(_("FS ID: %s"), "TODO")
+          format(_("Partition ID: %s"), "TODO")
```
ancorgs: Please update the comment for the translators

src/lib/y2storage/planned/md.rb
joseivanlopez: It is a `Planned::Md` and it has not `#size`.

ancorgs:
LGTM, let's go for it!

PR #469, opened 2017-12-15
SpaceMaker: less STUPID caching of @distribution

ancorgs:
DO NOT MERGE. Just for discussing purposes about https://trello.com/c/NTSIjyPk/788-3-osdistribution-p2-1066386-partitioner-tries-to-delete-windows


ancorgs:
@wfeldt This outlines a very conservative fix (changing almost no code) for the problem you pointed about caching values for different calls in the same variable.

PR #470, opened 2017-12-18
AutoYaST: Improve disklabel handling

imobachgs:
* Fixes [bsc#1073307](https://bugzilla.suse.com/show_bug.cgi?id=1073307)

## Expected behavior

* When the disk does not have a partition table:
  * If disklabel is defined in the profile -> it will use the given value.
  * If disklabel is not defined in the profile -> it will fallback to "msdos" (like in the past).

* When the disk has a partition table:
  * If the type defined in the profile (disklabel) is different -> change only if there are no partitions.
  * If the type matches -> do nothing.
  * If initialize is set to true -> always wipe the partition table

### DIFF
--- a/src/lib/y2storage/autoinst_proposal.rb
+++ b/src/lib/y2storage/autoinst_proposal.rb
@@ -129,7 +129,7 @@ module Y2Storage
       end
 
       disk_ptable_type = disk.partition_table ? disk.partition_table.type : nil
-      ptable_type || disk_ptable_type || Y2Storage::PartitionTables::Type::MSDOS
+      ptable_type || disk_ptable_type || disk.preferred_ptable_type
     end
 
     # Update partition table
--- a/src/lib/y2storage/autoinst_proposal.rb
+++ b/src/lib/y2storage/autoinst_proposal.rb
@@ -101,7 +101,9 @@ module Y2Storage
 
     # Add partition tables
     #
-    # This method create partitions table for disks which are missing one.
+    # This method create/change partitions tables according to information
+    # specified in the profile. Disks containing any partition will be ignored.
+    #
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -101,24 +101,48 @@ def propose_devicegraph(devicegraph, drives)
...
+    # @param drive_spec  [Y2Storage::AutoinstProfile::DriveSection] Drive section from the profile
+    # @return [Y2Storage::PartitionTables::Type] Partition table type
+    def suitable_ptable_type(disk, drive_spec)
+      ptable_type = nil
+      if drive_spec.disklabel
+        ptable_type = Y2Storage::PartitionTables::Type.find(drive_spec.disklabel)
       end
+
+      disk_ptable_type = disk.partition_table ? disk.partition_table.type : nil
+      ptable_type || disk_ptable_type || Y2Storage::PartitionTables::Type::MSDOS
```
joseivanlopez: Only for curiosity, what happens when we are using UEFI or DASD? Does it fallback to msdos? In that case, probably it fails when tries to create the partition table.

joseivanlopez: In fact, we have a method `Partitionable#preferred_ptable_type`

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/partitionable.rb#L246

imobachgs: Oh, I didn't new about that method. Maybe AutoYaST should rely on it. Thanks!

imobachgs:
Thanks!

PR #471, opened 2017-12-19
Add find by name

jreidinger:


jreidinger:
travis and friends are in good mood, so lets merge it.

PR #472, opened 2017-12-19
rewrite SpaceMaker::resize_and_delete! to be more human-readable

wfeldt:
This also includes making the exit criterion explicit even if this makes the function code so long rubocop complains.

### DIFF
--- a/src/lib/y2storage/proposal/space_maker.rb
+++ b/src/lib/y2storage/proposal/space_maker.rb
@@ -84,15 +84,32 @@ module Y2Storage
           partitions.delete(part)
         end
 
-        # To make sure we are not freeing space in a useless place, let's
-        # first restrict the operations to disks with particular requirements
+        # To make sure we are not freeing space in useless places first
+        # restrict the operations to disks with particular disk
+        # requirements.
+        #
+        # planned_partitions_by_disk() returns all partitions restricted to
+        # a certain disk. Most times partitions are free to be created
+        # anywhere but sometimes it is known in advance on which disk they
+        # should be created.
+        #
+        # Start by assigning space to them.
+        #
+        # The result (if successful) is kept in @distribution.
+        #
         planned_partitions_by_disk(partitions).each do |disk, parts|
           resize_and_delete!(parts, keep, lvm_helper, disk: disk)
         end
+
         # Doing something similar for #max_start_offset is more difficult and
         # doesn't pay off (#max_start_offset is used just in one case)
 
-        # Now with the full set of planned partitions and all the candidate disks
+        # Now repeat the process with the full set of planned partitions and all the candidate
+        # disks.
+        #
+        # Note that the result of the run above is not lost as already
+        # assigned partitions are taken into account.
+        #
         resize_and_delete!(partitions, keep, lvm_helper)
 
         {
@@ -155,8 +172,11 @@ module Y2Storage
       #
       # @return [Boolean]
       def success?(planned_partitions)
-        spaces = free_spaces(new_graph)
-        @distribution ||= dist_calculator.best_distribution(planned_partitions, spaces)
+        # Once a distribution has been found we don't have to look for another one.
+        if !@distribution
+          spaces = free_spaces(new_graph)
+          @distribution = dist_calculator.best_distribution(planned_partitions, spaces)
+        end
         !!@distribution
       rescue Error => e
         log.info "Exception while trying to distribute partitions: #{e}"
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -413,13 +470,14 @@ def delete_disk_content(planned_partitions, lvm_helper, disk)
...
 
         disks_for(new_graph, disk).each do |dsk|
-          break if success?(planned_partitions)
           log.info "Checking if the disk #{dsk.name} has a partition table"
 
-          if dsk.has_children? && dsk.partition_table.nil?
-            log.info "Found something that is not a partition table"
-            remove_content(dsk, lvm_helper)
-          end
+          next unless dsk.has_children? && dsk.partition_table.nil?
```
wfeldt: that's only for rubocop's pleasure

jreidinger: well, it makes some sense, to show only what is interesting, so early skip non-interesting parts. Maybe I just will write it like
```
next unless dsk.has_children?
next if dsk.partition_table
# so we have children but no partition table
```

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -413,13 +470,14 @@ def delete_disk_content(planned_partitions, lvm_helper, disk)
...
+
+          log.info "Found something that is not a partition table"
+          remove_content(dsk, lvm_helper)
+
+          break if success?(planned_partitions)
```
jreidinger: btw, now this success is called only when content is removed. Is it intention?

src/lib/y2storage/proposal/space_maker.rb
wfeldt: Well, **I** would have kept it the way it was. rubocop only started complaining due to a change someplace else.

src/lib/y2storage/proposal/space_maker.rb
wfeldt: yes, only when we changed something

jreidinger: ok, I just want to be sure.

src/lib/y2storage/proposal/space_maker.rb
jreidinger: it complain, because success was moved. So now only problematic disk is processed and rest can be early quit ( theory say that early quits in code safe some brain processing, but in reality it helps only sometimes and sometimes make it worse ).

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -173,24 +193,63 @@ def success?(planned_partitions)
...
 
+        # restart evaluation
         @distribution = nil
 
-        # Initially, resize only if there are no Linux partitions in the disk
+        # maybe it works already...
+        return if success?(planned_partitions)
+
+        # Try various methods to free space. Stop when a valid partition
+        # layout has been found.
```
ancorgs: If we slightly change the arguments order for `delete_partitions!` and `delete_disk_content`, we can write this part like this. That should please both rubocop and the human readers (as a human, I even prefer it).

```ruby
# Try various methods to free space. Stop when a valid partition
# layout has been found.
methods_and_args = [
  # step 1 - resize Windows partitions only if there are no Linux partitions on the disk
  [:resize_windows!,     [{ force: false }]],
  # step 2 - # delete Linux partitions
  [:delete_partitions!,  [:linux, keep]], 
  # step 3 - delete other (non-Windows) partitions
  [:delete_partitions!,  [:other, keep]], 
  # step 4 - resize Windows partitions
  [:resize_windows!,     [{ force: true }]],
  # step 5 - delete Windows partitions
  [:delete_partitions!,  [:windows, keep]], 
  # step 6 - if deleting partitions was not enough, maybe there is no
  # partition table and we have to wipe the disk
  [:delete_disk_content, [lvm_helper]]
]

methods_and_args.each_with_index do |method_and_args, idx|
  method, extra_args = method_and_args

  log.info "Step #{idx + 1} - #{method}"
  send(method, planned_partitions, disk, *extra_args)

  break if @distribution
end

raise Error unless @distribution
```

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -173,23 +193,59 @@ def success?(planned_partitions)
...
+          # step 2 - # delete Linux partitions
+          [:delete_partitions!,  [:linux, keep]],
+          # step 3 - delete other (non-Windows) partitions
+          [:delete_partitions!,  [:other, keep]],
+          # step 4 - resize Windows partitions
+          #
+          # ** Note **
+          #
+          # There are two steps where we try to resize a Windows partition.
+          # Both should be mutually exclusive. But the 'force' argument
```
ancorgs: I don't get why this is a problem and why both must be mutually exclusive.

ancorgs:
Code-wise it LGTM. I simply fail to see why changing the desired size of the windows partition is a problem. The comment states that is wrong and that looks too alarmist to me.

src/lib/y2storage/proposal/space_maker.rb
wfeldt: There's no point in resizing the same partition twice. This led (in the broken code) to the strange situation where the second step tried to enlarge windows. In general, do the resize before or after deleting partitions, but not twice. I would have actually expected this force parameter to guarantee exactly this - but it doesn't.
I've added this comment to remind future developers what to expect here.

ancorgs: But there is no resize operation happening twice in the disk or nothing like that. We are just changing the size of the windows region in the devicegraph in memory. As long as we don't call commit, I still don't see the problem with that.

Strictly speaking, it was not enlarging windows, it was just deciding that there was no need to reduce it as much as originally thought. That's not wrong. There is nothing like a real shrink and a real enlarge happening in the disk.

PR #473, opened 2017-12-19
BlkDevice.find_by_any_name and improved documentation 

ancorgs:
This is part of https://trello.com/c/lOQcX484/806-bug-1073254-storage-ng-staginge-fails-to-test-upgrade-a-131-installation-to-tw

Depends on this other PR being merged before: https://github.com/openSUSE/libstorage-ng/pull/414

### DIFF
--- a/test/y2storage/blk_device_test.rb
+++ b/test/y2storage/blk_device_test.rb
@@ -637,4 +637,65 @@ describe Y2Storage::BlkDevice do
       end
     end
   end
+
+  # This is mostly a direct forward to libstorage-ng, but this test is here
+  # because Storage::BlkDevice#find_by_any_name does not follow the usual
+  # libstorage-ng convention for exceptions.
+  describe "find_by_any_name" do
+    # The libstorage-ng counterpart of this method performs a real lookup in the
+    # system, so this can hardly be tested without a lot of mocking.
+    let(:storage_class) { Storage::BlkDevice }
+    let(:name) { "/dev/something" }
+
+    context "when libstorage-ng returns an object" do
+      let(:storage_device) do
+        Storage::BlkDevice.find_by_name(fake_devicegraph.to_storage_value, "/dev/sda")
+      end
+
+      it "returns the Y2Storage wrapped version of that object" do
+        expect(storage_class).to receive(:find_by_any_name).with(fake_devicegraph, name)
+          .and_return(storage_device)
+
+        result = described_class.find_by_any_name(fake_devicegraph, name)
+        expect(result).to be_a described_class
+        expect(result.to_storage_value).to eq storage_device
+      end
+    end
+
+    context "when libstorage-ng throws a DeviceNotFoundByName exception" do
+      before do
+        allow(storage_class).to receive(:find_by_any_name) do
+          raise Storage::DeviceNotFoundByName.new("A libstorage-ng error")
+        end
+      end
+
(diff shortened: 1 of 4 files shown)


test/y2storage/blk_device_test.rb
```diff
@@ -637,4 +637,65 @@
       end
     end
   end
+
+  # This is mostly a direct forward to libstorage-ng, but this test is here
+  # because Storage::BlkDevice#find_by_any_name does not follow the usual
+  # libstorage-ng convention for exceptions.
```
aschnell: What usual convention and why does it not apply here?

ancorgs: Raising `Storage::DeviceHasWrongType` or `Storage::DeviceNotFound` when looking for something that is not there. Raising `Storage::Exception` is kind of a... well... exception. :-)

aschnell: ```Storage::Exception``` is raised when the devicegraph is not the probed devicegraph.

ancorgs: Ah, ok. That was not clear to me from the libstorage-ng doc. I will update this.

ancorgs: @aschnell I have one question about that restriction of being only used in the probed devicegraph. And I guess @jreidinger opinion is useful here.

Take a look at this code currently present in yast2-bootloader:
```ruby
devices = Y2Storage::BlkDevice.all(staging)
# As wires devices have identical udev devices as its multipath device,
# we remove possible multipath wires from the list since we want to translate
# udev device to multipath device and not wire
devices.reject! { |d| d.is?(:disk) && d.descendants.any? { |i| i.is?(:multipath) } }
device = devices.find { |i| i.udev_full_all.include?(dev) }
return device.name if device
```
Most of it could be basically replaced by a call to `find_by_any_name` so it would work for all device names (right now it fail for virtio-, like it happened to yast2-update). BUT that code in yast2-bootloader is executed at the end of installation, after the libstorage commit.

Does still make sense in that moment to enforce the usage of probed? At the end of installation, the set of block devices in probed may not match the system (and on the other hand, `staging` should match).

A second and less important question, what about that issue with the udev names of multipath and their wires? Is handled by `find_by_any_name`?

ancorgs: For the update case, we can implement a hack in which yast2-update (which runs before the commit) stores in the userdata of each BlkDevice the udev name that was used for matching it (if any). Then the bootloader could use that in addition to `udev_full_all`.

But that sounds too much like a workaround.

aschnell: find_by_any_name only works on the probed devicegraph since the devices may
not exist in any other devicegraph and even those existing in other
devicegraphs but not in probed cannot be found (due to the system lookup).

For the duplictate use of udev names with multipath: It is the responsibility
of udev to generate the correct link in /dev.

ancorgs: _find_by_any_name only works on the probed devicegraph since the devices may not exist in any other devicegraph and even those existing in other devicegraphs but not in probed cannot be found (due to the system lookup)._

So that's basically what I said. The real requisite is that the blk devices in the devicegraph are the same than the block devices in the system. That's true for probed ONLY before the commit. After a successful commit is the other way around. The list of devices in probed may be different from the system but the list on staging so match.

PR #475, opened 2017-12-20
fix miscalculation of sizes when resizing a Windows partition (bsc#1066386)

wfeldt:
What otherwise happens is this: our planning strategy will typically resize
an existing Windows partition *twice*. First for the root partition, then
again for the rest.

Without this patch, after the first (correct) resize, the Windows partition
will leave a free space block that appears to be bigger than can actually be
used (due to the alignment issue). As a result, the second resize would be
planned too small.

### DIFF
--- a/test/data/devicegraphs/output/windows-pc.yml
+++ b/test/data/devicegraphs/output/windows-pc.yml
@@ -6,7 +6,7 @@
     partitions:
 
     - partition:
-        size: 738.00 GiB
+        size: 755711 MiB
         name: /dev/sda1
         id: 0x7
         file_system: ntfs
@@ -26,6 +26,9 @@
         id: swap
         file_system: swap
         mount_point: swap
+
+    - free:
+       size: 1 MiB
 
     - partition:
         size: unlimited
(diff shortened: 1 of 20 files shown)


ancorgs:
The change looks good. Good catch! We now just need the tests and all that. ;-)

test/data/devicegraphs/output/windows-pc.yml
```diff
@@ -27,6 +27,9 @@
         file_system: swap
         mount_point: swap
 
+    - free:
```
ancorgs: Checking for alignment instead of blindly adding that min_grain should save us from these cases in which we introduce a gap. But that's for a second iteration. Before the Christmas break, fixing the main bug and not killing Windows should be enough.

ancorgs:
As commented above, it looks good enough for the time being. Is MUCH better to be too pesimistic and leave a gap in some situations than to miscalculate and free to little space.

Just increase version and add changelog.

I will add regression tests later for the original bug (likely not today).

test/data/devicegraphs/output/windows-pc.yml
wfeldt: That will not be enough; we haven't even started to test for cases where we start with a misaligned setup.

PR #476, opened 2017-12-20
Partitioner: action to delete an LVM VG

ancorgs:
This implements https://trello.com/c/qu4XutJH/739-1-expert-partitioner-delete-a-vg

**Note:** It should be easier to review if done commit per commit.

delete_vg_confirm

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -24,5 +24,5 @@ Rake::Task["package"].prerequisites.delete("check:changelog")
 
 Yast::Tasks.configuration do |conf|
   conf.skip_license_check << /.*/
-  conf.documentation_minimal = 80
+  conf.documentation_minimal = 83
 end
--- a/src/lib/y2storage/devicegraph.rb
+++ b/src/lib/y2storage/devicegraph.rb
@@ -99,6 +99,8 @@ module Y2Storage
     # e.g., PartitionTable.delete_partition() or LvmVg.delete_lvm_lv().
     #
     # @see #remove_md
+    # @see #remove_lvm_vg
+    # @see #remove_btrfs_subvolume
     #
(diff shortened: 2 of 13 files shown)


src/lib/y2storage/devicegraph.rb
```diff
@@ -361,5 +389,35 @@ def device_tree
...
+      raise(ArgumentError, "No device provided") if device.nil?
+      raise(ArgumentError, "Device not found") unless device.exists_in_devicegraph?(self)
+
+      children = device.children
+      until children.empty?
+        remove_with_dependants(children.first, keep: keep + [device])
+        children = device.children
+      end
+
+      orphans = device.respond_to?(:potential_orphans) ? device.potential_orphans : []
```
jreidinger: is it needed? all devices should have this method, so why to ask for it? and above you check it is not nil.

Rakefile
```diff
@@ -24,5 +24,5 @@ Rake::Task["package"].prerequisites.delete("check:changelog")
 
 Yast::Tasks.configuration do |conf|
   conf.skip_license_check << /.*/
-  conf.documentation_minimal = 80
+  conf.documentation_minimal = 83
```
jreidinger: nice ;)

src/lib/y2storage/devicegraph.rb
ancorgs: It's maybe a little bit too defensive, but just imagine something changes in libstorage-ng and, for whatever reason we get a non-wrapped object. I.e. we get a `Storage::Whatever` because we have still not created the corresponding `Y2Storage::Whatever` and `Storage::Whatever` is not a subclass of `Storage::Device`. It may sounds paranoid, but better be safe from libstorage-ng changes.

PR #477, opened 2017-12-21
Export enable snapshots

imobachgs:
Export `<enable_snapshots>` element. It solves [bsc#1073544](https://bugzilla.suse.com/show_bug.cgi?id=1073544). I am not incrementing the version number because I plan to merge another fix today and I want to avoid submitting the same package twice.

### DIFF
--- a/src/lib/y2storage/autoinst_profile/drive_section.rb
+++ b/src/lib/y2storage/autoinst_profile/drive_section.rb
@@ -147,7 +147,7 @@ module Y2Storage
       # As usual, it keeps the behavior of the old clone functionality, check
       # the implementation of this class for details.
       #
-      # @param disk [Disk, Dasd]
+      # @param disk [Y2Storage::Disk, Y2Storage::Dasd] Disk
       def init_from_disk(disk)
         return false if disk.partitions.empty?
 
@@ -158,6 +158,7 @@ module Y2Storage
         @partitions = partitions_from_disk(disk)
         return false if @partitions.empty?
 
+        @enable_snapshots = enabled_snapshots?(disk)
         @partitions.each { |i| i.create = false } if reuse_partitions?(disk)
 
         # Same logic followed by the old exporter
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -302,6 +303,18 @@ def use_value_from_string(use)
...
+
+      # Determine whether snapshots are enabled for the current drive
+      #
+      # Currently AutoYaST does not support enabling/disabling snapshots
+      # for a partition but for the whole disk.
+      #
+      # @param disk [Y2Storage::Disk, Y2Storage::Dasd] Disk
+      # @return [Boolean] true if snapshots are enabled
+      def enabled_snapshots?(disk)
+        filesystems = disk.partitions.map(&:filesystem)
```
ancorgs: I guess we don't have to worry here about the case of a disk directly formatted (no partitions)

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -312,6 +312,14 @@ def subvolumes_prefix
         children.size == 1 ? children.first.path : ""
       end
 
+      # Determines whether the snapshots (snapper) are activated
+      #
+      # @return [Boolean] true if snapshots are present
+      def snapshots?
```
ancorgs: For already existing filesystems maybe this is our best chance to know whether the snapshots are configured. But for new filesystems we have `Filesystems::Btrfs.configure_snapper`

ancorgs: `Filesystems::Btrfs.configure_snapper` can be checked if there is no `.snapshots` subvolume.

ancorgs:
Is LVM completely out of scope here? I mean, what about `/` being an LV?

imobachgs:
LVM will be tracked at https://trello.com/c/Yt3qD65F/2001-export-snapshots-settings-for-lvm-devices.

PR #478, opened 2017-12-22
fix logic in GuidedProposal::calculate_proposal (bsc#1058027)

wfeldt:


src/lib/y2storage/guided_proposal.rb
```diff
@@ -119,6 +119,10 @@ def calculate_proposal
 
       exception = nil
       [:desired, :min].each do |target|
+
+        # reset candidate_devices, else #candidate_roots will just use it
+        populated_settings.candidate_devices = nil;
```
ancorgs: sure you didn't want to reset `root_device` instead of `candidate_devices`?

test/y2storage/guided_proposal_test.rb
```diff
@@ -161,7 +161,8 @@
 
       it "creates the needed partitions with correct device names" do
         proposal.propose
-        raid = proposal.devices.dm_raids.first
+        # note: potentially order dependent; thera are two raids defined
```
ancorgs: `Devicegraph#dm_raids` should return the list sorted by device name. So it shouldn't be any order dependency here.

test/y2storage/proposal_low_space_test.rb
```diff
@@ -40,4 +40,40 @@
       include_examples "LVM-based proposed layouts"
     end
   end
+
+  # Regression tests
```
ancorgs: We should add the bug number for the regression test here.

test/y2storage/guided_proposal_test.rb
wfeldt: Well, according to the log it tries the second raid first, for whatever reason. This is just a note for the next developer.

test/y2storage/guided_proposal_test.rb
```diff
@@ -140,7 +140,7 @@
     end
 
     context "when installing in a DM RAID" do
-      let(:scenario) { "empty-dm_raids.xml" }
+      let(:scenario) { "empty-dm_raids_no_sda.xml" }
```
ancorgs: Since the test seems to be there just to check than the partitions are correctly created and so on, wouldn't have been easier to restrict the proposal scope than creating a new xml file? Maintaining all those devicegraph definition is a PITA.

Something like this (I haven't tested it) could save us from adding just another blob:
```ruby
 before do
  settings.candidate_devices = ["/dev/mapper/isw_ddgdcbibhd_test1", "/dev/mapper/isw_ddgdcbibhd_test2"]
end
```

So sda is out in a more diplomatic way than killing it. :-)

wfeldt: I find this even harder to understand. That's a setup mixture of xml files and code in rspec. This entangles unrelated test cases even more.

PR #480, opened 2017-12-22
to_human_string do not raises an exception

imobachgs:
It fixes [bsc#1068087](https://bugzilla.suse.com/show_bug.cgi?id=1068087).

### DIFF
--- a/src/lib/y2storage/partition_id.rb
+++ b/src/lib/y2storage/partition_id.rb
@@ -58,7 +58,7 @@ module Y2Storage
     def to_human_string
       textdomain "storage"
 
-      string = TRANSLATIONS[to_i] or raise "Unhandled Partition ID '#{inspect}'"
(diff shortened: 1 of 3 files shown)


ancorgs:
The point is... I basically agree with this change. I always though an exception is too much. But when the exception was originally introduced it was pointed to be intentional following the "fail early" approach.

So to please everybody we should, at least, print a warning in the logs or something like that.

imobachgs:
I makes sense. Updated.

src/lib/y2storage/partition_id.rb
```diff
@@ -58,7 +59,8 @@ class PartitionId
     def to_human_string
       textdomain "storage"
 
-      string = TRANSLATIONS[to_i] or raise "Unhandled Partition ID '#{inspect}'"
+      log.info "Unhandled Partition ID '#{inspect}'"
```
ancorgs: You are logging always. Not only on failure. And using info instead of warn?

PR #482, opened 2018-01-04
Don't rely on a previous swap partition being formatted (bsc#1071515)

shundhammer:
https://bugzilla.suse.com/show_bug.cgi?id=1071515

Now checking if that previous swap partition actually did have a filesystem (in this case a swap signature) and if it is of type swap.

Not sure if in both cases it shouldn't rather be `next` rather than `break`; trying to be consistent here.

Please notice that this is untested as of now.

### DIFF
--- a/src/lib/y2storage/proposal/devicegraph_generator.rb
+++ b/src/lib/y2storage/proposal/devicegraph_generator.rb
@@ -155,6 +155,7 @@ module Y2Storage
         new_swap_volumes.each_with_index do |swap_volume, idx|
           deleted_swap = deleted_swaps[idx]
           break unless deleted_swap
+          break unless deleted_swap.formatted_as?(:swap)
 
           swap_volume.uuid = deleted_swap.filesystem.uuid
           swap_volume.label = deleted_swap.filesystem.label


src/lib/y2storage/proposal/devicegraph_generator.rb
```diff
@@ -155,6 +155,7 @@ def refine_planned_partitions!(planned_partitions, deleted_partitions)
         new_swap_volumes.each_with_index do |swap_volume, idx|
           deleted_swap = deleted_swaps[idx]
           break unless deleted_swap
+          break unless deleted_swap.formatted_as?(:swap)
```
ancorgs: This will raise an exception if `new_swap_volumes.size > deleted_swaps.size` (very common scenario, I would say).

Also in this scenario, this change will produce the UUID to NOT be reused:
```ruby
planned_partitions =  [a_swap_without_reuse]
deleted_partitions = [an_ext4_with_wrong_id_of_swap, a_real_swap]
```

A more robust fix (and more understandable IMO) would be to change the first line like this:
```ruby
deleted_swaps = deleted_partitions.select { |part| part.id.is?(:swap) && part.formatted_as?(:swap) }
```

Isn't it?

shundhammer:
We talked about this; ancor knows the code better. Closing this; it was just a suggestion anyway.

PR #483, opened 2018-01-08
Avoid exception when running the proposal in a BIOS RAID 

ancorgs:
Fix for https://bugzilla.suse.com/show_bug.cgi?id=1067349

Regression test included.


src/lib/y2storage/disk_device.rb
```diff
@@ -0,0 +1,45 @@
...
+# find current contact information at www.suse.com.
+
+module Y2Storage
+  # Mixin for classes that are usually treated like disks by YaST.
+  #
+  # @see Devicegraph::disk_devices
+  #
+  # Includes several methods usually checked by the partitioner, the proposal
+  # and other components. This mixing offers a default conservative
+  # implementation that can be overridden or extended in any of the classes.
```
joseivanlopez: NP: s/mixing/mixin

PR #484, opened 2018-01-08
Avoid error caused by unformatted swap partitions

ancorgs:
Fix for https://bugzilla.suse.com/show_bug.cgi?id=1071515

Includes regression test.

### DIFF
--- a/src/lib/y2storage/proposal/devicegraph_generator.rb
+++ b/src/lib/y2storage/proposal/devicegraph_generator.rb
@@ -149,7 +149,9 @@ module Y2Storage
       # @param deleted_partitions [Array<Partition>] partitions
       #     deleted from the initial devicegraph
       def refine_planned_partitions!(planned_partitions, deleted_partitions)
-        deleted_swaps = deleted_partitions.select { |part| part.id.is?(:swap) }
+        deleted_swaps = deleted_partitions.select do |part|
+          part.id.is?(:swap) && part.formatted_as?(:swap)
+        end
         new_swap_volumes = planned_partitions.select { |vol| !vol.reuse && vol.mount_point == "swap" }
 
         new_swap_volumes.each_with_index do |swap_volume, idx|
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/proposal/devicegraph_generator.rb
```diff
@@ -149,7 +149,9 @@ def new_physical_volumes(old_devicegraph, new_devicegraph)
       # @param deleted_partitions [Array<Partition>] partitions
       #     deleted from the initial devicegraph
       def refine_planned_partitions!(planned_partitions, deleted_partitions)
-        deleted_swaps = deleted_partitions.select { |part| part.id.is?(:swap) }
+        deleted_swaps = deleted_partitions.select do |part|
+          part.id.is?(:swap) && part.formatted_as?(:swap)
```
shundhammer: OK, but what will happen with the partition this bug was all about? It had a partition ID "swap", but it was not formatted at all. It will then no longer be in this `deleted_swaps` array. What will happen with that partition? I guess its disk space just goes to the free space and will be reused somehow, right?

ancorgs: This does not influence how/whether the partitions are deleted. Indeed, that partition was already deleted and the space was already used for whatever it was good for.

This code just does a final round through all those already deleted partitions (taking them from `probed`, of course, they are not in the result) in order to reuse their UUIDs and labels in the new swaps (which may save headaches to other Linux installations).

If the partition had partition id swap but no filesystem (that's `/dev/sda1` in our new unit test, so we are covered), there is nothing to reuse, so it's irrelevant for this piece of code (which, again, means nothing about it being deleted or not).

shundhammer: OK then.

ancorgs: Anyway, since I had to please rubocop anyways, I added a comment in the test.

PR #485, opened 2018-01-09
Added Mountable#persistent? (needed for bsc#1073696)

lslezak:
- See also https://github.com/yast/yast-packager/pull/308/
- 4.0.68

### DIFF
--- a/src/lib/y2storage/mountable.rb
+++ b/src/lib/y2storage/mountable.rb
@@ -88,6 +88,13 @@ module Y2Storage
 
     alias_method :mount_point=, :mountpoint=
 
+    # Is the mount persistent?
+    # @return [Boolen] true if the mount point is saved to /etc/fstab
+    #   (and will be mounted at boot again), false otherwise
+    def persistent?
+      to_storage_value.mount_point.in_etc_fstab?
+    end
+
     # Sets the options to use in /etc/fstab for a newly created filesystem.
     #
     # @param new_options [Array<String>]
--- a/src/lib/y2storage/mountable.rb
+++ b/src/lib/y2storage/mountable.rb
@@ -89,10 +89,11 @@ module Y2Storage
     alias_method :mount_point=, :mountpoint=
 
     # Is the mount persistent?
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/mountable.rb
```diff
@@ -88,6 +88,13 @@ def mountpoint=(path)
 
     alias_method :mount_point=, :mountpoint=
 
+    # Is the mount persistent?
+    # @return [Boolen] true if the mount point is saved to /etc/fstab
+    #   (and will be mounted at boot again), false otherwise
+    def persistent?
+      to_storage_value.mount_point.in_etc_fstab?
```
ancorgs: This can raise an exception if there is no mount point. This should be more robust:
```ruby
to_storage_value.has_mount_point && to_storage_value.mount_point.in_etc_fstab?
```

src/lib/y2storage/mountable.rb
```diff
@@ -88,6 +88,13 @@ def mountpoint=(path)
 
     alias_method :mount_point=, :mountpoint=
 
+    # Is the mount persistent?
+    # @return [Boolen] true if the mount point is saved to /etc/fstab
```
ancorgs: typo: Boolean

ancorgs:
Adding a simple unit test should not be that hard. For example, assigning a mountpoint to an existing filesystem and verifying that it becomes persistent. I will try to provide some code as starting point, give me a sec.

src/lib/y2storage/mountable.rb
```diff
@@ -88,6 +88,14 @@ def mountpoint=(path)
 
     alias_method :mount_point=, :mountpoint=
 
+    # Is the mount persistent?
+    #
+    # @return [Boolean] true if the mount point is saved to /etc/fstab
+    #   (and will be mounted at boot again), false otherwise
```
joseivanlopez: NP: I would add a "TODO" to say that this should be adapted once `MountPoint` wrapper class is there.

lslezak:
@ancorgs @joseivanlopez thank you very much for the review and the improvements! :+1: :1st_place_medal:

PR #486, opened 2018-01-11
More object-oriented handling of partitions resizing

ancorgs:
Related to https://trello.com/c/ymdimYtb/91-3-lvm-expert-partitioner-resize-a-lv

Reviewing commit by commit should be easier

* Moves the resizing, aligning and similar stuff the the right place (class `Partition`) instead of it being scattered around all the code.
* This should make the current partition dialogs way more reusable for LVs.
* Improved documentation about alignment in several places. Specially the meaning of policies and types.
* Bonus: we can now reuse this to resize in proposal honoring the alignment (separate PR will come).
* We can now use `AlignType::REQUIRED` in any operation (in addition to the default  `AlignType::OPTIMAL`).
 * Includes workaround to a libstorage-ng bug on alignment.


test/y2partitioner/dialogs/partition_resize_test.rb
```diff
@@ -157,18 +165,98 @@ def find_label(contents, text)
...
+          end
+
+          context "and the max size causes a not end-aligned partition" do
+            let(:max_size) { not_aligned_max_size }
+
+            it "updates the partition with the max aligned size" do
+              subject.store
+              expect(partition.size).to eq adjusted_size
+              expect(partition.end_aligned?).to eq(true)
+              expect(adjusted_size).to be <= max_size
```
joseivanlopez: NP: I know it is the same, but I would replace the last `expect` by `expect(partition.size).to be <= max_size` (but it LGTM at it is now)

joseivanlopez:
LGTM

PR #487, opened 2018-01-17
Remove TODOs in partitioner views

joseivanlopez:
PBI: https://trello.com/c/bHeInnsF/199-3-partitionerng-fix-hard-disk-overview


src/lib/y2partitioner/widgets/blk_device_attributes.rb
```diff
@@ -1,92 +1,183 @@
...
+        # which the device is connected to (e.g., SATA or ATA)
+        format(_("Bus: %s"), blk_device.bus || "")
+      end
+
+      # Information about number of sectors in the device
+      #
+      # @return [String]
+      def device_sectors
+        # TRANSLATORS: Number of sectors in the device, where %s is replaced by a number
+        format(_("Number of Sectors: %s"), blk_device.region.length)
```
ancorgs: If the placeholder is substituted by a number, shouldn't it be `%i`?

src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -197,6 +195,96 @@ def end_value(device)
...
+        return type_icon(device.plain_blk_device) if device.is?(:lvm_pv)
+
+        type = DEVICE_ICONS.keys.find { |k| device.is?(k) }
+        icon = type.nil? ? Icons::DEFAULT_DEVICE : DEVICE_ICONS[type]
+
+        small_icon(icon)
+      end
+
+      DEVICE_LABELS = {
+        disk:          _("DISK"),
```
ancorgs: If you use `_("DISK")` here, at class definition level, these strings will not change in the system language is changed at some later point, when the class is already defined. Check how `_N("DISK")` is used in other parts with the same problem.

src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -197,6 +195,96 @@ def end_value(device)
...
+        dasd:          _("DISK"),
```
ancorgs: Why the usage of UPCASE in terms like "DISK" or "PARTITION"?

On the other hand, maybe a comment to translators to advise to keep this sort (to be displayed in a column of a big list).

ancorgs: Well MD RAID and LVM are correctly capitalized in all upcase because they are abbreviations of names (e.g. Multiple Device Redundant Array of Independent Disks). So not the same case. 

On the other hand, I don't see how `PARTITION` can end up being used. And maybe the same applies to others. So please add a comment to those values that are there just for completeness but that are impossible to reach.

src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -197,6 +195,95 @@ def end_value(device)
...
+      #
+      # @see DEVICE_LABELS
+      #
+      # @param device [Y2Storage::BlkDevice]
+      # @return [String]
+      def default_type_label(device)
+        type = DEVICE_LABELS.keys.find { |k| device.is?(k) }
+        return "" if type.nil?
+
+        DEVICE_LABELS[type]
```
ancorgs: Shouldn't this be `_(DEVICE_LABELS[type])`?

PR #488, opened 2018-01-18
Add default fstab options

shundhammer:
https://trello.com/c/5zK5B9J6/94-5-betasles15-p2-1066076-installer-default-data-journaling-mode-in-ext34-differ-from-sles12

This adds fstab options (mount options to be written to /etc/fstab) to filesystems we create or mount, very much like in the old storage stack.

If the user or the application logic explicitly sets any fstab options, those are used, of course. If not, this now uses default fstab options which are different for each filesystem type: ext2/3/4, vfat, ntfs.

I extended the table that already existed for similar things that are different for each filesystem type.

In most cases it's a very simple list of strings, but there are also the Microsoft filesystems that need some locale-specific parameters such as `iocharset` and `codepage` where we need to figure out the appropriate values based on locale information. Those settings are used to convert filenames with special (non-ASCII) characters to and from Linux filenames.

This is not done in libstorage-ng for several reasons:

- Some day soon, this might have to become configurable for each product we make.
- It should be easily accessible and easy to change.
- It should be at one central place, not scattered over a dozen files.
- The locale information that is needed for `iocharset` and `codepage` has to be fetched from the `Encoding` Ruby module. We don't want to clutter the APIs to pass that information back and forth.

There is also the special case of the YAML reader (the FakeDeviceFactory): It should not implicitly set values that are not explicitly in the YAML file that is read.

------------------

After one very much failed approach that tried to overload `BlkDevice::create_blk_filesystem()` to fetch the default fstab options per filesystem type (and which had that problem that the YAML reader was then doing magic things without being asked to do them), now the default fstab options are explicitly added at the appropriate places.

This needs to cover:
- The proposal
- The guided setup (which is little more than the proposal)
- The expert partitioner

and it needs to cover both when we explicitly create a filesystem _and_ when we reuse an existing one and just add a new entry in /etc/fstab for it.

-------------------

## New fstab options in partitioner

ext4-fstab-options

### Resulting /etc/fstab

ext4-fstab

Notice the last entry with _ext4_ that shows the new default options. They are now correctly initialized and passed to the expert partitioner (which the screenshot shows). If the user doesn't open that dialog, they are just used as they are; if he does, he can change them, of course.

The proposal uses those defaults as well; see the corresponding unit tests.

### DIFF
--- a/test/y2storage/filesystems/type_test.rb
+++ b/test/y2storage/filesystems/type_test.rb
@@ -23,6 +23,8 @@
 require_relative "../spec_helper"
 require "y2storage"
 
+Yast.import "Encoding"
+
 describe Y2Storage::Filesystems::Type do
   describe "#to_human_string" do
     it "returns the description of the type" do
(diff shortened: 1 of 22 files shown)


test/y2storage/filesystems/type_test.rb
```diff
@@ -105,4 +107,80 @@
...
+
+    it "return the correct values in a utf8 ja_JP locale" do
+      Yast::Encoding.SetUtf8Lang(true)
+      Yast::Encoding.SetEncLang("ja_JP")
+      expect(described_class::VFAT.iocharset).to eq "utf8"
+      expect(described_class::VFAT.codepage).to eq "932"
+    end
+  end
+
+  describe "#default_fstab_options" do
```
ancorgs: Nitpicking: I would shuffle the organization of contexts and examples a little bit, so the unit tests really check that the locale-independent types are indeed not affected by the locale (and that it stays like that in the future).

Something in this direction:

```ruby
RSpec.shared_examples "locale-independent filesystem types" do
  it "returns the correct options for ext2" do
    expect(described_class::EXT2.default_fstab_options).to eq ["acl", "user_xattr"]
  end

  it "includes no encoding options for ext2" do
    options = described_class::EXT2.default_fstab_options
    expect(options).to_not include(a_string_starting_with("iocharset"))
    expect(options).to_not include(a_string_starting_with("codepage"))
  end

  it "returns the correct options for ext3" do
    expect(described_class::EXT3.default_fstab_options).to eq  ["data=ordered", "acl", "user_xattr"]
  end

  it "includes no encoding options for ext3" do
    options = described_class::EXT3.default_fstab_options
    expect(options).to_not include(a_string_starting_with("iocharset"))
    expect(options).to_not include(a_string_starting_with("codepage"))
  end
end

context "in a utf8 locale" do
  before do
    Yast::Encoding.SetUtf8Lang(true)
    Yast::Encoding.SetEncLang("de_DE")
  end

  include_examples "locale-independent filesystem types" 

  it "returns the correct options for vfat" do
    expect(described_class::VFAT.default_fstab_options).to eq ["utf8=true"]
  end
end

context "in a non-utf8 cs_CZ locale" do
  before do
    Yast::Encoding.SetUtf8Lang(false)
    Yast::Encoding.SetEncLang("cs_CZ")
  end

  include_examples "locale-independent filesystem types" 

  it "returns the correct options for vfat" do
    expect(described_class::VFAT.default_fstab_options).to eq ["iocharset=iso8859-2", "codepage=852"]
  end
end
```

In fact, even more combinations can be achieved with less code, but that's a different topic. My main point is that the way to probe extX are locale-independent is actually testing then with several locales as it's done with the locale-dependant filesystems.

shundhammer: OK, I'll see about that, but in the next PBI; I'd like to get this fix shipped with the upcoming next Beta by the end of tomorrow.

PR #491, opened 2018-01-19
Fix array subtracting

jreidinger:


src/lib/y2storage/device.rb
```diff
@@ -36,15 +36,31 @@ class Device
...
-    #   @note Devices are equal if they have the same {#sid storage id}.
-    #   @see sid
+    storage_forward :storage_eql, to: :==
+    protected :storage_eql
+
+    #  compare two devices.
+    #  @note devices are equal if they have the same {#sid storage id}.
+    #  @see sid
+    #
+    #  @param device [device]
```
ancorgs: The names and types in the code don't match the yardoc. Shouldn't this be?

```ruby
# @param other [Device]
# @return [Boolean]  false if compared to different class
```

src/lib/y2storage/device.rb
```diff
@@ -36,15 +36,31 @@ class Device
...
+
+    #  compare two devices.
+    #  @see ==
     #
-    #   @param device [Device]
-    #   @return [Boolean]
-    storage_forward :==
-    storage_forward :!=
+    #  @param device [device]
+    #  @return [boolean] true if compared to different class
```
ancorgs: Same than above, the yardoc looks inconsistent to me.

src/lib/y2storage/device.rb
```diff
@@ -36,15 +36,37 @@ class Device
...
-    storage_forward :!=
+    #  @param other [Device]
+    #  @return [Boolean] true if compared to different class
+    def !=(other)
+      !(self == other)
+    end
+
+    alias_method :eql?, :==
+
+    # redefine hash method to return same result for same devices indenpendently
```
ancorgs: typo: independently

On the other hand, since this definition of `#hash` is kind of unexpected and somebody could feel tempted to replace it with something more colorful, I would appreciate if you mention here that this is done in order to make the `Array` operations work. Maybe even mentioning the bug number.

jreidinger: ok

PR #493, opened 2018-01-22
raise exception properly when there are no disks (bsc#1057430)

wfeldt:
Actually the description of #calculate_proposal is wrong. The space
calculator doesn't raise NoDiskSpaceError but Error if things fail.

Anyway, sticking to the docs and raising NoDiskSpaceError if disks are missing.

### DIFF
--- a/src/lib/y2storage/guided_proposal.rb
+++ b/src/lib/y2storage/guided_proposal.rb
@@ -113,7 +113,7 @@ module Y2Storage
 
     # Calculates the proposal
     #
-    # @raise [NoDiskSpaceError] if there is no enough space to perform the installation
+    # @raise [Error, NoDiskSpaceError] if there is no enough space to perform the installation
     def calculate_proposal
       settings.freeze
 
@@ -143,7 +143,7 @@ module Y2Storage
         end
       end
 
-      raise exception
+      raise exception || NoDiskSpaceError.new("No usable disks detected")
     end
 
     # @return [Array<Planned::Device>]
--- a/test/y2storage/guided_proposal_no_disks.rb
+++ b/test/y2storage/guided_proposal_no_disks.rb
@@ -23,9 +23,6 @@
 require_relative "spec_helper"
 require "y2storage"
 
-# stop rspec warning
-RSpec::Expectations.configuration.on_potential_false_positives = :nothing
-
 describe Y2Storage::GuidedProposal do
   # Regression test (bsc#1057430)
   describe ".initial" do
--- a/test/y2storage/guided_proposal_no_disks.rb
+++ b/test/y2storage/guided_proposal_no_disks.rb
@@ -1,0 +1,44 @@
+#!/usr/bin/env rspec
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
+#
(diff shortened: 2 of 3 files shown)


src/lib/y2storage/guided_proposal.rb
```diff
@@ -139,11 +139,11 @@ def calculate_proposal
             exception = error
           end
 
-          return true unless exception
+          return unless exception
```
ancorgs: What was wrong with returning nil?

ancorgs:
In general it looks ok, apart from the rubocop offence, of course.

src/lib/y2storage/guided_proposal.rb
wfeldt: Returning `true` implies that the return value matters. Then you have to check the code what's really expected. That's just annoying.

wfeldt: I see - rubocop dislikes it. That's just insane.

ancorgs: And does it really matter? I mean, with the current code it can only return `nil` (or `true` in your version) or raise an exception. So with the current method API, the returned value looks indeed irrelevant to me.

Don't get me wrong, I'm not against returning nil, true or `:I_did_it` :-) I simply don't see why the returned value is important with the current definition in which this is more a "procedure" than a "function", sort to say.

wfeldt: Exactly. That's why I wanted to get **rid** of `true`.

ancorgs: Oh, damn. I have been reading it the other way around all the time, like if you were ADDING the true. Sorry. Looks like I can't distinguish red from green (has happened to me twice this month).

test/y2storage/guided_proposal_no_disks.rb
```diff
@@ -0,0 +1,44 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require_relative "spec_helper"
+require "y2storage"
+
+# stop rspec warning
+RSpec::Expectations.configuration.on_potential_false_positives = :nothing
```
ancorgs: Why is this needed?

Moreover, I assume than placing it here actually affects the whole testsuite (or any test run after this). In that case, it's quite hidden here.

ancorgs: Writing something like `to_not raise_error Y2Storage::Error` below looks less aggressive than disabling the warning (potentially) in the whole testsuite.

wfeldt: Hm, I had expected it to affect only the single test file.

Without this rspec will not let you the error type you catch unspecified without a big warning.

Any better idea?

wfeldt: It's a TypeError, actually. And this looked a bit too specific in a test to me.

ancorgs: Well, if you want a completely generic check, I guess you can check for `Exception` (the root class for any exception)

PR #494, opened 2018-01-22
Consider all spaces

imobachgs:
Unused spaces are taken into account when deciding which partitions distribution is better. Somehow definition of `gap` is changed. Previously, it was meant to be a space produced by the partitioning algorithm (for instance, putting a partition into a free space but leaving some space at the end). Now, any free space is considered a gap (no matter how it was created).


imobachgs:
Thanks!

PR #495, opened 2018-01-22
Resize LVM LV

ancorgs:
Since it implies renaming of files/classes, better review commit by commit.

This implements the main part of https://trello.com/c/ymdimYtb/91-3-lvm-expert-partitioner-resize-an-lv


src/lib/y2storage/lvm_lv.rb
```diff
@@ -55,6 +55,37 @@ class LvmLv < BlkDevice
...
+      return unless can_resize?
+
+      # The sizes in resize_info are already rounded to the extent size
+      self.size =
+        if new_size > resize_info.max_size
+          resize_info.max_size
+        elsif new_size < resize_info.min_size
+          resize_info.min_size
+        else
+          new_size
```
joseivanlopez: Should not the size be rounded to extent size?

joseivanlopez: Ok, in tests I see it is automatically rounded.

PR #497, opened 2018-01-23
Partitioner: treat International System units as power of two

ancorgs:
This should make this test succeed, since "1Mb" will be considered equivalent to "1Mib" (we can't educate the whole world):

https://openqa.opensuse.org/tests/588406

### DIFF
--- a/src/lib/y2partitioner/size_parser.rb
+++ b/src/lib/y2partitioner/size_parser.rb
@@ -1,0 +1,46 @@
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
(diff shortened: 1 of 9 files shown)


src/lib/y2partitioner/size_parser.rb
```diff
@@ -0,0 +1,46 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2storage"
```
joseivanlopez: NP: I would only include "y2storage/disk_size".

src/lib/y2partitioner/size_parser.rb
```diff
@@ -0,0 +1,46 @@
...
+  # considering International System units as base 2 units) and error handling
+  # (e.g. the user leaves the field empty or enters some crazy stuff).
+  module SizeParser
+    # Converts a string to DiskSize, returning nil if the conversion
+    # it is not possible
+    #
+    # @see Y2Storage::DiskSize#from_human_string
+    #
+    # @return [Y2Storage::DiskSize, nil]
+    def parse_user_size(string)
```
joseivanlopez: NP: we are tending to document all the param with `@param` (although the name is quite descriptive)

mvidner:
@ancorgs I disagree. Surely the whole world is a lot of people to educate, but
1. our QA department is teachable
2. this introduces an inconsistency, where MB means something else on input than on output
3. we will get a report
    > I have 10GB of space, create a "10GB" partition and then yast tells me it will not fit, WTF.

ancorgs:
@mvidner Well, it was discussed with @jsrain, @kobliha and others and the conclusion was to do it this way.

mvidner:
OK then, thanks.

PR #500, opened 2018-01-29
[WIP] NFS section of the partitioner

ancorgs:
Main PR for https://trello.com/c/jzlH7N52/72-5-expert-partitioner-nfs

Unit tests missing.


src/lib/y2partitioner/yast_nfs_client.rb
```diff
@@ -0,0 +1,264 @@
...
+    end
+
+    # Mark the client configuration as already read.
+    #
+    # @see .client_configured?
+    def self.mark_client_configured
+      @client_configured = true
+    end
+
+    # Resets the information checked by {.client_configured?}
```
imobachgs: Mix of 1st/3rd person across different methods.

src/lib/y2partitioner/yast_nfs_client.rb
```diff
@@ -0,0 +1,264 @@
...
+      # non-fstab settings of nfs-client (firewall, sysconfig, idmapd) get
+      # written on closing partitioner
+      process_result(widget_id, client_result)
+    end
+
+    # Name of the package containing the YaST client
+    #
+    # @return [String]
+    def package_name
+      PACKAGE.dup
```
imobachgs: The string is frozen and I would say that's caller responsability to *dup* it if needed. We are creating a new instance everytime we call this method.

src/lib/y2partitioner/yast_nfs_client.rb
```diff
@@ -0,0 +1,264 @@
...
+      log.info "Reading NFS settings"
+      call_client("Read")
+      self.class.mark_client_configured
+    end
+
+    # Checks whether the partitioner is being executed in the inst-sys (i.e.
+    # during system installation)
+    #
+    # @return [Boolean]
+    def in_inst_sys?
```
imobachgs: Why not to use `Yast::Stage.initial`? I know it looks better, but just for consistence with other YaST code...

src/lib/y2storage/filesystems/legacy_nfs.rb
```diff
@@ -0,0 +1,230 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2storage/filesystems/nfs"
+
+module Y2Storage
+  module Filesystems
+    # Class to represent an NFS mount, as defined in the old y2-storage.
```
imobachgs: *a NFS mount*

src/lib/y2storage/filesystems/legacy_nfs.rb
```diff
@@ -0,0 +1,230 @@
...
+      # Creates an {Nfs} object, equivalent to this one, in the devicegraph
+      #
+      # @raise [ArgumentError] if no devicegraph is given and no default
+      #   devicegraph has been previously defined
+      #
+      # @param devicegraph [Devicegraph, nil] if nil, the default devicegraph
+      #   will be used
+      # @return [Nfs] the new device
+      def create_nfs_device(devicegraph = nil)
+        graph = check_devicegraph(devicegraph)
```
imobachgs: np: I would replace `check` with `find`. At a first sight, I though that you were doing some "health check" to the devicegraph :)

src/lib/y2storage/filesystems/legacy_nfs.rb
```diff
@@ -0,0 +1,230 @@
...
+      # from its two components
+      def share_string(svr, dir)
+        "#{svr}:#{dir}"
+      end
+
+      # Ensures there is a devicegraph to work with, raising an exception
+      # otherwise
+      def check_devicegraph(devicegraph)
+        result = devicegraph || default_devicegraph
+        if result.nil?
```
imobachgs: np: You could just do `return result if result` and avoid nesting. Just a matter of preference.

src/lib/y2partitioner/yast_nfs_client.rb
ancorgs: Well, exactly because of that - it looks better. :-)

src/lib/y2storage/filesystems/legacy_nfs.rb
ancorgs: Sure? I had a similar discussion not long ago about "an LVM volume" vs "a LVM volume" and turned out the former was decided to be the right one. So I applied the same (lack of) rule here. Is supposed to be based on how the "average speaker" (if such thing exists) pronounces it.

src/lib/y2storage/filesystems/legacy_nfs.rb
ancorgs: Because it actually checks the validity of the `devicegraph` argument (and raises an exception if none provided). So I would rather change it to `check_devicegraph_arg` or `check_devicegraph_param` than to `find_devicegraph`

imobachgs: Ok, I am fine with any solution, but not `check_devicegraph` :)

src/lib/y2partitioner/yast_nfs_client.rb
imobachgs: In that case, I would rather prefer having that method in some place where other YaST code could use it (consistency + good looking). But for the time being, let's keep this as it is.

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jan 31 09:25:07 UTC 2018 - ancor@suse.com
+
+- Partitioner: initial support for NFS
+- Partitioner: removed useless tmpfs option
```
imobachgs: Please, add some bugzilla/fate nr (I am not that interested, but our bots are :) ).

PR #502, opened 2018-01-30
Improve error message

joseivanlopez:
Part of PBI https://trello.com/c/dsn6CQmd/204-3-lvm-expert-partitioner-resize-a-vg.

### DIFF
--- a/src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
+++ b/src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
@@ -97,6 +97,7 @@ module Y2Partitioner
       def selected_devices_validation
         return true if controller.devices_in_vg.size > 0
 
+        # TRANSLATORS: Error message when no device is selected
         Yast::Popup.Error(_("Select at least one device."))
         false
       end
@@ -110,6 +111,8 @@ module Y2Partitioner
         return true if controller.vg_size >= controller.lvs_size
 
         Yast::Popup.Error(
+          # TRANSLATORS: Error message when the resulting volume group size is not
+          # enough. %s is replaced by a size (e.g., 4 GiB).
           format(
             _("The volume group size cannot be less than %s."), controller.lvs_size
           )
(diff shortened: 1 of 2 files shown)


src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
```diff
@@ -129,12 +132,23 @@ def check_for_committed_devices(sids)
...
+          # %{pvs} is replaced by device names (e.g., /dev/sda1, /dev/sda2) and %{vg} is
+          # replaced by the volume group name (e.g., system)
+          _("Removing physical volumes %{pvs} from the volume group %{vg}\n" \
+            "is not supported because the physical volume is already in use")
         else
-          _("Physical volume %s cannot be removed because it already exists on disk")
+          # TRANSLATORS: Error message when a physical volume cannot be removed, where
+          # %{pvs} is replaced by device name (e.g., /dev/sda1) and %{vg} is replaced
+          # by the volume group name (e.g., system)
+          _("Removing the physical volume %{pvs} from the volume group %{vg}\n" \
```
ancorgs: Physical volumes, in plural

src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
```diff
@@ -129,12 +132,23 @@ def check_for_committed_devices(sids)
...
+            "is not supported because the physical volume is already in use")
```
ancorgs: Also plural at the end of the sentence. On the other hand, for both messages I guess it's more precise to say "the physical volumes may be already in use".  Strictly speaking, we are doing it to stay safe, but maybe there is no information allocated.

src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
joseivanlopez: Here it is the case of only one PV.

PR #505, opened 2018-01-31
Implement "arbitrary options" in the partitioner's fstab options page

shundhammer:
This uses introspection to check the other widgets on that same dialog page which values are already handled there (both VALUES and REGEXP constants) and populates the input field only with those that are not.

Upon storing the value, it accepts all options from the user. If any of them is already handled by another widget, it will still be set, but when entering the dialog again, that widget will pick up the value and handle it.

There is no sanity checking which values are valid. The background is to avoid keeping track of every possible mount option for all types of filesystems; new ones might come without us taking notice, so we rather trust the user than our ability to keep any such tables up-to-date for all future releases.

### DIFF
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -315,7 +315,7 @@ module Y2Partitioner
       VALUES = ["user", "nouser"].freeze
 
       def label
-        _("Mountable by user")
+        _("Mountable by User")
       end
 
       def help
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -361,7 +361,7 @@ module Y2Partitioner
       def help
         _("<p><b>Enable Quota Support:</b>\n" \
           "The file system is mounted with user quotas enabled.\n" \
-          "Default is false.</p>\n")
+          "Default is false.</p>")
       end
 
       def init
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -66,7 +66,10 @@ module Y2Partitioner
 
       def add_fstab_options(*options)
         # The options can only be modified using BlkDevice#fstab_options=
-        filesystem.fstab_options = filesystem.fstab_options + options
+        # so we can't just append options with << or push
+        new_options = filesystem.fstab_options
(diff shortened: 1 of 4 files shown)


src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -66,7 +66,10 @@ def delete_fstab_option!(option)
 
       def add_fstab_options(*options)
         # The options can only be modified using BlkDevice#fstab_options=
-        filesystem.fstab_options = filesystem.fstab_options + options
+        # so we can't just append options with << or push
+        new_options = filesystem.fstab_options
+        options.each { |opt| new_options << opt }
+        filesystem.fstab_options = new_options
```
joseivanlopez: To be honest, I did not understand why the previous version did not work.

src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -335,8 +360,8 @@ def label
 
       def help
         _("<p><b>Enable Quota Support:</b>\n" \
-        "The file system is mounted with user quotas enabled.\n" \
-        "Default is false.</p>\n")
+          "The file system is mounted with user quotas enabled.\n" \
+          "Default is false.</p>\n")
```
joseivanlopez: As HTML code is generated, I would say the last \n is leftover because `</p>` implies new line. An also, maybe we should use `<br />` instead of \n ?

shundhammer: That "\n" is unnecessary, right. "<p>...</p>" takes care of that just fine.

src/lib/y2partitioner/widgets/fstab_options.rb
shundhammer: It didn't work if multiple options were passed as a function argument.

And since that filesystem.fstab_options is closely connected to those SWIG bindings, i.e. the underlying C++ object is a std::list<std::string>, not a real Ruby Array<String>, it doesn't provide the Ruby _Enum_ methods which makes it a bit awkward to handle.

ancorgs: Wasn't that automatically handled by `Filesystem#fstab_options`? In theory it does internally basically what you are doing here.

shundhammer: I tried, and I got another very weird error that boiled down to "no clue what to do with a `std::list<std::string>` IIRC.

PR #506, opened 2018-02-02
add format option dialog

wfeldt:
changelog still needs to be updated


src/lib/y2partitioner/widgets/mkfs_options.rb
```diff
@@ -0,0 +1,161 @@
...
+      def handle(event)
+        case event["ID"]
+        when :help
+          Yast::Wizard.ShowHelp(help)
+        end
+      end
+
+      def contents
+        # FIXME: add some VSpacing(1)?
+        # ???: contents is called 3 times for each dialog, so cache it
```
ancorgs: Yes, the object oriented machinery on top of CWM does unexpected stuff in several cases. I also had to introduce some caches for the same reason... some of them way more weird than simply `@contents ||= `. :-)

src/lib/y2partitioner/widgets/mkfs_optiondata.rb
```diff
@@ -0,0 +1,254 @@
+require "yast"
+
+module Y2Partitioner
+  module Widgets
+    # a class
+    class MkfsOptiondata
```
ancorgs: If you never create an instance of this think (and you doesn't), then this should not be a class. I'm still trying to wrap my head around the whole thing to decide what it should be. :-)

ancorgs: It would make more sense from the object orientation POV if every instance of this class would correspond to each entry of the `FOO` array. So you could do something like:
```ruby
options = MkfsOptiondata.options_for(filesystem) # Array of objects, no array of hashes
option.first.default #=> "none"
options.first.default_is_boolean? #=> More readable than that case condition you have down there in other part of the code
...
```

ancorgs: All this seems to be designed to, somehow, avoid too much subclassing. For example, two comboboxes should be of the corresponding ComboBox class, even if they access different properties or have an slightly different validation. But the philosophy of the CWM OOP layer is (as weird as it sounds at first) to have a separate class for each widget. So you are expected to actually define something like `BlockSize < ComboBox` and `BytesPerInode < ComboBox`.

If you have a different class for each widget everything will flow better and you shouldn't need that many complex static methods, custom handlers for `help` or hacks like `handle_all_events = true`.

So, making a real class from each one of the entries in that FOO array would be the correct `CWM` approach. IMHO.

Josef is the authority in OOP-CWM though, if you want a second opinion. :wink:

wfeldt: > I'm still trying to wrap my head around the whole thing to decide what it should be. :-)

Well, in `widgets/help.rb` you made it a module. :-)

I'm not convinced I should turn the hashes into classes just for the nicer look  ATM the code is small and still easy to follow.

> But the philosophy of the CWM OOP layer is [...] to have a separate class for each widget

To have a unique `#widget_id` seems to be the requirement.

Getting the help button to work was strange, though. The code is copied from the fstab dialog. I haven't seen any other help button in the expert partitioner so far.

src/lib/y2partitioner/widgets/mkfs_optiondata.rb
```diff
@@ -0,0 +1,465 @@
...
+
+      # Read current option value from file system class.
+      #
+      # @param filesystem [Y2Storage::Filesystems]
+      #
+      # @return [String, Boolean]
+      #
+      def get(filesystem)
+        # current filesystem settings
+        fs_str = filesystem.send program + "_options"
```
ancorgs: The SUSE style guide encourages to use something like this here:
```ruby
fs_str = filesystem.send(program + "_options)
```
As much as I personally like omitting parenthesis in method calls, most people coming from other languages can be uttered confused by that practice.

src/lib/y2partitioner/widgets/mkfs_optiondata.rb
```diff
@@ -0,0 +1,465 @@
...
+
+        value = default
+
+        # find option args in current settings; if it's a non-boolean
+        # option get the value from the match
+        m = fs_str.match(/(^|\s)#{Regexp.escape(option_str)}(\S*)/)
+        value = boolean? ? !default : m[2] if m
```
ancorgs: Do we really need to make this line so cryptic combining `?` with an `if` at the end? I know there is a perl developer in your heart but... :wink:

Is this the same? Because I think it's, be far, more self-explanatory
```ruby
def get(filesystem)
  # current filesystem settings
  fs_str = filesystem.send(program + "_options)

  # find option args in current settings; if it's a non-boolean
  # option get the value from the match
  m = fs_str.match(/(^|\s)#{Regexp.escape(option_str)}(\S*)/)

  return default if m.nil?
  boolean? ? !default : m[2]
end
```

src/lib/y2partitioner/widgets/mkfs_optiondata.rb
```diff
@@ -0,0 +1,465 @@
...
+      #
+      # This calls the validation function if there is one defined _and_ an
+      # error message exists. Else it just returns true.
+      #
+      # @param val [String, Boolean]
+      #
+      # @return [Boolean]
+      #
+      def validate?(val)
+        return true unless validate && error
```
ancorgs: Hmmm, `#error` will always return something true-ish, isnt'it? As it's defined below, it will at least return `"\n\n"`

src/lib/y2partitioner/widgets/mkfs_optiondata.rb
wfeldt: The code is rubocop's suggestion, not mine, btw. So I assumed this is how rubyists like it. :-)

src/lib/y2partitioner/widgets/mkfs_optiondata.rb
wfeldt: And this is not enforced by rubocop?

src/lib/y2partitioner/widgets/mkfs_optiondata.rb
```diff
@@ -0,0 +1,465 @@
...
+    private
+
+      # Make option hash entries readable via methods.
+      #
+      # @param foo [Symbol]
+      #
+      # @return [Object]
+      #
+      def method_missing(foo)
+        @value[foo]
```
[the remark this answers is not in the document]
wfeldt: I cannot. It's expected to return nil in cases where the hash key does not exist.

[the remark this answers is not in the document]
ancorgs: Oh. So in this case hiding the programmer typos is a feature instead of a bug?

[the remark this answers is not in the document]
wfeldt: No, it's about this class representing a Hash with some added functionality. A missing entry is expected to return nil.

ancorgs:
I would normally would be WAY more demanding about the unit tests :wink:, but due to time pressures, this LGTM.

I will merge #519 as soon as Travis passes again, so you can rebase and add changelog entry and version bump.

PR #507, opened 2018-02-05
Unit test showcasing bug#1078691

ancorgs:
QA found a scenario in which we propose no "/"!!  We try to create a partition for it, but it somehow fails and, instead of reporting that, we continue with no "/". This test case should allow to reproduce it and fix it.

See https://bugzilla.suse.com/show_bug.cgi?id=1078691


wfeldt:
the minimal test case is here: https://github.com/yast/yast-storage-ng/pull/522

ancorgs:
Closed in favor of #522

PR #510, opened 2018-02-06
Add mount point class

joseivanlopez:
PBI: https://trello.com/c/NYy7AklQ/144-5-use-new-libstorage-ng-api-for-mount-points


src/lib/y2partitioner/actions/controllers/filesystem.rb
```diff
@@ -137,65 +152,49 @@ def apply_role
...
+          mount_path = current_value_for(:mount_point)
           mount_by = current_value_for(:mount_by)
           label = current_value_for(:label)
 
           if type.is?(:swap)
-            mount_point = "swap"
-          elsif mount_point == "swap"
-            mount_point = ""
+            mount_path = "swap"
+          elsif mount_path == "swap"
```
shundhammer: This part I don't understand. I can see that it was very much like this before, but still, does this really work?
If we have a type :swap, we set a mount_path to "swap". But if it was not :swap, but already had a mount_path "swap", the mount_path is reset to nil. Isn't that contradictory? Shouldn't this behave the same way in both cases - when the type is :swap or when the mount_path is "swap"?

src/lib/y2partitioner/actions/controllers/filesystem.rb
```diff
@@ -228,25 +227,84 @@ def partition_id=(partition_id)
...
+        #   options are given (see {Y2Storage::Filesystems::Type#default_mount_options}).
+        #
+        #   Take into account that modifying the mount point can have side effects
+        #   if the filesystem doesn't exist yet, like changing the list of
+        #   subvolumes if the new or old mount point is "/".
+        #
+        # @param path [String]
+        # @param options [Hash] options for the mount point (e.g., { mount_by: :uuid } )
+        def create_mount_point(path, options = {})
+          return if filesystem.nil? || !mount_point.nil?
```
shundhammer: IMHO this is a bit hard to read: It checks for two completely different things:
- Do we already have a filesystem? Return if we don't.
- Do we already have a mount point? Return if we do.
I know that it is correct in the logical sense, but you have to do mental acrobatics to figure out exactly when this returns.

src/lib/y2partitioner/actions/controllers/filesystem.rb
```diff
@@ -344,21 +405,24 @@ def create_filesystem(type)
...
-          self.mount_point = mount_point
+          filesystem.label = label if label
+
+          restore_mount_point(mount_path, mount_by: mount_by)
         end
 
-        def assign_fs_attrs(attrs = {})
-          attrs.each_pair do |attr, value|
-            filesystem.send(:"#{attr}=", value) unless value.nil?
+        def apply_mount_point_options(options)
```
shundhammer: Since it's no longer obvious like in the old code what type "options" is, please document it.

src/lib/y2partitioner/actions/controllers/filesystem.rb
```diff
@@ -369,19 +433,50 @@ def current_value_for(attribute)
...
+        # @return [Hash]
+        def role_values
+          fs_type = mount_path = mount_by = nil
+
+          case role
+          when :swap
+            part_id = Y2Storage::PartitionId::SWAP
+            fs_type = Y2Storage::Filesystems::Type::SWAP
+            mount_path = "swap"
+            mount_by = Y2Storage::Filesystems::MountByType::DEVICE
```
shundhammer: Not sure: Does this always enforce MountByType::DEVICE for swap? Is there still a chance to set LABEL or UUID despite this? Is this only a default, or is it the only possible value?

src/lib/y2partitioner/actions/controllers/lvm_vg.rb
```diff
@@ -419,14 +419,13 @@ def used?(device)
           true
         end
 
-        # Checks whether a device is mounted
+        # Checks whether a device has mount point
```
shundhammer: NP: "has **a** mount point"

src/lib/y2partitioner/actions/controllers/md.rb
```diff
@@ -249,8 +249,7 @@ def available?(partition)
           return false if partition.md
           return true if partition.filesystem.nil?
 
-          mount = partition.filesystem.mountpoint
-          mount.nil? || mount.empty?
+          partition.filesystem.mount_point.nil?
```
shundhammer: Is it possible that we have a mount point, but that mount point doesn't have a path or the path is empty? What would happen in that case?

src/lib/y2partitioner/actions/controllers/lvm_vg.rb
```diff
@@ -419,14 +419,13 @@ def used?(device)
...
         #
         # @return [Boolean] true if it is mounted; false otherwise
         def mounted?(device)
           return false unless formatted?(device)
 
-          mount_point = device.filesystem.mount_point
-          mount_point && !mount_point.empty?
+          !device.filesystem.mount_point.nil?
```
shundhammer: Is it possible that we have a mount point, but that mount point doesn't have a path or the path is empty? What would happen in that case?

src/lib/y2partitioner/actions/controllers/md.rb
shundhammer: Same as above: Can there be a mount point, but with a nil/empty path?

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -183,7 +204,7 @@ def refresh
         if filesystem
           Yast::UI.ChangeWidget(Id(:mount_device), :Enabled, true)
 
-          if filesystem.mountpoint.nil? || filesystem.mountpoint.empty?
+          if filesystem.mount_path.nil? || filesystem.mount_path.empty?
```
shundhammer: If we need that more often, how about a function `mount_path?` that checks if non-nil and non-empty?

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -250,26 +271,29 @@ def handle(event)
...
+        @controller.remove_mount_point
         @fstab_options_widget.disable
         @mount_point_widget.disable
       end
 
       def mount_point_change
-        mount_point = @mount_point_widget.value.to_s
-        @controller.mount_point = mount_point
-        if mount_point.nil? || mount_point.empty?
+        if mount_path.empty?
```
shundhammer: Is this safe since at other places we also check for `nil?`  ?

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -517,7 +541,7 @@ def duplicated_mount_point?
       def mounted_devices
         fs_sids = filesystem_devices.map(&:sid)
         devices = Y2Storage::Mountable.all(device_graph)
-        devices = devices.select { |d| !d.mount_point.nil? && !d.mount_point.empty? }
+        devices = devices.select { |d| !d.mount_path.nil? && !d.mount_path.empty? }
```
shundhammer: A `mount_point?` method would again come very handy here ;-)

src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -60,29 +77,40 @@ def ui_term_with_vspace(widget)
...
-        filesystem.fstab_options = new_options
+        # The options can only be modified using MountPoint#mount_options=
+        mount_point.mount_options = mount_point.mount_options + options
       end
 
       alias_method :add_fstab_option, :add_fstab_options
 
     private
 
+      # Currently editing filesystem
```
shundhammer: NP: A bit easy to misunderstand... how about "filesystem currently being edited" ?

src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -60,29 +77,40 @@ def ui_term_with_vspace(widget)
...
+
       # Common regexp checkbox widgets init.
       def init_regexp
-        i = filesystem.fstab_options.index { |o| o =~ self.class::REGEXP }
+        i = mount_point.mount_options.index { |o| o =~ self.class::REGEXP }
 
         self.value =
           if i
-            filesystem.fstab_options[i].gsub(self.class::REGEXP, "")
+            mount_point.mount_options[i].gsub(self.class::REGEXP, "")
```
shundhammer: :thumbsup: for bringing more consistency into that naming that has somewhat gone out of hand (mount_options vs. fstab_options)

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -295,11 +295,12 @@ def init_filesystem_fields(partition)
 
       # @param fs [Filesystem::BlkFilesystem]
       def init_mount_options(fs)
-        if fs.mountpoint && !fs.mountpoint.empty?
-          @mount = fs.mountpoint
-          @mountby = fs.mount_by.to_sym
+        if !fs.mount_point.nil?
+          @mount = fs.mount_point.path
```
shundhammer: Since we are dealing with so many mount-related things here, is it really a good idea to name a member variable @mount and not something more precise like @mount_path?

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -295,11 +295,12 @@ def init_filesystem_fields(partition)
...
+          @mountby = fs.mount_point.mount_by.to_sym
```
shundhammer: Consistency: @mountby here vs. mount_by everywhere else in the code

src/lib/y2partitioner/actions/controllers/filesystem.rb
joseivanlopez: Well, the idea is always set mount_path to "swap" when selected fs is swap. In case that selected fs is other different to swap, we leave the current mount_path (if there is one). In this last case, we should ensure that the reused mount path is not "swap", because that does not make sense.

src/lib/y2partitioner/actions/controllers/filesystem.rb
joseivanlopez: Totally agree. I will split it.

src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -142,20 +141,27 @@ def setup_filesystem(filesystem)
         filesystem.label = label if label
         filesystem.uuid = uuid if uuid
         filesystem.mkfs_options = mkfs_options if mkfs_options
-        filesystem.mount_by = mount_by if mount_by
-        setup_fstab_options(filesystem)
+        setup_mount_point(filesystem)
+      end
+
+      def setup_mount_point(filesystem)
+        assign_mountpoint(filesystem)
```
shundhammer: Consistency: assign_mountpoint() vs. xxx_mount_point() everywhere.

NP: And it's also `set_up_mount_point()`; "setup" is the noun. The verb is "to set up".
</wiseass_mode>  ;-)

src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -142,20 +141,27 @@ def setup_filesystem(filesystem)
...
       end
 
       # Set the fstab options, either those that were explicitly set, or the
       # defaults for this filesystem type
       #
-      # @param filesystem [Filesystems::BlkFilesystem]
-      def setup_fstab_options(filesystem)
-        return unless filesystem
+      # @param mount_point [MountPoint]
+      def setup_fstab_options(mount_point)
```
shundhammer: NP: Similar: `set_up_fstab_options`

src/lib/y2partitioner/actions/controllers/filesystem.rb
shundhammer: OK.

src/lib/y2partitioner/actions/controllers/filesystem.rb
joseivanlopez: It is only a default. It could be changed using the button for "fstab options".

src/lib/y2partitioner/actions/controllers/md.rb
joseivanlopez: The same applies here.

src/lib/y2partitioner/widgets/format_and_mount.rb
joseivanlopez: I think it is not needed. In fact, that line can be replaced by `filesystem.mount_point.nil?`

src/lib/y2partitioner/widgets/format_and_mount.rb
joseivanlopez: In this case, `mount_path` is a method belonging to the widget. It simply returns the string given by the user for the mount point, so we only check that the string is not empty.

src/lib/y2partitioner/widgets/format_and_mount.rb
joseivanlopez: Once again, it can be replaced by `!d.mount_point.nil?`

src/lib/y2storage/autoinst_profile/partition_section.rb
joseivanlopez: Checking it with Imo, that names correspond to the names in the autoyast xml. For now, I would not change them because we would have to adapt a lot of code.

src/lib/y2storage/autoinst_profile/partition_section.rb
joseivanlopez: The same than above.

src/lib/y2storage/autoinst_profile/partition_section.rb
shundhammer: OK then.

src/lib/y2storage/autoinst_profile/partition_section.rb
shundhammer: OK

shundhammer:
Completely LGTM now.

PR #511, opened 2018-02-06
Reuse already existing (and even formatted) partitions as RAID or PV

imobachgs:
It should fix [bsc#1077277](https://bugzilla.suse.com/show_bug.cgi?id=1077277).

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Wed Feb  7 10:57:26 UTC 2018 - igonzalezsosa@suse.com
 
-- Support reuse of already existing partitions as LVM physical
-  volumes or MD RAIDs (bsc#1077277).
+- AutoYaST: support reuse of already existing partitions as LVM
+  physical volumes or MD RAIDs (bsc#1077277).
 - 4.0.85
 
 -------------------------------------------------------------------
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Feb  7 10:57:26 UTC 2018 - igonzalezsosa@suse.com
+
+- Support reuse of already existing partitions as LVM physical
+  volumes or MD RAIDs (bsc#1077277).
+- 4.0.85
+
+-------------------------------------------------------------------
 Wed Feb  7 00:18:34 UTC 2018 - ancor@suse.com
 
 - Partitioner: fixed 'Device Graph' section (part of fate#318196).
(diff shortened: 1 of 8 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Feb  7 10:57:26 UTC 2018 - igonzalezsosa@suse.com
+
+- Support reuse of already existing partitions as LVM physical
+  volumes or MD RAIDs (bsc#1077277).
```
ancorgs: You should explicitly mention AutoYaST here.

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -102,8 +102,11 @@ def create_volume_group(planned_vg, devicegraph)
       # @param devicegraph [Devicegraph] to fetch the partitions
       def assign_physical_volumes(volume_group, part_names, devicegraph)
         partitions = part_names.map { |n| BlkDevice.find_by_name(devicegraph, n) }
+        used_devices = volume_group.lvm_pvs.map(&:blk_device)
         partitions.each do |partition|
+          next if used_devices.include?(partition) # already a PV for this VG
```
ancorgs: So here we only check for partitions that are directly a PV (no encryption) but in the next line, we seem to cope more or less transparently with possible encryption layer. My arachnid sense screams.

ancorgs: Wouldn't it be more consistent using?
```ruby
used_devices = volume_group.lvm_pvs.map(&:plain_blk_device)
```

ancorgs: Another alternative would be to not have `used_devices` and do this check instead (I don't have a preference).
```ruby
next if partition.lvm_pv && partition.lvm_pv.lvm_vg == volume_group
```

src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -102,8 +102,11 @@ def create_volume_group(planned_vg, devicegraph)
       # @param devicegraph [Devicegraph] to fetch the partitions
       def assign_physical_volumes(volume_group, part_names, devicegraph)
         partitions = part_names.map { |n| BlkDevice.find_by_name(devicegraph, n) }
+        used_devices = volume_group.lvm_pvs.map(&:plain_blk_device)
         partitions.each do |partition|
           device = partition.encryption || partition
+          next if used_devices.include?(device) # already a PV for this VG
+          device.remove_descendants
```
ancorgs: You fixed it twice... so now it's wrong again:-) `used_device` is now never an `Encryption` but `device` can be.

Either you use `map(&:blk_device)` again or you move the `next if` again to it's original location.

PR #515, opened 2018-02-08
AutoYaST: use find_by_any_name

imobachgs:
This change adds support to use devices names like `/dev/disk/by-label/some-label` in `drive/device` element.


imobachgs:
Thanks!

PR #516, opened 2018-02-08
Partitioner: fix for bsc#1078721

ancorgs:
This fixes https://trello.com/c/ahhE9izy/260-3-bsc1078721-p1-openqa-test-fails-in-partitioningfulllvm-partitions-not-deleted-after-creating-partition-table

* Fixes the bug
* Modify unit tests to actually test something meaningful

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/partition_table.rb
+++ b/src/lib/y2partitioner/actions/controllers/partition_table.rb
@@ -49,33 +49,34 @@ module Y2Partitioner
         # @return [String]
         attr_reader :disk_name
 
-        # The disk we are working on
-        # @return [Y2Storage::Disk] or [Y2Storage::Dasd]
-        attr_reader :disk
-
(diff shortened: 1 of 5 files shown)


src/lib/y2partitioner/actions/controllers/partition_table.rb
```diff
@@ -49,33 +49,34 @@ class PartitionTable
...
 
           @disk_name = disk_name
-          @disk = find_disk(disk_name)
-          log.error("Can't find disk #{@disk_name}") if @disk.nil?
+          log.error("Can't find disk #{@disk_name}") if disk.nil?
           @type = possible_partition_table_types.first
         end
 
+        # The disk we are working on
+        # @return [Y2Storage::Disk] or [Y2Storage::Dasd]
```
joseivanlopez: NP: It could be any partitionable.

PR #517, opened 2018-02-09
New finish client to enable multipathd

ancorgs:
Implements https://trello.com/c/dXnAZX6L/261-sles15-p1-1076183-multipathdservice-isnt-enabled-after-installation-on-multipath

Already tested manually

Bonus: update dependency on libstorage-ng to fix another reported bug

### DIFF
--- a/src/lib/y2storage/clients/finish.rb
+++ b/src/lib/y2storage/clients/finish.rb
@@ -41,12 +41,6 @@ module Y2Storage
         _("Saving file system configuration...")
       end
 
-      # Modes in which the client should run
-      # @return [Array<Symbol>]
-      def modes
-        [:installation, :update, :autoinst, :autoupg, :live_installation]
-      end
-
       # Performs the final actions in the target system
       def write
         return unless multipath?
--- a/src/lib/y2storage/clients/finish.rb
+++ b/src/lib/y2storage/clients/finish.rb
@@ -1,0 +1,67 @@
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/clients/finish.rb
```diff
@@ -0,0 +1,67 @@
...
+require "installation/finish_client"
+
+module Y2Storage
+  module Clients
+    # Finish client for storage-related tasks
+    class Finish < ::Installation::FinishClient
+      # Constructor
+      def initialize
+        textdomain "storage"
+        Yast.import "Service"
```
jreidinger: NP: just matter of taste, but Yast.import is almost same as `require` so can be together unless there is reason to import it only in constructor

src/lib/y2storage/clients/finish.rb
```diff
@@ -0,0 +1,67 @@
...
+      # @return [String]
+      def title
+        # progress step title
+        _("Saving file system configuration...")
+      end
+
+      # Modes in which the client should run
+      # @return [Array<Symbol>]
+      def modes
+        [:installation, :update, :autoinst, :autoupg, :live_installation]
```
jreidinger: why not keep default? To run it always?

src/lib/y2storage/clients/finish.rb
ancorgs: I checked it and most finish clients do it that way. So I decided to stay consistent.

src/lib/y2storage/clients/finish.rb
ancorgs: Oh, good point. I overlooked that in the documentation. I will change it and retry.

PR #519, opened 2018-02-12
Better sorting of boot-related partitions

ancorgs:
See https://trello.com/c/7NcmMuGq/264-sles15-p1-shipstopper-1076851-build-4201-and-4211-ppc64le-unable-to-boot-after-installation

When creating several of our proposed partitions in the same disk chunk, this ensures:
 - Any boot-related partition goes before (in the disk layout) than partitions coming from control.xml (e.g. '/').
 - PReP goes before `/boot` if both are proposed

This is only about using a reasonable order for adjacent partitions created by the proposal. This guarantees nothing about a given partition being at the beginning of the disk. So this is a partial fix for:

- [bsc#107685](https://bugzilla.suse.com/show_bug.cgi?id=1076851) which claims that PReP should be the first partition in the disk
- [bsc#1073680](https://bugzilla.suse.com/show_bug.cgi?id=1073680) which claims the same about `/boot/efi`... to please buggy EFI implementations


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Feb 12 13:13:35 UTC 2018 - ancor@suse.com
+
+- More reasonable location in the disk for the partitions proposed
+  to make the system bootable (bsc#1073680 and bsc#1076851).
+
```
joseivanlopez: Increase version?

ancorgs: Well, I wanted to coordinate with other changes being probably merged today

wfeldt:
Just one thought: could you collect the referenced bugs in boot-partition.md or some similar place?

doc/boot-partition.md
```diff
@@ -1,13 +1,24 @@
...
 
 #### Notes:
 
 - outright uncertain issues are tagged with '???', please try to clear them up
 - partition sizes are given as triples: [minimal, optimal, maximal]
 
+### Relevant bugs during SLE15 beta phase
+
+- In [bsc#107685](https://bugzilla.suse.com/show_bug.cgi?id=1076851) and
+  [bsc#1076851](https://bugzilla.suse.com/show_bug.cgi?id=1076851) it has been
```
wfeldt: that's the same as one line above

wfeldt: ... and a type in the bug number a line above

wfeldt:
lgtm

PR #520, opened 2018-02-12
Fix sanity checks

joseivanlopez:
PBI: https://trello.com/c/qtgGEDBm/259-1-efi-weaken-the-expert-partition-checks

### DIFF
--- a/test/y2storage/reusing_efi.rb
+++ b/test/y2storage/reusing_efi.rb
@@ -1,0 +1,177 @@
+#!/usr/bin/env rspec
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require_relative "spec_helper"
+require "y2storage"
+
+describe Y2Storage::BootRequirementsChecker do
+  using Y2Storage::Refinements::SizeCasts
+
+  before do
+    Y2Storage::StorageManager.create_test_instance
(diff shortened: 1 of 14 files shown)


test/y2storage/reusing_efi.rb
```diff
@@ -0,0 +1,120 @@
...
+
+  RSpec.shared_examples "reuses EFI" do
+    it "reuses the existing EFI partition" do
+      expect(checker.needed_partitions).to contain_exactly(
+        an_object_having_attributes(mount_point: "/boot/efi", reuse: "/dev/sda1")
+      )
+    end
+  end
+
+  context "when checking for an EFI partition" do
```
ancorgs: Not sure if I like this test in its current form. The output now looks something like:
```markdown
- if there are no EFI partitions
  - **it requires a new /boot/efi partition**

- if there is already an EFI partition
  - and it is too small
    - **it requires a new /boot/efi partition**
  - and it is too big (FIXME: it should not reuse EFI)
    - **it requires a new /boot/efi partition**
  - and it has a valid size
    - and its partition id is not ESP (FIXME: it should reuse EFI)
      - **it requires a new /boot/efi partition**
    - and its partition id is ESP
      - **it reuses the existing EFI partition**
```

That:
  - Leaves many combinations without checking. Too few scenarios to cover the three variables we have here: id (ESP/other), filesystem (none/vfat/other) and size (big/small/okish).
  - It seems to suggest we should discard big partitions (which I don't think is always true) or reuse all partitions with a proper size (which is false).

I would prefer something like this (not that much work to code):

```markdown
- if there is a partition without filesystem
  - and it is smaller than the proposal min
    - and the id is ESP
      - **it does not reuse the partition**
    - and the id is not ESP
      - **it does not reuse the partition**
  - and it is bigger than the proposal max
    - _# exactly the same than above (two examples)_
  - and it has a proper size
    - _# exactly the same than above (two examples)_

- if there is a partition with a non VFAT filesystem
  - _# exactly the same than above (six examples)_

- if there is a partition formatted as VFAT
  - and it is smaller than the proposal min
    - and the id is ESP
      - **it does not reuse the partition**
    - and the id is not ESP
      - **it does not reuse the partition**
  - and it is bigger than the proposal max
    - and the id is ESP
      - **it reuses the partition (FIXME: should we check content?)**
    - and the id is not ESP
      - **it does not reuse the partition (FIXME: should we check content?)**
  - and it has a proper size
    - and the id is ESP
      - **it reuses the partition (FIXME: should we check content?)**
    - and the id is not ESP
      - **it does not reuse the partition (FIXME: should we check content?)**
```

Which (in addition to fixing the mentioned "issues"):

 - Introduces a unit test for something we were doing wrong in the past (trusting the ID without checking for a filesystem)
 - Makes quite easy to remove ESP from the test (if we want), once it becomes irrelevant

In short, this is just my usual complain: if we don't test all the possible combination of the involved variables we are mainly repeating the implementation in RSpec format.

PR #521, opened 2018-02-14
[WIP] first version of fatal errors

jreidinger:
plan is to also add all current checks in bootloader itself.
screenshot from manual testing:

snimek obrazovky_2018-02-15_09-44-19

and warnings still works:

snimek obrazovky_2018-02-15_09-48-26

### DIFF
--- a/src/lib/y2partitioner/setup_errors_presenter.rb
+++ b/src/lib/y2partitioner/setup_errors_presenter.rb
@@ -44,13 +44,13 @@ module Y2Partitioner
     #
     # @return [String]
     def to_html
-      fatal_errors = fatal_errors_html
-      return fatal_errors if fatal_errors
+      errors = errors_html
+      return errors if errors
 
-      errors = [boot_errors_html, product_errors_html].compact
-      return "" if errors.empty?
+      warnings = [boot_warnings_html, product_warnings_html].compact
+      return "" if warnings.empty?
 
-      errors.join(Yast::HTML.Newline)
+      warnings.join(Yast::HTML.Newline)
     end
 
   private
@@ -58,37 +58,37 @@ module Y2Partitioner
     # @return [SetupChecker] checker for the current setup
     attr_reader :setup_checker
 
-    # HTML representation for boot errors
+    # HTML representation for boot warnings
     #
-    # @return [String, nil] nil if there is no boot error
-    def boot_errors_html
-      errors = setup_checker.boot_errors
+    # @return [String, nil] nil if there is no boot warning
+    def boot_warnings_html
+      warnings = setup_checker.boot_warnings
       # TRANSLATORS
       header = _("The system might not be able to boot:\n")
 
-      errors_html(header, errors)
+      create_html(header, warnings)
     end
 
-    # HTML representation for mandatory product errors
+    # HTML representation for mandatory product warnings
     #
-    # @return [String, nil] nil if there is no product error
-    def product_errors_html
-      errors = setup_checker.product_errors
+    # @return [String, nil] nil if there is no product warning
+    def product_warnings_html
+      warnings = setup_checker.product_warnings
       # TRANSLATORS
-      header = _("The system could not work properly because the following errors were found:\n")
+      header = _("The system could not work properly because the following warnings were found:\n")
 
-      errors_html(header, errors)
+      create_html(header, warnings)
     end
 
     # HTML representation for fatal booting errors
     #
-    # @return [String, nil] nil if there is no product error
-    def fatal_errors_html
-      errors = setup_checker.fatal_errors
+    # @return [String, nil] nil if there is no error
+    def errors_html
+      errors = setup_checker.errors
       # TRANSLATORS
       header = _("The system cannot be installed because the following errors were found:\n")
 
-      errors_html(header, errors)
+      create_html(header, errors)
     end
 
     # HTML representation for a set of errors
@@ -99,7 +99,7 @@ module Y2Partitioner
     # @param errors [Array<SetupError>] list of errors
     #
     # @return [String, nil] nil if there is no error
-    def errors_html(header, errors)
+    def create_html(header, errors)
       return nil if errors.empty?
 
       error_messages = errors.map(&:message)
--- a/src/lib/y2partitioner/setup_errors_presenter.rb
+++ b/src/lib/y2partitioner/setup_errors_presenter.rb
@@ -44,7 +44,7 @@ module Y2Partitioner
     #
     # @return [String]
     def to_html
-      errors_html || warnings_html || nil
+      errors_html || warnings_html || ""
     end
 
   private
--- a/src/lib/y2partitioner/setup_errors_presenter.rb
+++ b/src/lib/y2partitioner/setup_errors_presenter.rb
@@ -44,19 +44,23 @@ module Y2Partitioner
     #
     # @return [String]
     def to_html
-      errors = errors_html
-      return errors if errors
-
-      warnings = [boot_warnings_html, product_warnings_html].compact
-      return "" if warnings.empty?
-
-      warnings.join(Yast::HTML.Newline)
+      errors_html || warnings_html || nil
     end
 
   private
 
     # @return [SetupChecker] checker for the current setup
     attr_reader :setup_checker
+
+    # HTML representation for boot warnings
+    #
+    # @return [String, nil] nil if there is no boot warning
+    def warnings_html
+      warnings = [boot_warnings_html, product_warnings_html].compact
+      return nil if warnings.empty?
+
+      warnings.join(Yast::HTML.Newline)
+    end
 
     # HTML representation for boot warnings
     #
@@ -75,7 +79,7 @@ module Y2Partitioner
     def product_warnings_html
       warnings = setup_checker.product_warnings
       # TRANSLATORS
-      header = _("The system could not work properly because the following warnings were found:\n")
+      header = _("The system could not work properly because the following product requirements were not fulfilled:\n")
 
       create_html(header, warnings)
     end
--- a/src/lib/y2partitioner/setup_errors_presenter.rb
+++ b/src/lib/y2partitioner/setup_errors_presenter.rb
@@ -79,7 +79,10 @@ module Y2Partitioner
     def product_warnings_html
       warnings = setup_checker.product_warnings
       # TRANSLATORS
-      header = _("The system could not work properly because the following product requirements were not fulfilled:\n")
+      header = _(
+        "The system could not work properly because the following product " \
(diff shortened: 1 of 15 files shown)


src/lib/y2storage/boot_requirements_strategies/prep.rb
```diff
@@ -60,6 +56,9 @@ def errors
     protected
 
       def boot_partition_needed?
+        # FIXME: I (JR) think this is needed only for powernv where firmware have
+        #        to load grub2 configuration and kernel itself. For prep case
+        #        grub2 loads it itself.
```
joseivanlopez: Yes. I am fixing this here https://github.com/yast/yast-storage-ng/pull/520

src/lib/y2storage/boot_requirements_strategies/zipl.rb
```diff
@@ -48,6 +46,17 @@ def errors
         errors
       end
 
+      def fatal_errors
+        res = super
+        return res unless res.empty?
+
+        if missing_partition_for?(zipl_volume)
```
joseivanlopez: Why to check with a volume and report the error with another?

jreidinger: @joseivanlopez bug...btw in bootloader check is that /boot/zip *path* is on ext4. So we are fine if /boot is ext4 and no separate /boot/zipl needed or even / as ext4. how I can express it with volume?

src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -77,9 +77,29 @@ def needed_partitions(target)
       #
       # @return [Array<SetupError>]
       def errors
+        raise Error, "There are fatal errors, so cannot detect soft errors." unless fatal_errors.empty?
```
joseivanlopez: Raise an exception here looks me a little bit strong.

joseivanlopez:
A general comment. In some cases we could have both: "fatal errors" and "errors". For example, root is missing (fatal error) and also /boot/efi is missing (error). We could show both errors at the same time, and so the use knows what he needs to fix his configuration. With current version, we show first only fatal errors, and then errors.

Moreover, what about renaming it as "errors" and "warnings"?

I would show all of them together (what makes sense) in the same popup, and I will activate the proper buttons depending on if there are only warnings or not.

jreidinger:
@joseivanlopez I am considering that having together warning and errors and it have one problem that it confuse users. We show popup where is:

1 fatal entry
X boot entries ( e.g. when / missing or partition table I expect at least 5 another warnings )
Y product entries ( with missing / I expect another bunch of warnings )
and there is only 1 fatal.

So I decide to show when we have blocker only blocker to really focus on it. and when we have set of warnings, show all of them, as user have to agree with all of them.

Renaming sounds reasonable for me.

src/lib/y2storage/setup_checker.rb
```diff
@@ -55,43 +55,43 @@ def initialize(devicegraph)
...
-      boot_errors.empty?
+    # @return [Array<SetupError>]
+    def errors
+      boot_requirements_checker.errors
     end
 
-    # All storage errors detected in the setup, for example, when a /boot/efi partition
-    # is missing in a UEFI system, or a valid partition for root is missing.
+    # All storage warnings detected in the setup, for example, when a /boot/efi partition
+    # is missing in a UEFI system, or missing separate /boot.
```
joseivanlopez: NP: I would mention a case where a required product volume is missing: `in a UEFI system, or a required product volume (e.g., swap) is missing.`

src/lib/y2storage/setup_checker.rb
```diff
@@ -138,5 +138,9 @@ def mandatory?(volume)
     def missing?(volume)
       BlkDevice.all(devicegraph).none? { |d| d.match_volume?(volume) }
     end
+
+    def boot_requirements_checker
```
joseivanlopez: NP: I know that it is pretty clear, but please, comment at least the return type.

src/lib/y2partitioner/setup_errors_presenter.rb
```diff
@@ -44,37 +44,51 @@ def initialize(setup_checker)
...
+    # HTML representation for mandatory product warnings
     #
-    # @return [String, nil] nil if there is no product error
-    def product_errors_html
-      errors = setup_checker.product_errors
+    # @return [String, nil] nil if there is no product warning
+    def product_warnings_html
+      warnings = setup_checker.product_warnings
+      # TRANSLATORS
+      header = _("The system could not work properly because the following warnings were found:\n")
```
joseivanlopez: NP: I would avoid the word `warnings` there. An option could be: `The system could not work properly because the following issues were found:`

src/lib/y2partitioner/setup_errors_presenter.rb
```diff
@@ -44,37 +44,51 @@ def initialize(setup_checker)
     #
     # @return [String]
     def to_html
```
joseivanlopez: NP: I would simplify the method a little bit by defining a `#warnings_html` in a similar way to `#errors_html`:

```
def to_html
  errors_html || warnings_html || ""
end
```

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -209,6 +222,9 @@
       allow_any_instance_of(Y2Storage::BootRequirementsStrategies::ZIPL)
         .to receive(:zipl_volume).and_return(zipl_volume)
 
+      allow_any_instance_of(Y2Storage::BootRequirementsStrategies::ZIPL)
+        .to receive(:minimal_zipl_volume).and_return(zipl_volume)
```
joseivanlopez: Is this needed?

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -268,7 +284,7 @@
         context "when boot device has no partition table" do
           let(:boot_partition_table) { nil }
 
-          it "contains an error for unknown partition table" do
+          it "contains an fatal error for unknown partition table" do
```
joseivanlopez: NP: typo `a fatal`

src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -68,18 +68,35 @@ def needed_partitions(target)
...
+      #
+      # @note This method can be overloaded for derived classes.
+      #
+      # @see SetupError
+      #
+      # @return [Array<SetupError>]
+      def errors
+        res = []
+        if root_filesystem_missing?
+          error_message = _("There is no device mounted to '/'")
```
joseivanlopez: NP: `There is no device mounted at '/'` ??

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -177,8 +183,7 @@ def errors_on_msdos
       def unknown_boot_partition_table_error
         # TRANSLATORS: error message
         error_message = _(
-          "Boot requirements cannot be determined because " \
-          "boot disk has not partition table"
+          "Boot disk is without partition table"
```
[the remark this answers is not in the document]
jreidinger: I do not like too formal "Boot requirements cannot be determined" it sounds like our error that we cannot determine it. I think we should make it more user oriented. How you will see
`Boot disk is without partition table and we does not support booting from it`

BTW that reminds me that it would be nice to also add to end of message how to fix it. E.g. `You can fix it by creating partition table on disk`.

src/lib/y2storage/setup_checker.rb
jreidinger: :+1:

src/lib/y2partitioner/setup_errors_presenter.rb
jreidinger: makes sense. Maybe we should also mention that it is product specific requirements?

test/y2storage/boot_requirements_errors_test.rb
jreidinger: no longer

src/lib/y2storage/boot_requirements_strategies/base.rb
jreidinger: :+1:

src/lib/y2partitioner/setup_errors_presenter.rb
joseivanlopez: Yes, it makes sense.

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -37,23 +37,29 @@ def needed_partitions(target)
...
-      def errors
-        errors = super
-
-        if root_filesystem_missing?
-          errors << unknown_boot_disk_error
-        elsif boot_partition_table_missing?
-          errors << unknown_boot_partition_table_error
-        elsif boot_ptable_type?(:gpt)
-          errors += errors_on_gpt
+      def warnings
```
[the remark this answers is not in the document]
jreidinger: warnings is in fact never called when error is there ( that is reason why original I had exception there )

src/lib/y2partitioner/setup_errors_presenter.rb
```diff
@@ -44,20 +44,24 @@ def initialize(setup_checker)
...
-      errors = errors_html
-      return errors if errors
-
-      warnings = [boot_warnings_html, product_warnings_html].compact
-      return "" if warnings.empty?
-
-      warnings.join(Yast::HTML.Newline)
+      errors_html || warnings_html || nil
```
joseivanlopez: NP: Strictly speaking, last `|| nil` is not needed, but I can live with it.

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -183,7 +188,8 @@ def errors_on_msdos
       def unknown_boot_partition_table_error
         # TRANSLATORS: error message
         error_message = _(
-          "Boot disk is without partition table"
+          "Boot disk is without partition table and we does not support " \
+            "booting from it. You can fix it by creating partition table on disk."
```
joseivanlopez: NP: I will improve the text a little bit more.

"Boot disk has no partition table and it is not possible to boot from it." \
"You can fix it by creating a partition table on the boot disk."

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -406,7 +406,18 @@
         context "with a partitions-based proposal" do
           let(:use_lvm) { false }
 
-          include_examples "PReP partition"
```
joseivanlopez: Why is this change needed?

PR #522, opened 2018-02-14
minimal test case demonstrating bsc#1078691

wfeldt:


ancorgs:
The order of the commits puzzled me while trying to review in a commit-per-commit basis. How it is than the commit moving `test/y2storage/xxx.rb` to the proper location comes before the commit creating `test/y2storage/xxx.rb`?

ancorgs:
I believe that adding two kind-of-independent `reuse_name` and `reuse_sid` and checking always by both is kind of wrong.  Shouldn't the `Planned::Device` object ensure that both are consistent to each other so you use either of them? I mean:

```ruby
module Planned
  module Device
    # reader, not accessor
    attr_reader :reuse_name
    # same here, just reader
    attr_reader :reuse_sid

   # Yes, I suck at naming. The main point is that both internal attributes are set together
    def select_reuse_object(blk_dev)
      @reuse_name = blk_dev.name
      @reuse_sid = blk_dev.sid
    end
  end
end
```

And then we change all the places in which now there is something like `planed.reuse_name = my_dev.name` with a call to the new method.

That is. Now you KNOW there is always a `reuse_sid` in sync with `reuse_name` and you can simplify `#device_to_reuse` which is now offending CodeClimate, and you can simply rely on `reuse_sid` being correctly set beforehand in `SpaceMaker`.

ancorgs:
Moreover, if both are attributes, I could do this (which is a recipe for disaster):
```ruby
my_planned_device.reuse_name = one_dev.name
my_planned_device.reuse_sid = another_dev.sid
```

Or I could do
```ruby
if my_planned_device.reuse?
  puts "Reusing #{my_planned_device.sid}"
end
```
Which seems to make a lot of sense... but would only work as expected if the device to reuse was assigned using `#reuse_sid=` and not if it was assigned using `#reuse_name=`.

wfeldt:
> The order of the commits puzzled me while trying to review

Sorry, that was github's doing. The real commit order is in fact different. No idea hot to fix this. But at least the re-ordered commits don't overlap.

> I believe that adding two kind-of-independent reuse_name and reuse_sid attributes and checking
> always by both is kind of wrong.

Yes, agreed. I tried the last days to implement a clean solution. But this turned out to be far more disruptive than I expected. This patch is the minimal change I could come up with that fixes the problem with the least risk to break all kinds of things.

The are two issues
  1. The concept to refer to device objects by name is so far spread in the code that you have to change really a lot. Ideally you would refer to the device object directly and use neither name nor sid - but that is defeated by many places where we collect references to already destroyed objects (which shows that there is logically something wrong, imho).
  2. And this is where I gave up: our unit test mockup relies entirely on device names. sids are just not possible. If you start to wrap the name/sid pair into a new class for purity's sake you will just not be able to rewrite all the tests in any useful manner.

And yes, name and sid can diverge. This happens exactly in the bug this fixes.

In short, I would leave it as it is.

ancorgs:
> Sorry, that was github's doing. The real commit order is in fact different

Not a big deal. I was just curious.

> In short, I would leave it as it is.

Ok. I trust you. :-) But since you have to add a changelog entry anyways, would you mind to add a FIXME/TODO comment about the problem so we fix it next time we touch that? Something like (for example):

```ruby
# FIXME: having separate setters for reuse_name and reuse_sid is dangerous
# and makes some parts of the code cumbersome.
# We need to fix this class to ensure they are in sync, but adapting the mocks in
# the test cases was too much work.
attr_accessor :reuse_sid
```

PR #525, opened 2018-02-15
Adjust sizes and weights in the boot-related partitions

ancorgs:
Main PR for https://trello.com/c/7NcmMuGq/264-sles15-p1-shipstopper-1076851-build-4201-and-4211-ppc64le-unable-to-boot-after-installation

- Ensures the boot-related partitions get a sensible weight, so they don't stick always to its initial (minimum) size.
- Updates min/desired/max for all the partition types, after several conversations.
- Adds rationale to the RSpec tests and `doc/boot-requirements.md` for future discussions

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Feb 15 16:16:19 UTC 2018 - ancor@suse.com
+
+- Adjusted the suggested and minimum sizes of all the booting
+  partitions, both in the storage proposal and in the Partitioner
+  validations (bsc#1076851 and fate#318196).
+- 4.0.97
+
+-------------------------------------------------------------------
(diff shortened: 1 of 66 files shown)


wfeldt:
the values look reasonable enough to me

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Thu Feb 15 16:16:19 UTC 2018 - ancor@suse.com
+
+- Adjusted the suggested and minimum sizes of all the booting
+  partitions, both in the storage proposal and in the Partitioner
+  validations (bsc#1076851 and fate#318196).
```
joseivanlopez: NP: comment something about weight for booting partitions?

ancorgs: Weight is a rather low-level detail. The important user-facing effect is that the suggested sizes have changed. And that is mentioned.

PR #526, opened 2018-02-15
Fix hwinfo parsing

imobachgs:
Fixes [bsc#1080999](https://bugzilla.suse.com/show_bug.cgi?id=1080999).

### DIFF
--- a/src/lib/y2storage/hwinfo_reader.rb
+++ b/src/lib/y2storage/hwinfo_reader.rb
@@ -33,8 +33,9 @@ module Y2Storage
   #
   # @example Get information for "/dev/sda"
   #   hwinfo = HWInfo.instance.for_device(name)
-  #   hwinfo.driver #=> ["ahci", "sd"]
-  #   hwinfo.device_file #=> "/dev/sda"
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/hwinfo_reader.rb
```diff
@@ -104,21 +105,54 @@ def data_from_body(body)
...
+    # Parse the device_file key
+    #
+    # The device_file can contain up to two different values. See hwinfo sources
+    # for further details:
+    # https://github.com/openSUSE/hwinfo/blob/b3b2757b3633cde7f49c30757b9664defc773c86/src/hd/hdp.c#L468
+    #
+    # @param [String] value
+    # @return [Array<String>] array containing all values
+    def parse_key_device_file(value)
+      value.split(" ")
```
joseivanlopez: NP: `value.split` is the same.

imobachgs: In this case, I prefer to leave the argument as a reminder of the default value to my faulty memory :)

PR #529, opened 2018-02-19
Set default mount_by

joseivanlopez:
PBI: https://trello.com/c/rsjOOfOt/66-3-partitioner-settings-mount-by


src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -281,6 +373,20 @@ def selected_mount_by
       def value
         Yast::UI.QueryWidget(Id(:mt_group), :Value)
       end
+
+    private
+
+      def select_default_mount_by
+        # Mount point has always a value for mount_by. By default, MountPoint#mount_by
+        # backwards to the Storage#default_mount_by value, taking into account
```
ancorgs: I guess you mean "fallbacks" instead of "backwards". But even though, I still don't think the explanation about `#default_mount_by` fits here. This gets the value from the current mount_point, where that value comes from is not much relevant here.

src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -281,6 +373,20 @@ def selected_mount_by
...
+        # special cases (e.g., LVM LVs or NFS).
+        Yast::UI.ChangeWidget(Id(:mt_group), :Value, mount_point.mount_by.to_sym)
+      end
+
+      def disable_not_possible_mount_bys
+        not_possible_mount_bys = Y2Storage::Filesystems::MountByType.all - mount_point.possible_mount_bys
+        not_possible_mount_bys.map { |m| Yast::UI.ChangeWidget(Id(m.to_sym), :Enabled, false) }
```
ancorgs: I guess you meant `#each` instead of `#map`.

src/lib/y2partitioner/widgets/pages/settings.rb
```diff
@@ -0,0 +1,118 @@
...
+            _("Default Mount by")
+          end
+
+          def init
+            self.value = Y2Storage::StorageManager.instance.default_mount_by.to_s
+          end
+
+          def items
+            [
+              [Y2Storage::Filesystems::MountByType::DEVICE.to_s, _("Device Name")],
```
ancorgs: Shouldn't this translation go to the `Y2Storage::Filesystems::MountByType::DEVICE` class so we here just have something like this?
```ruby
Y2Storage::Filesystems::MountByType.all.map { |m| [m.to_s, m.to_human_string] }
```

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -268,6 +268,25 @@ def handle(event)
...
+      # when the label is entered (see {Dialogs::FstabOptions}).
+      #
+      # @return [Boolean] true if the label is not required or it is required and given; false
+      #   otherwise.
+      def validate
+        return true if !formatted? || !mounted? || !mounted_by_label? || !empty_label?
+
+        # TRANSLATORS: Error message when a device should be mounted by label but no label
+        # is given.
+        Yast::Popup.Error(_("Provide a volume label to mount by label."))
```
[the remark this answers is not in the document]
joseivanlopez: It is exactly the same message used in the old partitioner.

src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -232,6 +250,80 @@ def store
...
+      # @note An error popup is presented when the filesystem is mounted by label
+      #   but a label is not given.
+      #
+      # @return [Boolean]
+      def presence_validation
+        return true unless mounted_by_label?
+        return true unless value.empty?
+
+        # TRANSLATORS: Error messagge when the label is not given.
+        Yast::Popup.Error(_("Provide a volume label to mount by label."))
```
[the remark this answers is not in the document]
joseivanlopez: Same than before again ;)

src/lib/y2partitioner/widgets/fstab_options.rb
joseivanlopez: Ok, I will simply remove the comment.

PR #530, opened 2018-02-19
if there is any item in table, ensure at least something is selected

jreidinger:


src/lib/y2partitioner/widgets/configurable_blk_devices_table.rb
```diff
@@ -50,7 +50,10 @@ def opt
       # @macro seeAbstractWidget
       def init
         initial_sid = UIState.instance.row_sid
-        self.value = row_id(initial_sid) if initial_sid
+
+        # if we do not have valid sid, then pick first available device.
+        # Reason is to allow e.g. chain of delete like described in bsc#1076318
+        self.value = row_id(valid_sid?(initial_sid) ? initial_sid : devices.first)
```
joseivanlopez: What happens when the last element is deleted? I guess we should do something like:

```
initial_sid =  UIState.instance.row_sid
initial_sid = devices.first.sid if !valid_sid?(initial_sid) && !devices.empty?
```
```

Comment (jreidinger) [reply src/lib/y2partitioner/widgets/configurable_blk_devices_table.rb]: well, it works, but it prints to log warning that widget with id "table:device:0". I will add that check.

File: test/y2partitioner/widgets/configurable_blk_devices_table_test.rb [hunk 2]
```diff
@@ -33,6 +33,39 @@
...
+
+        subject.init
+      end
+    end
+
+    context "UIState contain sid that is not in table" do
+      it "selects any valid device in table" do
+        Y2Partitioner::UIState.instance.select_row("999999999")
+
+        expect(subject).to receive(:value=)
```
ancorgs: I would check here that the argument is different from `999999999` and is one of `devices.map(&:sid)`

jreidinger: makes sense

PR #532, opened 2018-02-19
Special mount option handling for certain mount paths

shundhammer:
https://trello.com/c/9c9qZoIn/269-special-mount-options-for-boot-efi-bsc1080731

This is a bit of a mess; it reintroduces special handling we had in yast-storage-old for mount options for certain mount points. We had meant to leave this kind of thing behind, but it turns out we still need some of the old workarounds for some quirks or limitations of some kernel filesystem modules or low-level tools.

I tried to keep this localized in one file so anybody needing to change yet another one of those special cases doesn't need to hunt a half dozen files and a half dozen classes for all the places where this needs to be adapted.

At some points, the code may seem a little overkill, but this is meant to be extensible and, equally important, easy to understand.

The downside is that this is all in the base class where some of it could be moved to derived classes (ext3, ext4, vfat), and that the caller now needs to pass an additional `mount_path` parameter to the `default_fstab_options` method (but this parameter is optional).

So now we have some of the messy part of the old code back, but that's life; at least the mess should now be clear to read, understand and modify.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Feb 19 18:08:01 UTC 2018 - shundhammer@suse.com
+
+- Special handling for mount options for / and /boot/*
+  (bsc#1080731, bsc#1061867, bsc#1077859)
+- 4.0.99
+
+-------------------------------------------------------------------
 Mon Feb 19 14:19:30 UTC 2018 - jreidinger@suse.com
 
 - Ensure that there is always selected item in table, if it is not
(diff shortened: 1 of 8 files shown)


ancorgs:
Hmmm, this produces the correct default options (that is, for the proposal, AutoYaST and initial settings in the expert partitioner). But... what happens if the user clicks on the "Fstab Options" for an ext4 root and then clicks "ok". Wouldn't be `data=ordered` inserted in that case?

ancorgs:
Or maybe, `data=ordered` just works (since it's the default). If so, we could leave it as it is for the time being. I really don't know.

shundhammer:
We just agreed in IRC to merge this first part for now.

The UI part in the partitioner will disable those specific options for certain mount points.

PR #533, opened 2018-02-19
Partitioner: bringing back old partitioner logic about mount points

ancorgs:
Main PR for two cards: https://trello.com/c/FBPlQOBf/262-1-partitioner-apply-role-correctly and https://trello.com/c/EzGZXxXN/274-sles15-several-reports-irritating-mount-point-suggestions-for-existing-swap-efi-zipl

This adds three commits fixing a total of 4 bugs.

- First commit: bring back the original list of mount points instead of a fixed list
  - Fixes half of [bsc#1076167](https://bugzilla.suse.com/show_bug.cgi?id=1076167) by bringing `/boot/efi' back
  - Fixes [bsc#1081200](https://bugzilla.suse.com/show_bug.cgi?id=1081200) by bringing `/boot/zipl' back
- Second commit: bring back original logic for the "Mount Device" option in the roles "Operating System" and "Data".
  - Fixes [bsc#1078975](https://bugzilla.suse.com/show_bug.cgi?id=1078975) and [bsc#1073854](https://bugzilla.suse.com/show_bug.cgi?id=1073854)
- Third commit: offer "swap" as mount point for preexisting swap partitions
  - Fixes second half of [bsc#1076167](https://bugzilla.suse.com/show_bug.cgi?id=1076167) (and several duplicates about swap reusing).

Everything manually tested.

The code climate issues are not really new. The `Actions::Controllers::Filesystem` class had already some disabled cops. We need to break it into pieces... but not now.

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/filesystem.rb
+++ b/src/lib/y2partitioner/actions/controllers/filesystem.rb
@@ -387,7 +387,9 @@ module Y2Partitioner
         #
         # @return [Array<String>]
         def mount_paths
-          all_mount_paths - mounted_paths
+          mount_paths = all_mount_paths - mounted_paths
+          mount_paths.unshift("swap") if filesystem && filesystem.type.is?(:swap)
+          mount_paths
         end
 
         # All paths used by the preexisting subvolumes (those that will not be
--- a/src/lib/y2partitioner/actions/controllers/filesystem.rb
+++ b/src/lib/y2partitioner/actions/controllers/filesystem.rb
@@ -497,6 +497,9 @@ module Y2Partitioner
           else
             part_id = DEFAULT_PARTITION_ID
             fs_type = (role == :system) ? DEFAULT_FS : DEFAULT_HOME_FS
+            # Behavior of the old SingleMountPointProposal (behavior introduced
+            # back in 2005 with unknown rationale)
+            mount_path = mount_paths.first unless Yast::Mode.normal
           end
 
           {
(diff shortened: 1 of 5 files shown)


src/lib/y2partitioner/actions/controllers/filesystem.rb
```diff
@@ -373,6 +374,33 @@ def configure_snapper
...
+        # Paths that are mounted in the current device graph, excluding
+        # subvolumes
+        #
+        # @return [Array<String>]
+        def mounted_paths
+          devices = mounted_devices.reject { |d| d.is?(:btrfs_subvolume) }
+          devices.map(&:mount_path)
+        end
+
+        # Sorted list of the default mount paths to offer to the user
```
joseivanlopez: Are they actually sorted?

src/lib/y2partitioner/actions/controllers/filesystem.rb
```diff
@@ -552,6 +584,85 @@ def btrfs?
...
+
+        # This implements the code that used to live in
+        # Yast::Filesystems::SuggestMPoints (whatever the rationle behind those
+        # mount paths was back then) with the only exception mentioned in
+        # {#booting_paths}
+        #
+        # @return [Array<String>]
+        def all_mount_paths
+          @all_mount_paths ||=
+            if Yast::Stage.initial
```
joseivanlopez: Question: `#initial` means during installation?

ancorgs: Yes. First stage (so installation or first phase of the autoinstallation)

src/lib/y2partitioner/actions/controllers/filesystem.rb
ancorgs: Yes, following the same criteria than the old partitioner (which contained a really complicated logic to guarantee this same sorting).

PR #534, opened 2018-02-19
AutoYaST default filesystem

imobachgs:
This PR adds support to AutoYaST to select a filesystem when it is not specified in the profile. It's still a WIP as, for instance, we should handle the "legacy" configuration.

### DIFF
--- a/src/lib/y2storage/volume_specification_builder.rb
+++ b/src/lib/y2storage/volume_specification_builder.rb
@@ -1,0 +1,227 @@
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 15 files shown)


src/lib/y2storage/volume_specification_builder.rb
```diff
@@ -0,0 +1,227 @@
...
+    end
+
+    # Return a volume specification for a given mount point
+    #
+    # It will use the volume specification found within the list of volumes
+    # in the proposal settings. If it is not found, it will try to propose
+    # a fallback.
+    #
+    # @param mount_point [String] Volume mount point
+    # @return [VolumeSpecification] Volume specification;, nil if the
```
joseivanlopez: NP: typo, comma is leftover.

src/lib/y2storage/volume_specification_builder.rb
```diff
@@ -0,0 +1,227 @@
...
+    # Return a volume spec from proposal settings for the given mount point
+    #
+    # @param mount_point [String] Volume mount point
+    # @return [VolumeSpecification,nil] Volume specification if found; otherwise,
+    #   it returns nil.
+    def proposal_spec(mount_point)
+      case proposal_settings.format
+      when :ng
+        ng_proposal_spec(mount_point)
+      when :legacy
```
joseivanlopez: Question: is it necessary to continue supporting legacy format?

imobachgs: That's a good question. And, now that you have mentioned, I would say that the legacy format is deprecated. So I will happily remove those bits :)

PR #535, opened 2018-02-20
Special path handling for mount options in partitioner

shundhammer:
Second part of  https://trello.com/c/9c9qZoIn/269-special-mount-options-for-boot-efi-bsc1080731

## Partitioner Part

This makes sure to not force mount options that should not be used for some special paths ("/", "/boot/*") back into the mount options when the user enters the "mount options" dialog in the partitioner.

For ext3/ext4, this removes the "journal" combo box, for vfat, it removes the "iocharset" combo box.

First I tried to just disable them, but this looks so unnatural and awkward that it raises a lot of questions: At least in my Qt widget style, only the combo box label is greyed out, the combo box itself is not; yet, you cannot click it, and it displays a value that is no longer applicable.

So, removing those combo boxes in just that scenario worked a lot better (and it is consistent with what we do otherwise: Simply not offer options that are not applicable in the current scenario).

If the user insists, he can still enter that option manually in the "arbitrary mount options" field.

I also changed the controller to make sure it always performs the special path handling whenever the mount path or the filesystem type changes; before, it would sometimes keep mount options which may or may not be desirable anymore.

## Special Handling for VFAT /boot/EFI

mount-opt-boot-efi

_Notice no "Charset for file names" ("iocharset=...") combo box_

## Normal Handling for VFAT

mount-opt-vfat

## Special Handling for Ext4 Root Filesystem

mount-opt-root-ext4

_Notice no "Data Journaling Mode" ("data=...") combo box_

## Normal Handling for Ext4

mount-opt-ext4

### DIFF
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -104,6 +104,14 @@ module Y2Partitioner
         filesystem.mount_point
       end
 
+      # Mount path of the current filesystem
+      #
+      # @return [String]
+      def mount_path
+        return nil if filesystem.mount_point.nil?
+        filesystem.mount_point.path
+      end
+
       # Common regexp checkbox widgets init.
       def init_regexp
         i = mount_point.mount_options.index { |o| o =~ self.class::REGEXP }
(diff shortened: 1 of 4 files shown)


src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -104,6 +104,14 @@ def mount_point
         filesystem.mount_point
       end
 
+      # Mount path of the current filesystem
+      #
+      # @return [String]
+      def mount_path
```
ancorgs: Isn't this redundant with `Y2Storage::Mountable#mount_path`?

PR #536, opened 2018-02-20
Improve checks

jreidinger:


src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -116,6 +121,14 @@ def boot_partition_needed?
         false
       end
 
+      def too_small_boot?
+        filesystem = devicegraph.filesystems.find { |f| f.mount_point == "/boot" }
+        return false unless filesystem
+
+        available_size = filesystem.blk_devices.map(&:size).reduce(:+)
+        return available_size < DiskSize.MiB(100) # for number see below minimal size for boot volume
```
ancorgs: Well, 100MiB is the minimal size the proposal will use as starting point, which doesn't imply is the minimal size to contain a kernel .

jreidinger: I think current kernel require around 67MiB, but usually you need to have here also some boot code. and with kernel multiversioning it takes more. And kernel is still growing, so for me 100 MiB is reasonable size.

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -612,6 +612,17 @@ def refresh
...
+      # validates if it can be encrypted
+      def validate
+        return true unless value
+
+        # size number is from bsc1065071#c5
+        return true if @controller.blk_device.size > ::Y2Storage::DiskSize.MiB(2)
+
+        ::Yast::Popup.Error(_("For encryption partition have to be bigger then 2 MiB."))
+        false
+      end
```
joseivanlopez: I see a problem with this approach. The minimum size depends on several aspects, not only on the fact the device is formatted. For example, different filesystem types could require different min sizes.

A part of that, there are more checks that have to be done an the end of this screen. For example, we have to check that the user gives a filesystem label when the default mount_by is set to label, see [my PR](https://github.com/yast/yast-storage-ng/pull/529/commits/435abbc51a204d4af9da85aa831f3ac9fc72014f).

Maybe it is time to create a class (e.g. `Controllers::Filesystem::Validations` or something similar) to manage all validations needed in that screen, reporting the list of errors.

jreidinger: well, goal of this is to check real minimal for optimistic approach. I do not opposite that it can be improved, but I need to know how to improve.

ancorgs: In theory, `blk_device.resize_info` can be used to know the minimum size based on everything the block device contains (filesystem of a given type, encryption and so on). If we go for the "cheap" solution (independent validations placed in random widgets with hardcoded limits), I would like to see a big TODO/FIXME next to each one. Explaining that it should be centralized and mentioning `blk_device.resize_info`.

jreidinger: @ancorgs well, problem with resize_info is that I have to first create that encryption device to know it, but I worry it is too late. Of course we can add generic checkk if resize_info for each partition is bigger then size, but for this one I want quick check in widget to prevent continue.
So do you think it makes sense to also add generic check that check if all partition have size at least as big as minimal in resize_info?

ancorgs: Of course, we may need to change a little bit how some things are done to achieve that. But in the mid term I want **only one** check for the size of the device relying on the library. For the future, I don't want this check (with a narrow-scoped limit just hard-coded somewhere in a widget) and **additionally** the one based on `resize_info`.

In the future, the current approach can lead to a warning saying the 2MiB is too small (only local widget check). Then the user changes it to 4MiB and it gets a second warning coming from the filesystem_type widget saying that XFS requires at least 5 MiB (I just made up the number). Then he changes it to 6MiB and that's still wrong because the combination of 2 and 5 means the real minimum is 7MiB.

_Beware: the rest is actually another more broad topic. I just couldn't resist_

That problem illustrate want of the things I dislike about the CWM "default philosophy" of having domain-related logic (like validation and similar stuff) spread over a collection of widgets and subwidgets. In the partitioner we are trying to limit that by placing as much domain-related logic as possible into the classes within the `Actions` namespace. So we have only one place where we validate the different aspects of the device being created, instead of having that scattered over a dozen of CWM classes. So our widgets are as dumb as possible, instead of being super-smart components.

ancorgs: Just to clarify. For the time being, I'm just asking for a FIXME/TODO mentioning a more centralized and library-based solution for the future. Not to change anything right now.

src/lib/y2storage/boot_requirements_strategies/base.rb
ancorgs: I think that, if you rebase on top of #534 you could use this here:
```ruby
def too_small_boot?
  #... [stripped] ...
  available_size < min_boot_size
end

def min_boot_size
  @min_boot_size ||= VolumeSpecificationBuilder.new.for("/boot").min_size
end
```

Instead of repeating the value here.

src/lib/y2partitioner/widgets/format_and_mount.rb
jreidinger: @ancorgs to comment last paragraph. CWM validation is useful when you have invalid value that is local to one widget. What you want is probably due to complexity check that verify whole partition setup after whole wizard and this cannot be done on CWM level and more probably on sequencer level which checks it as last step and allow going back is something is wrong.

ancorgs: Yes, in this case the reason is that the check is complex and implies several aspects. But even in simpler cases, I believe that having the validations about a given device in a single place makes more sense than a lot of local "validate" methods scattered all around in widgets.

So is not "we have to centralize here because is a complex check". For me is more like "we have to centralize because is the sane approach, period" :wink:

Anyway, we are getting too much out of scope. My fault. :wink: Let's leave this for some beer-powered discussion.

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -612,6 +612,19 @@ def refresh
...
+      # validates if it can be encrypted
+      def validate
+        return true unless value
+
+        # FIXME: have generic check when new partition is created if it is valid
+        # like e.g. if fs fits to partition, and others
+        # size number is from bsc1065071#c5
+        return true if @controller.blk_device.size > ::Y2Storage::DiskSize.MiB(2)
+
+        ::Yast::Popup.Error(_("For encryption partition have to be bigger then 2 MiB."))
```
ancorgs: I little bit hard to read. What about "Only devices bigger than 2 MiB can be encrypted"

src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -95,6 +95,11 @@ def errors
           res << SetupError.new(message: error_message)
         end
 
+        if too_small_boot?
+          error_message = _("Device mounted at '/boot' is too small to contain kernel.")
```
ancorgs: Missing determinants "**The** device mounted at '/boot' is too small to contain **a** kernel"

src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -117,6 +122,17 @@ def boot_partition_needed?
         false
       end
 
+      def too_small_boot?
+        # for other partitions it is not needed as packager check disk usage, but /boot is special
+        # as it contain initrd that is generated and also bootloader code.
+        filesystem = devicegraph.filesystems.find { |f| f.mount_point == "/boot" }
```
ancorgs: this should use `#mount_path` (which returns a string) instead of `#mount_point` which returns a `MountPoint` object.

src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -117,6 +122,17 @@ def boot_partition_needed?
...
+        return false unless filesystem
+
+        available_size = filesystem.blk_devices.map(&:size).reduce(:+)
+        # FIXME: it does not count fs size and other and it is just hard coded.
+        return available_size < DiskSize.MiB(100) # for number see below minimal size for boot volume
```
ancorgs: Instead of a comment with "see below minimal size"... why don't you really use that minimal size?
```ruby
available_size < boot_volume.min_size
```

By the way, `return` is surplus and anti-ruby :wink:

jreidinger: makes sense. BTW about return, usually rubocop reminds me that :)

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -257,6 +257,21 @@
 
     let(:zipl_partition) { partition_double("/dev/sda1") }
 
+    let(:efiboot) { true }
+
+    it "contains an error when there is /boot that is not big enough" do
```
ancorgs: IMHO there is too much mocking here. As probed by the fact that you were using the wrong method to find the mount point and it was still passing.

jreidinger: @ancorgs it is in whole this test. To be honest I prefer here if real graph from xml or yaml is used instead, but currently it will be quite big change.

test/y2partitioner/widgets/format_and_mount_test.rb
```diff
@@ -204,6 +204,50 @@
     include_examples "CWM::ComboBox"
   end
 
+  describe Y2Partitioner::Widgets::EncryptBlkDevice do
+    subject { described_class.new(controller) }
+
+    include_examples "CWM::CheckBox"
+
+    describe "#validate" do
```
ancorgs: I find the output a little bit confusing

- encrypt is not selected
  - it returns true
- partition is bigger than 2MiB
  - it returns true
- partition is equal or smaller than 2MiB
  - it returns false
  - it shows error popup

Just imagine we change execution order into:

- partition is equal or smaller than 2MiB
  - it returns false
  - it shows error popup
- encrypt is not selected
  - it returns true
- partition is bigger than 2MiB
  - it returns true

What happens then if the partition is small but encryption is NOT checked? Should I understand it returns false?

My minimum proposed fix:
- encrypt is not selected
  - it returns true
- encrypt is checked and partition is bigger than 2MiB
  - it returns true
- encrypt is checked and partition is equal or smaller than 2MiB
  - it returns false
  - it shows error popup

My preferred fix (I believe that exploring scenarios in width are one of the main advantages of RSpec, as much the produced nesting can be ugly to some people):

- if the partition is bigger than 2MiB
  - and encrypt is not selected
    - it returns true
  - and encrypt is selected
    - it returns true

- if the partition is 2MiB large
  - and encrypt is not selected
    - it returns true
  - and encrypt is selected
    - it returns true

- if the partition is smaller than 2MiB
  - and encrypt is not selected
    - it returns true
  - and encrypt is selected
    - it returns false
    - it shows error popup

src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -117,6 +122,17 @@ def boot_partition_needed?
...
       end
 
+      def too_small_boot?
+        # for other partitions it is not needed as packager check disk usage, but /boot is special
+        # as it contain initrd that is generated and also bootloader code.
+        filesystem = devicegraph.filesystems.find { |f| f.mount_path == "/boot" }
+        return false unless filesystem
+
+        available_size = filesystem.blk_devices.map(&:size).reduce(:+)
+        # FIXME: it does not count fs size and other and it is just hard coded.
```
ancorgs: We should actually check for available space in the filesystem. Not for the size of the container device. Is not only more precise, but also more straightforward than summing up sizes.

Just imagine we are reusing a boot partition without having explicitly marked it to be formatted - the device would be big enough but potentially full with whatever garbage already.

ancorgs: BTW, a filesystem spread over several blk devices is a deprecated feature in storage-ng

ancorgs: Even further, `Devicegraph#filesystems` return `Array<Filesystems::Base>` and `#blk_devices` is only defined in `Filesystems::BlkFilesystem`, not in the base class.

test/y2storage/boot_requirements_errors_test.rb
ancorgs: Ok, let's leave it as it is. But please do some manual test with the `partitioner_testing` client

src/lib/y2storage/boot_requirements_strategies/base.rb
jreidinger: so I need to use detect_size_info, right?

joseivanlopez: Apart of commented by ancor, I see that we could check `/boot` twice. Here the available size is checked, but `/boot` in also checked in cases where it is a required separate partition, for example [here](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/boot_requirements_strategies/legacy.rb#L170).

I don't think that it is a problem, but a little weird. We could show an initial error because `boot` is too small and then other warning because the `boot` has not correct filesystem type, for example.

test/y2partitioner/widgets/format_and_mount_test.rb
jreidinger: well, my problem with later one is that so much nesting, but it is so much info that reading it can make you more confuse. Like there is always and encrypt is not selected, can it sometime return false? so minimum proposed fix given better clue and for me is better documentation as it contain less glue and just important info.

src/lib/y2storage/boot_requirements_strategies/base.rb
jreidinger: @joseivanlopez as this is error warning won't be seen. We first force to fix fatal errors and then user can confirm warnings.

jreidinger: and if there are more requirements for /boot partition I am not sure how we can handle it differently? join our related part for it?

joseivanlopez: @jreidinger I guess a wrong filesystem type should also be a stopper error, isn't it? In that case we should add something as `wrong_boot_fs_type?` and check that `!Filesystems::Type.root_filesystems.include?(boot.filesystem.type)`.

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -67,13 +69,19 @@
...
+
     RSpec.shared_examples "missing boot partition" do
       it "contains an error for missing boot partition" do
         expect(checker.warnings.size).to eq(1)
         expect(checker.warnings).to all(be_a(Y2Storage::SetupError))
 
-        missing_volume = checker.warnings.first.missing_volume
-        expect(missing_volume).to eq(boot_volume)
+        message = checker.warnings.first.message
+        expect(message).to match(/Missing device for \/boot/)
```
joseivanlopez: Now we have a new class `VolumeSpecificationBuilder` [1]. Maybe it would be better to check that there is a warning for the missing volume:

```
boot_volume = Y2Storage::VolumeSpecificationBuilder.new.for("/boot")
expect(checker.warnings).to include(an_object_having_attributes(missing_volume: boot_volume))
```

The same for the rest of test.

[1] https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/volume_specification_builder.rb#L58

jreidinger: well, even when I really like reusability and DRY in production code, for tests I prefer more explicit checks which is trivial, so we are sure tests are correct.

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -404,66 +318,86 @@
...
+            include_examples "missing prep partition"
+          end
         end
 
-        context "with a Software RAID proposal" do
-          let(:use_raid) { true }
-          include_examples "PReP partition"
+        # sorry, but I won't write it in xml and yaml does not support it
+        # scenario generator would be great
+        xcontext "with a Software RAID proposal" do
```
joseivanlopez: I guess you could use whatever devicegraph we already have with a MD RAID and manually create a PReP partition in before section.

test/support/boot_requirements_context.rb
```diff
@@ -29,57 +29,22 @@ def find_vol(mount_point, volumes)
     volumes.find { |p| p.mount_point == mount_point }
   end
 
-  subject(:checker) { described_class.new(devicegraph) }
```
joseivanlopez: We should not touch this context because it affects to boot requirements tests, which are used for documenting booting requirements. Probably this is too sensitive to be modified at this moment. I would keep it untouched.

Moreover, I would not use it in `boot_requirements_errors_test.rb`. Instead, I would directly add what is necessary for those tests instead of using this context.

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -472,39 +406,30 @@
...
+      xcontext "using a zfcp disk as boot disk" do
         let(:dasd) { false }
         let(:type) { Y2Storage::DasdType::UNKNOWN }
         let(:format) { Y2Storage::DasdFormat::NONE }
 
         include_examples "zipl partition"
       end
 
-      context "using a FBA DASD disk as boot disk" do
+      xcontext "using a FBA DASD disk as boot disk" do
```
joseivanlopez: For this test, I guess we could use a dasd scenario and simply mock the type of the dasd to be FBA.

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -472,39 +406,30 @@
...
         let(:dasd) { true }
         let(:type) { Y2Storage::DasdType::FBA }
         include_examples "unsupported boot disk"
       end
 
-      context "using a (E)CKD DASD disk as boot disk" do
+      xcontext "using a (E)CKD DASD disk as boot disk" do
```
joseivanlopez: We have a devicegraph that contains CKD Dasd with both formats: `kinds-of-dasd.xml`. We can use it and create needed partition in before section.

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -472,39 +406,30 @@
...
-        let(:root_filesystem) { nil }
+        let(:scenario) { "false-swaps" }
+
         include_examples "unknown boot disk"
       end
 
-      context "using a zfcp disk as boot disk" do
+      # TODO: find coresponding xml files for it
+      # and ideally from real usage
+      xcontext "using a zfcp disk as boot disk" do
```
joseivanlopez: Probably, in the scenario `kinds-of-dasd.xml`, we could add a new DASD with `<transport>FCoE</transport>`. This should be enough.

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -119,33 +127,33 @@
...
-          missing_volume = checker.warnings.first.missing_volume
-          expect(missing_volume).to eq(efi_volume)
+          message = checker.warnings.first.message
+          expect(message).to match(/Missing device for \/boot\/efi/)
         end
       end
 
       context "when there is a /boot/efi partition in the system" do
+        let(:scenario) { "efi" }
         let(:partitions) { [boot_partition, efi_partition] }
```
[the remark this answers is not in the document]
jreidinger: no, I have to clean it whole, it is a lot of changes.

test/y2storage/boot_requirements_errors_test.rb
jreidinger: this is too artificial for me. My goal is to have clear scenario and then apply test on top of it. Having hacks that modify partition in code makes it just harder to see point of test and verify it is correct.

test/support/boot_requirements_context.rb
jreidinger: well, plan is to remove overmocking also in boot_requirements test to operate on real device graph.

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -171,105 +179,41 @@
...
-
-      context "and there is a PReP partition in the system" do
-        let(:partitions) { [boot_partition, prep_partition] }
+        before do
+          allow(checker.send(:strategy).boot_disk).to receive(:mbr_gap).and_return(256.KiB)
+        end
 
-        it "does not contain errors" do
-          expect(checker.warnings).to be_empty
+        xit "it contains error if there is no separate /boot" do
```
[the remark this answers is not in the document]
jreidinger: Problem here is that it is shared example. So I cannot define here scenario for each case. My plan is to kill this shared example and just add both context directly to define proper scenario for each case ( lvm, encrypted, raid ). So scenario is not the same.

test/y2storage/boot_requirements_errors_test.rb
jreidinger: ok, will try.

joseivanlopez:
Another not so important thing. This PR adds a lot of new devicregraph files and all of them are for boot checking purposes. Why not to have all those devicegraphs in its own subdirectory (e.g, `data/devicegraphs/boot_checks`) ?

jreidinger:
@joseivanlopez well, I think having such device graphs makes sense not only for boot checks as it represent common scenario and I think we will need it in more locations. In general I like testing with devicegraph as it is end to end testing as device graph is what libstorage gives you and also what it gets back.

test/y2storage/boot_requirements_checker_x86_test.rb
```diff
@@ -137,53 +109,27 @@
...
-              it "requires only a /boot partition" do
-                expect(checker.needed_partitions).to contain_exactly(
-                  an_object_having_attributes(mount_point: "/boot")
-                )
-              end
-            end
-          end
+              let(:scenario) { "dos_btrfs_lvm_min_gap" }
 
-          context "in an encrypted proposal" do
```
joseivanlopez: Is it not needed a separate boot on Legacy + MSDOS + Encrypted + MBR without additional space? The proposal is requiring a separate boot now in that scenario, so: a) we should document this important change (rationale or something like that), b) the proposal should be modified to fulfill new behaviour.

test/y2storage/boot_requirements_checker_x86_test.rb
```diff
@@ -219,18 +161,7 @@
...
-            let(:use_lvm) { true }
-
-            it "raises an exception" do
-              expect { checker.needed_partitions }.to raise_error(
-                Y2Storage::BootRequirementsChecker::Error
-              )
-            end
-          end
-
-          context "in an encrypted proposal" do
```
joseivanlopez: Here the same comment than above.

test/y2storage/boot_requirements_checker_x86_test.rb
```diff
@@ -82,7 +59,7 @@
           end
 
           context "if there is already a GRUB partition" do
-            let(:grub_partitions) { [grub_partition] }
+            let(:scenario) { "missing_bios_boot" }
```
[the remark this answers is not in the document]
joseivanlopez: Ok, fixed in the next commit :)

joseivanlopez:
I am worried about the disabled tests for s390 ('test/y2storage/boot_requirements_checker_s390_test.rb`). I think we can take .xml scenarios from libstorage-ng for each case, or try to create them. Those tests are very important, and also are used to generate the documentation that we use to refer when we are talking about the proposal requirements.

Also, we should synchronize the documentation with the code. We have a rake task for that (`rake doc:bootspecs`).

And also I see several disabled tests in `test/y2storage/boot_requirements_errors_test.rb`. Cannot they be activated easily?

test/y2storage/boot_requirements_checker_aarch64_test.rb
```diff
@@ -27,31 +27,14 @@
...
-      allow(dev_sda).to receive(:mbr_gap).and_return mbr_gap_size
-      allow(dev_sda).to receive(:efi_partitions).and_return efi_partitions
-      allow(dev_sda).to receive(:partitions).and_return(efi_partitions)
-      allow(dev_sdb).to receive(:efi_partitions).and_return other_efi_partitions
-      allow(dev_sdb).to receive(:partitions).and_return(other_efi_partitions)
-    end
-
-    include_context "plain UEFI"
+  let(:architecture) { :aarch64 }
+  let(:architecture) { :aarch64 }
```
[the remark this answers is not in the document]
jreidinger: thanks for catch

jreidinger:
@joseivanlopez well, writing s390 xml is quite tricky and what is worse for me is that it is not so easy to see what scenario is. Also I think there is some discussion and we plan to change this boot requirements.

about disabled test in requirements_error - skipped is basically s390 and raid. Only remaining one is mbr gap shared example, which I will fix.

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Feb 23 14:57:51 UTC 2018 - jreidinger@suse.com
+
+- Do not allow to encrypt too small partition (bsc#1065071)
+- Check that if separated "/boot" is used that it is big enough
```
joseivanlopez: NP: I see it a bit difficult to read. Probably some like this?: `- Check size for separate /boot`.

test/y2partitioner/widgets/format_and_mount_test.rb
```diff
@@ -275,6 +275,50 @@
...
+
+      context "encrypt is not checked" do
+        it "returns true" do
+          allow(subject).to receive(:value).and_return(false)
+
+          expect(subject.validate).to eq true
+        end
+      end
+
+      context "encrypt is checked and the partition is bigger then 2MiB" do
```
joseivanlopez: NP: "bigger than" instead of "bigger then".

test/y2partitioner/widgets/format_and_mount_test.rb
```diff
@@ -275,6 +275,50 @@
...
+        it "returns true" do
+          expect(subject.validate).to eq true
+        end
+      end
+
+      context "encrypt is checked and the partition is equal or smaller then 2MiB" do
```
joseivanlopez: NP: "smaller than" instead of "smaller then".

test/y2storage/boot_requirements_errors_test.rb
joseivanlopez: Anyway, I see we have support for Dasd in [fake_device_factory.rb](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/fake_device_factory.rb#L213). It should be easy to create an scenario with Dasd (even with a specific type and format).

PR #539, opened 2018-02-21
Default for partition table type is now GPT, no longer MSDOS

shundhammer:
https://trello.com/c/3LaKjLaH/248-1-sles15-p2-1078709-build-4321storage-ng-ms-dos-is-pre-selected-when-creating-new-partition-table

https://fate.suse.com/323457

See also the discussion in the PBI:

The old default coming from libstorage-ng was MSDOS unless it wouldn't work (because it's size limited to MaxUInt32 sectors or something like that). That default is overridden for certain classes in our Ruby part which preferred GPT everywhere.

The partitioner was the only part of the code simply using the default type coming from libstorage-ng; now this changed to GPT as well.

My changes in this PR reflect that changed default (in the unit tests), and it now build-requires the latest libstorage-ng that introduced that change to make sure the unit tests don't fail.

### DIFF
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:        4.0.103
+Version:        4.0.104
 Release:	0
 BuildArch:	noarch
 
--- a/test/y2partitioner/actions/controllers/partition_table_test.rb
+++ b/test/y2partitioner/actions/controllers/partition_table_test.rb
@@ -63,6 +63,12 @@ describe Y2Partitioner::Actions::Controllers::PartitionTable do
       end
     end
 
+    describe "#default_partition_table_type" do
+      it "returns type GPT" do
+        expect(subject.default_partition_table_type).to eq Y2Storage::PartitionTables::Type::GPT
+      end
+    end
+
     describe "#can_create_partition_table?" do
       it "detects that it can create a partition table" do
         expect(subject.can_create_partition_table?).to eq true
@@ -117,6 +123,12 @@ describe Y2Partitioner::Actions::Controllers::PartitionTable do
     describe "#possible_partition_table_types" do
       it "returns only type DASD" do
         expect(subject.possible_partition_table_types).to eq [Y2Storage::PartitionTables::Type::DASD]
+      end
+    end
+
+    describe "#default_partition_table_type" do
+      it "returns type DASD" do
+        expect(subject.default_partition_table_type).to eq Y2Storage::PartitionTables::Type::DASD
       end
     end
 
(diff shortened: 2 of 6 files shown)


package/yast2-storage-ng.spec
```diff
@@ -26,12 +26,12 @@ Source:		%{name}-%{version}.tar.bz2
...
-# BlkDevice#usable_as_blk_device? and Partitionable#usable_as_partitionable?
-Requires:	libstorage-ng-ruby >= 3.3.160
+# New partition table default is now GPT, no longer MSDOS
+Requires:	libstorage-ng-ruby >= 3.3.169
 
 BuildRequires:	update-desktop-files
-# BlkDevice#usable_as_blk_device? and Partitionable#usable_as_partitionable?
-BuildRequires:	libstorage-ng-ruby >= 3.3.160
+# New partition table default is now GPT, no longer MSDOS
+BuildRequires:	libstorage-ng-ruby >= 3.3.169
```
ancorgs: As commented on the trello card, synchronizing the value of both defaults (libstorage-ng and Y2Storage) may be a nice to have bonus. But I don't think it's the real fix. The real fix would be to use the Y2Storage default in the partitioner, so we don't rely on it being in sync with the library default (no other part of YaST relies on that).

test/y2partitioner/actions/controllers/partition_table_test.rb
```diff
@@ -34,7 +34,7 @@
 
     describe "#new" do
       it "has an initial type" do
-        expect(subject.type).to eq Y2Storage::PartitionTables::Type::MSDOS
+        expect(subject.type).to eq Y2Storage::PartitionTables::Type::GPT
```
ancorgs: The output of this test is
```
describe Y2Partitioner::Actions::Controllers::PartitionTable
  With a PC with 2 disks with some partitions
    it has an initial type
```
But:
 1) The fact that is a windows PC with 2 disks is as descriptive as irrelevant from the specification point of view 
 2) More important. This is not testing that "it has an initial type". This is testing than "the initial type is GPT".

This would be the real check for the  "has an initial type" check.

```ruby
expect(subject.type).to be_a Y2Storage::PartitionTables::Type
```

shundhammer: Yes, that actually makes a lot more sense; it does not depend on the default in the library.

package/yast2-storage-ng.spec
shundhammer: Yes, the partitioner does use the default - whatever comes from the library.

I changed those version numbers only for the unit tests, but your solution below is the better approach, so I can revert that new requirement to avoid too close coupling of those versions without really being forced to it.

PR #541, opened 2018-02-22
Added missing textdomain

shundhammer:
https://trello.com/c/B6Cx73jd/285-multiple-p1s-add-missing-textdomain-in-y-storage-ng
https://bugzilla.suse.com/show_bug.cgi?id=1081454 and more

## Strategy

We need one `textdomain` per file that uses translations so the _gettext_ tools can pick up the translations and put them into the correct .pot file.

We also need one `textdomain` in the execution path that uses the translations so the corresponding .po file is loaded. For that, one `textdomain` in some constructor in the inheritance hierarchy is sufficient.

What I did here was to add one `textdomain` in each file where they were missing at the most convenient place: Either in `def initialize` if there is one, or in a similar init-like method, or, if there is none, in the first method that has `_(...)` translation markers.

I tried to check as good as possible that in the latter case some parent class has a `textdomain` in its constructor. One is enough in this case, since it is always the same `textdomain "storage"`.

It might still be not 100% correct, though, but probably a lot better than having 39 files where it's completely missing.

### DIFF
--- a/src/lib/y2partitioner/widgets/lvm_edit_button.rb
+++ b/src/lib/y2partitioner/widgets/lvm_edit_button.rb
@@ -36,8 +36,6 @@ module Y2Partitioner
 
       # @macro seeAbstractWidget
       def label
-        textdomain "storage"
-
         _("Edit...")
       end
 
--- a/src/lib/y2partitioner/widgets/lvm_edit_button.rb
+++ b/src/lib/y2partitioner/widgets/lvm_edit_button.rb
@@ -29,6 +29,11 @@ module Y2Partitioner
   module Widgets
     # Button for editing a volume group or logical volume
     class LvmEditButton < DeviceButton
+      def initialize(*args)
+        super
+        textdomain "storage"
+      end
+
       # @macro seeAbstractWidget
       def label
         textdomain "storage"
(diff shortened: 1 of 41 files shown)


ancorgs:
Those `textdomain` in all kind of random places look too misplaced to me. If a class don't have a `initialize` method I would rather add it instead of just looking for another (sometimes very weird) place to put it.

```ruby
def initialize
  super
  textdomain "storage"
end
```

shundhammer:
You are right in principle, but that means first figuring out what (if any) parameters the parent constructor takes and duplicating that for the derived class.

ancorgs:
If you fear that, you can use something like `def initialize(*args)`.

An example to clarify what I mean:
```ruby
class TheBase
  def initialize(name)
    @name = name
  end 

  def hello
    puts "Hello, I'm #{@name}"
  end 
end

class TheSub < TheBase
  def initialize(*args)
    super
    @name = @name + "!!"
  end 
end

class AnotherSub < TheBase
end

test = TheSub.new("Fred")
test.hello

test2 = AnotherSub.new("Astaire")
test2.hello
```

shundhammer:
Ok, that sounds feasible.

src/lib/y2partitioner/widgets/lvm_edit_button.rb
```diff
@@ -29,8 +29,15 @@ module Y2Partitioner
...
     # Button for editing a volume group or logical volume
     class LvmEditButton < DeviceButton
+      def initialize(*args)
+        super
+        textdomain "storage"
+      end
+
       # @macro seeAbstractWidget
       def label
+        textdomain "storage"
```
ancorgs: Leftover?

shundhammer: Ah, right. Removing

PR #544, opened 2018-02-22
Prevent to modify used devices

joseivanlopez:
PBI: https://trello.com/c/YYSoL44u/191-1-partitioner-prevent-editing-of-used-partitions

### DIFF
--- a/src/lib/y2partitioner/actions/edit_blk_device.rb
+++ b/src/lib/y2partitioner/actions/edit_blk_device.rb
@@ -116,9 +116,8 @@ module Y2Partitioner
       def used_device_error
         return nil unless device.part_of_lvm_or_md?
 
-        # TRANSLATORS: Error message when trying to edit an used device. %{name} is
-        # replaced by the device name (e.g., /dev/sda1).
         format(
+          # TRANSLATORS: %{name} is replaced by a device name (e.g., /dev/sda1).
           _("The device %{name} is in use. It cannot be\n" \
             "edited. To edit %{name}, make sure it is not used."),
           name: device.name
--- a/src/lib/y2partitioner/actions/edit_blk_device.rb
+++ b/src/lib/y2partitioner/actions/edit_blk_device.rb
@@ -93,11 +93,34 @@ module Y2Partitioner
       #
       # @return [Boolean] true if the edit action can be performed; false otherwise.
       def run?
-        error = extended_partition_error || lvm_thin_pool_error
-        return true unless error
+        return true if errors.empty?
 
-        Yast::Popup.Error(error)
+        # Only first error is shown
+        Yast::Popup.Error(errors.first)
+
         false
+      end
+
+      # Errors when trying to edit a device
+      #
+      # @return [Array<Strings>]
+      def errors
+        [used_device_error, extended_partition_error, lvm_thin_pool_error].compact
+      end
+
+      # Error when trying to edit an used device
+      #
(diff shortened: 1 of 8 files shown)


src/lib/y2partitioner/actions/edit_blk_device.rb
```diff
@@ -93,13 +93,38 @@ def title
...
+      # Error when trying to edit an used device
+      #
+      # @note A device is being used when it forms part of an LVM or MD RAID.
+      #
+      # @return [String, nil] nil if the device is not being used.
+      def used_device_error
+        return nil unless device.part_of_lvm_or_md?
+
+        # TRANSLATORS: Error message when trying to edit an used device. %{name} is
+        # replaced by the device name (e.g., /dev/sda1).
```
ancorgs: Not sure if this line will be added to the .pot file. Please place it in the line below, right before the call to `_()`

ancorgs: Not to mention than I don't think our translators need a note explaining them than the string "The device is in use. It cannot be used"  is an error about the user trying to edit an used device. :wink:

src/lib/y2partitioner/actions/resize_blk_device.rb
```diff
@@ -80,18 +73,49 @@ def resize
...
+      #
+      # @note A device is being used when it forms part of an LVM or MD RAID.
+      #
+      # @return [String, nil] nil if the device is not being used.
+      def used_device_error
+        return nil unless device.part_of_lvm_or_md?
+
+        # TRANSLATORS: Error message when trying to resize an used device. %{name} is
+        # replaced by the device name (e.g., /dev/sda1).
+        format(
```
ancorgs: Same than above. To be sure I would move the comment (I keep only the part about the placeholder).

src/lib/y2partitioner/actions/resize_blk_device.rb
```diff
@@ -80,18 +73,49 @@ def resize
...
+      # Error when the device cannot be resized
+      #
+      # TODO: Distinguish the reason why it is not possible to resize, for example:
+      # * extended partition with committed logical partitions
+      #
+      # @return [String, nil] nil if the device can be resized.
+      def cannot_be_resized_error
+        return nil if device.resize_info.resize_ok?
+
+        # TRANSLATORS: a generic error message when a device cannot be resized.
```
ancorgs: Once again, translators usually read what they are translating. :wink:

src/lib/y2storage/blk_device.rb
```diff
@@ -320,6 +320,13 @@ def direct_md
       descendants.detect { |dev| dev.is?(:md) && dev.devices.include?(self) }
     end
 
+    # Whether the device forms part of an LVM or MD RAID
+    #
+    # @return [Boolean] true if it forms part of an LVM or MD RAID; false otherwise.
```
ancorgs: This looks like the over-explained pull request, not only to translators. :wink:

test/y2storage/blk_device_test.rb
```diff
@@ -582,6 +582,71 @@
     end
   end
 
+  describe "#part_of_lvm_or_md?" do
+    context "for a device not used in an LVM neither a MD RAID" do
```
ancorgs: https://thewritepractice.com/how-to-use-either-neither-or-and-nor-correctly/

PR #546, opened 2018-02-23
 Better handling of errors in the callbacks 

ancorgs:
This implements callbacks for https://trello.com/c/lo1jjvQDv/225-2-sles15-p1-1057869-build2011storage-ng-disk-will-be-encrypted-if-change-partition-settings-twice

probing

If the user clicks on abort the exception is still raised. That will be fixed in a separate pull request. But I wanted to ensure this to make it into RC1.

### DIFF
--- a/src/lib/y2storage/callbacks/commit.rb
+++ b/src/lib/y2storage/callbacks/commit.rb
@@ -21,7 +21,6 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "yast"
 require "y2storage/callbacks/libstorage_callback"
 
 module Y2Storage
--- a/src/lib/y2storage/callbacks/libstorage_callback.rb
+++ b/src/lib/y2storage/callbacks/libstorage_callback.rb
@@ -1,0 +1,76 @@
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "storage"
+require "yast"
+
+Yast.import "Report"
+Yast.import "Popup"
+Yast.import "Label"
+
+module Y2Storage
+  module Callbacks
+    # Mixin with methods that are common to all kind of callbacks used for the
+    # interaction with libstorage-ng.
+    module LibstorageCallback
+      include Yast::Logger
+      include Yast::I18n
+
+      # Callback for libstorage-ng to show a message to the user.
+      #
+      # Currently it performs no action, we don't want to bother the regular
+      # user with information about every single step. Libstorage-ng is
+      # already writing that information to the YaST logs.
+      #
+      # See Storage::Callbacks#message in libstorage-ng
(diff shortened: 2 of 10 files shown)


src/lib/y2storage/callbacks/libstorage_callback.rb
```diff
@@ -0,0 +1,76 @@
...
+      # @param message [String] error title coming from libstorage-ng
+      # @param what [String] details coming from libstorage-ng
+      # @return [Boolean] true will make libstorage-ng ignore the error, false
+      #   will result in a libstorage-ng exception
+      def error(message, what)
+        textdomain "storage"
+        log.info "libstorage-ng reported an error, asking the user whether to continue"
+        log.info "Error details. Message: #{message}. What: #{what}."
+
+        question = _("Continue despite the error?")
```
jreidinger: Hmm, when I see screenshot, I think we should more focus on user. What is risks when he continue? When probing failed, does it mean disk is not available? info about partition is missing? partition is missing? What is consequence of that failure and what is risk when user continue?

BTW second idea, why we do not have "retry" here? some users can fix on console or it can be just some race condition problem, so user can retry failed probing.

src/lib/y2storage/callbacks/libstorage_callback.rb
```diff
@@ -0,0 +1,76 @@
...
+        result = Yast::Report.ErrorAnyQuestion(
+          Yast::Popup.NoHeadline, "#{message}\n\n#{what}\n\n#{question}",
+          Yast::Label.ContinueButton, Yast::Label.AbortButton, :focus_no
```
jreidinger: So in autoyast it by default cancel when any probing failed? not sure if customers will be happy.

src/lib/y2storage/callbacks/probe.rb
```diff
@@ -21,26 +19,14 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "storage"
+require "yast"
```
jreidinger: why yast is required here? I do not see any Yast usage.

src/lib/y2storage/callbacks/commit.rb
```diff
@@ -21,61 +21,14 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "storage"
 require "yast"
```
jreidinger: same here, why it is needed?

src/lib/y2storage/callbacks/libstorage_callback.rb
ancorgs: The information we have and the options we can offer (like retrying) are limited what libstorage-ng offers us. So your comments should target the library.

src/lib/y2storage/callbacks/libstorage_callback.rb
ancorgs: Yes. That's what we decided so far. A failure in probing is something quite serious, so stopping is the less risky option. Let's see how much people complain about it.

src/lib/y2storage/callbacks/commit.rb
ancorgs: Is needed for the `require "y2storage/callbacks/libstorage_callback"` below to be correct. The `require "yast"` changes the loading paths, so y2storage is found under `/usr/share/YaST/lib`.

src/lib/y2storage/callbacks/probe.rb
ancorgs: Same than above. Yast usage is there, even if you don't see it. :wink:

src/lib/y2storage/callbacks/libstorage_callback.rb
jreidinger: well, I do not do any reviews for libstorage-ng. So maybe something for @jsrain or @aschnell to notice.

src/lib/y2storage/callbacks/commit.rb
jreidinger: that is wrong, lib/y2storage/callbacks/commit.rb is already in special load path, so how you load it without modifyed LOAD_PATH?

ancorgs: Fair enough, I will remove it.

PR #547, opened 2018-02-24
Better check for encryption password in the Partitioner

ancorgs:
Part of https://trello.com/c/lo1jvQDv/225-sles15-p5-1057869-build2011storage-ng-disk-will-be-encrypted-if-change-partition-settings-twice

Empty passwords are not longer accepted


imobachgs:
Manually tested.

PR #548, opened 2018-02-26
Fixed remaining translation

shundhammer:
https://trello.com/c/B6Cx73jd/285-multiple-p1s-add-missing-textdomain-in-y-storage-ng

https://bugzilla.suse.com/show_bug.cgi?id=1081454

The reason why this file was not picked up by y2makepot was that all translatable messages contain `#`. But that doesn't work for `_(...)` anyway since Ruby will expand the string inside the message before the message gets to `_(...)`, so the string is not constant, so it cannot be looked up as a `msgid` in the `.po` file.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Feb 26 16:11:12 UTC 2018 - shundhammer@suse.com
+
+- Use format(), not Ruby variable expansion for translated messages
+  (bsc#1081454)
+- 4.0.112
+
+-------------------------------------------------------------------
 Mon Feb 26 14:54:59 UTC 2018 - ancor@suse.com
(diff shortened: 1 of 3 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Feb 26 16:11:12 UTC 2018 - shundhammer@suse.com
+
+- Use format(), not Ruby variable expansion for translated messages
+  (bsc#1081454)
+- 4.0.112
```
joseivanlopez: Version should be 4.0.113

shundhammer: Fixed. Thanks for spotting this!

shundhammer:
BTW after `rake pot` the `storage.pot` file now also contains those two messages.

PR #552, opened 2018-02-27
Clone disk

joseivanlopez:
PBI: https://trello.com/c/QYrAzDGf/67-3-storageng-expert-partitioner-allow-to-clone-a-disk

### DIFF
--- a/src/lib/y2partitioner/actions/clone_disk.rb
+++ b/src/lib/y2partitioner/actions/clone_disk.rb
@@ -1,0 +1,115 @@
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "yast/i18n"
+require "y2partitioner/dialogs/disk_clone"
+require "y2partitioner/actions/controllers/disk_device"
+
+Yast.import "Popup"
+
+module Y2Partitioner
+  module Actions
+    # Action for cloning a disk
+    #
+    # To clone a disk means to copy the partition table and its partitions. Filesystems,
+    # encryptions, LVM or MD RAID over the partitions are not copied.
+    class CloneDisk
+      include Yast::I18n
+
+      # Constructor
+      #
+      # @param device [Y2Storage::Disk, Y2Storage::Dasd, Y2Storage::Multipath, Y2Storage::DmRaid]
+      def initialize(device)
+        textdomain "storage"
+
+        @controller = Controllers::DiskDevice.new(device)
+      end
+
+      # Checks whether it is possible to clone the device, and if so, the action is performed.
+      #
+      # @note An error popup is shown when the device cannot be cloned. In case the device
+      #   can be cloned, a dialog is presented to select over which disks to clone. Selected
+      #   disks are stored into the controller, see
+      #   {Controllers::DiskDevice#selected_devices_for_cloning}).
+      #
+      # @return [Symbol] :finish if the action is performed; :back or dialog result otherwise.
+      def run
+        return :back unless validate
+
+        result = Dialogs::DiskClone.run(controller)
+        return result if result != :ok
+
+        clone_disk
+        :finish
+      end
+
+    private
+
+      # @return [Controllers::DiskDevice]
+      attr_reader :controller
+
+      # Checks whether the clone action can be performed
+      #
+      # @note A disk can be cloned if it has a partition table and there are other
+      #   suitable disks where to clone it. In case the disk cannot be cloned, an
+      #   error message is shown.
+      #
+      # @return [Boolean] true if the clone action can be performed; false otherwise.
+      def validate
+        error = partition_table_error || suitable_devices_error
+        return true if error.nil?
+
+        Yast::Popup.Error(error)
+
+        false
+      end
+
+      # Error message when the disk has no partition table
+      #
+      # @return [String, nil] nil if the disk has partition table.
+      def partition_table_error
+        return nil if controller.partition_table?
+
+        _("There are no partitions on this disk, but a clonable\n" \
+          "disk must have at least one partition.\n" \
+          "Create partitions before cloning the disk.")
+      end
+
+      # Error message when there are no suitable devices where to clone the disk
+      #
+      # @return [String, nil] nil if there are suitable devices.
+      def suitable_devices_error
+        return nil if controller.suitable_devices_for_cloning?
+
+        _("This disk cannot be cloned. There are no suitable\n" \
+          "disks that could have the same partitioning layout.")
+      end
+
+      # Performs the cloning over the selected disks
+      #
+      # @see Controllers::DiskDevice#clone_to_device
+      def clone_disk
+        controller.selected_devices_for_cloning.each { |d| controller.clone_to_device(d) }
+      end
+    end
+  end
+end
--- a/src/lib/y2partitioner/actions/clone_disk.rb
+++ b/src/lib/y2partitioner/actions/clone_disk.rb
@@ -21,10 +21,9 @@
 
 require "yast"
 require "yast/i18n"
+require "yast2/popup"
 require "y2partitioner/dialogs/disk_clone"
 require "y2partitioner/actions/controllers/disk_device"
-
-Yast.import "Popup"
 
 module Y2Partitioner
   module Actions
@@ -78,7 +77,7 @@ module Y2Partitioner
         error = partition_table_error || suitable_devices_error
         return true if error.nil?
 
-        Yast::Popup.Error(error)
+        Yast2::Popup.show(error, headline: :error)
 
         false
       end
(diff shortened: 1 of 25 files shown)


src/lib/y2partitioner/actions/clone_disk.rb
```diff
@@ -0,0 +1,115 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "yast/i18n"
+require "y2partitioner/dialogs/disk_clone"
+require "y2partitioner/actions/controllers/disk_device"
+
+Yast.import "Popup"
```
jreidinger: JFYI yast2 now have more modern Popup class - https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/yast2/popup.rb

joseivanlopez: Yes, probably we should start to use it everywhere.

src/lib/y2partitioner/actions/clone_disk.rb
```diff
@@ -0,0 +1,115 @@
...
+      # Checks whether the clone action can be performed
+      #
+      # @note A disk can be cloned if it has a partition table and there are other
+      #   suitable disks where to clone it. In case the disk cannot be cloned, an
+      #   error message is shown.
+      #
+      # @return [Boolean] true if the clone action can be performed; false otherwise.
+      def validate
+        error = partition_table_error || suitable_devices_error
+        return true if error.nil?
```
jreidinger: what about `return true unless error`?

src/lib/y2partitioner/actions/controllers/disk_device.rb
```diff
@@ -0,0 +1,196 @@
...
+module Y2Partitioner
+  module Actions
+    module Controllers
+      # This class stores needed information to perform certain actions over a disk device,
+      # for example, the disks over which to clone the device. It also takes care of
+      # updating the devicegraph when needed.
+      class DiskDevice
+        # Current disk device
+        #
+        # @return [Y2Storage::BlkDevice] it should be a disk device,
```
jreidinger: should or have to be?

src/lib/y2partitioner/actions/controllers/disk_device.rb
```diff
@@ -0,0 +1,196 @@
...
+
+        # @return [Array<Y2Storage::Partitionable>]
+        attr_accessor :selected_devices_for_cloning
+
+        # Constructor
+        #
+        # @param device [Y2Storage::BlkDevice] it should be a disk device,
+        #   (i.e., device#is?(:disk_device) #=> true).
+        def initialize(device)
+          @device = device
```
jreidinger: please verify it is really disk_device otherwise we get nasty failures later

src/lib/y2partitioner/actions/clone_disk.rb
joseivanlopez: Well, I agree `unless error` is more readable, but in general I prefer to distinguish between nil and false cases. With `error.nil?` we explicitly express that there is no error. With `unless error` we are not sure at first glance if error is boolean or not.

Only a matter of taste.

src/lib/y2partitioner/actions/controllers/disk_device.rb
joseivanlopez: Yes, better "has to be".

src/lib/y2partitioner/actions/controllers/disk_device.rb
joseivanlopez: Ok.

src/lib/y2partitioner/actions/controllers/disk_device.rb
```diff
@@ -0,0 +1,199 @@
...
+
+        # Clones the partition table of the current device into the given device
+        #
+        # @note All partitions of the current device are cloned into the given
+        #   device, see {#clone_partition}.
+        #
+        # @param device [Y2Storage::Partitionable] device where to clone
+        def clone_partition_table(device)
+          device.create_partition_table(self.device.partition_table.type)
+          sorted_partitions = self.device.partitions.sort_by(&:name)
```
jreidinger: why self is needed?

src/lib/y2partitioner/actions/controllers/disk_device.rb
```diff
@@ -0,0 +1,199 @@
...
+          sorted_partitions.each { |p| clone_partition(device, p) }
+        end
+
+        # Clones a partition into the given device
+        #
+        # @note Cloned partition will have a region and partition id equal to the given
+        #   partition. Note that lucks, filesystems or other stuff are not cloned.
```
jreidinger: lucks typo, it should be LUKS

src/lib/y2partitioner/actions/controllers/disk_device.rb
joseivanlopez: `Controllers::DiskDevice#device` is the device associated to the controller. In the method we have also a param `device`. We need self here to take the partitions from the device associated to the controller and not from the param device.

src/lib/y2partitioner/confirm_recursive_delete.rb
```diff
@@ -0,0 +1,112 @@
...
+        button_box)
+    end
+
+    # @param headline [String]
+    # @param label_before [String]
+    # @param rich_text [String]
+    # @param label_after [String]
+    # @param button_term [Yast::UI::Term]
+    # @return [Boolean]
+    def fancy_question(headline, label_before, rich_text, label_after, button_term)
```
jreidinger: maybe this label before and label after can be added to Yast2::Popup so we can use common method.

src/lib/y2partitioner/dialogs/disk_clone.rb
```diff
@@ -0,0 +1,193 @@
...
+          controller.selected_devices_for_cloning = selected_devices
+        end
+
+        # FIXME: The help handle does not work without a wizard
+        #
+        # This handle should belongs to the dialog
+        def handle(event)
+          return nil if event["ID"] != :help
+
+          Yast::Wizard.ShowHelp(help)
```
jreidinger: I think this is bad idea, what if more widgets have help? it means that everyone shows his own help and only one can rule them all.

src/lib/y2partitioner/confirm_recursive_delete.rb
joseivanlopez: Yes, that would be nice.

joseivanlopez: Is there any way to do exactly the same with current Yast2::Popup ?

src/lib/y2partitioner/dialogs/disk_clone.rb
```diff
@@ -0,0 +1,193 @@
...
+        # @return [Y2Storage::BlkDevice]
+        def device
+          controller.device
+        end
+
+        # Whether there is any selected device
+        #
+        # @return [Boolean]
+        def selected_devices?
+          !selected_devices.empty?
```
jreidinger: other way can be `selected_devices.any?` which is a bit more readable for me.

src/lib/y2partitioner/widgets/disk_expert_menu_button.rb
```diff
@@ -43,30 +48,41 @@ def label
       #
       # @return [Array<[Symbol, String]>]
       def items
-        [[:create_partition_table, _("Create New Partition &Table...")]]
+        [
+          [:create_partition_table, _("Create New Partition &Table")],
```
jreidinger: why you remove that "..."?

src/lib/y2partitioner/widgets/disk_expert_menu_button.rb
```diff
@@ -43,30 +48,41 @@ def label
...
 
+      # @see Y2Partitioner::Actions::CreatePartitionTable
       def create_partition_table
         log.info("User selected 'create new partition table' for #{@disk.name}")
         Actions::CreatePartitionTable.new(@disk.name).run
       end
+
+      # @see Y2Partitioner::Actions::CloneDisk
+      def clone_disk
+        log.info("User selected 'clone this disk' for #{@disk.name}")
```
jreidinger: just my 2c but when debugging I prefer disk.inspect over disk.name as it shows quickly important info.

src/lib/y2partitioner/widgets/disk_expert_menu_button.rb
joseivanlopez: The string is in the old partitioner without dots.

src/lib/y2partitioner/actions/clone_disk.rb
jreidinger: well, for old code now it is not ideal time, but for new code it makes sense.

src/lib/y2partitioner/actions/clone_disk.rb
jreidinger: well, for me explicit nil? usually indicates that value can be `false` which is not true in this case.

src/lib/y2partitioner/actions/controllers/disk_device.rb
jreidinger: ok, I suggest to try nicer name here to avoid errors when modifying this method. What about param name `partitionable`?

src/lib/y2partitioner/confirm_recursive_delete.rb
jreidinger: I worry not, you can have headline + richtext, but not before/after label. We already face it before with some other dialog, so I should add it to my todo list.

src/lib/y2partitioner/widgets/disk_expert_menu_button.rb
jreidinger: OK. JFYI in past we had convention that button that leads to another wizard had this "...", but we weren't much consistent with it.

src/lib/y2partitioner/actions/clone_disk.rb
joseivanlopez: Different point of view. I use it to check that the object is pointing to "nothing".

src/lib/y2partitioner/actions/controllers/disk_device.rb
joseivanlopez: In general, we are avoiding `partitionable` concept, so I would not use that name. I can change it to `target_device`.

PR #553, opened 2018-02-28
- adapted to new activate callbacks in libstorage-ng

aschnell:
See https://github.com/openSUSE/libstorage-ng/pull/481.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Feb 28 14:26:21 CET 2018 - aschnell@suse.com
+
+- adapted to new activate callbacks in libstorage-ng (see
+  bsc#1082542)
+- 4.0.116
+
+-------------------------------------------------------------------
 Tue Feb 27 09:47:17 UTC 2018 - igonzalezsosa@suse.com
 
 - Keep encryption when adding a device to a LVM volume group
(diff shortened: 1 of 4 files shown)


ancorgs:
Not sure if I get the intention/meaning of the new argument.

aschnell:
You can now activate multipath even if it does not look to be multipath. As it turns out the current multipath detection is not working always (e.g. when only one path is there while the detection is done).

ancorgs:
But with this proposed PR we don't do it. So nothing has changed from the original bug point of view. Isn't it?

What are the implications of trying to activate multipath when there are no real multipath devices?

aschnell:
Yes, right now we do not do it. But we might have to do it, e.g. depending on a boot parameter as suggested in the bug. I currently want to provide the possibility to avoid API changes later on, e.g. in an L3.

joseivanlopez:
@ancorgs If I understood correctly, this PR only opens the possibility to use the param from the clients, but it does not try to change the current behavior. So, if something does not look a multipath, then multipath is not activated.

But maybe, even when something does not look a multipath, we should ask to the user, due to in some cases, the multipath is not correctly detected, so the user can decide.

ancorgs:
@aschnell Ok. makes sense

@joseivanlopez No, I would not change the behavior now. Let's see how the discussion evolves.

joseivanlopez:
@ancorgs Yes, I only comment it as a possibility. For now, I would keep the same behavior.

PR #554, opened 2018-02-28
ensure proper hierarchy when creating btrfs subvolumes (bsc#1078732)

wfeldt:
When creating btrfs subvolumes the proper hierarchy must be preserved. It's not possible to create `/foo/bar` when a subvolume `/foo/bar/xxx` already exists.

If the filesystem hasn't been created yet we can just rearrange all subvolumes to fit the new one in. For an existing filesystem there's nothing we can do except notify the user.

### DIFF
--- a/src/lib/y2storage/filesystems/btrfs.rb
+++ b/src/lib/y2storage/filesystems/btrfs.rb
@@ -82,16 +82,14 @@ module Y2Storage
 
       # Convert path to canonical form.
       #
-      # That is, a single slash between elements. No final slash except for "/".
+      # That is, a single slash between elements. No leading or final slashes.
       #
       # @param path [String] subvolume path
       #
       # @return [String] sanitized subvolume path
       #
       def canonical_subvolume_name(path)
-        path = path.squeeze("/")
-        path = path.chomp("/") if path != "/"
-        path
+        path.squeeze("/").chomp("/").sub(/^\//, "")
       end
 
       # Check if a subvolume can be created.
@@ -121,8 +119,8 @@ module Y2Storage
       #
       def subvolume_descendants(path)
         path = canonical_subvolume_name(path)
-        path += "/" unless path.end_with?("/")
-        btrfs_subvolumes.find_all { |sv| sv.path.start_with?(path) }
+        path += "/" unless path.empty?
+        btrfs_subvolumes.find_all { |sv| sv.path.start_with?(path) && sv.path != path }
       end
 
       # Returns the default subvolume, creating it when necessary
--- a/src/lib/y2storage/filesystems/btrfs.rb
+++ b/src/lib/y2storage/filesystems/btrfs.rb
@@ -80,6 +80,36 @@ module Y2Storage
         true
       end
 
+      # Check if a subvolume can be created.
+      #
+      # It can always be created if we're going to create the whole file
+      # system anyway.
+      #
+      # If the file system exists already there must at least be nothing
+      # else below path.
+      #
+      # @param path [String] subvolume path
+      #
+      # @return [Boolean]
+      #
+      def subvolume_can_be_created?(path)
+        return true unless exists_in_probed?
+        !subvolume_descendants_exist?(path)
+      end
+
+      # List of subvolumes hierarchically below path
+      #
+      # That is, subvolumes starting with path.
+      #
+      # @param path [String] subvolume path
+      #
+      # @return [Array<BtrfsSubvolume>]
+      #
+      def subvolume_descendants(path)
+        path += "/"
+        btrfs_subvolumes.find_all { |sv| sv.path.start_with?(path) }
+      end
+
       # Returns the default subvolume, creating it when necessary
       #
       # If a specific default subvolume path is requested, returns a subvolume with
@@ -119,7 +149,7 @@ module Y2Storage
         devicegraph.remove_btrfs_subvolume(subvolume)
       end
 
-      # Creates a new btrfs subvolume for the filesystem
+      # Create a new btrfs subvolume for the filesystem
       #
       # @note The subvolume mount point is generated from the filesystem mount point
       # and the subvolume path.
@@ -128,12 +158,23 @@ module Y2Storage
       #
       # @param path [string] absolute subvolume path
       # @param nocow [Boolean] no copy on write property
+      #
+      # @return [BtrfsSubvolume, nil] new subvolume
+      #
       def create_btrfs_subvolume(path, nocow)
-        subvolume = default_btrfs_subvolume.create_btrfs_subvolume(path)
-        subvolume.nocow = nocow
-        subvolume_mount_path = btrfs_subvolume_mount_point(path)
-        subvolume.create_mount_point(subvolume_mount_path) unless subvolume_mount_path.nil?
-        subvolume
+        subvolume = find_btrfs_subvolume_by_path(path)
+        return subvolume if subvolume
+
+        if !subvolume_can_be_created?(path)
+          log.error "cannot create subvolume #{path}"
+          return
+        end
+
+        if subvolume_descendants_exist?(path)
+          create_btrfs_subvolume_full_rebuild(path, nocow)
+        else
+          create_btrfs_subvolume_nochecks(path, nocow)
+        end
       end
 
       # Adds btrfs subvolumes defined from a list of specs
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -119,7 +149,7 @@ def delete_btrfs_subvolume(devicegraph, path)
         devicegraph.remove_btrfs_subvolume(subvolume)
       end
 
-      # Creates a new btrfs subvolume for the filesystem
+      # Create a new btrfs subvolume for the filesystem
```
ancorgs: Any reason to change this? In fact, the whole class (except the methods you added and this one you changed) are documented in third person (i.e. with the "s"), following the Ruby convention.

It's not a big deal if some people prefer to not follow the convention when they document a new method. But I would avoid a war in which each side "fixes" the wording used by the other side on every modification of a documented method. :wink:

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -128,12 +158,23 @@ def delete_btrfs_subvolume(devicegraph, path)
...
+      #
+      # @return [BtrfsSubvolume, nil] new subvolume
+      #
       def create_btrfs_subvolume(path, nocow)
-        subvolume = default_btrfs_subvolume.create_btrfs_subvolume(path)
-        subvolume.nocow = nocow
-        subvolume_mount_path = btrfs_subvolume_mount_point(path)
-        subvolume.create_mount_point(subvolume_mount_path) unless subvolume_mount_path.nil?
-        subvolume
+        subvolume = find_btrfs_subvolume_by_path(path)
```
ancorgs: So now there is a chance that `#create_btrfs_subvolume` returns a subvolume that it didn't actually create. Please, document that. It's completely unexpected to me.

src/lib/y2storage/filesystems/btrfs.rb
wfeldt: A quick (not representative) check shows it's not consistent; not even your comments within a class cf. `fake_device_factory.rb`.

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -80,6 +80,36 @@ def supports_btrfs_subvolumes?
...
+      # List of subvolumes hierarchically below path
+      #
+      # That is, subvolumes starting with path.
+      #
+      # @param path [String] subvolume path
+      #
+      # @return [Array<BtrfsSubvolume>]
+      #
+      def subvolume_descendants(path)
+        path += "/"
```
ancorgs: I know we never use paths like "/var/" (trailing slash) internally. But I fear something like that can end up here via autoyast profile or lack of validation in the Partitioner. So better be safe with something like
```ruby
path += "/" unless path.end_with?("/")
```

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -391,6 +432,98 @@ def path_without_prefix(subvolume_path)
...
+
+      # Find the most suitable parent for a new subvolume
+      #
+      # That is, the one with the longest path that is part of path.
+      #
+      # @param path [String] subvolume path
+      #
+      # @return [BtrfsSubvolume] parent subvolume
+      #
+      def best_match(path)
```
ancorgs: I'm the less suitable person to complain about names but `Btrfs#best_match` doesn't really look like the best name for a method to search a candidate parent.

src/lib/y2storage/filesystems/btrfs.rb
wfeldt: If that's the case the path should be put into a canonical form elsewhere. (I actually expected this.) Otherwise you'd have to deal with strange forms like `/foo///bar/` or whatever all over the place. At the very least provide a method for it.

BTW, just checked, and the path above is acceptable to the partitioner and it fails miserably when committing the change.

test/y2storage/filesystems/btrfs_test.rb
```diff
@@ -194,25 +194,114 @@
...
+      end
+
+      it "can insert a subvolume into an existing hierarchy" do
+        filesystem.create_btrfs_subvolume(path2, nocow)
+        filesystem.create_btrfs_subvolume(path3, nocow)
+        expect(filesystem.btrfs_subvolumes.map(&:path)).to include(path3)
+      end
+    end
+
+    context "when the filesystem already exists" do
```
ancorgs: It also makes sense to have another test checking that it returns nil if the subvolume couldn't be inserted

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -80,6 +80,36 @@ def supports_btrfs_subvolumes?
...
+      #
+      # If the file system exists already there must at least be nothing
+      # else below path.
+      #
+      # @param path [String] subvolume path
+      #
+      # @return [Boolean]
+      #
+      def subvolume_can_be_created?(path)
+        return true unless exists_in_probed?
```
ancorgs: I prefer to restrict the usage of `#exists_in_probed?` as much as possible. In general we would like to have code that can be used to calculate stuff taking any devicegraph as starting point, not only TheOneAndOnlyProbed . It's more testable and flexible.

For example, [BlkFilesystem#to_be_formatted?](https://github.com/yast/yast-storage-ng/blob/21a499e61698ffcf4d73d1ad532e501062371dea/src/lib/y2storage/blk_device.rb#L396) gets the initial devicegraph as argument. In the Partitioner it will get `Y2Partitioner::DeviceGraphs.system`. In the proposal that would be whatever was the starting point for the proposal (which strictly speaking doesn't HAVE to be probed). In a test it can be another thing.

wfeldt: `#create_btrfs_subvolume` doesn't know a devicegraph. So `in_probed?` is the best to do, and the semantics is correct. It compares to the situation the devicegraph commit will encounter later.

src/lib/y2storage/filesystems/btrfs.rb
ancorgs: `fake_device_factory.rb` is one of the classes mainly created and documented by @shundhammer, who also doesn't follow the Ruby convention. As said, I'm fine with some people not following it. I just want to avoid that everyone touches what others did to adjust it to his version.

test/y2storage/filesystems/btrfs_test.rb
```diff
@@ -230,15 +230,33 @@
...
       it "can not insert a subvolume into an existing hierarchy" do
         filesystem.create_btrfs_subvolume(path2, nocow)
-        filesystem.create_btrfs_subvolume(path3, nocow)
+        result = filesystem.create_btrfs_subvolume(path3, nocow)
+        expect(result).to be_nil
         expect(filesystem.btrfs_subvolumes.map(&:path)).to_not include(path3)
       end
+
+      it "and returns nil" do
+        expect(result).to be_nil
```
ancorgs: This test is doing nothing. `#create_btrfs_subvolume` has not been called here at all. I don't think you got how `let` works.

test/y2storage/filesystems/btrfs_test.rb
```diff
@@ -230,15 +230,33 @@
...
+      end
+    end
+  end
+
+  describe "#canonical_subvolume_name" do
+    let(:devicegraph) { Y2Storage::StorageManager.instance.staging }
```
ancorgs: Why is this `let(:devicegraph)` here? I fail to see what is used for

test/y2storage/filesystems/btrfs_test.rb
ancorgs: This is probably what you wanted to express:

```ruby
context "when the filesystem already exists and the subvolume does not fit in the existing hierarchy" do
  before do
    allow(filesystem).to receive(:exists_in_probed?) .and_return(true)
    filesystem.create_btrfs_subvolume(path2, nocow)
  end

  it "does not create the subvolume" do
    filesystem.create_btrfs_subvolume(path3, nocow)
    expect(filesystem.btrfs_subvolumes.map(&:path)).to_not include(path3)
  end

  it "returns nil" do
    result filesystem.create_btrfs_subvolume(path3, nocow)
    expect(result).to be_nil
  end
end
```

ancorgs:
LGTM. This is ready for the changelog entry and version bump

PR #556, opened 2018-02-28
add smarter space info for checking size of separate /boot (bsc#1078774)

jreidinger:


src/lib/y2storage/filesystems/base.rb
```diff
@@ -55,6 +56,29 @@ class Base < Mountable
       #   @return [SpaceInfo]
       storage_forward :detect_space_info, as: "SpaceInfo"
 
+      # smarted space info including caching and handling not probed devices.
+      #   @return [SpaceInfo]
+      def space_info
```
ancorgs: As commented on IRC. If we are only interested right now in knowing the free space, I would only implement a `#free_size` or `#free_space` method. Handcrafting a whole `SpaceInfo` object implies we should adapt this method in the future every time a new attribute is added to `SpaceInfo` in libstorage-ng. And maybe we don't have a good way to infer such new attribute.

So it's better to not over-promise.

jreidinger: so we just return free_size if detect pass and estimation if detect failed, right?

src/lib/y2storage/filesystems/base.rb
```diff
@@ -65,6 +82,17 @@ def in_network?
       def types_for_is
         super << :filesystem
       end
+
+      FREE_SPACE_FALLBACK = DiskSize.new(0)
+      def detect_free_space
+        return FREE_SPACE_FALLBACK unless is?(:blk_filesystem)
+
+        size = blk_devices.map(&:size).reduce(:+)
+        used = detect_resize_info.min_size
```
ancorgs: You could use `resize_info` instead of `detect_resize_info` here. In some scenarios it may be faster:
https://github.com/yast/yast-storage-ng/blob/97207385177e05fcd46afe48e66bfbc58c07391b/src/lib/y2storage/device.rb#L242

src/lib/y2storage/filesystems/base.rb
```diff
@@ -65,6 +82,17 @@ def in_network?
...
+        size - used
+      rescue Storage::Exception
+        return FREE_SPACE_FALLBACK
```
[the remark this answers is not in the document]
ancorgs: Why don't you add that as a comment? It's a good point and it will be lost after the PR is merged.

test/y2storage/filesystems/base_test.rb
```diff
@@ -0,0 +1,80 @@
...
+            expect(filesystem.free_space).to eq(blk_device.size - size)
+          end
+        end
+
+        context "detect_resize_info failed" do
+          it "returns zero" do
+            size = Y2Storage::DiskSize.MiB(0)
+            allow(filesystem).to receive(:detect_resize_info).and_raise(Storage::Exception, "Error")
+
+            expect(filesystem.free_space).to eq size
```
ancorgs: Doesn't it work as `expect(filesystem.free_space).to be_zero`?

test/y2storage/filesystems/base_test.rb
```diff
@@ -0,0 +1,80 @@
...
+          end
+        end
+      end
+
+      context "it is not block filesystem" do
+        it "returns zero" do
+          allow(filesystem).to receive(:is?).and_return(false)
```
ancorgs: Too much mocking. Wouldn't it be better to use a real Nfs object here?

ancorgs: Which boils down to another example in which is obvious that the RSpec tests are written while reading the source code (same workflow). Without reading the structure of the `if` and `rescue` I would have written:

- For an NFS share
  - if it can be mounted (#detect_space_info succeeds)
  - if it cannot be mounted (#detect_space_info fails)

- For a BlkFilesystem
  - if it can be mounted (#detect_space_info succeeds)
  - if it cannot be mounted (#detect_space_info fails)
    - if #detect_resize_info succeeds
    - if #detect_resize_info fails

Because for me NFS and blk filesystems are completely different scenarios, no matter whether they share same code path.

src/lib/y2storage/filesystems/base.rb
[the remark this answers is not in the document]
joseivanlopez: If `detect_resize_info` fails, probably the library cannot do much with that device. I would return zero.

src/lib/y2storage/filesystems/base.rb
```diff
@@ -55,6 +55,23 @@ class Base < Mountable
       #   @return [SpaceInfo]
       storage_forward :detect_space_info, as: "SpaceInfo"
 
+      # smart detection of free space
+      # it try to use detect_space_info and cache it. But if it failed, it try
+      # to compute it from detect_resize_info. If it failed or filesystem is
+      # not block filesystem, then it return zero size.
```
joseivanlopez: NP: incorrect usage of third person:

```
      # Smart detection of free space
      #
      # It tries to use detect_space_info and caches it. But if it fails, it tries
      # to compute it from detect_resize_info. If it fails again or filesystem is
      # not a block filesystem, then it returns zero size.
```

src/lib/y2storage/filesystems/base.rb
```diff
@@ -65,6 +82,17 @@ def in_network?
       def types_for_is
         super << :filesystem
       end
+
+      FREE_SPACE_FALLBACK = DiskSize.new(0)
+      def detect_free_space
```
joseivanlopez: NP: Probably the name does not fit very well. This method is simply proposing a fallback value, why not to call it `free_space_fallback` ? Only an idea.

src/lib/y2storage/filesystems/base.rb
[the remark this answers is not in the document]
jreidinger: if we cannot resize it, it does not mean it does not have enough space. So I add comment as ancor suggest.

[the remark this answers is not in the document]
joseivanlopez: Yes, I also would add a comment. I am only saying that from the library point of view, if resize_info fails, mos likely that device could not be correctly managed by the library (only thinking aloud)

test/y2storage/filesystems/base_test.rb
jreidinger: problem with your approach is that it makes hard to extend it. E.g. when there is third type of Filesystem, it will be completely ignore. Which I am not sure if it is good. So now after changing implementation maybe something like if filesystem have block devices or not, is determinant and not its type anymore.

src/lib/y2storage/filesystems/base.rb
```diff
@@ -65,6 +84,20 @@ def in_network?
...
+      def compute_free_space
+        # e.g. nfs where blk_devices cannot be queried
+        return FREE_SPACE_FALLBACK unless respond_to?(:blk_devices)
+
+        size = blk_devices.map(&:size).reduce(:+)
+        used = resize_info.min_size
+        size - used
+      rescue Storage::Exception
+        # it is questionable if this is correct behavior when resize_info failed,
+        # but there is high chance we can use it with libstorage, so better act like zero device.
```
ancorgs: I guess you meant "we can't"

src/lib/y2storage/filesystems/base.rb
```diff
@@ -65,6 +84,20 @@ def in_network?
...
+        return FREE_SPACE_FALLBACK
```
ancorgs: NP: surplus `return`

jreidinger:
manual tests works, so lets merge it.

PR #559, opened 2018-03-01
Proposal tries to use preferred partition type

joseivanlopez:
PBI: https://trello.com/c/cgDMlMWF/283-2-proposal-change-existing-partition-table-to-the-preferred-type


ancorgs:
I don't get why `GuidedSetup#clean_graph` and `PartitionKiller` create partition tables. They are cleaning/killing, so IMO they should only destroy the old partition table.

We already have code to propose a preferred partition table in completely empty disks, there is no reason to duplicate that logic/responsibility.

Moreover, if we finally don't use the disk, I believe that killing the partition table is a less aggressive side effect than replacing it with something different. And it also makes easier to recover the previous partition table if we finally don't use the disk (as soon as the library allows us to do so).

src/lib/y2storage/guided_proposal.rb
```diff
@@ -167,9 +167,45 @@ def space_maker
...
+    # @return [Y2Storage::Devicegraph]
     def clean_graph
-      @clean_graph ||= space_maker.delete_unwanted_partitions(initial_devicegraph)
+      return @clean_graph if @clean_graph
+
+      new_devicegraph = initial_devicegraph.dup
+
+      # FIXME: This could change partition tables of devices that are not used
+      # for the proposal.
+      fix_empty_partition_tables(new_devicegraph)
```
ancorgs: Didn't we discuss that this method should...?

(A) only delete the current partition table without proposing the new one (that's already done in another part)
(B) return the list of the affected devices (sids or names), so the deleted partition tables can be restored at the end

About (B) I believe we agreed to place a comment here for the future like:
```
# TODO: remember the list of affected devices so we can restore their partition tables at
# the end of the process for those devices that were not used, as soon as libstorage-ng
# allows us to copy sub-graphs.
```

src/lib/y2storage/partitionable.rb
```diff
@@ -262,11 +262,20 @@ def preferred_ptable_type
...
-    # @return [PartitionTable]
+    # @return [PartitionTables::Base]
     def ensure_partition_table
       partition_table || create_partition_table(preferred_ptable_type)
     end
 
+    # Deletes the partition table
+    #
+    # @note All dependent devices will be deleted (MD RAIDs, LVM, etc).
+    def delete_partition_table
```
ancorgs: It would be nice to be more consistent with these names.

On one hand, we have `Device#remove_descendants`, `Mountable#remove_mount_point`, `Md#remove_device`, `LvmVg#remove_lvm_pv` or `BlkDevice#remove_encryption`.

On the other hand, we have `PartitionTables#Base.delete_partition`, `PartitionTables#Base.delete_all_partitions` and `LvmVg#delete_lvm_lv`, `BlkDevice#delete_filesystem`.

Not something for this PR, I know... but cases like `BlkDevice#remove_encryption` vs `BlkDevice#delete_filesystem` makes me wanna cry. Looks like the PHP API.

joseivanlopez: Yes, I also hate it. In this case I have to use on of both options, and the closer was `#delete_partition`, so ...

test/y2storage/proposal_scenarios_x86_test.rb
```diff
@@ -133,7 +133,7 @@
     end
 
     context "in a PC with an empty partition table" do
-      let(:scenario) { "empty_hard_disk_mbr_50GiB" }
+      let(:scenario) { "empty_hard_disk_gpt_50GiB" }
```
ancorgs: This in fact changes the starting point, which looks a bit like cheating. What about this?

```ruby
let(:scenario) { "empty_hard_disk_mbr_50GiB" }
let(:expected_scenario) { "empty_hard_disk_gpt_50GiB" }
```

That would indeed prove it behaves as we want. With your change we are actually doing the same tests twice (there is another "_in a PC with an empty GPT partition table_" below in the same file).

joseivanlopez: Yes, that was I thought, but then I do it wrong.

PR #560, opened 2018-03-02
 Ensure PReP as primary (bsc#1082468) 

ancorgs:
Implements https://trello.com/c/LKmwH0cb/286-3-sles15-p1-1082468-sles15-beta7-error-during-yast-installation-bootloader-prep

- General support to enforce primary planned partitions.
- Ensures the new PReP partition suggested by BootRequirementsChecker to be primary.
- Updated documentation.

### DIFF
--- a/src/lib/y2storage/volume_specification_builder.rb
+++ b/src/lib/y2storage/volume_specification_builder.rb
@@ -156,7 +156,6 @@ module Y2Storage
         v.min_size = DiskSize.MiB(2)
         v.desired_size = DiskSize.MiB(4)
         v.max_size = DiskSize.MiB(8)
-        # So far we are always using msdos partition ids
         v.partition_id = PartitionId::PREP
       end
     end
--- a/doc/boot-partition.md
+++ b/doc/boot-partition.md
@@ -1,7 +1,7 @@
 Boot Partition Layout / Restrictions For Storage Proposal
 =========================================================
 
-[Revision 2018-02-15]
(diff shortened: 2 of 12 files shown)


src/lib/y2storage/volume_specification_builder.rb
```diff
@@ -156,7 +156,6 @@ def fallback_for_prep
         v.min_size = DiskSize.MiB(2)
         v.desired_size = DiskSize.MiB(4)
         v.max_size = DiskSize.MiB(8)
-        # So far we are always using msdos partition ids
```
jreidinger: why primary is not here, but at boot requirements? I have to say it is a bit confusing to have it at two places. What is rationale?

ancorgs: `VolumeSpecification` and `Planned::Partition` are in fact different beasts. The first one can represent both a LV and a partition. It's the OOP direct representation of the volumes as they are specified in the control file.

A `Planned::Partition` is a more down-to-earth representation of something that is indeed being to be created, and contains additional information that is only relevant for partitions.

ancorgs: BTW, in the past we hadn't `Planned::Partition` and `Planned::LvmLv`, we did the whole process with a more general `PlannedVolume` class... and the code turned to be VERY hard to follow. 

Creating more concrete and specialized `Planned::Partition` and `Planned::LvmLv` objects helped a lot making understandable code.

PR #561, opened 2018-03-05
show specialized warning when /boot/efi is on MD raid level 1

jreidinger:


src/lib/y2storage/boot_requirements_strategies/uefi.rb
```diff
@@ -39,10 +39,16 @@ def needed_partitions(target)
       # @return [Array<SetupError>]
       def warnings
         res = super
+        textdomain "storage"
```
ancorgs: Isn't this a weird location to call textdomain?

src/lib/y2storage/boot_requirements_strategies/uefi.rb
```diff
@@ -39,10 +39,16 @@ def needed_partitions(target)
...
 
+        # EFI in RAID can work, but it is not much reliable. see bsc#1081578#c9
+        if efi_in_md_raid1?
+          msg = _("/boot/efi is inside MD RAID. In general it is discouraged to use this setup,
+            but it can work.")
```
ancorgs: Well, according to https://fate.suse.com/322485 and the discussions in the bug, I don't think this is really discouraged. We can't ensure it will work in all cases, but it's a reasonable approach. The user """just""" has to be careful setting the flags (and I bet that will come back as a feature request for YaST).

What about something like this?

```
"/boot/efi is in a software RAID. That setup is not guaranteed to boot.\n" \
"Proceed only if you know the implications."
```

But I'm not sure. Maybe @mvidner or @kobliha have better ideas about the wording.

src/lib/y2storage/boot_requirements_strategies/uefi.rb
jreidinger: well, as it is only place calling localized message it is fine for me, but I can create initialize there if needed.

src/lib/y2storage/boot_requirements_strategies/uefi.rb
shundhammer: That message LGTM.

jreidinger: @ancorgs it sounds good. About discouraging, problem is with any software writting there some content. Problem is also when we start using newer md raid metadata, even 1.2 which keep 4k for boot code is not enough for /boot/efi I worry. but message sounds good enough for me.

kobliha: That setup is not guaranteed to boot `in all cases`.

I'd maybe try to point to possible failures, but that could make the text toooooo long to read and understand.

kobliha: PLS IMO we should have a check for version, if it's not 1.0 then it needs to be an error.
Or Exception?

ancorgs: libstorage-ng always uses version 1.0. It was requested explicitly to support scenarios like this. In theory, if we adopt 1.X in the future (with X > 1 in any case), such decision will be taken having things like this into account. So IN THEORY, trusting libstorage-ng to choose the appropriate metadata version should be okish.

mvidner: Josef's English is OK, but the message is unclear, saying, "don't do this, or maybe do". Ancor's message is better, saying "don't do this".

But after reading the Fate item (and there is no "bug"?!) I still don't understand. Suppose that I made a SW RAID1, and manually put /boot/efi on it, thinking, "OK, since this is a mirror, it will surely work". Then I see a message like this. I would happily adjust the setup manually if the message only told me.

Maybe we don't want to expose the user to the whole gory story but let's explain the concrete problems, or hint at them, in the comments.

jreidinger: @mvidner related entries - https://bugzilla.suse.com/show_bug.cgi?id=1081578#c9 and https://fate.suse.com/322485 and old one for SLE12 SP2 https://fate.suse.com/314829

mvidner: Thanks! So what about this for a comment near the message?

```rb
# See bsc#1081578#c9, FATE#322485, FATE#314829
# - RAID metadata must be somewhere where it will not interfere with UEFI reading
#   the disk. YaST currently uses "mdadm --metadata=1.0" which is OK
# - The partition boot flag must be on
# - The partition RAID flag must be on (but setting it resets the boot flag)
```

src/lib/y2storage/boot_requirements_strategies/uefi.rb
mvidner: This will break as soon as you add another message, and forget to update the textdomain since a grep will show it is already there... please move it.

kobliha:
Well, I thought we talked about `Warning`, not `Error`?

jreidinger:
@kobliha well, we use same header, it is same for all warnings in expert partitioner.

src/lib/y2storage/boot_requirements_strategies/uefi.rb
```diff
@@ -40,9 +45,16 @@ def needed_partitions(target)
       def warnings
         res = super
 
+        # EFI in RAID can work, but it is not much reliable. see bsc#1081578#c9
+        if efi_in_md_raid1?
+          msg = _(
+            "/boot/efi is in a software RAID. That setup is not guaranteed to boot.\n" \
```
ancorgs: @kobliha suggested "That setup is not guaranteed to boot in all cases".

kobliha:
/me likes it, PLS create a new screenshot for the blog :)

jreidinger:
@kobliha screenshot updated. So only wait for @ancorgs for this title change to review.

PR #562, opened 2018-03-05
AY: retry without automatically added boot devices

imobachgs:
Fixes [bsc#1082999](https://bugzilla.suse.com/show_bug.cgi?id=1082999).

This happens when trying to reuse an existing layout which does not contain
a boot partition. Instead of just crashing, AutoYaST should warn the user
and try to go ahead.


src/lib/y2storage/autoinst_issues/could_not_create_boot.rb
```diff
@@ -0,0 +1,52 @@
...
+    # There is no enough disk space to build the storage proposal
+    class CouldNotCreateBoot < Issue
+      def initialize(*args)
+        super
+        textdomain "storage"
+      end
+
+      # Fatal problem
+      #
+      # @return [Symbol] :warning
```
joseivanlopez: NP: it returns `:warn`, no `:warning`.

PR #563, opened 2018-03-06
Proposal with implicit partition table

joseivanlopez:
PBI: https://trello.com/c/HSxXeicS/298-3-proposal-with-implicit-partition-table-fba-or-ldl


src/lib/y2storage/free_disk_space.rb
```diff
@@ -50,6 +50,13 @@ def initialize(disk, region)
       @region = Y2Storage::Region.new(region)
     end
 
+    # Whether this free space belongs to an implicit partition table
+    #
+    # @return [Boolean]
+    def in_implicit_partition_table?
```
ancorgs: I would not mention the implicit partition table as part of the name. That's the "why" (the only "why" we have so far but maybe not the only reason to reuse partitions in the future).

The "what" is the fact that this space is not really an unpartitioned disk chunk, but a partition that needs to be reused as-is.

So I would go with something as explicit as `#partition_to_reuse?`

src/lib/y2storage/partition_tables/implicit_pt.rb
```diff
@@ -0,0 +1,69 @@
...
+      #
+      # @raise [Y2Storage::Error] if there is no partition
+      # @return [Y2Storage::Partition]
+      def partition
+        raise Error, "Implicit partition table must have one partition" if partitions.empty?
+
+        partitions.first
+      end
+
+      # Free spaces in the inplicit partition table
```
ancorgs: Typo: inplicit

src/lib/y2storage/partition_tables/implicit_pt.rb
```diff
@@ -0,0 +1,69 @@
...
+      #
+      # @note The single partition cannot be removed. The whole partition is
+      #   considered as freespace when it is not in use (has not filesystem).
```
ancorgs: typo: freespace

ancorgs: "has no filesystem"... or is not a PV, or is not part of a software RAID....

src/lib/y2storage/free_disk_space.rb
ancorgs: And maybe we should add some information at the top class definition. Explaining that one of these objects can represent an unpartitioned chunk (maybe within and extended partition) or the space of a partition to be reused.

src/lib/y2storage/planned/assigned_space.rb
```diff
@@ -188,6 +189,20 @@ def require_end_alignment?
...
 
+      # Whether there are too many partitions to allocate in this space
+      #
+      # @note A disk space holded by an implicit partition table cannot
+      #   allocate more than one partition.
+      #
+      # @see Y2Storage::PartitionTables::ImplicitPt
+      #
+      # @return [Boolean]
+      def too_many_partitions?
```
ancorgs: This does not detect all the cases in which there are too many partitions. So the name is confusing. What you are really doing is checking wrong usage of spaces with a true `#partition_to_reuse?` (sorry, I used my suggested name here). So the method name should say that and not pretend to be smarter.

src/lib/y2storage/planned/partitions_distribution.rb
```diff
@@ -348,16 +348,18 @@ def too_many_primary_with_extended?(primary_spaces, ptable)
       def too_many_primary?(spaces, ptable)
         primary_spaces = spaces.select { |s| s.partition_type == :primary }
 
-        if !ptable.extended_possible?
+        if ptable.type.is?(:implicit)
+          num_partitions(primary_spaces) > ptable.max_primary
```
ancorgs: We should never get here. Isn't it?

Anyway, I would compare `num_partitions(spaces) > ptable.max_primary`

src/lib/y2storage/proposal/partition_creator.rb
```diff
@@ -158,6 +160,28 @@ def create_partition(planned_partition, free_space, primary)
...
+      # @raise [Y2Storage::NoMorePartitionSlotError] if the single implicit partition is
+      #   already in use.
+      #
+      # @param ptable [Y2Storage::PartitionTables::ImplicitPt]
+      # @return [Y2Storage::Partition] single implicit partition
+      def use_implicit_partition(ptable)
+        partition = ptable.partition
+        return partition unless implicit_partition_in_use?(partition)
+
+        raise NoMorePartitionSlotError
```
ancorgs: Please add a string describing what the problem was when raising a relatively generic exception.

src/lib/y2storage/proposal/partition_killer.rb
```diff
@@ -69,10 +69,12 @@ def find_partition(sid)
...
-      # If the partition was the only remaining logical one, it also deletes the
-      # now empty extended partition. The partition table is also deleted when
-      # the last partition is deleted. In case of AutoYaST, deletion of the partition
-      # table is avoided because AutoYaST uses its own logic to reuse partition tables.
+      #
+      # @note If the partition was the only remaining logical one, it also deletes the
+      #   now empty extended partition. The partition table is also deleted when
+      #   the last partition is deleted. In case of AutoYaST, deletion of the partition
+      #   table is avoided because AutoYaST uses its own logic to reuse partition tables.
+      #   In case of a sigle implicit parition, the partition is not deleted, but only wiped.
```
ancorgs: typos: sigle, parition.

Please clarify what "wiped" means in this context.

src/lib/y2storage/planned/partitions_distribution.rb
joseivanlopez: Why not?

Ok, better to use `spaces`.

PR #565, opened 2018-03-07
Consistent mount_by for subvolumes

shundhammer:
https://trello.com/c/c4V8FLoO/276-1-sles15-p1-1080408-hps-bug-sles15-beta6-installer-writes-inconsistent-fstab-entries-for-btrfs-subvolumes

https://bugzilla.suse.com/show_bug.cgi?id=1080408

When creating subvolumes or their mount points, copy _mount_by_ from the parent Btrfs.

Also, when changing _mount_by_ for a Btrfs in the expert partitioner, apply that change to all subvolumes (all that have a mount point) as well; this is the best we can do right now because we don't have much of a user interface for any more detailed settings for subvolumes.

### DIFF
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -107,6 +107,13 @@ module Y2Partitioner
         @controller.filesystem
       end
 
+      # Check if the underlying filesystem is a btrfs.
+      #
+      # @return [Boolean]
+      def btrfs?
+        @controller.btrfs?
+      end
+
       # Mount point of the current filesystem
       #
       # @return [Y2Storage::MountPoint]
@@ -337,6 +344,10 @@ module Y2Partitioner
 
       def store
         mount_point.mount_by = selected_mount_by
+        filesystem.copy_mount_by_to_subvolumes if btrfs?
+        # TODO: If some day we get more a more sophisticated UI for editing
+        # subvolumes, don't always simply overwrite the subvolumes' mount_by
+        # with whatever is set here - ask the user if this is desired.
       end
 
       def init
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -317,7 +317,11 @@ module Y2Partitioner
       # @return [Boolean] true if the label is duplicated; false otherwise.
       def duplicated_label?
         return false if value.empty?
-        working_graph.filesystems.any? { |f| f.sid != filesystem.sid && f.label == value }
+        working_graph.filesystems.any? do |fs|
+          next false if fs.sid == filesystem.sid
+          next false unless fs.respond_to?(:label) # NFS doesn't support labels
+          fs.label == value
+        end
       end
 
       # Widget to select the mount by option
--- a/test/y2storage/filesystems/btrfs_test.rb
+++ b/test/y2storage/filesystems/btrfs_test.rb
@@ -867,25 +867,37 @@ describe Y2Storage::Filesystems::Btrfs do
 
   describe "#copy_mount_by_to_subvolumes" do
     let(:subvol_mount_points) { filesystem.btrfs_subvolumes.reject { |s| s.mount_point.nil? } }
-    let(:subvol_mount_bys) { subvol_mount_points.map { |m| m.mount_by.to_sym } }
-    let(:btrfs_mount_by) { filesystem.mount_by.to_sym }
 
-    it "starts with both the btrfs and all subvolumes mounted by label" do
+    # Need some real methods here to avoid 'let' blocks being cached between
+    # the 'before' block and the 'it' block. Notice that 'let!' does NOT
+    # prevent that in this case: It would only reevaluate the block between
+    # different examples, not between the 'before' block and the 'it' block
+    # within the same example.
+    def subvol_mount_bys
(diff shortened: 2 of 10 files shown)


test/y2storage/filesystems/btrfs_test.rb
```diff
@@ -864,4 +864,28 @@
       end
     end
   end
+
+  describe "#copy_mount_by_to_subvolumes" do
+    let(:subvol_mount_points) { filesystem.btrfs_subvolumes.reject { |s| s.mount_point.nil? } }
+    let(:subvol_mount_bys) { subvol_mount_points.map { |m| m.mount_by.to_sym } }
+    let(:btrfs_mount_by) { filesystem.mount_by.to_sym }
+
+    it "starts with both the btrfs and all subvolumes mounted by label" do
```
ancorgs: Something like this does not make much sense in RSpec:

```ruby
describe "#whatever_method" do
  it "foo bar" do
    # Something that doesn't call #whatever_method
  end
end
```

Following the convention, if you use a name of a method as title for the "describe", then each "it" block is supposed to describe one aspect of the behavior of such method. So if the method is not called... :thinking: Moreover the output should look like a description of how the method works. Something like this describes how the methods works:

```
Btrfs
  - #copy_mount_by_to_subvolumes
    * copies the mount_by of the Btrfs to all its subvolumes

MountPoint
  - #mount_by=
    * in a Btrfs device
      - sets the mount_by of the Btrfs
      - does not affect the mount_by of the subvolumes
```

Or maybe, to have everything as close as possible:

```
Btrfs
  - #copy_mount_by_to_subvolumes
    * copies the mount_by of the Btrfs to all its subvolumes

  - changing the mount_by with #mount_point.mount_by=
    * sets the mount_by of the Btrfs
    * does not affect the mount_by of the subvolumes
```

But something like this doesn't describe the method Btrfs#copy_mount_by_to_subvolumes (as the two first lines state):
```
Btrfs
  - #copy_mount_by_to_subvolumes
    * starts with both the btrfs and all subvolumes mounted by label
    * changing only the btrfs mount_by does not affect the subvolumes
```

At most, it's a description of the starting point of your test. That's what "let" and "before" are for (setting the starting point). Not the role of "it". And take into account the three "its" could be executed in any order.

src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -337,6 +348,10 @@ def label
 
       def store
         mount_point.mount_by = selected_mount_by
+        filesystem.copy_mount_by_to_subvolumes if btrfs?
```
ancorgs: I'm not sure if having to call `copy_mount_by_to_subvolumes` every time we change the mount_by of the Btrfs is robust enough. I would sleep better if that would be a responsibility of `MountPoint#mount_by=`.

99.99% of the times you want the Btrfs to be on sync with the subvolumes, so I think it's better if the models already guarantee that implicitly. Maybe is more cumbersome in the (quite weird) case in which you really want to enforce the Btrfs and its subvolumes to be out of sync (if that's technically safe) but... Do we want to make this more error prone for the regular case just to make the exception easier?

ancorgs: On the other hand, the idea of having a separate `Action` (even a separate `Controller` in this case of a multi-step action) is to have clean widgets with as little logic as possible. So the business logic is not scattered through the UI code.

So I would suggest something like:
```ruby
  def store
    controller.mount_by = selected_mount_by
  end
```

And taking the decisions in the controller:

```ruby
def mount_by=(value)
  mount_point.mount_by = value
  filesystem.copy_mount_by_to_subvolumes if btrfs?
end
```

So the widget stays as a widget, and not as an intelligent thingie that needs to know whether it's in a btrfs context or not and act in consequence.

test/y2storage/filesystems/btrfs_test.rb
shundhammer: The point of this is to assert the correct starting conditions. I spent **hours** trying to debug this stuff, and it turned out several times that you cannot assume _anything_ - from no `mount_by` set to having two subvolumes without a mount point to whatever.

src/lib/y2partitioner/widgets/fstab_options.rb
shundhammer: So you think that hiding that snippet in the controller is a better approach?

I disagree. Nobody will ever look there. Everybody is just looking at the place where the user changes the value, and that's in the widget.

Right now, we have the relevant code scattered among a half-dozen places; this is what took me so long to figure out where to put what. Adding yet another place that will surprise anybody doing maintenance on the code is clearly not the way to go IMHO.

Add to that the fact that this particular controller is way too large and way too complex already - it is the one controller that serves 3 or 4 different dialogs already. We already talked about the need to split it up, so the places where the relevant code is scattered will multiply even more.

test/y2storage/filesystems/btrfs_test.rb
shundhammer: ...all of which just makes _RSpec_ just barf and crash with utterly useless error messages that are completely undebuggable, of course.

ancorgs: If you want to ensure the starting condition for all the "its", you can use `expect` in a "before" block. So you can sleep well because you ensure the starting point. That's declarative enough without having to misuse "it".

shundhammer: Ok then.

PR #567, opened 2018-03-08
Fix for bsc#1084213 and bsc#1084261

ancorgs:
First step to fix https://trello.com/c/MSBg5L2S/304-sles15-p1-1084213-storage-ng-home-subvolume-not-created-if-user-opts-to-not-have-a-home-partition

### DIFF
--- a/src/lib/y2storage/planned/can_be_formatted.rb
+++ b/src/lib/y2storage/planned/can_be_formatted.rb
@@ -39,8 +39,12 @@ module Y2Storage
       # @return [String] UUID to enforce in the filesystem
       attr_accessor :uuid
 
-      # @return [Array<SubvolSpecification>] Btrfs subvolume specifications
-      attr_accessor :subvolumes
+      # Btrfs subvolume specifications.
+      #
+      # @see #subvolumes= which doesn't work exactly as the standard Ruby setter
+      #
+      # @return [Array<SubvolSpecification>]
+      attr_reader :subvolumes
 
       # @return [String] Parent for all Btrfs subvolumes (typically "@")
       attr_accessor :default_subvolume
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -418,6 +418,8 @@ def remove_shadowed_subvols(planned_devices)
           subvols_added =
             device.respond_to?(:mount_point) && device.mount_point == "/" && @default_subvolumes_used
 
+          # Use #dup to avoid bsc#1084213
```
joseivanlopez: I would `dup` the list just when it is assigned from the settings:
https://github.com/ancorgs/yast-storage-ng/blob/b59dafb1b2e7c1d58d35e914907ea7fe7b0ef7dc/src/lib/y2storage/proposal/autoinst_devices_planner.rb#L233

This would avoid possible similar errors in other places.

src/lib/y2storage/proposal/devices_planner_strategies/base.rb
```diff
@@ -105,10 +105,13 @@ def remove_shadowed_subvolumes(planned_devices)
           planned_devices.each do |device|
             next unless device.respond_to?(:subvolumes)
 
+            # Use #dup to avoid bsc#1084213
```
joseivanlopez: Same here:
https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/proposal/devices_planner_strategies/ng.rb#L126

src/lib/y2storage/proposal/autoinst_devices_planner.rb
ancorgs: There are at least three assignations of that. So duping all of them looks wrong.

At most, I would define a setter. Like
```ruby
module Planned::CanBeFormatted
  def subvolumes=(value)
    @subvolumes = value.dup
  end
end
```

That was indeed my first option, but I somehow though it would have been more surprising. Less rubyist, sort to say.

I don't care much, either solution (the one on this PR or redefining the setter) are acceptable to me.

ancorgs:
Verified manually that this fixes the reported bug.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
ancorgs: So:

- Solution 1: the one currently on this PR. It's correct but @joseivanlopez pointed that we can make the same error in the future in another part.
- Solution 2: redefining the setter. It's also correct and doesn't have the mentioned problem, although it has slightly surprising implications (from the OOP rubyist POV).
- Solution 3: a `dup` on each assignation as suggested by @joseivanlopez. I don't like it. Very error prone.

I discussed offline with @imobachgs and he would go for 2. Since 2 is also pretty aligned with @joseivanlopez's concerns, I will switch from 1 to 2.

ancorgs:
New solution pushed with `-f`. I will re-test manually now

ancorgs:
New solution also tested manually

src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -60,6 +64,29 @@ def initialize_can_be_formatted
...
+      # duplicated, the contained objects are still shared.
+      #
+      # @example
+      #   planned.subvolumes = my_list
+      #   my_list << a_new_one
+      #   # planned.subvolumes doesn't contain a_new_one now. This is not the
+      #   # most common ruby behavior.
+      #   my_list.first.path = "changed" # This change affects planned.subvolumes
+      #   # because the object is also in that collection (not a deep copy).
+      #
```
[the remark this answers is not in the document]
ancorgs: Well, a deep copy would be even more unexpected. When I pass a list of objects I'm actually passing those objects, not a copy of them. Two objects that are not the same object but represent the same thing are a constant source of bugs (they don't work with `include?`, `==`, etc.)

PR #568, opened 2018-03-09
Allow to modify patition id

joseivanlopez:
PBI: https://trello.com/c/r6sk4mQB/270-2-sles15-p2-1077868-storage-ng-expert-partitioner-has-several-issues-with-existing-partition-use


joseivanlopez:
Current appearance with "Partition Id" aligned to left and with more vertical space. Button for Btrfs subvolumes is also more separate from above element.

partition_id

PR #569, opened 2018-03-12
Read-only volumes

imobachgs:
This PR contains a couple of related changes.

## Improving the VolumeSpecification API

Some time ago, we introduced a new `VolumeSpecificationBuilder` class which is responsible for building `VolumeSpecification` objects. It allowed us to move the logic about boot properties calculation (sizes, filesystem types, etc.) to a central place to be used by the boot requirements checker and the AutoYaST proposal.

In this PR I've added a new `VolumeSpecificaton.for` method which relies on `VolumeSpecificationBuilder` but prevents the rest of the code to know about that class. Additionally, it keeps a cache with already calculated volumes.

## Bug #1079000

This bug is about setting the root filesystem as "ro" for CaaSP/Kubic. It has two different parts:

* Setting the "ro" mount option using the *guided proposal*, *AutoYaST proposal* and in the *Expert Partitioner*.
* Setting the read-only property (yast2-installation package) at the end of the installation. See https://github.com/yast/yast-installation/pull/668.

See [bsc#1079000](https://bugzilla.suse.com/show_bug.cgi?id=1079000).

## Bug #bsc#1077866

The second bug is about setting the right subvolumes list in the expert partitioner.

See [bsc#1077866](https://bugzilla.suse.com/show_bug.cgi?id=1077866)

### DIFF
--- a/src/lib/y2storage/subvol_specification.rb
+++ b/src/lib/y2storage/subvol_specification.rb
@@ -151,7 +151,9 @@ module Y2Storage
     # @note The new subvolume is set as 'can be auto deleted'.
     #
     # @param filesystem [Filesystems::Btrfs]
-    # @return [BtrfsSubvolume]
+    # @return [BtrfsSubvolume,nil] New BtrfsSubvolume; nil if the subvolume could not
+    #   be created.
+    # @see FileSystems::Btrfs#create_btrfs_subvolume
     def create_btrfs_subvolume(filesystem)
       subvolume_path = filesystem.btrfs_subvolume_path(path)
       subvolume = filesystem.create_btrfs_subvolume(subvolume_path, !copy_on_write)
--- a/src/lib/y2storage/subvol_specification.rb
+++ b/src/lib/y2storage/subvol_specification.rb
@@ -155,6 +155,7 @@ module Y2Storage
     def create_btrfs_subvolume(filesystem)
       subvolume_path = filesystem.btrfs_subvolume_path(path)
       subvolume = filesystem.create_btrfs_subvolume(subvolume_path, !copy_on_write)
+      return if subvolume.nil?
       subvolume.can_be_auto_deleted = true
       subvolume
     end
--- a/src/lib/y2storage/volume_specification_builder.rb
+++ b/src/lib/y2storage/volume_specification_builder.rb
@@ -22,7 +22,6 @@
 require "y2storage/filesystems/type"
 require "y2storage/partition_id"
 require "y2storage/proposal_settings"
-require "y2storage/subvol_specification"
 
 module Y2Storage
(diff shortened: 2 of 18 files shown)


src/lib/y2storage/subvol_specification.rb
```diff
@@ -155,6 +155,7 @@ def shadowed?(fs_mount_point, other_mount_points)
     def create_btrfs_subvolume(filesystem)
       subvolume_path = filesystem.btrfs_subvolume_path(path)
       subvolume = filesystem.create_btrfs_subvolume(subvolume_path, !copy_on_write)
+      return if subvolume.nil?
```
joseivanlopez: NP: Please, document in which cases this returns nil.

src/lib/y2storage/volume_specification_builder.rb
```diff
@@ -172,5 +174,17 @@ def fallback_for_swap
         v.max_size = DiskSize.GiB(2)
       end
     end
+
+    # Volume specification fallback for root (/)
+    #
+    # @return [VolumeSpecification]
+    def fallback_for_root
```
joseivanlopez: As commented offline, there should not be a fallback for root volume.

test/y2storage/volume_specification_test.rb
```diff
@@ -51,6 +50,17 @@
...
+    context "when a volume specification was already calculated" do
+      before do
+        described_class.for("/")
+      end
+
+      it "retrieves the value from the internal cache" do
+        expect(Y2Storage::VolumeSpecificationBuilder).to_not receive(:new)
+        described_class.for("/")
+      end
+    end
```
joseivanlopez: It would be nice to have a test for `#clear_cache`.

test/y2storage/proposal/devices_planner_strategies/ng_test.rb
```diff
@@ -639,6 +642,36 @@
...
+          let(:fs_type) { :ext4 }
+
+          it "is not set as read only" do
+            expect(planned_device.read_only).to eq(false)
+          end
+        end
+      end
+
+      context "and the volume is not set as read only" do
+        it "is not set as read only"
```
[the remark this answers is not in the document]
imobachgs: oops, yes, I forgot :)

src/lib/y2storage/volume_specification_builder.rb
imobachgs: Ok, I've reverted this change.

test/y2storage/volume_specification_test.rb
imobachgs: I was taking a shortcut because it is used in the before block for the `for` method. But you are right, a proper test should be used.

PR #570, opened 2018-03-13
Add btrfs_read_only documentation

imobachgs:
Related to #569.


imobachgs:
Good point. I've fixed that example too.

PR #571, opened 2018-03-13
Warn if reusing an existing system mount point in the partitioner during installation

shundhammer:
https://trello.com/c/BgAmje4M/

- Only during installation (not in the installed system, not during upgrade)
- Only if creating or editing a mount point that is not to be formatted
- Only for a small number of well-known system mount points:
  - /
  - /usr
  - /boot

...post a warning:

warn-reuse-part

Upon "Yes" (keep unformatted), apply the changes and close this dialog.
Upon "No", go back and let the user choose something else (format or change mount point).

### DIFF
--- a/src/lib/y2partitioner/widgets/format_and_mount.rb
+++ b/src/lib/y2partitioner/widgets/format_and_mount.rb
@@ -268,6 +268,12 @@ module Y2Partitioner
         nil
       end
 
+      def validate
+        validate_system_mount_points && validate_mount_by_label
+      end
+
+    private
+
       # It is necessary to prevent an empty filesystem label when the option mount_by is set
       # to label by default. The label value is validated when the user gives one value
       # (by editing the fstab options), but the user could mount a device without entering
@@ -278,7 +284,7 @@ module Y2Partitioner
       #
       # @return [Boolean] true if the label is not required or it is required and given; false
       #   otherwise.
-      def validate
+      def validate_mount_by_label
         return true if !formatted? || !mounted? || !mounted_by_label? || !empty_label?
 
         # TRANSLATORS: Error message when a device should be mounted by label but no label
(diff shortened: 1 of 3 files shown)


src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -287,7 +293,57 @@ def validate
...
+          _(
+            "If in doubt, better go back and mark this partition for\n" \
+            "formatting, especially if it is assigned to one of the standard mount points\n" \
+            "like /, /boot, /opt or /var.\n"
+          ) +
+          # continued popup text
+          _(
+            "If you decide to format the partition, all data on it will be lost.\n" \
+            "\n" \
+            "Really keep the partition unformatted?\n"
```
ancorgs: NP: The question is clear enough thanks to the previous explanation, but I would say something like "Really skip formatting the partition?".

I mean "keep the partition unformatted" sounds confusing when you actually are trying to reuse a formatted partition keeping the current format. There is nothing unformatted in such setup. Its simply not formatted by us.

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -287,7 +293,57 @@ def validate
...
 
-    private
+      # Check if a system mount point is reused without formatting the
+      # partition during installation and warn the user if it is.
+      #
+      # @return [Boolean] true if okay, false if not
+      def validate_system_mount_points
+        return true unless Yast::Mode.installation
+        return true if to_be_formatted?
+        return true if !mounted?
```
ancorgs: NP: This is one of those cases in which "unless" really helps to save CPU cycles in my brain. :-)

src/lib/y2partitioner/widgets/format_and_mount.rb
ancorgs: Oh, I saw now the trick below about keeping the old text. So never mind.

ancorgs:
Just one concern, what if "ReiserFS" or "Btrfs" are changed as part of the translation? Just imagine that in Korean, Arabic or whatever language those are turned into the corresponding abbreviation of the target alphabet.

joseivanlopez:
It would be easy to add some unit tests for this new validation, see https://github.com/yast/yast-storage-ng/blob/master/test/y2partitioner/widgets/format_and_mount_test.rb#L55

shundhammer:
Well, okay, then we can't recycle the message.

The I might as well simply write the message that I'd really want to post here.

shundhammer:
A trailing slash in the mount point will cause many more problems; I just checked the code. The MountPoint widget should sanitize the content when it is stored (and maybe also show an error); it's not only trailing slashes, also leading or trailing whitespace that comes to mind.

PR #573, opened 2018-03-15
Partitioner works with implicit partition tables

joseivanlopez:
PBI: https://trello.com/c/1AGO9Dkf/312-1-partitioner-work-with-implicit-partition-tables-s390

### DIFF
--- a/src/lib/y2partitioner/actions/delete_device.rb
+++ b/src/lib/y2partitioner/actions/delete_device.rb
@@ -21,13 +21,11 @@
 
 require "yast"
 require "yast/i18n"
+require "yast2/popup"
 require "y2partitioner/device_graphs"
 require "y2partitioner/confirm_recursive_delete"
 require "y2storage/filesystems/btrfs"
 require "abstract_method"
-
-Yast.import "Popup"
-Yast.import "HTML"
 
 module Y2Partitioner
   module Actions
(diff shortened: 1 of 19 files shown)


src/lib/y2partitioner/actions/delete_partition.rb
```diff
@@ -53,12 +44,42 @@ def delete
...
+
+      # Error when the device contains an implicit partition table
+      #
+      # @return [String, nil] nil if the device does not contain an implicit
+      #   partition table.
+      def implicit_partition_table_error
+        return nil if !device.implicit_partition_table?
+
+        _("This partition cannot be deleted. Implicit partition table\n" \
+          "must have one partition.")
```
ancorgs: Same than before, the "implicit partition table" is probably a non-sense outside storage-ng. So I would say something like "This is an implicit partition and cannot be deleted.".

src/lib/y2storage/partition.rb
```diff
@@ -244,6 +244,15 @@ def aligned_min_size(align_type = nil)
       min
     end
 
+    # Whether the partition belongs to an implicit partition table
+    #
+    # @see Y2Storage::PartitionTables::ImplicitPt
+    #
+    # @return [Boolean]
+    def implicit_partition_table?
```
ancorgs: Why don't call this method simply `#implicit?`. Again, what is implicit is the partition, we are adding the partition table to the mix just because we need it (I don't think it's going to disappear in the short term, but still)... and `#implicit?` is shorter :-)

src/lib/y2partitioner/actions/delete_device.rb
```diff
@@ -82,17 +80,36 @@ def device_graph
...
+      # @return [Boolean]
       def confirm
-        Yast::Popup.YesNo(
-          # TRANSLATORS %s is the name of the device to be deleted (e.g., /dev/sda1)
-          format(_("Really delete %s?"), device.name)
-        )
+        # TRANSLATORS %s is the name of the device to be deleted (e.g., /dev/sda1)
+        message = format(_("Really delete %s?"), device.name)
+
+        result = Yast2::Popup.show(message, buttons: :yes_no)
```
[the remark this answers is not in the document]
joseivanlopez: @jreidinger encourages us to use it in new code, instead the old `Yast::Popup`. So, I took the opportunity to do it in this part of the code.

PR #574, opened 2018-03-15
Max prep warning

jreidinger:


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Mar 15 15:29:56 UTC 2018 - jreidinger@suse.com
+
+- Warn if user creates too big prep partition which firmware can
```
ancorgs: NP: shouldn't it capitalized as PReP?

jreidinger: yes, you are right

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -223,6 +223,14 @@ def efi_in_md_raid1?
         return raid.md_level.is?(:raid1)
       end
 
+      # Method to return all prep partitions - newly created and also reused.
+      # It is useful to do checks on top of that partitions
+      def prep_partitions
```
ancorgs: `BootRequirements::Analyzer` is designed in a way that most of their methods analyze the full set of devices (i.e. the planned devices + the devices on the devicegraph). Methods that only search in part of the whole set (this one search the devicegraph but ignores planned_devices) should have a name that makes that clear.

In other words, I would expect a method `Analyzer#prep_partitions` to return a superset of `Analyzer#planned_prep_partitions" and the PReP partitions in the devicegraph.

For the partitioner that's irrelevant because planned_partitions is empty in that scope. But if the proposal or AutoYaST initialize the analyzer object as `Analyzer.new(devicegraph, some_planned_devices)` they expect methods to check both unless the name of the method says otherwise.

jreidinger: so should I rename it, move it or add also planned ones? what do you suggest?

src/lib/y2storage/boot_requirements_strategies/prep.rb
```diff
@@ -55,6 +64,21 @@ def warnings
...
+      # PReP partition is needed, so return any warning related to it.
+      def prep_warnings
+        res = []
+
+        if missing_partition_for?(prep_volume)
+          res << SetupError.new(missing_volume: prep_volume)
+        elsif prep_too_big?
+          msg = _("PReP partition is too big. " \
+           "Some firmwares can refuse to load it and prevent booting.")
+          res << SetupError.new(message: msg)
```
ancorgs: Well, we should tell the users which is the max size unless we want them to start a trial and error process. Something like "PReP partition is too big. Some firmwares are unable to load PReP partitions bigger than %s. If that happens the system will not boot." Just an idea. Any sentence mentioning the limit should make it.

ancorgs: Moreover, I just saw down there that we check for any PReP partition being too big. Even those that are not located in the same disk than "/". In that case maybe we should include the name of the partition in the message or something. Maybe I want to install in sda and I create a proper "/" and PReP there. I would like to know if the warning is actually referring to sdb. In that case, I don't care.

test/y2storage/boot_requirements_errors_test.rb
```diff
@@ -355,6 +355,18 @@
             include_examples "no errors"
           end
 
+          context "there is too big PReP partition" do
```
ancorgs: Why is this tested only in the partitions-based case and not in LVM?

In fact, somebody could expect the check to behave differently in both scenarios. For partition-based installations in which several disks contain a PReP, we know which one is the relevant one (same disk than "/"), so we could fine-tune the check.

I'm not asking you to fine-tune the check, I'm just saying that it makes sense to include the test in both scenarios. In the first place to be better test-covered and in the second to specify that we don't do any extra check in the partition-based scenario.

jreidinger: well, to be honest I do not get what is reason for different context as it is irrelevant to PREP proposal. in all case you need PReP that is directly on primary partition. So other tests are basically bogus that test identical stuff. There is also RAID and encrypted scenario and none of them make sense for non-power nv scenario. I am not sure if we really want complete matrix of various scenarios even that are not related to code. Because it start growing in too much directions.

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
jreidinger: what about rename it to something like `graph_prep_partitions` ? and if you want both you do `graph_prep_partitions + planned_prep_partitions`

ancorgs: Fine for me

test/y2storage/boot_requirements_errors_test.rb
ancorgs: I can live without it, it's not a blocker for merging.

That being said "are not related to code" is exactly the mindset I usually try to fight when we talk about RSpec tests. I think about them from the perspective of the output they produce and the scenarios and behaviors that outputs describe. I don't think about the code when writing tests for a component. So this looked kind of confusing to me (it makes me wonder whether too big PReP is not checked with LVM):

```
with a partitions-based proposal
   there is a PReP partition
      does not contain an error
   there is too big PReP partition
      contains a warning for too big PReP partition
   PReP partition missing
      contains an error for missing PReP partition
with a LVM-based proposal
   there is a PReP partition
      does not contain an error
   PReP partition missing
      contains an error for missing PReP partition
```

In my opinion we think too much in terms of covering all the code and too little in terms of covering scenarios. But that's a philosophical topic for the workshop. As I said, I'm fine with the current tests, they are not a stopper.

jreidinger: @ancorgs my problem with this approach is that for non-power PReP we should write something like "prep is always needed" and not writing all scenarios just to repeat that PReP is needed. It really does not help with reading output of rspec or documenting behavior. For me saying that prep is needed document also behavior not just code.

src/lib/y2storage/boot_requirements_strategies/prep.rb
```diff
@@ -55,6 +64,34 @@ def warnings
...
+        msg = format(
+          n_("Following PReP partition is too big: %s. ",
+            "Following PReP partitions are too big: %s.",
+            big_partitions.size
+          ),
+          big_partitions.map(&:name).join(", ")
+        )
+        # TRANSLATORS: %s is human readable partition size like 8 MiB.
+        msg += format(_("Some firmwares can refuse to load PReP partition " \
+          "that is bigger %s and thus prevent booting."), MAX_PREP_SIZE)
```
ancorgs: "refuse to load PReP partition that is bigger 8 MiB" sounds strange. I would go for "refuse to load PReP partitions bigger **than** 8 MiB". Or being closer to your idea "refuse to load **any** PReP Partition that is bigger **than** 8 MiB"

src/lib/y2storage/boot_requirements_strategies/prep.rb
```diff
@@ -55,6 +62,36 @@ def warnings
...
+            n_(
+              "Following PReP partition is too big: %s. ",
+              "Following PReP partitions are too big: %s.",
+              big_partitions.size
+            ),
+            big_partitions.map(&:name).join(", ")
+          )
+        # TRANSLATORS: %s is human readable partition size like 8 MiB.
+        msg + format(_("Some firmwares can fail to load PReP partitions " \
+          "bigger %s and thus prevent booting."), MAX_PREP_SIZE)
```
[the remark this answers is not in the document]
jreidinger: damn, will fix it

src/lib/y2storage/boot_requirements_strategies/prep.rb
```diff
@@ -55,6 +62,36 @@ def warnings
...
+        res
+      end
+
+      def big_prep_warning(big_partitions)
+        # TRANSLATORS: %s is single or list of partitions that are too big.
+        msg =
+          format(
+            n_(
+              "Following PReP partition is too big: %s. ",
+              "Following PReP partitions are too big: %s.",
```
ancorgs: This can stand as it is because omitting the initial "the" in this kind of texts is becoming kind of usual in computer messages (that's why I didn't notice at first sight, or maybe I have been too exposed to you :wink:), but for it to be perfect English it should be "The following..."

PR #575, opened 2018-03-15
Check if partition resize is possible based on filesystem type

shundhammer:
https://trello.com/c/lD8N7IET/200-5-partitioner-show-specific-error-messages-when-resizing-partitions

This is the first line of defence: If the filesystem type doesn't support resizing at all, this is the most obvious reason; we don't need to check if other, more complex conditions are met that the user might try to fix first. 

Since a filesystem is the innermost thing that is affected by a resize, otherwise the user might get errors about
- the partition that cannot be resized for any other reason
- an encryption layer that cannot be resized

There might still be more reasons, also on the filesystem level, why resizing is currently (!) not possible. But if it can't be done for a specific filesystem type at all, we don't need to bother with anything else; this is checked here, and it is checked first before going into any more details. We don't even need any more details if it doesn't work in principle.

### DIFF
--- a/src/lib/y2partitioner/actions/resize_blk_device.rb
+++ b/src/lib/y2partitioner/actions/resize_blk_device.rb
@@ -85,7 +85,9 @@ module Y2Partitioner
       #
       # @return [Array<Strings>]
       def errors
-        [used_device_error, cannot_be_resized_error].compact
+        [used_device_error,
+         fstype_resize_support_error,
+         cannot_be_resized_error].compact
       end
 
       # Error when trying to resize an used device
@@ -102,6 +104,17 @@ module Y2Partitioner
             "resized. To resize %{name}, make sure it is not used."),
           name: device.name
         )
+      end
+
+      # Error when the filesystem type cannot be resized
+      #
+      # @return [String, nil] nil if the filesystem can be resized.
+      def fstype_resize_support_error
+        return nil unless device.formatted?
+        fs = device.blk_filesystem
+        return nil if fs.supports_grow? || fs.supports_shrink?
+
+        _("This filesystem type cannot be resized.")
       end
 
       # Error when the device cannot be resized
--- a/src/lib/y2partitioner/actions/resize_blk_device.rb
+++ b/src/lib/y2partitioner/actions/resize_blk_device.rb
@@ -111,8 +111,7 @@ module Y2Partitioner
       # @return [String, nil] nil if the filesystem can be resized.
       def fstype_resize_support_error
         return nil unless device.formatted?
-        fs = device.blk_filesystem
-        return nil if fs.supports_grow? || fs.supports_shrink?
+        return nil if device.blk_filesystem.supports_resize?
 
         _("This filesystem type cannot be resized.")
       end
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -58,6 +58,14 @@ class BlkFilesystem < Base
       storage_forward :uuid
       storage_forward :uuid=
 
+      # @!method supports_shrink?
+      #   @return [Boolean] whether the filesystem supports shrinking
+      storage_forward :supports_shrink?, to: :supports_shrink
+
+      # @!method supports_grow?
+      #   @return [Boolean] whether the filesystem supports shrinking
```
joseivanlopez: Fix doc: s/shrinking/grow

test/y2storage/filesystems/blk_filesystem_test.rb
```diff
@@ -30,26 +30,81 @@
...
+    end
+
+    context "for an NTFS filesystem" do
+      let(:dev_name) { ntfs_part }
+
+      it "returns true" do
+        expect(filesystem.supports_shrink?).to eq true
+      end
+    end
+  end
```
joseivanlopez: Well, we are not checking all filesystem types here, but anyway, we are only wrapping these methods, so strictly speaking, the logic about what can be shrinked/grown should be in the library part.

joseivanlopez:
Sorry, I have just realized that there is no test for the new check in the resize action. Please, add one.

src/lib/y2partitioner/actions/resize_blk_device.rb
```diff
@@ -104,6 +106,17 @@ def used_device_error
         )
       end
 
+      # Error when the filesystem type cannot be resized
+      #
+      # @return [String, nil] nil if the filesystem can be resized.
+      def fstype_resize_support_error
+        return nil unless device.formatted?
+        fs = device.blk_filesystem
+        return nil if fs.supports_grow? || fs.supports_shrink?
```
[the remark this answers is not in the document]
shundhammer: Yes, we can do that.

src/lib/y2storage/filesystems/blk_filesystem.rb
shundhammer: ah, right.

test/y2storage/filesystems/blk_filesystem_test.rb
shundhammer: I hope it's there as well. I added those unit tests mostly to make sure they are exported properly from C++, and the ? is added as desired in the wrapper layer. And there is at least one test that returns something different than "it works" (XFS cannot be shrunk) to make sure it returns _false_ and not _nil_.

shundhammer:
How do you propose to set up the scenario for testing in the expert partitioner? Mock everything to death as usual, thus spending endless time for a completely worthless test?
I tested it manually, of course.

joseivanlopez:
Well, we could ensure at least that an error message is presented when the filesystem do not support resizing.

shundhammer:
Setting up the scenario is the problem.

PR #576, opened 2018-03-16
Fix space distribution on LVM VGs

imobachgs:
See [bsc#1079369](https://bugzilla.suse.com/show_bug.cgi?id=1079369).

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Mar 16 10:06:25 UTC 2018 - igonzalezsosa@suse.com
+
+- AutoYaST: fixed space distribution on LVM volume groups when
+  using percentages (bsc#1079369).
+
+-------------------------------------------------------------------
 Thu Mar 15 16:29:56 UTC 2018 - jreidinger@suse.com
 
 - Warn if user creates too big PReP partition which firmware can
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/proposal/lvm_creator.rb
```diff
@@ -265,6 +266,23 @@ def name_taken?(name, root)
...
+
+      # Returns a list of planned logical volumes adjusting the size
+      #
+      # All logical volumes which sizes are specified as a percentage will get
+      # their minimal and maximal sizes adjusted.
+      #
+      # @param lvs [Array<Planned::LvmLv>] List of planned logical volumes
+      # @param vg  [Planned::LvmVg]        Volume group to place logical volumes on
+      # @return [Array<Planned::LvmLv] New list of planned logical volumes with adjusted sizes
+      def planned_lvs_in_vg(lvs, vg)
```
joseivanlopez: NP: I would clone all lvs or none. I mean, I would expect as result a list of completely new lvs or a list with all the original lvs but some of them modified. Anyway, it can be stay at it is.

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Mar 16 10:06:25 UTC 2018 - igonzalezsosa@suse.com
+
+- AutoYaST: fixed space distribution on LVM volume groups when
+  using percentages (bsc#1079369).
+
```
joseivanlopez: Why to not increase version? More merges expected during the day?

imobachgs: Because I have a second fix and I will bump the version later :)

src/lib/y2storage/proposal/lvm_creator.rb
imobachgs: Actually, it is a bug. Fixed.

PR #579, opened 2018-03-16
Methods to deal with /etc/fstab specs (part of bsc#1071454)

ancorgs:
See https://trello.com/c/Mz4nNca0/302-3-sles15-p1-1071454-upgrade-to-sles15-fails-if-var-is-separate and https://github.com/yast/yast-update/pull/96#issuecomment-373754382

### DIFF
--- a/test/y2storage/filesystems/nfs_test.rb
+++ b/test/y2storage/filesystems/nfs_test.rb
@@ -1,0 +1,59 @@
+#!/usr/bin/env rspec
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
+#
+# All Rights Reserved.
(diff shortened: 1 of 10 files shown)


ancorgs:
Coverage decreased because I didn't test the abstract `Filesystem::Base` class... because it's abstract.

test/y2storage/filesystems/nfs_test.rb
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env rspec
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
```
imobachgs: Please, update year.

PR #580, opened 2018-03-16
Fix errors with zero-size devices 

joseivanlopez:
PBI: https://trello.com/c/jPe9PnBi/308-3-sles15-p1-1083887-zero-size-disks-not-able-to-find-the-scsi-disks-while-trying-to-install-sles-15-on-vmax-luns

* Devices with zero-size are excluded to be used in Proposal and Partitioner. 
* Fixed tests to avoid creation of zero-size devices or partitions.
* Adapted yaml reader/writer to properly manage zero-size devices.
* Bonus: yaml reader/writer tests are re-written in a more rspec way.

### DIFF
--- a/src/lib/y2storage/yaml_writer.rb
+++ b/src/lib/y2storage/yaml_writer.rb
@@ -72,12 +72,19 @@ module Y2Storage
     # @return [Array<Hash>]
     def yaml_device_tree(devicegraph)
       yaml = []
-      devicegraph.disk_devices.each { |device| yaml << yaml_disk_device(device) }
+      devices(devicegraph).each { |device| yaml << yaml_disk_device(device) }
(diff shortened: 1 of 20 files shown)


src/lib/y2storage/yaml_writer.rb
```diff
@@ -72,13 +72,20 @@ def write(devicegraph, yaml_file)
...
+      devices(devicegraph).each { |device| yaml << yaml_disk_device(device) }
       devicegraph.lvm_vgs.each { |lvm_vg| yaml << yaml_lvm_vg(lvm_vg) }
       yaml
     end
 
   private
 
+    # Devices that will be converted to yaml format
+    #
+    # @return [Array<Y2Storage::Device>]
```
ancorgs: Maybe we need a more descriptive name here. Something like `root_devices` or `top_level_devices`...

PR #581, opened 2018-03-20
Remove incomplete vgs

joseivanlopez:
PBI: https://trello.com/c/7ighv7Fr/306-5-sles15-p1-1083672-storage-proposal-fails-if-volume-group-has-missing-physical-volumes


src/lib/y2partitioner/widgets/summary_text.rb
```diff
@@ -83,8 +83,11 @@ def refresh_value
       end
 
       # Updates the value of {#actions}
+      #
+      # @note Actions are obtained from the raw probed devicegraph
```
ancorgs: Not sure if this sentence is clear enough. It can look like you ask the list of actions to raw_probed instead of current_graph.  I would say something like "Actions are calculated with the raw probed devicegraph as starting point".

src/lib/y2storage/callbacks/sanitize.rb
```diff
@@ -0,0 +1,80 @@
...
+      #
+      # @param errors [Array<String>] Errors detected in probed devicegraph.
+      # @return [Boolean] true if the user decides to sanitize the devicegraph.
+      def sanitize?(errors)
+        log.info "probed devicegraph contains errors, asking the user whether to sanitize it"
+        log.info "Errors details: #{errors}"
+
+        header = _("The following errors were detected in the system:")
+        error_messages = errors.join("\n")
+        details = _("If you continue, those devices will be deleted.")
```
ancorgs: It sounds like the device are going to be deleted immediately, not as part of the commit phase. I would omit this sentence here and would include more information about the consequences in the concrete error reported by `DevicegraphSanitizer`.

And about that message... let me try to propose something (just a raw suggestion):

Message during installation:

```
The volume group %{name} is incomplete because some physical volumes are missing.
If you continue, the volume group will be deleted later as part of the installation process.
Moreover, incomplete volume groups are ignored by the partitioning proposal and are
not visible in the Expert Partitioner.
```

Message for stand-alone partitioner:

```
The volume group %{name} is incomplete because some physical volumes are missing.
Incomplete volume groups are not visible in the Partitioner and will be deleted at the
final step, when all the changes are performed in the system.
```

src/lib/y2storage/devicegraph.rb
```diff
@@ -128,6 +128,17 @@ def dup
     end
     alias_method :duplicate, :dup
 
+    # Copies the devicegraph into another one, but avoiding to copy into itself
+    #
+    # @retrun[Boolean] true if the devicegraph was copied; false otherwise.
```
ancorgs: typo: retrun

src/lib/y2storage/storage_manager.rb
```diff
@@ -366,6 +396,37 @@ def reset_staging_revision
       @staging_revision_after_probing = 0
     end
 
+    # Sanitizes the raw probed devicegraph
+    #
+    # @note The raw probed devicegraph can contain some errors (e.g., incomplete LVM VGs).
```
ancorgs: Why is all the description nested into the `@note` tag. I would not use `@note` here, this is just explaining this method.

src/lib/y2storage/storage_manager.rb
```diff
@@ -366,6 +396,37 @@ def reset_staging_revision
...
+    #   This method tries to fix all errors from raw probed.
+    #
+    #   With the default callbacks, the user is asked whether to sanitize the raw probed
+    #   devicegraph when it contains some errors.
+    #
+    #   The raw probed devicegraph remains untouched, and the new sanitized one is internally
+    #   saved and copied into the staging devicegraph.
+    #
+    # @raise [Y2Storage::Error] if the raw probed cannot be sanitized
```
ancorgs: Well, to be precise it raises the error if the callbacks (i.e. the user) decided to not sanitize. Is not that we cannot, is that we were told not to do it.

ancorgs: Is maybe worth mentioning that in that case `probed` and `staging` remain untouched. They only become useless for installation/partitioner purposes, but they are still there an normally accessible.

src/lib/y2partitioner/widgets/rescan_devices_button.rb
```diff
@@ -68,12 +69,17 @@ def rescan_message
...
       end
 
-      # Reprobes and updates devicegraphs for the partitioner
+      # Reprobes and updates devicegraphs for the partitioner.
       #
-      # @note A message is shown during the reprobing action
+      # @note A message is shown during the reprobing action.
+      #
+      # @raise [Y2Partitioner::ForcedAbortError] When the probed devicegraph contains errors
+      #   and the user decices to not sanitize
```
ancorgs: typo: decices.

Moreover, the exception is also raised if there is an error during probing itself and the user decides to abort (so not only sanitization is affected).

DimStar77:
This is my suspect to cause https://openqa.opensuse.org/tests/644836#step/upgrade_select/5

PR #582, opened 2018-03-21
Show specific error messages when resizing partitions

shundhammer:
https://trello.com/c/lD8N7IET/200-5-partitioner-show-specific-error-messages-when-resizing-partitions

Blog post: https://github.com/yast/yast.github.io/issues/147#issuecomment-375342963

This PR wraps the new resize blocker reasons from libstorage-ng to provide human-readable error messages for all the individual reasons that block a resize operation: 

There might be more than one problem at the same time. We want to show the user all of them, or at least all that make sense; it's no use to show him just one, and once he managed to fix that one, he might find out that there are more, so all his efforts (which might involve data loss due to deleting partitions and/for filesystems) were still in vain.

The first part of this uses introspection to look up constants starting with `RB_` and convert them to Ruby symbols for much better readability, and later to translated texts. The Ruby symbols might come handy some day to check for various common combinations of reasons that can be summarized in other ways that are better to understand, or where we might want to suppress some of the reasons that might tend to confuse the user.

resize-error-reasons-01

resize-error-reasons-02

resize-error-reasons-03

### DIFF
--- a/src/lib/y2storage/resize_info.rb
+++ b/src/lib/y2storage/resize_info.rb
@@ -20,6 +20,7 @@
 # find current contact information at www.suse.com.
 
 require "y2storage/storage_class_wrapper"
+require "yast/i18n"
 
 module Y2Storage
   # Information about the possibility of resizing a given device
@@ -32,7 +33,32 @@ module Y2Storage
   # This is a wrapper for Storage::ResizeInfo
   class ResizeInfo
     include StorageClassWrapper
+    include Yast::I18n
+    extend Yast::I18n
     wrap_class Storage::ResizeInfo
+
+    # rubocop:disable Metrics/LineLength
+    REASON_TEXTS =
+      {
+        RB_RESIZE_NOT_SUPPORTED_BY_DEVICE:       N_("Resizing is not supported by this device."),
+        RB_MIN_MAX_ERROR:                        N_("Combined limitations of partition and filesystem prevent resizing."),
+        RB_SHRINK_NOT_SUPPORTED_BY_FILESYSTEM:   N_("This filesystem does not support shrinking."),
+        RB_GROW_NOT_SUPPORTED_BY_FILESYSTEM:     N_("This filesystem does not support growing."),
+        RB_FILESYSTEM_INCONSISTENT:              N_("Filesystem consistency check failed."),
+        RB_MIN_SIZE_FOR_FILESYSTEM:              N_("This filesystem already has the minimum possible size."),
+        RB_MAX_SIZE_FOR_FILESYSTEM:              N_("This filesystem already has the maximum possible size."),
+        RB_FILESYSTEM_FULL:                      N_("The filesystem is full."),
+        RB_NO_SPACE_BEHIND_PARTITION:            N_("There is no space behind this partition."),
+        RB_MIN_SIZE_FOR_PARTITION:               N_("This partition already has the minimum possible size."),
+        RB_EXTENDED_PARTITION:                   N_("Extended partitions cannot be resized."),
+        RB_ON_IMPLICIT_PARTITION_TABLE:          N_("The partition on an implicit partition table cannot be resized."),
+        RB_SHRINK_NOT_SUPPORTED_FOR_LVM_LV_TYPE: N_("Shrinking of this type of LVM logical volumes is not supported."),
+        RB_RESIZE_NOT_SUPPORTED_FOR_LVM_LV_TYPE: N_("Resizing of this type of LVM logical volumes is not supported."),
+        RB_NO_SPACE_IN_LVM_VG:                   N_("No space left in the LVM volume group."),
+        RB_MIN_SIZE_FOR_LVM_LV:                  N_("The LVM logical volume already has the minimum possible size."),
+        RB_MAX_SIZE_FOR_LVM_LV_THIN:             N_("The LVM thin logical volume already has the masimum size.")
+      }.freeze
+    # rubocop:enable Metrics/LineLength
 
     # @!method resize_ok?
     #   @return [Boolean] whether is possible to resize the device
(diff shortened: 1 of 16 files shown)


src/lib/y2storage/resize_info.rb
```diff
@@ -32,8 +33,33 @@ module Y2Storage
...
+        RB_FILESYSTEM_FULL:                      N_("The filesystem is full."),
+        RB_NO_SPACE_BEHIND_PARTITION:            N_("There is no space behind this partition."),
+        RB_MIN_SIZE_FOR_PARTITION:               N_("This partition already has the minimum possible size."),
+        RB_EXTENDED_PARTITION:                   N_("Extended partitions cannot be resized."),
+        RB_ON_IMPLICIT_PARTITION_TABLE:          N_("The partition on an implicit partition table cannot be resized."),
+        RB_SHRINK_NOT_SUPPORTED_FOR_LVM_LV_TYPE: N_("Shrinking of this type of LVM logical volumes is not supported."),
+        RB_RESIZE_NOT_SUPPORTED_FOR_LVM_LV_TYPE: N_("Resizing of this type of LVM logical volumes is not supported."),
+        RB_NO_SPACE_IN_LVM_VG:                   N_("No space left in the LVM volume group."),
+        RB_MIN_SIZE_FOR_LVM_LV:                  N_("The LVM logical volume already has the minimum possible size."),
+        RB_MAX_SIZE_FOR_LVM_LV_THIN:             N_("The LVM thin logical volume already has the masimum size.")
```
jreidinger: type masimum

src/lib/y2storage/resize_info.rb
```diff
@@ -32,8 +33,33 @@ module Y2Storage
...
+    extend Yast::I18n
     wrap_class Storage::ResizeInfo
 
+    # rubocop:disable Metrics/LineLength
+    REASON_TEXTS =
+      {
+        RB_RESIZE_NOT_SUPPORTED_BY_DEVICE:       N_("Resizing is not supported by this device."),
+        RB_MIN_MAX_ERROR:                        N_("Combined limitations of partition and filesystem prevent resizing."),
+        RB_SHRINK_NOT_SUPPORTED_BY_FILESYSTEM:   N_("This filesystem does not support shrinking."),
+        RB_GROW_NOT_SUPPORTED_BY_FILESYSTEM:     N_("This filesystem does not support growing."),
```
jreidinger: I think it would be nice to write user which filesystem we think does not support shrinking like "The XFS filesystem does not support shrinking."

src/lib/y2storage/resize_info.rb
```diff
@@ -32,8 +33,33 @@ module Y2Storage
...
+        RB_FILESYSTEM_INCONSISTENT:              N_("Filesystem consistency check failed."),
+        RB_MIN_SIZE_FOR_FILESYSTEM:              N_("This filesystem already has the minimum possible size."),
+        RB_MAX_SIZE_FOR_FILESYSTEM:              N_("This filesystem already has the maximum possible size."),
+        RB_FILESYSTEM_FULL:                      N_("The filesystem is full."),
```
jreidinger: Maybe extend it to "The filesystem is full. The shring is not possible. If you need shrink, consider freeing some space there."

src/lib/y2storage/resize_info.rb
```diff
@@ -32,8 +33,33 @@ module Y2Storage
...
+        RB_NO_SPACE_BEHIND_PARTITION:            N_("There is no space behind this partition."),
```
jreidinger: In general I think we should help user to deal with it if we can so here write something like "There is no space behind this partition. If there is partition behind this one, then deleting should help."

src/lib/y2storage/resize_info.rb
```diff
@@ -32,8 +33,33 @@ module Y2Storage
...
+        RB_MIN_SIZE_FOR_PARTITION:               N_("This partition already has the minimum possible size."),
```
jreidinger: I have problem to recognize what is difference from filesystem full. So this happen when filesystem metadata needs this size?

src/lib/y2storage/resize_info.rb
```diff
@@ -32,8 +33,33 @@ module Y2Storage
...
+        RB_EXTENDED_PARTITION:                   N_("Extended partitions cannot be resized."),
+        RB_ON_IMPLICIT_PARTITION_TABLE:          N_("The partition on an implicit partition table cannot be resized."),
+        RB_SHRINK_NOT_SUPPORTED_FOR_LVM_LV_TYPE: N_("Shrinking of this type of LVM logical volumes is not supported."),
+        RB_RESIZE_NOT_SUPPORTED_FOR_LVM_LV_TYPE: N_("Resizing of this type of LVM logical volumes is not supported."),
```
jreidinger: here maybe write given type. So user can easily google that it is really not possible.

src/lib/y2storage/resize_info.rb
```diff
@@ -32,8 +33,33 @@ module Y2Storage
...
+        RB_NO_SPACE_IN_LVM_VG:                   N_("No space left in the LVM volume group."),
```
jreidinger: I expect this is for increase of size. So maybe hint them that they need to shrink another logical volume before increasing size of this one?

src/lib/y2storage/resize_info.rb
shundhammer: There is always only one involved: The one you selected when you clicked the button.

src/lib/y2storage/resize_info.rb
shundhammer: At that point, it isn't even clear that this was about shrinking. All we get from the library are the flags which are converted to those Ruby symbols (looking very much like constants) here. It may be the common thing, but it's not necessarily only shrinking that is prevented here; from all we know at this point, it might be any operation.

src/lib/y2storage/resize_info.rb
shundhammer: This is much too long at this point, and we would be lost in too many details very quickly. In this case, there might be another partition behind the one the user wanted to resize, or it might also be the end of the disk. This is not the place to do this kind of analysis; we just don't know any more than what the library gives us.

And besides, we are already very much overengineering this.

src/lib/y2storage/resize_info.rb
shundhammer: This if for the partition; the other one was for the filesystem. That's two different things. Plus, there might also be an encryption layer in between (or, in very pathological cases, several nested ones).

src/lib/y2storage/resize_info.rb
shundhammer: Since those errors are only reported in the expert partitioner and the user just clicked on that object, he really should know what kind this is. Again, at this point we don't have any more information; this is what the library gives us.

src/lib/y2storage/resize_info.rb
shundhammer: Or add another physical volume. Or delete a logical volume. We give reasons why the operation the user requested doesn't work; he will have to figure out what to do about it.

This is an error message, not a troubleshooting guide or a generic manual about storage. ;-)

src/lib/y2storage/resize_info.rb
shundhammer: Ah, right, thx!

shundhammer: Fixed.

src/lib/y2storage/resize_info.rb
shundhammer: Also consider that there might be a number of those messages that are reported at the same time, and we might run into screen space limitations (think about NCurses). Once we start giving more advanced tips, we'll need to make this scrollable which means RichText which means no good default for the message popup's size, so scrolling (in the worst case in both dimensions) would be the common thing.

And since the user needs to close the dialog to start fixing the problem, any more detailed instructions won't be visible on the screen for long so the user can refer to them later; this is the situation where less is really more (more useful).

src/lib/y2storage/resize_info.rb
jreidinger: Yes, I am just trying to comment it from user side ( user who understand a bit storage, but is not expert).

src/lib/y2storage/resize_info.rb
jreidinger: I am not saying to give always bullet proof manual, but at least some hint would be nice. It is always better to get "We cannot do this because ABC. Doing DEF can help." than "We cannot do this because ABC". But it is not blocker for me, it is mainly Nice to Have.

PR #583, opened 2018-03-22
Avoid data loss in the partitioner

ancorgs:
See https://trello.com/c/DIOx0i1I/176-5-partitioner-prevent-data-loss-by-honoring-previous-state-when-editing

### DIFF
--- a/src/lib/y2partitioner/blk_device_restorer.rb
+++ b/src/lib/y2partitioner/blk_device_restorer.rb
@@ -1,0 +1,209 @@
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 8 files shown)


src/lib/y2partitioner/blk_device_restorer.rb
```diff
@@ -0,0 +1,209 @@
...
+
+    # Equivalent to {#device} in the system devicegraph, if any
+    #
+    # @return [Y2Storage::BlkDevice, nil]
+    def system_device
+      DeviceGraphs.instance.system.find_device(device.sid)
+    end
+
+    # Equivalent to {#device} in its checkpoint devicegraph, if any
+    #
```
joseivanlopez: NP: I would explain that the device is searched into system when there is no checkpoint, and it returns nil when the device does not exist (nor in system).

joseivanlopez:
Another thing, I don't see a new entry in changelog and version bumped.

PR #586, opened 2018-03-27
Add snapshots size check 

joseivanlopez:
PBI: https://trello.com/c/pvMqokfm/321-not-planned-sles15-p1-1085131-build-4864storage-ng-there-is-no-warning-when-btrfs-root-partition-is-too-small-for-snapshot

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Apr  2 11:53:31 UTC 2018 - jlopez@suse.com
+
+- Partitioner: add check for min size when using snapshots
+  (bsc#1085131).
+- 4.0.144
+
+-------------------------------------------------------------------
 Mon Apr  2 11:04:31 UTC 2018 - ancor@suse.com
 
 - If a duplicate PV is found, show an specific error message with
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Mon Apr  2 11:53:31 UTC 2018 - jlopez@suse.com
 
-- Partitioner: add check for min size when using snapshots
+- Partitioner: add check for minimum size when using snapshots
   (bsc#1085131).
 - 4.0.144
 
--- a/src/lib/y2partitioner/dialogs/blk_device_resize.rb
+++ b/src/lib/y2partitioner/dialogs/blk_device_resize.rb
@@ -1,13 +1,34 @@
-require "y2storage"
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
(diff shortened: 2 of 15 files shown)


src/lib/y2partitioner/filesystem_errors.rb
```diff
@@ -0,0 +1,133 @@
...
+    # @return [String, nil] nil if size is ok or no check is performed.
+    def small_size_for_snapshots_error(filesystem, new_size: nil)
+      textdomain "storage"
+
+      return nil unless installing? && small_size_for_snapshots?(filesystem, new_size: new_size)
+
+      format(
+        _("Your %{name} device is very small for snapshots.\n" \
+          "We recommend to increase the size of the %{name} device\n" \
+          "to %{min_size} or more or to disable snapshots."),
```
imobachgs: np: what about something like *to at least %{min_size} or to disable snapshots* (to avoid "*or* more *or* to disable...").

test/y2partitioner/filesystem_errors_test.rb
```diff
@@ -0,0 +1,194 @@
...
+    let(:root_spec_min_size) { 10.GiB }
+
+    let(:partition_size) { 20.GiB }
+
+    let(:new_size) { nil }
+
+    let(:partition) { fake_devicegraph.find_by_name("/dev/sda1") }
+
+    shared_examples "no snapshots error" do
+      it "does not contain 'small device for snapshots' error" do
```
imobachgs: np: 'small for snapshots'. Same below.

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Mar 28 17:39:56 UTC 2018 - jlopez@suse.com
+
+- Partitioner: add check for min size when using snapshots
```
imobachgs: I would replace `min` with the full word.

src/lib/y2partitioner/dialogs/blk_device_resize.rb
```diff
@@ -188,28 +211,35 @@ def help
...
+        # min and max possible sizes.
         #
         # @note An error popup is shown when the given size is not valid.
+        #   A warning popup is shown if there are some warnings.
         #
-        # @return [Boolean] true if the given size is valid; false otherwise.
+        # @see #errors
+        # @see #validation_warnings
+        #
+        # @return [Boolean] true if the given size is valid and the user
```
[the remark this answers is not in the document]
joseivanlopez: Well, it is tricky. In case of no errors, the size is valid, but there could be warnings. In that case, the user is asked. So, I would say that the method returns true ""if there are no errors and the user accepts the warnings (if someone)". I will improve it.

src/lib/y2partitioner/dialogs/blk_device_resize.rb
```diff
@@ -1,13 +1,34 @@
-require "y2storage"
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
```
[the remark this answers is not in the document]
joseivanlopez: Just adding missing incenses, so I think it should be 2017.

src/lib/y2partitioner/dialogs/format_and_mount.rb
```diff
@@ -1,5 +1,28 @@
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
```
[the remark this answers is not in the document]
joseivanlopez: Same here, just adding missing license.

src/lib/y2partitioner/filesystem_errors.rb
joseivanlopez: Agree. I was simply reusing old texts, but I think your propose is better.

joseivanlopez:
I agree. Probably some kind of mixin for that `#validate` would avoid the duplication, but for now they are only three lines, so I guess it doesn't worth.

PR #587, opened 2018-03-28
Specific error for duplicate PVs

ancorgs:
Last step of https://trello.com/c/T5rxuHfg/300-2-sles15-p1-1082542-build4731storage-ng-delete-lvmlv-failed-when-reinstall-on-encrypted-lvm

report_yesno

### DIFF
--- a/src/lib/y2storage/callbacks/libstorage_callback.rb
+++ b/src/lib/y2storage/callbacks/libstorage_callback.rb
@@ -64,12 +64,18 @@ module Y2Storage
         log.info "libstorage-ng reported an error, asking the user whether to continue"
         log.info "Error details. Message: #{message}. What: #{what}."
 
+        desc = error_description(what)
+        hint = _("Click below to see more details (English only).")
         question = _("Continue despite the error?")
-        focus = default_answer_to_error ? :focus_yes : :focus_no
-        result = Yast::Report.ErrorAnyQuestion(
-          Yast::Popup.NoHeadline, "#{message}\n\n#{what}\n\n#{question}",
-          Yast::Label.ContinueButton, abort_button_label, focus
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/callbacks/libstorage_callback.rb
```diff
@@ -90,6 +96,49 @@ def abort_button_label
...
+          "The same LVM physical volume was found in several devices.\n" \
+          "Maybe there are multipath devices in the system but multipath support\n" \
+          "was not enabled.\n\n"
+        )
+        if Yast::Mode.auto
+          result << _(
+            "Use 'start_multipath' in the AutoYaST profile to enable multipath."
+          )
+        else
+          result << _(
```
joseivanlopez: NP: I think we can detect if the user already has used `LIBSTORAGE_MULTIPATH_AUTOSTART` and, in that case, change the message a little. Otherwise, the user could think that he is not doing it correctly.

src/lib/y2storage/callbacks/libstorage_callback.rb
```diff
@@ -67,7 +67,7 @@ def error(message, what)
         desc = error_description(what)
         hint = _("Click below to see more details (English only).")
         question = _("Continue despite the error?")
-        msg = "#{message}\n\n#{desc}\n\n#{hint}\n\n#{question}"
+        msg = "#{message}\n\n#{desc}\n#{hint}\n\n#{question}"
```
joseivanlopez: Why this change? I would keep it as it was, and simply add `desc = error_description(what).chop` in case we want to ignore the possible `\n`. Otherwise, we have to remember to add a `\n` in all possible results of `#error_description`.

PR #591, opened 2018-04-03
Recover method #exists_in_probed

joseivanlopez:
See https://bugzilla.suse.com/show_bug.cgi?id=1087818.


src/lib/y2storage/device.rb
```diff
@@ -193,6 +197,19 @@ def hash
...
 
+    # Check whether the device exists in the probed devicegraph
+    #
+    # @note The probed devicegraph is the sanitized version of the
+    #   initial raw probed.
+    #
+    # @see StorageManager#raw_probed
+    # @see exists_in_devicegraph?
+    #
+    # @return [Boolean]
```
jreidinger: to bring some light into bootloader usage of this method. It is used mainly to detect which partition is "new", so not exists in probed graph. It is used to determine which partition create specifically for booting purpose like new PREP partition and in such case bootloader prefer this newly created partition. So for bootloader, I think it does not matter much if it is sanitized version or not.

joseivanlopez: Ok, so in yast-bootloader we can keep the code as it is now, but do you prefer to use raw version there?

jreidinger: well, I do not care. As long as method allow me to identify if partition is newly created ( so not exists in probed ).

PR #592, opened 2018-04-05
Handle lock system

joseivanlopez:
PBI: https://trello.com/c/jItiacnG/335-readd-system-wide-lock-in-libstorage-ng

Notes for reviewers:
* [These changes](https://github.com/yast/yast-ruby-bindings/pull/215) in `yast-ruby-bindings`are required for this PR.
* Clients from other packages using storage-ng do not need to be adapted. With the default callbacks, an `AbortException` is raised if the user aborts and the process will finish.

* codeclimate is complaining for a couple of methods (high complexity), but it seems difficult to improve them without a less readable way.
* Class StorageManager starts to be too large, and codeclimate also detects it. Another class to lighten up in future.

### DIFF
--- a/src/clients/partitioner_testing.rb
+++ b/src/clients/partitioner_testing.rb
@@ -5,6 +5,11 @@ require "yast"
 require "y2partitioner/clients/main"
 require "y2storage"
 
+# Uncomment next line and run the file with root privileges to test system lock
+# exit unless Y2Storage::StorageManager.setup(mode: :rw)
+
+Y2Storage::StorageManager.create_test_instance
+
 arg = Yast::WFM.Args.first
 case arg
 when /.ya?ml$/
--- a/src/clients/partitioner_testing.rb
+++ b/src/clients/partitioner_testing.rb
@@ -5,7 +5,7 @@ require "yast"
 require "y2partitioner/clients/main"
 require "y2storage"
 
-# Uncomment next line and run the file with root privileges to test system lock
+# Comment next line and run the file with root privileges to test system lock
 Y2Storage::StorageManager.create_test_instance
 
 arg = Yast::WFM.Args.first
--- a/src/clients/partitioner_testing.rb
+++ b/src/clients/partitioner_testing.rb
@@ -6,17 +6,15 @@ require "y2partitioner/clients/main"
 require "y2storage"
 
 # Uncomment next line and run the file with root privileges to test system lock
-# exit unless Y2Storage::StorageManager.setup(mode: :rw)
-
 Y2Storage::StorageManager.create_test_instance
 
 arg = Yast::WFM.Args.first
 case arg
 when /.ya?ml$/
-  Y2Storage::StorageManager.instance.probe_from_yaml(arg)
+  Y2Storage::StorageManager.instance(mode: :rw).probe_from_yaml(arg)
 when /.xml$/
   # note: support only xml device graph, not xml output of probing commands
-  Y2Storage::StorageManager.instance.probe_from_xml(arg)
+  Y2Storage::StorageManager.instance(mode: :rw).probe_from_xml(arg)
 else
   raise "Invalid testing parameter #{arg}, expecting foo.yml or foo.xml."
 end
--- a/src/lib/y2storage/callbacks/initialize.rb
+++ b/src/lib/y2storage/callbacks/initialize.rb
@@ -21,8 +21,8 @@
 
 require "yast"
 require "yast/i18n"
-require "yast2/popup"
 
+Yast.import "Report"
 Yast.import "Mode"
 Yast.import "Label"
 
@@ -73,11 +73,11 @@ module Y2Storage
 
         buttons = { yes: _("Retry"), no: abort_button_label }
 
-        answer = Yast2::Popup.show(message, headline: headline, buttons: buttons)
+        answer = Yast::Report.yesno_popup(message, headline: headline, buttons: buttons)
 
         log.info "User answer: #{answer}"
 
-        answer == :yes
+        answer
       end
 
     private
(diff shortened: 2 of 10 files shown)


src/lib/y2storage/callbacks/initialize.rb
```diff
@@ -0,0 +1,111 @@
...
+      #
+      # @return [Boolean] true if the user decides to retry.
+      def retry?
+        log.info "Storage subsystem is locked by process #{locker_pid}, asking to user whether to retry"
+
+        if locker_name
+          message = format(
+            # TRANSLATORS: %{name} is replaced by the name of a process (e.g., yast2)
+            # and %{pid} by the pid of a process (e.g., 5032).
+            _("The storage subsystem is locked by the application \"%{name}\" (%{pid})."),
```
ancorgs: It's not clear from the message to the user that the number in parenthesis corresponds to the pid of the process.

src/lib/y2storage/callbacks/initialize.rb
```diff
@@ -0,0 +1,111 @@
...
+            name: locker_name,
+            pid:  locker_pid
+          )
+        else
+          message = _("The storage subsystem is locked by an unknown application.")
```
ancorgs: If we don't know the name we don't know the pid either? Just asking because if the pid is always known, it should be always displayed.

src/lib/y2storage/callbacks/initialize.rb
```diff
@@ -0,0 +1,111 @@
...
+        end
+
+        message << "\n"
+        message << _("You must quit that application before you can continue.")
```
ancorgs: Just a matter of personal taste, but I prefer to have long texts like this in a single `_()` call including the line breaks, so

- the translators have the full picture while translating
- the translators can change the exact location of `\n` or even replace them by other character that makes more sense in their language

src/lib/y2storage/storage_manager.rb
```diff
@@ -460,20 +467,64 @@ def configure_snapper?
...
+      #   created instance (i.e., current instance is ro but rw is requested).
+      #
+      # @param mode [Symbol, nil] :ro, :rw. If nil, a default mode is used, see
+      #   {.default_storage_mode}.
+      # @param callbacks [Callbacks::Initialize, nil]
+      #
+      # @return [Boolean] true if the storage instance was correctly created for
+      #   the given mode; false otherwise.
+      def setup(mode: nil, callbacks: nil)
+        mode ||= default_storage_mode
```
ancorgs: Why is this done here if it's also done at the beginning of `.instance`? I find it confusing.

src/lib/y2storage/storage_manager.rb
```diff
@@ -460,20 +467,64 @@ def configure_snapper?
...
-        create_instance unless @instance
-        @instance
+      def instance(mode: nil, callbacks: nil)
+        return @instance if @instance && mode.nil?
+
+        mode ||= default_storage_mode
+
+        if @instance
+          return @instance if valid_instance?(mode)
+          raise Error, "Unexpected storage mode: current is #{@instance.mode}, requested is #{mode}"
```
ancorgs: Maybe we should have a more specific exception here. Not a big deal, but...

src/lib/y2storage/callbacks/initialize.rb
```diff
@@ -0,0 +1,111 @@
...
+        message << "\n\n"
+        message << _("Would you like to abort or try again?")
+
+        headline = _("Accessing the storage subsystem failed")
+
+        buttons = { yes: _("Retry"), no: abort_button_label }
+
+        answer = Yast2::Popup.show(message, headline: headline, buttons: buttons)
```
ancorgs: Shouldn't all callbacks use `Report` instead of directly `Popup`. After all, they can be raised during AutoYaST.

src/lib/y2storage/storage_manager.rb
```diff
@@ -516,6 +582,54 @@ def create_logger
...
         ::Storage.logger = @logger
       end
+
+      # Default access mode (read-only or read-write)
+      #
+      # @note During installation, access mode should be read-write.
+      #
+      # @return [Symbol] :ro, :rw
+      def default_storage_mode
+        Yast::Mode.installation || Yast::Mode.autoinst ? :rw : :ro
```
ancorgs: This is redundant. If `Mode.autoinst` is true then `Mode.installation` is also true. So you don't need the second part of the condition. Check the table at the Mode documentation (unfortunately not rendered properly in rubydocs).

src/lib/y2storage/callbacks/initialize.rb
joseivanlopez: It copied from old storage.

src/lib/y2storage/callbacks/initialize.rb
joseivanlopez: The pid is 0 (or negative number) in that case. Again, this mimics the old behavior.

src/clients/partitioner_testing.rb
```diff
@@ -5,13 +5,16 @@
 require "y2partitioner/clients/main"
 require "y2storage"
 
+# Uncomment next line and run the file with root privileges to test system lock
+Y2Storage::StorageManager.create_test_instance
```
ancorgs: Uncomment? Is not commented!

test/y2partitioner/clients/main_test.rb
```diff
@@ -14,7 +14,22 @@
...
+    context "when storage system cannot be initialized as read-write" do
+      before do
+        allow(Y2Storage::StorageManager).to receive(:setup).with(mode: :rw).and_return(false)
+      end
+
+      it "returns nil" do
+        expect(subject.run).to be_nil
+      end
+
+      it "does not run the client" do
```
ancorgs: Does not run the **dialog**

PR #594, opened 2018-04-09
Force UTF-8 encoding for (most) strings coming from libstorage-ng (bsc#1088067)

ancorgs:
See https://trello.com/c/UCbdJhFn/346-urgent-sles15-p1-1088067-installation-error-if-selecting-non-english-language

Review with care. Not much tested.

### DIFF
--- a/src/lib/y2storage/storage_class_wrapper.rb
+++ b/src/lib/y2storage/storage_class_wrapper.rb
@@ -306,6 +306,8 @@ module Y2Storage
       end
 
       def object_for(wrapper_class, storage_object)
+        # As workaround for bsc1088067, force encoding like the old
+        # yast-storage did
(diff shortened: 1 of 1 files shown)


src/lib/y2storage/storage_class_wrapper.rb
```diff
@@ -308,7 +306,11 @@ def processed_storage_result(result, class_name)
       end
 
       def object_for(wrapper_class, storage_object)
-        if wrapper_class.respond_to?(:downcasted_new)
+        if storage_object.is_a?(String)
```
jreidinger: please add comment why.

ancorgs: Good point

PR #597, opened 2018-04-09
Make disable_order affect all configurable aspects of the volume

ancorgs:
The behavior was intentionally implemented in a different way to what was documented at https://github.com/yast/yast-storage-ng/blob/master/doc/old_and_new_proposal.md

Now code and documentation match.

### DIFF
--- a/src/lib/y2storage/proposal/initial_strategies/ng.rb
+++ b/src/lib/y2storage/proposal/initial_strategies/ng.rb
@@ -107,25 +107,26 @@ module Y2Storage
 
         # Returns the first volume (according to disable_order) whose settings could be modified.
         #
-        # @note Volumes without disable order are left to the end.
+        # @note Volumes without disable order are never altered.
         #
         # @param settings [ProposalSettings]
         # @return [VolumeSpecification, nil]
         def first_configurable_volume(settings)
-          volumes = settings.volumes.select { |v| configurable_volume?(v) }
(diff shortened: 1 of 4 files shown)


ancorgs:
The only source file modified is still 100% covered with tests. So the coverage drop (0.0004%) is not significant. Simply there is less code.

src/lib/y2storage/proposal/initial_strategies/ng.rb
```diff
@@ -107,39 +107,37 @@ def try_without_proposed(proposal, current_settings, volume, devicegraph, disk_a
...
-
-          (with_disable_order.sort_by(&:disable_order) + without_disable_order).first
+          settings.volumes.select { |v| configurable_volume?(v) }.sort_by(&:disable_order).first
         end
 
         # A volume is configurable if it is proposed and its settings can be modified.
         # That is, #adjust_by_ram, #snapshots or #proposed are true and some of them
         # could be change to false.
         #
+        # @note A volume is considered configurable if it has disable order. Otherwise,
```
joseivanlopez: NP: Maybe we can have only one explanation about what it is considered a configurable volume. Right now, the rationale is splitted into the first paragraph and the `@note` entry.

ancorgs: Yeah, I just moved the former note from one method to another. On the other hand, the current structure makes more clear that the user can still alter volumes without `disable_order` using the UI. Integrating that in the sentence above maybe makes it harder to read. So I think I will leave it as it is.

PR #599, opened 2018-04-10
Partitioner: handle legacy entries which use nfs4

ancorgs:
Communicate properly with the embedded yast2-nfs-client so legacy entries using `nfs4` in the vfstype field can be detected and corrected.

This depends on https://github.com/openSUSE/libstorage-ng/pull/510 (I will add the dependency as soon as that gets merged).

nfs4-in-partitioner

### DIFF
--- a/src/lib/y2storage/filesystems/legacy_nfs.rb
+++ b/src/lib/y2storage/filesystems/legacy_nfs.rb
@@ -21,6 +21,7 @@
 
 require "yast"
 require "y2storage/filesystems/nfs"
+require "y2storage/filesystems/type"
 
 module Y2Storage
   module Filesystems
@@ -67,6 +68,10 @@ module Y2Storage
       # @return [String]
       attr_reader :fstopt
 
+      # Filesystem type used in fstab
+      # @return [Type] possible values are Type::NFS and Type::NFS4
+      attr_reader :fs_type
+
       # Creates a new object from a hash with the legacy fields used in
       # TargetMap-based code.
       #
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/filesystems/legacy_nfs.rb
```diff
@@ -95,9 +100,8 @@ def to_hash
           "device"  => share_string(server, path),
           "mount"   => mountpoint,
           "fstopt"  => fstopt,
-          # TODO: libstorage-ng does not distinguish different NFS versions, so
-          # for the time being, always use :nfs here
-          "used_fs" => :nfs
+          "used_fs" => fs_type.to_sym,
```
shundhammer: Please add a comment what the difference between the two is (which is which).

ancorgs: Great question :-)

test/y2partitioner/widgets/pages/nfs_mounts_test.rb
```diff
@@ -28,6 +28,8 @@
 describe Y2Partitioner::Widgets::Pages::NfsMounts do
   before do
     devicegraph_stub("nfs1.xml")
+    # Workaround. Will not be needed as soon as the xml format includes MountPoint#mount_type
```
ancorgs: I will work on it right now (proper support for that attribute when read from XML). But I didn't want to wait another 12+ hours to have a testable docker image and merge this.

PR #602, opened 2018-04-12
AutoYaST partition_type support

imobachgs:
Fixes [bsc#1081506](https://bugzilla.suse.com/show_bug.cgi?id=1081506).


src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -134,8 +134,10 @@ def best_distribution(planned_partitions, disk_names)
       # @return [PartitionCreatorResult]
       def create_partitions(new_partitions, disk_names)
         log.info "Partitions to create: #{new_partitions}"
+        primary, non_primary = new_partitions.partition(&:primary)
```
jreidinger: method name `partition` is a bit confusing in this context. But not your fault.

PR #603, opened 2018-04-12
Support missing settings

joseivanlopez:
PBI: https://trello.com/c/nujLPA7G/342-2-sles15-p1-1087486-storage-ng-various-storage-ng-related-controlxml-parameters-not-working

### DIFF
--- a/src/lib/y2partitioner/dialogs/main.rb
+++ b/src/lib/y2partitioner/dialogs/main.rb
@@ -21,10 +21,12 @@
 
 require "yast"
 require "cwm/dialog"
+require "yast2/popup"
 require "y2partitioner/device_graphs"
 require "y2partitioner/ui_state"
 require "y2partitioner/widgets/overview"
 require "y2partitioner/exceptions"
+require "y2storage/partitioning_features"
 
 Yast.import "Label"
 Yast.import "Mode"
@@ -33,9 +35,11 @@ Yast.import "Hostname"
 
 module Y2Partitioner
   module Dialogs
-    # main entry point to partitioner showing tree pager with all content
+    # Main entry point to Partitioner showing tree pager with all content
     class Main < CWM::Dialog
-      # @return [Y2Storage::DeviceGraph] device graph with all changes done in dialog
+      include Y2Storage::PartitioningFeatures
+
+      # @return [Y2Storage::Devicegraph] device graph with all changes done in dialog
       attr_reader :device_graph
 
       def initialize
@@ -67,28 +71,28 @@ module Y2Partitioner
         Yast::Mode.installation ? Yast::Label.AcceptButton : Yast::Label.FinishButton
       end
 
-      # runs dialog.
-      # @param system [Y2Storage::DeviceGraph] system graph used to detect if something
-      # is going to be formatted
-      # @param initial [Y2Storage::DeviceGraph] device graph to display
-      # @return [Symbol] result of dialog
+      # Runs the dialog
+      #
+      # @param system [Y2Storage::Devicegraph] system graph used to detect if something
+      #   is going to be formatted.
+      # @param initial [Y2Storage::Devicegraph] device graph to display.
+      #
+      # @return [Symbol] result of the dialog.
       def run(system, initial)
-        DeviceGraphs.create_instance(system, initial)
         UIState.create_instance
-        res = nil
+        DeviceGraphs.create_instance(system, initial)
+
+        return :back unless run_partitioner?
+
+        result = nil
+
         loop do
-          res = super()
-
-          next if res == :redraw
-          if res == :abort && Yast::Mode.installation
-            next unless Yast::Popup.ConfirmAbort(:painless)
-          end
-
-          break
+          result = super()
+          break unless continue_running?(result)
         end
-        @device_graph = DeviceGraphs.instance.current
 
-        res
+        @device_graph = DeviceGraphs.instance.current
+        result
       rescue Y2Partitioner::ForcedAbortError
         :abort
       end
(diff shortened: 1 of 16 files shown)


src/lib/y2partitioner/dialogs/main.rb
```diff
@@ -104,6 +106,63 @@ def run(system, initial)
...
+      end
+
+      # Popup to alert the user about the usage of the Partitioner
+      #
+      # @return [Symbol] user's answer (:yes, :no)
+      def partitioner_warning
+        # Warning popup about using the expert partitioner
+        message = _(
+          "This is for experts only.\n" \
+          "You might lose support if you use this!\n\n" \
```
wfeldt: "You might lose support if you use this!" is at one time a bold claim (loosing support for a product) and also very unspecific.

How about something along the lines of "Not all partition setups will be supported."?

wfeldt: Also, the openSUSE branch does not have support at all.

joseivanlopez: Yes, cbf641f is bsc#1088960. I will indicate it in the changelog (I didn't know we had already a bug for it).

joseivanlopez:
Yes, cbf641f is bsc#1088960. I will indicate it in the changelog (I didn't know we had already a bug for it).

src/lib/y2partitioner/dialogs/main.rb
```diff
@@ -67,28 +69,28 @@ def next_button
...
+      # @return [Symbol] result of the dialog.
       def run(system, initial)
-        DeviceGraphs.create_instance(system, initial)
         UIState.create_instance
-        res = nil
-        loop do
-          res = super()
+        DeviceGraphs.create_instance(system, initial)
+
+        return :back unless accepted_partitioner_warning?
```
ancorgs: The name is rather confusing. In most cases the warning will not be displayed (either no need to accept it). Calling something called `accepted_partitioner_warning?` and getting `true` in those cases... is not nice for the reader.

src/lib/y2partitioner/dialogs/main.rb
joseivanlopez: Fine with me, but the wording is copied from old storage. Do you think we should change it now?

src/lib/y2partitioner/dialogs/main.rb
```diff
@@ -104,6 +106,63 @@ def run(system, initial)
...
+        message = _(
+          "This is for experts only.\n" \
+          "You might lose support if you use this!\n\n" \
+          "Please refer to the manual to make sure your custom\n" \
+          "partitioning meets the requirements of this product."
+        )
+
+        buttons = { yes: Yast::Label.ContinueButton, no: Yast::Label.CancelButton }
+
+        Yast2::Popup.show(message, headline: :warning, buttons: buttons, focus: :no)
```
ancorgs: You can use `buttons: :continue_cancel`, so you don't need to construct your own `buttons` hash

src/lib/y2partitioner/dialogs/main.rb
wfeldt: Yes, I'm aware of it. It still strikes me as odd.

As said for the openSUSE part we don't provide support, so you can't really loose it. And for SLES the customer has a real contract. And I'm not sure we can just claim such a thing.

Just a thought. Treat it as NP, if you want.

wfeldt: I was just reminded that this warning is intended specifically for caasp and there it is true.

So, please ignore this.

src/lib/y2storage/clients/inst_disk_proposal.rb
```diff
@@ -111,9 +114,8 @@ def expert_partitioner(initial_graph)
           @proposal = nil
           @devicegraph = dialog.device_graph
         when :back
-          # Try to create a proposal when staging has been reseted to probed
-          # (i.e., after rescannig devices)
-          create_initial_proposal unless storage_manager.staging_changed?
+          # Try to create a proposal when the system was reprobed
+          create_initial_proposal if reprobed?
```
ancorgs: Good catch, please mention bsc#1088960 in the comment

ancorgs:
There is a fundamental thing I dislike. `expert_partitioner_warning` and `proposal_settings_editable` are not settings for the proposal (despite the name of the latter).

The `ProposalSettings` objects are there to be passed to every proposal attempt in order to configure such attempt. That class is not meant to be a jumble of all kinds of settings related to `Y2Storage`.

src/lib/y2partitioner/dialogs/main.rb
ancorgs: Well, know it will also be displayed for Kubic, I guess. In that context, the usage of "product" and "lose support" becomes tricky. @sysrich Will kubic use `expert_partitioner_warning`?

src/lib/y2storage/proposal_settings.rb
```diff
@@ -105,6 +105,15 @@ class ProposalSettings
     #   the root filesystem
     attr_accessor :subvolumes
 
+    # @note :ng format
```
ancorgs: This is not legacy or ng format. Is simply not a proposal setting. :-)  But in any case, it has always been there.

src/lib/y2storage/proposal_settings.rb
```diff
@@ -105,6 +105,15 @@ class ProposalSettings
...
+    # @return [Boolean] whether to show an extra warning pop-up when the user enters
+    #   the expert partitioner
+    attr_accessor :expert_partitioner_warning
+
+    # @note :ng format
+    # @return [Boolean] whether the user is allowed to change the proposal settings
```
ancorgs: Same than the other

src/lib/y2partitioner/dialogs/main.rb
sysrich: @ancorgs yes, Kubic will use that warning, primarily because it's also needed by CaaSP and Kubic is exploring new methods to be considered for the CaaSP installation

In both cases I'm fine with the wording because it just simply wont work if (for example) you setup Kubic or CaaSP without btrfs and snapshots and readonly rootfs :)

src/lib/y2storage/clients/inst_disk_proposal.rb
```diff
@@ -155,6 +165,35 @@ def new_proposal(settings)
...
+      #
+      # @return [Boolean]
+      def show_guided_setup?
+        Dialogs::GuidedSetup.allowed? || several_candidate_disks?
+      end
+
+      # Whether there are more than one candidate disks
+      #
+      # @return [Boolean]
+      def several_candidate_disks?
```
ancorgs: Not sure why the logic about several_candidate_disks? is here and not in the `GuidedSetup` class. That's kind of splitting the logic into two places.

src/lib/y2storage/dialogs/guided_setup.rb
```diff
@@ -35,6 +36,41 @@ module Dialogs
     #
     # Calculates the proposal settings to be used in the next proposal attempt.
     class GuidedSetup
+      extend PartitioningFeatures
+
+      class << self
+        # Whether is is allowed to use the Guided Setup
+        #
+        # @return [Boolean]
+        def allowed?
```
ancorgs: Isn't this adding one surplus indirection level?

joseivanlopez: Yes, but as commented offline, the method `settings_editable?` is meant to only read the value from the control file and to propose a default value. I would create a new class `PartitioningSettings` with the logic to read all partitioning section from control file, proposing the expected default values for each setting. Then, current class `ProposalSettings` and this `GuidedSetup` could use it instead of reading the setting by themselves. But we are in RC3 phase.

PR #605, opened 2018-04-16
Add missing translation

joseivanlopez:
PBI: https://trello.com/c/VjBb7Jxy/347-0-sles15-p3-1086892-slex15loc-yastalllangsuntranslated-modify-the-disks-confirmation-pop-up-after-finishing-changes-in-expert-pa

### Note for reviewers

Version number is not increased because I want to merge this PR before this one https://github.com/yast/yast-storage-ng/pull/603

### DIFF
--- a/src/lib/y2partitioner/clients/main.rb
+++ b/src/lib/y2partitioner/clients/main.rb
@@ -32,7 +32,10 @@ module Y2Partitioner
     # The entry point for starting partitioner on its own. Use probed and staging device graphs.
     class Main
       extend Yast::I18n
-      extend Yast::Logger
+
+      def initialize
+        textdomain "storage"
+      end
 
       # Run the client
       #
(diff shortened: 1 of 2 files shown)


src/lib/y2partitioner/clients/main.rb
```diff
@@ -58,15 +59,15 @@ def self.run(allow_commit: true)
...
 
-      # Ask whether to proceed with changing the disks;
-      # or inform that we will not do it.
+      # Ask whether to proceed with changing the disks; or inform that we will not do it.
+      #
       # @return [Boolean] proceed
       def self.should_commit?(allow_commit)
         if allow_commit
-          q = "Modify the disks and potentially destroy your data?"
+          q = _("Modify the disks and potentially destroy your data?")
```
ancorgs: Is this the same message we used to have in yast-storage? Just asking to make sure we do not promote some temporary message to be a final one, specially if we have a translated version already.

joseivanlopez: Well, it seems the old Partitioner does not show a similar warning. Instead, it has a last step to summarize all the actions to perform.

joseivanlopez:
Closed in favor of https://github.com/yast/yast-storage-ng/pull/610

PR #606, opened 2018-04-16
Missing help texts in partitioner

shundhammer:
https://trello.com/c/gZKY3Hth/265-3-sles15-p5-1079591-partitioner-empty-help-texts

Some of the help texts were taken from the old yast-storage to make use of existing translations. But this was not always possible, so many texts are new.

## Add LVM VG

help-add-lvm-vg

## Add LVM LV

help-add-lvm-lv-1

help-add-lvm-lv-2

## Add RAID

help-raid

## NFS

help-nfs

This was a bit problematic since the NFS module is actually embedded into the partitioner. That NFS module uses `Wizard::SetHelpText()`, but this is immediately overwritten by CWM and the tree / page logic in the partitioner. I tried to somehow recycle the original help texts from the NFS module, but since there is no `Wizard::GetHelpText()` counterpart, this proved to be nontrivial; so I chose to just add a new help text for the partitioner running the NFS module in the embedded mode. This only affects the list page; the "Add" page etc. all have their separate working help texts since they don't use the wizard (so the texts are not overwritten immediately by CWM).

## Device Graph

help-devicegraph

## Installation Summary

help-summary-installed-system

This is a bit different between the installed system and the installation to make it clear that during installation, nothing is changed until the complete installation summary is confirmed. This screenshot shows only the help text in the installed system.

## Settings

help-settings

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Apr 17 12:01:08 UTC 2018 - shundhammer@suse.com
+
+- Added missing help texts in the partitioner (bsc#1079591) 
+- 4.0.157
+
+-------------------------------------------------------------------
 Fri Apr 13 12:33:25 UTC 2018 - jlopez@suse.com
 
 - Added support for settings 'expert_partitioner_warning' and
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:        4.0.156
+Version:        4.0.157
 Release:	0
 BuildArch:	noarch
 


src/lib/y2partitioner/dialogs/lvm_lv_size.rb
```diff
@@ -148,6 +148,14 @@ def label
           _("Size")
         end
 
+        # @macro seeAbstractWidget
+        def help
+          _("<p><b>Size:</b> The size of this logical volume. " \
+            "Many filesystem types (e.g., Btrfs, XFS, Ext2/3/4) " \
+            "can be enlarged later if needed." \
+            "<p>")
```
mvidner: \</p>

src/lib/y2partitioner/dialogs/md.rb
```diff
@@ -156,6 +167,46 @@ def buttons
...
+            "This mode combines management of a larger number of disks " \
+            "and still maintains some redundancy. " \
+            "This mode can be used on three disks or more. " \
+            "If one disk fails, all data are still intact. " \
+            "If two disks fail simultaneously, all data are lost.")
+        end
+
+        def raid6_help
+          _("<b>RAID 6:</b>" \
+            "This is similar to RAID 5, but with even more reduncany. " \
```
mvidner: s/reduncany/redundancy/

shundhammer:
Thanks for spotting this. Both fixed.

src/lib/y2partitioner/dialogs/md.rb
```diff
@@ -142,6 +142,17 @@ def items
           ]
         end
 
+        def help
+          # TRANSLATORS: Help text heading
+          "<p><b>" + _("RAID Type:") + "</b><ul><li>" +
+            [raid0_help,
```
joseivanlopez: NP: `Yast::HTML.Para` and `Yast::HTML.List` could be used instead of write plain HTML (but I am fine as it is).

src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
```diff
@@ -38,6 +38,26 @@ def initialize(controller)
         super()
       end
 
+      def help
+        help_available_pvs + help_selected_pvs
+      end
+
+      def help_available_pvs
+        _("<p><b>Available Devices:</b> " \
+          "A list of available LVM physical volumes. " \
```
joseivanlopez: Strictly speaking, not only partitions can be used. In general, not used devices (disks, raids, etc) can be selected as PV. Maybe this description should be similar to the the description in "md_devices_selector". See https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/actions/controllers/lvm_vg.rb#L147

src/lib/y2partitioner/widgets/md_devices_selector.rb
```diff
@@ -128,6 +132,25 @@ def handle(event)
       # @return [Y2Partitioner::Actions::Controllers::Md]
       attr_reader :controller
 
+      def help_available
+        _("<p><b>Available Devices:</b> " \
+          "Unused disks and partitions that can be used for a RAID. " \
```
joseivanlopez: Right now, the MD RAID can only be created using partitions, so the description should be adapted. See https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/actions/controllers/md.rb#L75

src/lib/y2partitioner/widgets/md_devices_selector.rb
```diff
@@ -128,6 +132,25 @@ def handle(event)
...
+          "and no filesystem directly on the disk. " \
+          "A partition can be used if it is not mounted. " \
+          "It is recommended to use partition ID \"Linux RAID\" " \
+          "for those partitions." \
+          "</p>")
+      end
+
+      def help_selected
+        _("<p><b>Selected Devices:</b> " \
+          "The disks and partitions that are used for the RAID. " \
```
joseivanlopez: It should only be partitions according to current implementation.

src/lib/y2partitioner/dialogs/md.rb
shundhammer: Since we are using literal HTML tags all over the place everywhere else (in particular in the help texts), this would just be an exercise in futility IMHO.

src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
shundhammer: Strictly speaking, it does not say here that this is limited to partitions. It just shows one way to add devices if there are none to choose from, and since it's pretty hard to insert additional physical hard disks at this stage, the next paragraph suggest the one thing that is doable with a reasonable amount of work for the user: Create dedicated LVM PV partitions.

This help is not meant to be a complete substitute for the manual; it is supposed to give some hints. Unlike the old help texts that were pretty much useless and explained only the obvious things (which is of course a lot easier to write), my intention was to point the user into the right direction what can be done. Yes, this might mean that it does not cover every single corner case; that's life. But an understandable text is much better than a lawyer-like complete specification of all the possible scenarios and pathological cases; clarity wins over completeness.

src/lib/y2partitioner/widgets/md_devices_selector.rb
shundhammer: But that's a bug then, right? Wouldn't I really want to use plain disks for something like mirroring? If this is a limitation of the current code, will somebody adapt the help text when this limitation is lifted some day? (In my experience, this always gets forgotten).

shundhammer: BTW Arvin confirmed that libstorage-ng does not have that restriction, so this is purely a restriction of our yast-storage-ng Ruby code.

joseivanlopez: Probably it only uses partitions because it is what the old partitioner did. Ok, I am fine if we keep the current help text, but we also should create a new card in trello to correct the behavior of the code. Otherwise, this mismatch will come with bug reports, I guess.

src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
joseivanlopez: Completely agree, but I was not suggesting to add a fully complete description of all possible cases, only some description in the way you did for md devices. But I agree, at least it is told the easy way to have available PVs.

PR #608, opened 2018-04-18
Implement use_available for lvm_vg_strategy

ancorgs:
Main PR for https://trello.com/c/zQBrlwY3/332-3-proposal-implement-missing-lvm-strategies-useavailable

Review commit by commit.

It looks like tons of YAML files were added from scratch but that's not exactly true. The new files actually contain a 1:1 copy of the previous YAML files. We made `use_available` the default, but I wanted to keep those tests for which the usage of `use_needed` was significant. So I kept the YAML files (renaming then to `whatever-use_needed.xml`) and added tests for them.


ancorgs:
Codeclimate is re-detecting complexity issues that were already there. It complains I fixed 4 and produced 4 new ones. Obviously, what really happened is that the methods where relocated to another file.

src/lib/y2storage/proposal/phys_vol_strategies/use_available.rb
```diff
@@ -0,0 +1,138 @@
...
+        # @return [Array<Array<Planned::AssignedSpace>>] the sets with more
+        #   assigned spaces come first, because if they succeed there is no
+        #   point in checking any other combination
+        def space_combinations
+          [true, false].repeated_permutation(useful_spaces.size).map do |yes_no|
+            yes_no.each_with_object([]).with_index do |(value, array), idx|
+              array << useful_spaces[idx] if value
+            end
+          end.sort_by(&:size).reverse
+        end
```
ancorgs: As pointed by @joseivanlopez offline, this method could be as easy as
```ruby
def space_combinations
  useful_spaces.size.downto(1).each_with_object([]) do |size, result|
    result.concat(useful_spaces.combination(size).to_a)
  end
end
```

PR #610, opened 2018-04-19
Add partitioner confirmations

joseivanlopez:
PBI: https://trello.com/c/VjBb7Jxy/347-0-sles15-p3-1086892-slex15loc-yastalllangsuntranslated-modify-the-disks-confirmation-pop-up-after-finishing-changes-in-expert-pa

### DIFF
--- a/src/lib/y2partitioner/clients/main.rb
+++ b/src/lib/y2partitioner/clients/main.rb
@@ -107,6 +107,7 @@ module Y2Partitioner
 
       # Popup when commit is forbidden (e.g., when the client is used for manual testing)
       def forbidden_commit_warning
+        # TRANSLATORS: this comment is only for testing purposes. Do not care too much.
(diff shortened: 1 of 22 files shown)


src/lib/y2partitioner/clients/main.rb
```diff
@@ -19,57 +19,97 @@
...
+          "kernel, which will most likely lead to data loss.\n\n" \
+          "To continue despite this warning, click Yes."
+        )
+
+        Yast2::Popup.show(message, headline: :warning, buttons: :yes_no)
+      end
+
+      # Popup when commit is forbidden (e.g., when the client is used for manual testing)
+      def forbidden_commit_warning
+        message = _("Nothing gets written because commit is not allowed.")
```
ancorgs: NP: maybe a comment for translators (also useful for developers) so they are away this is not really a string to be displayed to normal users, just something for the test mode.

PR #613, opened 2018-04-23
Partitioner: default value for snapper checkbox (bsc#1084491)

ancorgs:
Main PR for https://trello.com/c/P3bHR7pD/369-not-estimated-p1-1084491-build-4897-storage-openqa-test-fails-in-installationsnapshots-snapper-is-not-configured-when-lvm-encryp

### DIFF
--- a/src/lib/y2partitioner/filesystem_role.rb
+++ b/src/lib/y2partitioner/filesystem_role.rb
@@ -1,0 +1,139 @@
+# encoding: utf-8
+
+# Copyright (c) [2017-2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2storage"
+
+Yast.import "Mode"
+
+module Y2Partitioner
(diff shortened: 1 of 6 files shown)


ancorgs:
I just verified manually it works as expected during SLE15 installation, i.e. the snapshots checkbox is checked by default for the role Operating System and devices >= 12GiB.

So I added version bump and changelog entry (rebased).

src/lib/y2partitioner/filesystem_role.rb
```diff
@@ -0,0 +1,139 @@
...
+module Y2Partitioner
+  # Class to represent each one of the roles presented to the user when creating
+  # a new partition, LVM logical volume or MD RAID.
+  class FilesystemRole
+    include Yast::I18n
+    extend Yast::I18n
+    textdomain "storage"
+
+    # Constructor
+    def initialize(id, name, part_id, fs_type)
```
joseivanlopez: Please, document parameters.

src/lib/y2partitioner/filesystem_role.rb
```diff
@@ -0,0 +1,139 @@
...
+      @id = id
+      @name = name
+      @partition_id = Y2Storage::PartitionId.find(part_id)
+      @filesystem_type = Y2Storage::Filesystems::Type.find(fs_type) if fs_type
```
joseivanlopez: It seems that `fs_type` is optional, isn't it? In that case, I would initialize it as nil in the parameters list.

src/lib/y2partitioner/filesystem_role.rb
```diff
@@ -0,0 +1,139 @@
...
+
+    # Sorted list of all possible roles
+    def self.all
+      ALL.dup
+    end
+
+    # Finds a role by its id
+    #
+    # @param id [Symbol, nil]
+    # @return [FilesystemRole, nil] nil if such nil does not exist
```
joseivanlopez: Fix text: nil if such "role" does not exist.

src/lib/y2partitioner/filesystem_role.rb
ancorgs: Is not that bad since the constructor is only used by the class itself to populate `ALL`. So to minimize risks after the successful manual tests, I will only perform the changes suggested to comments.

PR #619, opened 2018-04-25
Fix partitioner client bug

joseivanlopez:
Partitioner was failing after confirming changes in a running system.

The bug was introduced with PR#610. See also https://trello.com/c/VjBb7Jxy/347-2-sles15-p3-1086892-slex15loc-yastalllangsuntranslated-modify-the-disks-confirmation-pop-up-after-finishing-changes-in-expert-pa


shundhammer:
Fix confirmed working.

shundhammer:
LGTM again.

shundhammer:
I trust that test does what it is supposed to do; I couldn't say that I know for sure what excatly this does. ;-)

PR #621, opened 2018-04-26
WIP: provide an alternative to the old Storage.GetDetectedDiskPaths

ancorgs:
This is a WIP.

Needed for https://github.com/yast/yast-installation/pull/689/

Depends on https://github.com/openSUSE/libstorage-ng/pull/522


src/lib/y2storage/storage_manager.rb
```diff
@@ -339,6 +339,29 @@ def mode
...
+    # It performs a very simple check and returns true if there is any device
+    # of one of the acceptable types (basically disks or DASDs).
+    #
+    # It will never trigger a hardware probing. The method works even if
+    # such probing has not been performed yet.
+    #
+    # @return [Boolean]
+    def devices_for_installation?
+      if probed?
+        !probed.empty?
```
aschnell: probed.empty? is different from Storage.light_probe. E.g. the first is ```true``` if only NFS mount points are found, e.g. of the installation medium, while the second one is ```false```.

ancorgs:
Mandatory: The dependency to libstorage-ng should be increased to a version containing `light-probe`.

Nice to have: I would like to see some unit test.

In addition, one commit is called "WIP", the other "changes" and then we have two merge commits. Not the best example of a traceable and meaningful git history.

In any case, I cannot approve it. :-)

src/lib/y2storage/storage_manager.rb
ancorgs: Well, if we want a closer behavior in both cases we could check for `probed.disk_devices.empty?`

Although we would need to revisit the code after adding support for Xen virtual partitions. https://trello.com/c/EqN6jm0m/370-sles15-p3-1085134-cannot-install-when-using-xen-virtual-partitions-xvda1xvda2

PR #622, opened 2018-04-26
Import mount points

joseivanlopez:
PBI: https://trello.com/c/rM3nhW1o/352-5-storage-ng-partitioner-bring-import-mount-points-back

Tested manually and it works.

ng-import_mount_points

ng-import_mount_points_error2


src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -0,0 +1,184 @@
...
+        end
+
+        # Formats the device indicated in the fstab entry
+        #
+        # @param device [Y2Storage::BlkDevice]
+        # @param fs_type [Y2Storage::Filesystems::Type]
+        #
+        # @return [Y2Storage::Filesystems::Base]
+        def format_device(device, fs_type)
+          device.delete_filesystem
```
ancorgs: NP: Shouldn't this be `remove_descendants`? Just trying to cover weird use cases like using a partition that nowadays belongs to an LVM or it's encrypted, but that it used to be used as a plain partition in the imported fstab.

I admit that's a very weird scenario, but with `delete_filesystem` there is a chance we crash on those scenarios, while `#remove_descendants` should be more robust.

src/lib/y2partitioner/actions/import_mount_points.rb
```diff
@@ -0,0 +1,91 @@
...
+require "yast"
+require "yast/i18n"
+require "yast2/popup"
+require "y2partitioner/actions/controllers/fstabs"
+require "y2partitioner/dialogs/import_mount_points"
+
+module Y2Partitioner
+  module Actions
+    # Action for importing mount points from a fstab file
+    class ImportMountPoints
```
ancorgs: NP: Not sure if we need a separate Actions object and a separate Controller for this. At first sight, it may look like a bit too much for not so much logic (is not one of those actions with multiple interdependent steps and so on).

Just thinking aloud, not implying is necessarily  wrong.

joseivanlopez: Well, IMHO too much logic for only a button and a dialog. I guess the result is much clearer with the Action and the Controller. It is true that there aren't several steps, but there are several widgets accessing to common data. Moreover, the button would require info about the found fstabs to show an error popup or not, so it should read somehow the fstabs. I guess it is better to have all that stuff in a specific action instead of directly in the button.

src/lib/y2partitioner/dialogs/popup.rb
```diff
@@ -26,6 +26,22 @@ module Y2Partitioner
...
+      DEFAULT_MIN_WIDTH = 50
+      private_constant :DEFAULT_MIN_WIDTH
+
+      DEFAULT_MIN_HEIGHT = 18
+      private_constant :DEFAULT_MIN_HEIGHT
+
+      # @!method min_width=(value)
+      #   Sets the popup min width
+      #   @param value [Integer]
+      attr_writer :min_width
```
ancorgs: Why creating an accessor instead of redefining a `min_width` method in the subclasses? Just as it's done for `ok_button_label` for example.

src/lib/y2partitioner/widgets/fstab_selector.rb
```diff
@@ -0,0 +1,341 @@
...
+      end
+
+      # Checks whether the mount points of the selected fstab can be imported
+      #
+      # It shows a popup with the errors detected in the selected fstab.
+      #
+      # @see Actions::Controllers::Fstabs#selected_fstab_errors
+      #
+      # @return [Boolean]
+      def validate
```
ancorgs: Just curious: the "all or nothing" behavior was already there in the old partitioner or did the old partitioner perform partial imports if some mount points were importable and others weren't?

src/lib/y2partitioner/dialogs/popup.rb
joseivanlopez: In fact, the first implementation was a method, but I am not sure, I see these values as a kind of configuration but not a different behavior. I mean, I would not define a new subclass only because these values.

src/lib/y2partitioner/widgets/fstab_selector.rb
joseivanlopez: Good question. I will check.

joseivanlopez: I was checking the behavior of the old partitioner. It seems that the fstab entries for missing devices are not shown, so as consequence, it performs partial imports. But, when any device is found, it present an empty list, and when the user import it, simply the probed devicegraph is load.

I am not sure which behavior is better. Does it make sense to import the mount points if some of them cannot be imported?

src/lib/y2partitioner/dialogs/popup.rb
ancorgs: That's exactly the CWM approach. Subclasses for redefining mostly everything. Why should this be different?

src/lib/y2partitioner/widgets/fstab_selector.rb
ancorgs: The whole functionality is rather weird and is there to please some obscure use case buried in the YaST history. Since we don't have the original use cases at hand, I would reproduce the functionality as closely as possible. But maybe we should ask that question to @jsrain

src/lib/y2partitioner/actions/import_mount_points.rb
ancorgs: Ok. Fine with me.

src/lib/y2partitioner/actions/controllers/fstabs.rb
ancorgs: As discussed in another channels. Maybe a better fix is not to use `remove_descendants` (which is rather aggressive and can cause collateral damages to other mount points we want to import) but to ensure the code is only executed in cases in which the current `delete_filesystem` cannot fail and crash. That is, being more strict on deciding whether a mount point is importable or not.

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -207,10 +214,34 @@ def find_linux_systems(disk)
...
       filesystems.compact!
-      return [] if filesystems.empty?
       filesystems.map { |f| release_name(f) }.compact
     end
 
+    # Finds fstab files in all filesystems
+    #
+    # @see #suitable_root_filesystems
+    #
+    # @return [Array<Fstab>]
```
ancorgs: I would add a `TODO` comment somewhere around this functionality about (the lack of) caching. As we discussed offline, it's a pity that we are always re-mounting some partitions that were likely mounted already at a previous point of the installation by `#installed_systems`

src/lib/y2storage/fstab.rb
```diff
@@ -0,0 +1,64 @@
...
+    attr_reader :entries
+
+    # Constructor
+    #
+    # @param path [String]
+    # @param filesystem [Filesystems::Base]
+    def initialize(path = FSTAB_PATH, filesystem = nil)
+      @path = path
+      @filesystem = filesystem
+      @entries = StorageManager.fstab_entries(path)
```
ancorgs: I wouldn't delegate this to `StorageManager`. No need to pollute that class with something that definitely doesn't seem to fit its current responsibilities. I would have the call to `Storage.read_simple_etc_fstab` directly in the `Fstab` class.

src/lib/y2storage/storage_manager.rb
```diff
@@ -595,6 +599,18 @@ def create_test_instance
         create_instance(test_environment)
       end
 
+      # Reads a fstab file and returns its entries
+      #
+      # @param path [String] path to the fstab file to read
+      # @return [Array<SimpleEtcFstabEntry>]
+      def fstab_entries(path)
```
ancorgs: As commented above, I don't see any good reason for this to belong here.

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -0,0 +1,237 @@
...
+        # An entry can be imported when the device is known and it is not used
+        # by other device (e.g., used by LVM or MD RAID) or it is a known NFS.
+        #
+        # @param entry [Y2Storage::SimpleEtcFstabEntry]
+        # @return [Boolean]
+        def can_be_imported?(entry)
+          device = entry.device(system_graph)
+          return false unless device
+
+          !device.is?(:blk_device) || can_be_formatted?(device)
```
ancorgs: Why the `!device.is?(:blk_device)` check? I would appreciate an explanatory comment

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -0,0 +1,237 @@
...
+        # by other device (e.g., used by LVM or MD RAID).
+        #
+        # @return [String, nil] nil if all entries can be imported
+        def not_importable_entries_error
+          entries = not_importable_entries
+          return nil if entries.empty?
+
+          mount_points = entries.map(&:mount_point).join("\n")
+
+          format(_("The following mount points cannot be imported:\n%{mount_points}"),
```
ancorgs: Named placeholders like `%{mount_points}` are not very common outside the Ruby world, so better add a TRANSLATORS comment advising to not modify the content inside the `%{}`

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -0,0 +1,237 @@
...
+
+        # Formats the device indicated in the fstab entry
+        #
+        # @param device [Y2Storage::BlkDevice]
+        # @param fs_type [Y2Storage::Filesystems::Type]
+        #
+        # @return [Y2Storage::Filesystems::Base]
+        def format_device(device, fs_type)
+          device.delete_filesystem
+          device.create_filesystem(fs_type)
```
ancorgs: What happens if fs_type is 'auto'?

ancorgs: Or even 'none'

src/lib/y2partitioner/actions/controllers/fstabs.rb
joseivanlopez: I tried to explain it the description of the method: An entry can be imported when the device is known and it is not used by other device (e.g., used by LVM or MD RAID), or it is a known NFS.", but I put it more explicitly in the code line.

PR #623, opened 2018-04-27
Fix logger

jreidinger:


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Apr 27 05:57:25 UTC 2018 - jreidinger@suse.com
+
+- fix libstorage logging that do not expect printf expansion
```
jreidinger: expand not expect :)

jreidinger: no, in fact I want expect

src/lib/y2storage/storage_manager.rb
```diff
@@ -643,6 +643,10 @@ def retry_create_instance?(error, callbacks)
     class StorageLogger < ::Storage::Logger
       # rubocop:disable Metrics/ParameterLists
       def write(level, component, filename, line, function, content)
+        # libstorage pretent that it use same logging as y2_logger but y2_logger support also
```
aschnell: libstorage-ng does not pretend anything. If you look at Logger::write and tell me where the arguments to insert should some from I would be very happy.

PR #625, opened 2018-04-27
Proper partitioner checks for root on NFS (bsc#1090752)

ancorgs:
See https://trello.com/c/F1oQeriK/377-1-sles15-p1-1090752-internal-error-undefined-method-mbrgap-for-nilnilclass

The reporter found this crash installing on NFS-root in a diskless system
crash

Additionally, while writing the unit tests I found the boot requirement checks also misbehaved if there were local disks combined with root in NFS, just in a different way.

This PR adds fixes and unit tests for both situations.

Even after fixing the crash, the SetupChecker was still reporting this warning.
wrong-warning

Now that should also be fixed with the latest commit (added after the first approval of this PR). Everything up to that point tested manually in a diskless system and in a disk with a local disk (in addition to the NFS mount).

BUT, at the moment of writing bootloader fails if the user decides to continue. That's, of course, another story.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Apr 27 15:23:05 UTC 2018 - ancor@suse.com
+
+- Partitioner: fixed checks when the root filesystem is NFS
+  (bsc#1090752).
+- 4.0.169
+
+-------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


jreidinger:
bootloader failure is caused by missing check for nfs root ( it is old workaround ). Ideal solution for bootloader will be to use NONE bootloader, as for nfs root we do not install it anywhere.

PR #627, opened 2018-05-03
Properly handle subvolumes/subvolumes_prefix

imobachgs:
* Fixes [bsc#1091669](https://bugzilla.suse.com/show_bug.cgi?id=1091669) and [bsc#1090095](https://bugzilla.suse.com/show_bug.cgi?id=1091669)
* Related to https://github.com/yast/yast-storage-ng/pull/627
* Trello cards: https://trello.com/c/xjlFsCBz and https://trello.com/c/nTbvDnuU/

### DIFF
--- a/src/lib/y2storage/proposal/autoinst_devices_planner.rb
+++ b/src/lib/y2storage/proposal/autoinst_devices_planner.rb
@@ -58,7 +58,6 @@ module Y2Storage
       # @return [Array<Planned::Device>] List of planned devices
       def planned_devices(drives_map)
         result = []
-        @default_subvolumes_used = false
 
         drives_map.each_pair do |disk_name, drive_section|
           case drive_section.type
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -246,17 +245,26 @@ def add_snapshots(device, drive_section)
...
-          subvol_specs = proposal_settings.subvolumes
-        end
+        device.default_subvolume = section.subvolumes_prefix || defaults[:subvolumes_prefix]
 
-        device.default_subvolume = section.subvolumes_prefix ||
-          proposal_settings.legacy_btrfs_default_subvolume
-        device.subvolumes = subvol_specs
+        device.subvolumes =
+          if section.create_subvolumes
+            section.subvolumes || defaults[:subvolumes] || []
```
joseivanlopez: NP: Why to not always ensure useful defaults? I mean, here I would expect that `defaults[:subvolumes]` always has a proper default value, but sometimes it is `nil`, which isn't a valid default for subvolumes (it needs to be an array).

test/y2storage/autoinst_profile/partition_section_test.rb
```diff
@@ -498,9 +507,19 @@ def section_for(name)
...
+        section.filesystem = :btrfs
+        section.subvolumes = nil
+      end
+
+      it "does not export #subvolumes" do
+        expect(section.to_hashes.keys).to_not include "subvolumes"
+      end
+
+      it "does not export #create_subvolumes" do
+        expect(section.to_hashes.keys).to_not include "subvolumes"
```
joseivanlopez: I guess you want to check the key `create_subvolumes`

test/y2storage/proposal/autoinst_devices_planner_test.rb
```diff
@@ -508,6 +521,28 @@
         end
       end
 
+      context "when a subvolumes prefix is specified" do
```
joseivanlopez: NP: typo "when a subvolume"

src/lib/y2storage/proposal/autoinst_devices_planner.rb
imobachgs: Because if you use a `Btrfs` filesystem on a partition which is not defined in the list of volumes, it won't work.

PR #631, opened 2018-05-04
- work on disable MD auto assembly

aschnell:
For https://trello.com/c/rjLmkP8u/373-sles15-p2-1090690-storage-ng-error-encrypting-a-partition-that-was-part-of-an-md-raid.


package/yast2-storage-ng.spec
```diff
@@ -28,6 +28,8 @@ Requires:	yast2 >= 4.0.61
 Requires:	yast2-ruby-bindings >= 4.0.6
 # function light_probe
 Requires:	libstorage-ng-ruby >= 3.3.254
+# communicate with udisks
+Requires:	rubygem(dbus)
```
lslezak: Is it `rubygem(ruby-dbus)` ? And I think the `rubygem(%rb_default_ruby_abi:ruby-dbus)`  way should be preferred as that will enforce the system Ruby version. (Currently there is only a single Ruby version in SLE15/Leap15 but this will be future proof if Tumbleweed adds a new experimental Ruby version.)

src/lib/y2storage/inhibitors/mdadm_auto_assembly.rb
```diff
@@ -0,0 +1,49 @@
...
+module Y2Storage
+  # class to inhibit mdadm from doing auto assembly
+  class MdadmAutoAssembly
+    include Yast::Logger
+
+    def inhibit
+      log.info "set udev ANACONDA property"
+
+      begin
+        Yast::Execute.locally!("/sbin/udevadm", "control", "--propertyx=ANACONDA=yes")
```
lslezak: Is it really `propertyx`? There is `property` value used in the `uninhibit` method below...

src/lib/y2storage/inhibitors/udisks_inhibitor.rb
```diff
@@ -0,0 +1,62 @@
...
+      rescue DBus::Error => e
+        log.error "uninhibit udisks failed #{e.message}"
+      end
+    end
+
+  private
+
+    def dbus_object
+      system_bus = DBus::SystemBus.instance
+      service = system_bus.service("org.freedesktop.UDisks")
```
lslezak: How does it behave when Udisks is not installed in the system? E.g. the minimal SLES (not registered, without the packages DVD) does not contain the udisks2 package.

package/yast2-storage-ng.spec
aschnell: No idea about this Ruby packing stuff. I just copied it from the BuildRequires.

src/lib/y2storage/inhibitors/mdadm_auto_assembly.rb
aschnell: Ok, that was indeed a test to check error handling.

src/lib/y2storage/inhibitors/udisks_inhibitor.rb
aschnell: DBus exceptions are caught and logged.

package/yast2-storage-ng.spec
mvidner: The name of the gem is "ruby-dbus", the spec file needs to be fixed.

src/lib/y2storage/inhibitors/udisks_inhibitor.rb
lslezak: OK, thanks.

lslezak:
I do not consider a tool which finds duplicates in the code as "useless", but anyway... approved.

ancorgs:
180 new lines of code and no unit tests?

src/lib/y2partitioner/clients/main.rb
```diff
@@ -44,9 +45,16 @@ def initialize
       #
       # @param allow_commit [Boolean] whether the changes can be stored on disk
       def run(allow_commit: true)
```
ancorgs: Not sure how relevant it is, but before the change this method always returned either nil (if the user quited before attempting to commit) or the output of `StorageManager.commit` (which is a boolean indicating whether the commit was successful).

Now it always returns the output of `inhibitors.uninhibit`.

It would be better to keep the behavior.

src/lib/y2storage/inhibitors/udisks_inhibitor.rb
```diff
@@ -0,0 +1,62 @@
...
+      log.info "inhibit udisks"
+      begin
+        @dbus_cookie = dbus_object.Inhibit().first
+      rescue Errno::ENOENT, DBus::Error => e
+        log.error "inhibit udisks failed #{e.message}"
+      end
+    end
+
+    def uninhibit
+      return if !@dbus_object
```
ancorgs: Wouldn't it be useful to log that `#unhibit` was called but there was no `@dbus_object` (likely it means we used the methods in the wrong order).

ancorgs:
So this PR uses the new inhibitors when the partitioner runs standalone (i.e. in an installed system). What's the plan about calling `#inhibit` and `#unhinibit` during the installation? Maybe `#ihnibit` in system_analysis and `#uninhibit` at the end of `inst_prepdisks`?

src/lib/y2partitioner/clients/main.rb
ancorgs: Summarizing my two previous comments, this should be better in terms of respecting the current behavior and avoiding side effects:

```ruby
def run(allow_commit: true)
  return nil if !run_partitioner?

  begin
    inhibitors = Y2Storage::Inhibitors.new
    inhibitors.inhibit

    if partitioner_dialog.run == :next
      allow_commit ? commit : forbidden_commit_warning
    end
  ensure
    inhibitors.uninhibit
  end
end
```

Or maybe, to look closer to the original (and more explicit about returning nil):

```ruby
def run(allow_commit: true)
  return nil if !run_partitioner?

  begin
    inhibitors = Y2Storage::Inhibitors.new
    inhibitors.inhibit

    return nil if partitioner_dialog.run != :next
    allow_commit ? commit : forbidden_commit_warning
  ensure
    inhibitors.uninhibit
  end
end
```

aschnell:
And now there are already conflicts with the usual suspects: changes and spec.

ancorgs:
JFYI: the ruby way of having RAII would be offering block syntax, like it's done in the ruby stdlib for opening files or like we do for example [in SCR#change_scr_root](https://www.rubydoc.info/github/yast/yast-ruby-bindings/Yast%2FRSpec%2FSCR:change_scr_root).

ancorgs:
> The return value of run is not documented. So I must assume it is irrelevant, esp. since
> there is no return statement. We had this discussion before are agreed to disagree.

We agreed to relax Rubocop so you could use explicit `return` even when not needed. We didn't agree that NOT using the explicit `return` (that is, doing things in the usual Ruby way) means you don't care about the returned value so any other developer can change it.

PR #636, opened 2018-05-08
Partitioner: add configure button

ancorgs:
Main PR for https://trello.com/c/XH6Ig5kp/376-3-leap15-impossible-to-install-on-iscsi-or-fcoe

partitioner-configure

Tested manually in an installed system (including the automatic installation of required packages) and during installation (including deletion of some clients to verify its buttons are not longer displayed).

All the strings are reused from the old storage.


src/lib/y2partitioner/widgets/configure.rb
```diff
@@ -0,0 +1,276 @@
...
+    private
+
+      # @return [Array<Yast::WidgetTerm>]
+      def items
+        actions.map do |action|
+          Item(Id(action.id), Yast::Term.new(:icon, action.icon), action.label)
+        end
+      end
+
+      # @return [Yast::WidgetTerm]
```
jreidinger: please try to be consistent with CWM itself which also have opt and here it very confuse me why it have different syntex. See please https://github.com/yast/yast-yast2/blob/33d7484d4827421d11da27b23d6317ef7d2d5b95/library/cwm/src/lib/cwm/abstract_widget.rb#L92

so above in widget should be `Opt(*opt), label, items`

src/lib/y2partitioner/widgets/configure.rb
```diff
@@ -0,0 +1,276 @@
...
+      # Localized label for the menu button
+      #
+      # @return [String]
+      def label
+        # Translators: Configure menu in the initial Partitioner screen
+        _("Configure...")
+      end
+
+      # @macro seeCustomWidget
+      def ids_in_contents
```
jreidinger: please document why it have to be redefined.

ancorgs: I will.

But take into account that for MenuButton you always have to either redefine this or use `handle_all_events`, like we have been forced to do [here](https://github.com/yast/yast-storage-ng/blob/7fed3157279cbc04645e46f650f55525227a55af/src/lib/y2partitioner/widgets/lvm_add_button.rb#L38) or [here](https://github.com/yast/yast-storage-ng/blob/aab80d89db6fada311cae0d042aa579169363b20/src/lib/y2partitioner/widgets/disk_expert_menu_button.rb#L39)

jreidinger: yep, maybe we can add it as CWM object for MenuButton, so it is shared.

src/lib/y2partitioner/widgets/configure.rb
```diff
@@ -0,0 +1,280 @@
...
+      #   executed; nil otherwise.
+      def handle(event)
+        action = Action.find(event["ID"])
+
+        return nil unless action
+        return nil unless warning_accepted?(action) && availability_ensured?(action)
+
+        Yast::WFM.call(action.client)
+        reprobe
+        :reprobe
```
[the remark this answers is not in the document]
joseivanlopez: As commented offline, we can keep this as it is now, and then I will fix it in my PR.

src/lib/y2partitioner/widgets/configure.rb
```diff
@@ -0,0 +1,280 @@
...
+      def ids_in_contents
+        actions.map(&:id)
+      end
+
+      # Available actions to be offered to the user
+      #
+      # @return [Array<Action>]
+      def actions
+        @actions ||=
+          if Yast::Stage.initial
```
joseivanlopez: Why to use `Yast::Stage.initial` instead of `Yast::Mode.instalation` ?

src/lib/y2partitioner/widgets/configure.rb
```diff
@@ -0,0 +1,280 @@
...
+            Action.supported
+          end
+      end
+
+      # Displays the action warning to the user
+      #
+      # @param action [Action]
+      # @return [Boolean] whether the user confirmed the warning message
+      def warning_accepted?(action)
+        Yast::Popup.YesNo(action.warning_text)
```
joseivanlopez: I guess we should start to always use `Yast2::Popup` in new code.

test/y2partitioner/widgets/reprobe.rb
```diff
@@ -0,0 +1,53 @@
+# encoding: utf-8
```
joseivanlopez: NP: We could move this file to `test/support`, due to we have put there other examples files.

src/lib/y2partitioner/widgets/configure.rb
ancorgs: Because is what the old code for the same button in yast2-storage did. And same applies about the decision whether to re-activate or not, in the past `Stage` was used for it.

src/lib/y2partitioner/widgets/configure.rb
ancorgs: Well, in this particular case it provides no benefit (even the opposite, code is more verbose). Another example of being conservative.

src/lib/y2partitioner/widgets/configure.rb
```diff
@@ -0,0 +1,280 @@
...
+Yast.import "PackageSystem"
+
+module Y2Partitioner
+  module Widgets
+    # "Configure" menu button used to run the corresponding YaST clients to
+    # activate several storage technologies
+    #
+    # Most of the behavior of this widget is a direct translation of the old
+    # Yast::PartitioningEpAllInclude from yast-storage
+    class Configure < CWM::CustomWidget
```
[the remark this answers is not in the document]
ancorgs: Well, most of the times it returns a `MenuButton` and sometimes it returns an `Empty`. So I considered `ConfigureWidget` (since `ConfigureMenuButton` was not fully precise)... then I realized `Widgets::ConfigureWidget` was redundant... and I ended up with no suffix. :wink:

So typical twisted developer mind process.

src/lib/y2partitioner/widgets/configure.rb
ancorgs: And there are other good reasons. In an AutoYaST installation, the second stage runs already in the target system, after the first reboot (but I assume `Mode.installation` is still true). In that situation we want the Partitioner (and Y2Storage in general) to behave like if it's working in an already installed system.

PR #637, opened 2018-05-08
Rudimentary support in YAML writer for unknown device types (RAID, multipath, NFS)

shundhammer:
https://trello.com/c/w14Rh3JY/378-1-yaml-writer-inform-about-missing-stuff

For device types (more exactly: for block devices) where we don't have proper support in our YAML reader / writer, add at least a section to the YAML output that shows that there is something so the XML file can be used for more details.

Sample output:

```yaml
- disk:
    name: "/dev/sdc"
    size: 32 GiB
    block_size: 0.5 KiB
    io_size: 0 B
    min_grain: 1 MiB
    align_ofs: 0 B
- unsupported_device:
    type: Y2Storage::DmRaid
    name: "/dev/mapper/isw_ddgdcbibhd_test1"
    support: unsupported in YAML - check XML
- unsupported_device:
    type: Y2Storage::DmRaid
    name: "/dev/mapper/isw_ddgdcbibhd_test2"
    support: unsupported in YAML - check XML
- unsupported_device:
    type: Y2Storage::Filesystems::Nfs
    name: server:/path
    support: unsupported in YAML - check XML
```

## Notice

Right now, the YAML reader can**not** read such a file; it will be rejected because of those `unsupported_device` entries.  IMHO this is desired, but it is debatable if we should extend the YAML reader so it can at least ignore that part. OTOH this might be unsafe for some use cases.

### DIFF
--- a/src/lib/y2storage/yaml_writer.rb
+++ b/src/lib/y2storage/yaml_writer.rb
@@ -87,6 +87,7 @@ module Y2Storage
       yaml = []
       top_level_devices(devicegraph).each { |device| yaml << yaml_disk_device(device) }
       devicegraph.lvm_vgs.each { |lvm_vg| yaml << yaml_lvm_vg(lvm_vg) }
+      unsupported_devices(devicegraph).each { |device| yaml << yaml_unsupported_device(device) }
       yaml
     end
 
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/yaml_writer.rb
```diff
@@ -440,5 +441,70 @@ def yaml_btrfs_subvolume(subvol)
       content["nocow"] = "true" if subvol.nocow?
       { "subvolume" => content }
     end
+
+    # Return YAML for any unsupported objects in a devicegraph.
+    # Right now this is limited to block devices; this might change in the
```
ancorgs: This comment about only block devices is outdated. We also consider `Nfs` objects.

src/lib/y2storage/yaml_writer.rb
```diff
@@ -440,5 +441,70 @@ def yaml_btrfs_subvolume(subvol)
...
+      content = []
+      unsupported_devices(devicegraph).each { |device| content << yaml_unsupported_device(device) }
+      return nil if content.empty?
+
+      { "unsupported" => content }
+    end
+
+    # Return YAML for one unsupported device.
+    #
+    # @param  device [Device]
```
ancorgs: Maybe worth mentioning that it should be a device responding to `#name`. Which, BTW, in the usual Ruby/Yardoc notation can be specified as
```ruby
# @param device [#name]
```

src/lib/y2storage/yaml_writer.rb
shundhammer: Removing that method; it's now used anymore.

src/lib/y2storage/yaml_writer.rb
```diff
@@ -440,5 +441,56 @@ def yaml_btrfs_subvolume(subvol)
       content["nocow"] = "true" if subvol.nocow?
       { "subvolume" => content }
     end
+
+    # Return YAML for one unsupported device.
+    #
+    # @param  device [Device]
```
ancorgs: Repeating a comment that got lost in the push.

It's maybe worth mentioning that it must be a device responding to `#name`. BTW, in the usual Yardoc/Ruby notation that can be expressed like:
```ruby
# @param device [#name]
```

src/lib/y2storage/yaml_writer.rb
shundhammer: Oops, I was pretty sure that _name_ is defined for all devices. If that's not the case, I'll rather add a `if respond_to?(:name)`.

PR #638, opened 2018-05-08
Fix partitioner buttons

joseivanlopez:
PBI: https://trello.com/c/Lqos3IvN/238-2-partitioner-fix-abort-and-back-buttons

partitioner_quit

partitioner_cancel_button

Manually tested.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri May 11 10:31:10 UTC 2018 - jlopez@suse.com
+
+- Partitioner: fix buttons to abort and to go back
+  (part of fate#318196).
+- 4.0.179
+
+-------------------------------------------------------------------
 Fri May 11 10:08:24 UTC 2018 - jlopez@suse.com
 
 - Partitioner: allow to select only valid parity algorithms when
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -2,8 +2,9 @@
 Fri May 11 10:31:10 UTC 2018 - jlopez@suse.com
 
 - Partitioner: fix buttons to abort and to go back
-  (part of fate#318196).
-- 4.0.179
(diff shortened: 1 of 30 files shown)


src/lib/y2partitioner/device_graphs.rb
```diff
@@ -44,16 +44,31 @@ def create_instance(system, initial)
...
+    # Disk analyzer for the system graph
+    #
+    # @return [Y2Storage::DiskAnalyzer]
     attr_reader :disk_analyzer
 
     def initialize(system: nil, initial: nil)
       @system = system || storage_manager.probed
-      initial ||= storage_manager.staging
-      @current = initial.dup
+      @initial = initial || storage_manager.staging
```
jreidinger: you can have it in method itself, as it is evaluated lazy.

try it youtself in easy script:
```
class A
  def random
    rand(5)
  end

  def initialize(a: random)
    puts a
  end
end

A.new
A.new
A.new
A.new
```

joseivanlopez: Umm, I have realized that in this case we don't want only a default value, we want to use `storage` values even when `nil` is passed as value for `:system` and `:initial`.

jreidinger: then please document it as it is not much obvious.

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri May 11 10:31:10 UTC 2018 - jlopez@suse.com
+
+- Partitioner: fix buttons to abort and to go back
+  (part of fate#318196).
```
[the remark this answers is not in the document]
joseivanlopez: I don't guess so. It is incredible, ins't it?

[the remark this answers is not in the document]
ancorgs: On the other hand, it would be nice to mention here the fix done about reprobing and proposal calculation. It's not related to the buttons of the partitioner and it's quite buried in this PR.

PR #641, opened 2018-05-11
Install missing packages before comitting changes

imobachgs:
Fixes [bsc#1089508](https://bugzilla.suse.com/show_bug.cgi?id=1089508).


imobachgs:
If the user declines, the validation just fails and the user is still in the same screen. So he/she can modify the layout to, for instance, stop requiring that package.

PR #643, opened 2018-05-15
Fix detection of candidate disks

joseivanlopez:
PBI: https://trello.com/c/dqXhOfwl/379-2-sles15-p1-1091047-build-5881-openqa-test-fails-in-awaitinstall-cannot-umount-dev-vdb1

Manually tested.

### DIFF
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -30,6 +30,8 @@ Requires:	yast2-ruby-bindings >= 4.0.6
 Requires:	libstorage-ng-ruby >= 3.3.254
 # communicate with udisks
 Requires:	rubygem(ruby-dbus)
+# Y2Packager::Repository
+Requires:	yast2-packager >= 3.3.7
 
 BuildRequires:	update-desktop-files
 # function light_probe
@@ -47,6 +49,8 @@ BuildRequires:	rubygem(yast-rake)
 BuildRequires:	rubygem(rspec)
 # communicate with udisks
 BuildRequires:	rubygem(ruby-dbus)
+# Y2Packager::Repository
+BuildRequires:	yast2-packager >= 3.3.7
 PreReq:         %fillup_prereq
 
 Obsoletes:	yast2-storage
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:        4.0.181
+Version:        4.0.182
 Release:	0
 
 BuildRoot:	%{_tmppath}/%{name}-%{version}-build
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -49,8 +49,6 @@ BuildRequires:	rubygem(yast-rake)
 BuildRequires:	rubygem(rspec)
 # communicate with udisks
 BuildRequires:	rubygem(ruby-dbus)
-# Y2Packager::Repository
-BuildRequires:	yast2-packager >= 3.3.7
 PreReq:         %fillup_prereq
 
 Obsoletes:	yast2-storage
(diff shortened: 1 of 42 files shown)


package/yast2-storage-ng.spec
```diff
@@ -30,6 +30,8 @@ Requires:	yast2-ruby-bindings >= 4.0.6
 Requires:	libstorage-ng-ruby >= 3.3.254
 # communicate with udisks
 Requires:	rubygem(ruby-dbus)
+# Y2Packager::Repository
+Requires:	yast2-packager >= 3.3.7
```
ancorgs: Hmmm, yast2-packager depends on yast2-storage-ng. Is it safe/good to have such a cycle? Just asking.

jreidinger: in Requires it is not problem. You have to avoid it in BuildRequires

joseivanlopez: In fact, `Y2Packager::Repository` can be mocked during tests, so I will remove BuildRequires one.

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -254,105 +255,131 @@ def release_name(filesystem)
...
+      devices.prepend(device)
+    end
+
+    # Device names indicated in the URI of the installation repositories
+    #
+    # @see #local_repositories
+    #
+    # @return [Array<String>]
+    def repositories_devices
+      return @repositories_devices if @repositories_devices
```
ancorgs: Is this memoization worth it? Does it really save time or resources?

Just asking because I'm concerned about situations in which the user adds a new local repository (going back and forward in the installation process). Quite a corner case, I know.

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -254,105 +255,131 @@ def release_name(filesystem)
...
+
+    # Device names indicated in the URI of an installation repository
+    #
+    # For example:
+    #   "hd:/subdir?device=/dev/sda1&filesystem=reiserfs" => ["/dev/sda1"]
+    #   "dvd:/?devices=/dev/sda,/dev/sdb" => ["/dev/sda", "/dev/sdb"]
+    #
+    # @param repository [Y2Packager::Repository]
+    # @return [Array<String>]
+    def repository_devices(repository)
```
ancorgs: Not in this phase (so close to GMC1), but these are the kind of things that should be moved to the `Repository` class itself. There should be no need of knowing so much about the repository url structures in yast2-storage-ng or any other YaST module.

ancorgs: Maybe adding a TODO about it?

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -254,105 +255,131 @@ def release_name(filesystem)
...
-
-      disk.partitions.each do |partition|
-        if NO_INSTALLATION_IDS.include?(partition.id)
-          log.info "Skipping #{partition} (ID #{partition.id.inspect})"
-          next
-        elsif installation_volume?(partition)
-          return true
-        end
-      end
+      devicegraph.disk_devices.select { |d| candidate_disk?(d) }
```
ancorgs: So we are not longer discarding USB disks if a non-USB one is present? TBH, I never liked that idea much... but I don't feel we are in the right timeframe to change that behavior. It's not really related to the main topic (fixing detection of installation media).

joseivanlopez: Do you think it is dangerous? I was thinking about it, and I would expect to have all devices (usb or not) available. But anyway, if this change is too big, I can go back to the old behavior.

src/lib/y2storage/disk_analyzer.rb
joseivanlopez: As commented in IRC, this values are recalculated when going back and forward, due to the system is repobed and DiskAnalyzer is created again.

ancorgs: Actually there would be a corner case if the repositories can be added in a step that comes after the proposal (let's say the software selection that can be triggered in the summary screen) and then the user goes back to the guided setup (without reprobing).

But I admit that's really a wild corner case and the user deserves all the harm. :wink:

PR #644, opened 2018-05-15
don't reuse prep partitions larger than 8 MiB (bsc#1090019)

wfeldt:


doc/old_and_new_proposal.md
```diff
@@ -217,6 +217,7 @@ Besides these, there is another element:
   * `desired_size` *(disksize, default: `0 B`)*
   * `min_size` *(disksize, default: `0 B`)*
   * `max_size` *(disksize, default: `unlimited`)*
+  * `max_size_limit` *(disksize, default: `unlimited`)*
```
joseivanlopez: I guess it is not needed to make this value configurable via control file. It is used only for prep partitions, and that partitions are automatically proposed (not configurable in control file at all).

wfeldt:
You cannot simply use `max_size` for `max_size_limit`, it's really two different things. You'd have to raise `max_size` to `unlimited` for most of our partitions, then. And this would lead to strange proposals. I think it's best to keep them separate.

doc/old_and_new_proposal.md
wfeldt: It's currently in a state of limbo anyway - `control.rng ` would need to be adjusted. I'm fine with removing it from `FEATURES` in `VolumeSpecification` - I just thought it consistent to have all attributes configurable in this class.

src/lib/y2storage/boot_requirements_strategies/prep.rb
```diff
@@ -62,6 +58,11 @@ def warnings
...
     protected
 
+      # Maximum size firmware can handle
+      def max_prep_size
+        prep_volume.max_size_limit
+      end
+
       # PReP partition is needed, so return any warning related to it.
       def prep_warnings
         res = []
```
joseivanlopez: So, now, if a big prep exists, the detailed warning about that (`big_prep_warning`) is not used. Why to not switch `if` and `elsif` conditions?

doc/old_and_new_proposal.md
joseivanlopez: For me looks fine to have all attributes in `VolumeSpecification`, but not all volume specification attributes have to come from the control file. So, I would simply remove it from the documentation about the new settings.

src/lib/y2storage/boot_requirements_strategies/prep.rb
wfeldt: Sorry, does it relate to this code block?

joseivanlopez: This https://github.com/yast/yast-storage-ng/blob/4a77f3ce58d81d382d4b08418aef901b6b721dd7/src/lib/y2storage/boot_requirements_strategies/prep.rb#L71

I see you had to change a test due to the `if` conditions is now true when there is a big prep. As we have a specific message for that case, I would exchange `if` and `elsif` conditions.

PR #645, opened 2018-05-15
Proof of concept for bsc#1089274

ancorgs:
This is an early POC for https://bugzilla.suse.com/show_bug.cgi?id=1089274#c57 created for early review


ancorgs:
Yes, the class and the different methods I'm adding to its API (of course, adding only what we strictly need right now) are named in a generic way, so in the future we can have adjustments that are activations of a setting or that are not related to a concrete volume (e.g. changing resize_windows or something like that).

So far, the code is always about deactivating volume attributes, but the API is more agnostic.

src/lib/y2storage/dialogs/proposal.rb
```diff
@@ -125,24 +125,90 @@ def propose!
...
+        actions_source_html + _("<p>Changes to partitioning:</p>") + @actions_presenter.to_html
+      end
+
+      # @see #actions_html
+      def actions_source_html
+        return actions_source_for_partitioner unless proposal
+        return actions_source_for_guided_setup unless settings_adjustment
+        return actions_source_for_default_settings if settings_adjustment.empty?
+
+        para(_("Initial layout proposed after adjusting the Guided Setup settings:")) +
```
jreidinger: the wording is a bit strange to me. Why it mention "adjusting the Guided Setup". If I get it, user does not run guided setup ( because if so, then second return is in effect, right? ), so it should be more like "Initial layout proposed after adjusting the Product settings:"

src/lib/y2storage/dialogs/proposal.rb
```diff
@@ -125,24 +125,90 @@ def propose!
...
+      end
+
+      # @see #actions_source_html
+      def actions_source_for_guided_setup
+        para(_("Layout proposed by the Guided Setup with the settings provided by the user."))
+      end
+
+      # @see #actions_source_html
+      def actions_source_for_default_settings
+        para(_("Initial layout proposed with the default Guided Setup settings."))
```
jreidinger: same here. I believe here should be "default Product settings".

src/lib/y2storage/proposal/initial_strategies/ng.rb
```diff
@@ -55,56 +56,86 @@ def initial_proposal(settings: nil, devicegraph: nil, disk_analyzer: nil)
...
+        #
+        # @param proposal [GuidedProposal] initial proposal, its settings will
+        #   be adapted and reused (i.e. they will be modified)
+        # @param volume [VolumeSpecification] it must be one of the volumes from
+        #   the settings of the initial proposal
+        # @param attr [Symbol] name of the attribute to disable in the volume
+        # @param devicegraph [Devicegraph] see {#try_proposal}
+        # @param disk_analyzer [DiskAnalyzer] see {#try_proposal}
+        def retry_proposal(proposal, volume, attr, devicegraph, disk_analyzer)
+          volume.send(:"#{attr}=", false)
```
jreidinger: `public_send` cannot be used?

src/lib/y2storage/proposal/initial_strategies/ng.rb
```diff
@@ -55,56 +56,86 @@ def initial_proposal(settings: nil, devicegraph: nil, disk_analyzer: nil)
...
+          adjustment = proposal.auto_settings_adjustment.add_volume_attr(volume, attr, false)
```
jreidinger: NP: style note. Having three params and one of it boolean is really bad design decision. What false stands for? is it value for attr? why it is add?

For me more nature API would be `proposal.auto_settings_adjustment.attrs_for(volume)[attr] = false`

src/lib/y2storage/proposal/settings_adjustment.rb
```diff
@@ -0,0 +1,126 @@
...
+      # @param volume [VolumeSpecification] concrete volume that was configured
+      # @param attribute [Symbol] attribute of the volume that was adjusted
+      # @param value [Object] value set for the attribute
+      #
+      # @return [SettingsAdjustment]
+      def add_volume_attr(volume, attribute, value)
+        log.debug "Adjustments: add volume #{volume.mount_point} with #{attribute} == #{value}"
+        result = dup
+        result.volumes[volume.mount_point] ||= []
+        result.volumes[volume.mount_point] << attribute
```
jreidinger: where you use value? I see only adding attribute, but nowhere value. Confusing for me.

src/lib/y2storage/proposal/settings_adjustment.rb
```diff
@@ -0,0 +1,126 @@
...
+        log.debug "New adjustments: #{result.volumes}"
+        result
+      end
+
+      # User-oriented localized sentences explaining each aspect of the
+      # adjustment that have been performed on top of the default settings
+      #
+      # @return [Array<String>]
```
jreidinger: please mention that volumes is returned sorted or it is just implementation detail?

src/lib/y2storage/proposal/settings_adjustment.rb
```diff
@@ -0,0 +1,126 @@
...
+          _("do not propose a separate %s") % mount_point
+        elsif attrs.include?(:snapshots) && attrs.include?(:adjust_by_ram)
+          # TRANSLATORS: %s is a mount point like "/home"
+          _("disabled snapshots and RAM-based size adjustments for %s") % mount_point
+        elsif attrs.include?(:snapshots)
+          # TRANSLATORS: %s is a mount point like "/home"
+          _("disabled snapshots for %s") % mount_point
+        else
+          # TRANSLATORS: %s is a mount point like "/home"
+          _("do not adjust size of %s based on RAM size") % mount_point
```
jreidinger: WTF? Why we should base non-swap volume on RAM size?

src/lib/y2storage/proposal/settings_adjustment.rb
```diff
@@ -0,0 +1,126 @@
...
+      end
+
+      # @see #volume_description
+      def regular_vol_description(mount_point, attrs)
+        if attrs.include?(:proposed)
+          # TRANSLATORS: %s is a mount point like "/home"
+          _("do not propose a separate %s") % mount_point
+        elsif attrs.include?(:snapshots) && attrs.include?(:adjust_by_ram)
+          # TRANSLATORS: %s is a mount point like "/home"
+          _("disabled snapshots and RAM-based size adjustments for %s") % mount_point
```
jreidinger: same WTF as below. Does storage really change size of /home depending how much RAM system have???

src/lib/y2storage/proposal/settings_adjustment.rb
```diff
@@ -0,0 +1,126 @@
...
+        else
+          # TRANSLATORS: %s is a mount point like "/home"
+          _("do not adjust size of %s based on RAM size") % mount_point
+        end
+      end
+
+      # @see #volume_description
+      def swap_description(attrs)
+        if attrs.include?(:proposed)
+          _("do not propose a separate swap")
```
jreidinger: separate swap? what it is? System has swap or not . Of course user can create swap file on regular disk and loop mount it, but this sentence looks like we will create such file and add it to fstab, which I believe we don't do.

src/lib/y2storage/proposal/settings_adjustment.rb
```diff
@@ -0,0 +1,126 @@
...
+        else
+          _("do not enlarge swap to RAM size for suspend")
```
jreidinger: This is a bit hard to me to understand from user perspective. What about ""do not enlarge swap to RAM size (suspend may not work)"

src/lib/y2storage/proposal/settings_adjustment.rb
```diff
@@ -0,0 +1,126 @@
...
+      def regular_vol_description(mount_point, attrs)
+        if attrs.include?(:proposed)
+          # TRANSLATORS: %s is a mount point like "/home"
+          _("do not propose a separate %s") % mount_point
+        elsif attrs.include?(:snapshots) && attrs.include?(:adjust_by_ram)
+          # TRANSLATORS: %s is a mount point like "/home"
+          _("disabled snapshots and RAM-based size adjustments for %s") % mount_point
+        elsif attrs.include?(:snapshots)
+          # TRANSLATORS: %s is a mount point like "/home"
+          _("disabled snapshots for %s") % mount_point
```
jreidinger: it is a bit inconsistent with other messages. What about "do not propose snapshots for %s" ?

src/lib/y2storage/dialogs/proposal.rb
ancorgs: To give the users a hint so they know for sure than that initial proposal uses the same logic and settings they can reach by clicking on "Guided Setup".

src/lib/y2storage/proposal/initial_strategies/ng.rb
ancorgs: Sure. I can change it

src/lib/y2storage/dialogs/proposal.rb
jreidinger: But for me it looks like that installer think that user use guided setup. At least it is confusing for me.

src/lib/y2storage/proposal/initial_strategies/ng.rb
ancorgs: Well, strictly speaking from the POV from which the API is designed, the last argument is not necessarily a boolean. Although right now we only disable attributes, the method is meant to assign any value that makes sense for each attribute (attributes can be boolean, strings, arrays, DiskSize objects...).

Still you suggestion would apply, I know. But I'm not sure if I fully like it... specially for the kind of partial implementation we wanted to do at this point in time (so far we only allow settings attrs to false to keep stuff deadly simple in GMC phase).

src/lib/y2storage/proposal/settings_adjustment.rb
ancorgs: Explained in the documentation of the class above: "_Although the name of the class and methods are designed to support more scenarios in the future, the current implementation only supports disabling attributes in concrete volumes, since that's the only action that {GuidedProposal.initial} performs so far".

src/lib/y2storage/proposal/settings_adjustment.rb
ancorgs: I just wanted to ensure some kind of consistency since the strings will be displayed to the user, but the current sorting (by volume path) is actually just an arbitrary criterion. Nothing so relevant or well-though to be documented, IMHO.

src/lib/y2storage/proposal/settings_adjustment.rb
ancorgs: If configured to do so, it could be. We actually had a REAL request to make the root partition size adjustable based on the RAM size (for allocating kdump stuff). Fun, isn't it?

src/lib/y2storage/proposal/settings_adjustment.rb
ancorgs: Yep, read above

src/lib/y2storage/proposal/settings_adjustment.rb
jreidinger: then please verify that third argument is false and if not, raise ArgumentError or NotImplementedError . So at least it will be obvious in future if you use it with "true" or any other value.

src/lib/y2storage/dialogs/proposal.rb
ancorgs: And "Product settings" is not the right wording either. The settings are actually defined by the combination of product, module/extension, installation role and whatnot.

src/lib/y2storage/proposal/settings_adjustment.rb
ancorgs: I will document it here as well.

src/lib/y2storage/dialogs/proposal.rb
jreidinger: ok, so maybe "default settings" is better as it emphasize that it is default without any user modification, but with reduction needed by small disk.

src/lib/y2storage/proposal/settings_adjustment.rb
ancorgs: Well, actually the relationship between the size of swap and the possibilities for the system to suspend is a a complex topic. We have different levels/strategies of suspension right now. Thus, the exact wording and consequences of the related checkbox have been discussed several times... so I will change it to something even more conservative.

src/lib/y2storage/dialogs/proposal.rb
ancorgs: In other words, using "Guided Setup" was intentional here. As a way of telling the user than the initial proposal is just 100% equivalent to running the Guided Setup with the described settings.

src/lib/y2storage/proposal/settings_adjustment.rb
```diff
@@ -94,24 +96,28 @@ def regular_vol_description(mount_point, attrs)
...
+        elsif attrs == [:snapshots]
           # TRANSLATORS: %s is a mount point like "/home"
-          _("disabled snapshots for %s") % mount_point
-        else
+          _("do not enable snapshots for %s") % mount_point
+        elsif attrs == [:adjust_by_ram]
           # TRANSLATORS: %s is a mount point like "/home"
           _("do not adjust size of %s based on RAM size") % mount_point
+        else
+          raise "Unknown volume adjustment"
```
jreidinger: please include in exception also attrs, so we know what is there immediatelly. like "Unknown volume adjustment %{attrs.inspect}"

src/lib/y2storage/proposal/settings_adjustment.rb
```diff
@@ -94,24 +96,28 @@ def regular_vol_description(mount_point, attrs)
...
       # @see #volume_description
       def swap_description(attrs)
         if attrs.include?(:proposed)
-          _("do not propose a separate swap")
+          _("do not propose swap")
+        elsif attrs == [:adjust_by_ram]
+          _("do not enlarge swap to RAM size")
         else
-          _("do not enlarge swap to RAM size for suspend")
+          raise "Unknown adjustment for swap"
```
jreidinger: same here.

PR #649, opened 2018-05-30
WIP: Limit volume label length and permitted characters

shundhammer:
_**Superseded by https://github.com/yast/yast-storage-ng/pull/654**_

https://trello.com/c/pVvQMTRq/313-2-sles15-p3-1084867-installer-does-not-handle-xfs-label-length-limitations

This limits what a user can enter for volume labels:

- The length of the volume label
- The permitted characters

The length is provided by libstorage-ng directly with `BlkFilesystem::max_labelsize()`.

There is no such counterpart for the permitted characters, so this was added to `Filesystem::Type` on the yast-storage-ng Ruby level.

We could provide a different set of characters for each filesystem, and this is actually prepared in this PR, but so far it does what was requested in the PBI: Keep it simple.

We now restrict it to `A..Za..z0..9` and some very few special characters: `-_.` (minus, underscore, period) which all filesystem types appear to support. No weird non-ASCII characters since those volume labels have to work with system tools that all might have their specific restrictions. Some filesystem types may allow a little bit more, but that does not mean that we are required to let the user enter that here.

This PR uses widget properties of the underlying UI InputField widget that handles this directly, so there is no need for any error popups: The user simply cannot enter more than the permitted amount of characters, and he also cannot enter any characters that are not permitted.

### DIFF
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -256,6 +256,8 @@ module Y2Partitioner
 
       def init
         self.value = filesystem.label
+        Yast::UI.ChangeWidget(Id(widget_id), :ValidChars, valid_chars)
+        Yast::UI.ChangeWidget(Id(widget_id), :InputMaxLength, input_max_length)
       end
 
       # Validates uniqueness of the given label. The presence of the label is also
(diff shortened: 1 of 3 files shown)


ancorgs:
Don't we want this in the SLE-15-GA branch as well? I don't think there is a need to diverge branches/packages so quickly.

I mean, as long as we are talking only about bugfixes and not introducing anything disruptive, I would keep committing to SLE-15-GA (which does NOT create a maint. request automagically) and then merging to master from there. In that way we keep both code-streams (and versions) in sync.

In the future, if a maint. request is really required, it will include a synchronization of the SLE15 and SLE15SP1 packages. Which can only be good for us.

That's what we do with all the other YaST packages.

PR #652, opened 2018-06-04
Move partitions

joseivanlopez:
PBI: https://trello.com/c/CwguWHoI/174-5-partitioner-plain-partitioning-move

### DIFF
--- a/src/lib/y2partitioner/actions/move_partition.rb
+++ b/src/lib/y2partitioner/actions/move_partition.rb
@@ -1,0 +1,344 @@
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "yast/i18n"
+require "yast2/popup"
+require "y2partitioner/dialogs/partition_move"
+require "y2partitioner/device_graphs"
+require "y2partitioner/ui_state"
+
+module Y2Partitioner
+  module Actions
+    # Action for moving a partition
+    #
+    # To move a partition means to place it at the beginning or at the end of the
+    # previous or next adjacent free space, respectively.
+    class MovePartition
+      include Yast::I18n
+
+      # Constructor
+      #
+      # @param partition [Y2Storage::BlkDevice]
+      def initialize(partition)
+        textdomain "storage"
+
+        @partition = partition
+        UIState.instance.select_row(partition.sid)
+      end
+
+      # Checks whether it is possible to move the partition, and if so, the action is performed.
+      #
+      # @note An error popup is shown when the partition cannot be moved. In case the partition
+      #   can be moved, a dialog is presented to select whether to move forward or backward.
+      #
+      # @return [Symbol] :finish if the action is performed; :back or dialog result otherwise.
+      def run
+        return :back unless validate
+
+        answer = move_partition_question
+        return answer unless answer == :ok
+
+        move_partition
+        :finish
+      end
+
+    private
+
+      # @return [Y2Storage::BlkDevice]
+      attr_reader :partition
+
+      # @return [Symbol] :forward, :backward
+      attr_reader :movement
+
+      # Checks whether the partition can be moved
+      #
+      # @note When the partition cannot be moved, it shows a popup with the reason,
+      #   see {#error}.
+      #
+      # @return [Boolean] true if the partition can be moved; false otherwise.
+      def validate
+        return true unless error
+
+        Yast2::Popup.show(error, headline: :error)
+        false
+      end
+
+      # Why the partition cannot be moved
+      #
+      # @note A partition can be moved if is not an extended partition and it does
+      #   not exist on disk and there is adjancent free space where to move it. In
+      #   some cases, the move action could be called over a device that is not a
+      #   partition (i.e., when the selected element on a table is not a partition).
+      #   For that reason, the type of the device must be checked.
+      #
+      # @return [String, nil] nil when there is no error and the partition can be moved.
+      def error
+        device_type_error ||
+          extended_partition_error ||
+          existing_partition_error ||
+          no_space_error
+      end
+
+      # Error when a no partition device is tried to be moved
+      #
+      # @return [String, nil] nil the device is a partition
+      def device_type_error
+        return nil if partition.is?(:partition)
+
+        _("Hard disks, BIOS RAIDs and multipath\n" \
+          "devices cannot be moved.")
+      end
+
+      # Error when a extended partition is tried to be moved
+      #
+      # @return [String, nil] nil when the partition is not extended
+      def extended_partition_error
+        return nil unless partition.type.is?(:extended)
+
+        _("An extended partition cannot be moved.")
+      end
+
+      # Error when an existing partition is tried to be moved
+      #
+      # @return [String, nil] nil if the partition does not exist on disk
+      def existing_partition_error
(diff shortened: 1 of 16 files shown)


src/lib/y2partitioner/actions/move_partition.rb
```diff
@@ -0,0 +1,344 @@
...
+          no_space_error
+      end
+
+      # Error when a no partition device is tried to be moved
+      #
+      # @return [String, nil] nil the device is a partition
+      def device_type_error
+        return nil if partition.is?(:partition)
+
+        _("Hard disks, BIOS RAIDs and multipath\n" \
```
jreidinger: this looks like a bit insuficient to the above check. Can we write only that "Only partition can be moved."?

joseivanlopez: Well, I basically copied the text from the old partitioner.

src/lib/y2partitioner/actions/move_partition.rb
```diff
@@ -0,0 +1,344 @@
...
+      # Error when an existing partition is tried to be moved
+      #
+      # @return [String, nil] nil if the partition does not exist on disk
+      def existing_partition_error
+        return nil unless partition.exists_in_devicegraph?(DeviceGraphs.instance.system)
+
+        format(
+          # TRANSLATORS: error message where %{name} is replaced by a partition name (e.g. /dev/sda1)
+          _("The partition %{name} is already created on disk\n" \
+            "and cannot be moved."),
```
jreidinger: maybe mention that To move it, it have to be deleted and created again? So user knows what to do.

src/lib/y2partitioner/actions/move_partition.rb
```diff
@@ -0,0 +1,344 @@
...
+
+      # Previous free space, not necessary adjacent to the partition
+      #
+      # @note In case the partition to move is a primary partition, free spaces inside an extended
+      #   partition are not taken into account. Otherwise, when the partition is logical, only free
+      #   spaces inside the extended partition are considered.
+      #
+      # @return [Y2Storage::PartitionTables::PartitionSlot, nil] nil if there is no previous free space.
+      def previous_unused_slot
+        unused_slots(partition.type).select { |s| s.region.start < partition.region.start }.last
```
jreidinger: last is not needed, why not use find? like
```
  unused_slots(partition.type).reverse_each.find { |s| s.region.start < partition.region.start }
```

src/lib/y2partitioner/actions/move_partition.rb
```diff
@@ -0,0 +1,344 @@
...
+
+      # Next free space, not necessary adjacent to the partition
+      #
+      # @note In case the partition to move is a primary partition, free spaces inside an extended
+      #   partition are not taken into account. Otherwise, when the partition is logical, only free
+      #   spaces inside the extended partition are considered.
+      #
+      # @return [Y2Storage::PartitionTables::PartitionSlot, nil] nil if there is no next free space.
+      def next_unused_slot
+        unused_slots(partition.type).select { |s| s.region.start > partition.region.start }.first
```
jreidinger: same here, I expect find and not combination of select and first.

src/lib/y2partitioner/actions/move_partition.rb
```diff
@@ -0,0 +1,344 @@
...
+
+      # Partition placed before the partition to move
+      #
+      # @note In case the partition to move is a primary partition, logical partitions are not taken
+      #   into account. Otherwise, when the partition is logical, only other logical partitions are
+      #   considered.
+      #
+      # @return [Y2Storage::Partition, nil] nil if there is no partition before.
+      def previous_partition
+        partitions(partition.type).select { |p| p.region.start < partition.region.start }.last
```
jreidinger: same here

src/lib/y2partitioner/actions/move_partition.rb
```diff
@@ -0,0 +1,344 @@
...
+      # @see #moves_forward, #moves_backward
+      def move_partition
+        movement == :forward ? move_forward : move_backward
+      end
+
+      # Moves the patition by placing it at the beginning of the previous adjacent free space
+      def move_forward
+        return unless previous_adjacent_unused_slot
+
+        partition.region.start = previous_adjacent_unused_slot.region.start
```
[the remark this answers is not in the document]
joseivanlopez: It also confused me, but it is the way the old partitioner works. "Move forward" means to move thought the beginning of the disk and "move backwards" to the end.

src/lib/y2partitioner/actions/move_partition.rb
```diff
@@ -0,0 +1,344 @@
...
+
+      # Partition placed after the partition to move
+      #
+      # @note In case the partition to move is a primary partition, logical partitions are not taken
+      #   into account. Otherwise, when the partition is logical, only other logical partitions are
+      #   considered.
+      #
+      # @return [Y2Storage::Partition, nil] nil if there is no partition after.
+      def next_partition
+        partitions(partition.type).select { |p| p.region.start > partition.region.start }.first
```
jreidinger: same here. And btw maybe constant Proc can be used here for predicates, but just nice to have.

src/lib/y2partitioner/actions/move_partition.rb
joseivanlopez: Again, the text was simply copied. I also would like to improve that messages, but not sure if we want to do it now.

src/lib/y2partitioner/actions/move_partition.rb
```diff
@@ -0,0 +1,344 @@
...
+      end
+
+      # Error when there is no adjacent free space to move the partition
+      #
+      # @return [String, nil] nil when there is any adjacent free space
+      def no_space_error
+        return nil unless possible_movement.nil?
+
+        # TRANSLATORS: error message where %{name} is replaced by a partition name (e.g. /dev/sda1)
+        format(_("No space to move partition %{name}"), name: partition.name)
```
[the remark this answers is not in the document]
joseivanlopez: The same here.

src/lib/y2partitioner/dialogs/partition_move.rb
```diff
@@ -0,0 +1,125 @@
...
+          VBox(Label(format(_("Move partition %{name} backward?"), name: partition.name)))
+        when :forward
+          # TRANSLATORS: %{name} is replaced by a partition name (e.g. /dev/sda1)
+          VBox(Label(format(_("Move partition %{name} forward?"), name: partition.name)))
+        end
+      end
+
+    private
+
+      # Only shows ok and cancel buttons (help is excluded)
```
jreidinger: why help is excluded?

joseivanlopez: It is not shown in the old partitioner.

src/lib/y2partitioner/dialogs/partition_move.rb
```diff
@@ -0,0 +1,125 @@
...
+            ["backward", _("Backward")]
+          ]
+        end
+
+        def init
+          self.value = "forward"
+        end
+
+        def store
+          dialog.selected_movement = value.to_sym
```
jreidinger: hmm, why widget uses strings, just to later switch it to symbol? why not use symbol from the start?

src/lib/y2partitioner/actions/move_partition.rb
jreidinger: me neither, so please try to clarify it. As I think that it makes sense to give user some hint how to solve it.

src/lib/y2partitioner/dialogs/partition_move.rb
joseivanlopez: Because tests fail: `include_examples "CWM::RadioButtons"`

src/lib/y2partitioner/actions/move_partition.rb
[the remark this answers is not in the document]
jreidinger: so please at least document it in the code that it is forward to the start of disk and backward is towards to end of disk. I expect other readers will be also confused.

src/lib/y2partitioner/dialogs/partition_move.rb
jreidinger: ok, so please document it. ( and to be honest here for forward/backward help would be nice ).

src/lib/y2partitioner/dialogs/partition_move.rb
jreidinger: OK, so please document that CWM RadioButton expects strings as ids so in future it will be easy to see why.

src/lib/y2partitioner/actions/move_partition.rb
joseivanlopez: From the beginning, the first goal has been to mimic the old partitioner, so I would go with the messages just as the old partitioner. Anyway, I have added FIXME notes as you suggested.

PR #654, opened 2018-06-05
Limit volume label length and permitted characters

shundhammer:
_**This supersedes https://github.com/yast/yast-storage-ng/pull/649**_

https://trello.com/c/pVvQMTRq/313-2-sles15-p3-1084867-installer-does-not-handle-xfs-label-length-limitations

This limits what a user can enter for volume labels:

- The length of the volume label
- The permitted characters

The length is provided by libstorage-ng directly with `BlkFilesystem::max_labelsize()`.

There is no such counterpart for the permitted characters, so this was added to `Filesystem::Type` on the yast-storage-ng Ruby level.

We could provide a different set of characters for each filesystem, and this is actually prepared in this PR, but so far it does what was requested in the PBI: Keep it simple.

We now restrict it to `A..Za..z0..9` and some very few special characters: `-_.` (minus, underscore, period) which all filesystem types appear to support. No weird non-ASCII characters since those volume labels have to work with system tools that all might have their specific restrictions. Some filesystem types may allow a little bit more, but that does not mean that we are required to let the user enter that here.

This PR uses widget properties of the underlying UI InputField widget that handles this directly, so there is no need for any error popups: The user simply cannot enter more than the permitted amount of characters, and he also cannot enter any characters that are not permitted.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Tue Jun  5 10:24:23 UTC 2018 - shundhammer@suse.com
 
-- Partitioner: Handle limitations vor volume labels (bsc#1084867)
+- Partitioner: Handle limitations for volume labels (bsc#1084867)
 - 4.0.185
 
 -------------------------------------------------------------------
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jun  5 10:24:23 UTC 2018 - shundhammer@suse.com
+
+- Partitioner: Handle limitations vor volume labels (bsc#1084867)
+- 4.0.185
+
+-------------------------------------------------------------------
 Mon Jun  4 15:13:54 UTC 2018 - jlopez@suse.com
 
 - Partitioner: allow to move partitions (part of fate#318196).
--- a/test/y2storage/filesystems/type_test.rb
+++ b/test/y2storage/filesystems/type_test.rb
@@ -231,4 +231,27 @@ describe Y2Storage::Filesystems::Type do
       end
     end
   end
+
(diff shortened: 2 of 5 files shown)


test/y2storage/filesystems/type_test.rb
```diff
@@ -231,4 +231,27 @@
...
+    it "contains some non-alphanumeric characters" do
+      chars = Y2Storage::Filesystems::Type::EXT2.label_valid_chars
+      expect(chars).to include("-")
+      expect(chars).to include("_")
+      expect(chars).to include(".")
+    end
+
+    it "does not contain any whitespace" do
+      chars = Y2Storage::Filesystems::Type::EXT2.label_valid_chars
+      expect(chars).not_to include(" ")
```
jreidinger: you can in fact write `expect(chars).not_to match(/\s/)` and it will check that not whitespace is there.

jreidinger: just NP

test/y2storage/filesystems/type_test.rb
```diff
@@ -231,4 +231,27 @@
       end
     end
   end
+
+  describe "label_valid_chars" do
```
jreidinger: convention is to write "#label_valid_chars". "#" is used for instance methods and "." for class methods.

shundhammer: ok.

shundhammer:
Right.

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Jun  5 10:24:23 UTC 2018 - shundhammer@suse.com
+
+- Partitioner: Handle limitations vor volume labels (bsc#1084867)
```
ancorgs: typo: for

shundhammer: argh... fixing

PR #655, opened 2018-06-05
[WIP] [RFC] Allows to use a random encryption pwd when filesystem support it

dgdavid:
Hi there!

This PR tries to resolve the bug related to encryption of swap partitions [(see it at bugzilla#1088641)](https://bugzilla.suse.com/show_bug.cgi?id=1088641),  allowing user to set the encryption password as random when filesystem supports it.

I am opening it as a request for comments because I need your feedback, opinion, and hints to finish it and, at least in this case, it seems to me a better way that shooting questions in the IRC channel. I mean, at this point maybe async feedback is a better choice.

So, let me to leave right here some questions, doubts or alternatives that I thought for which I want to know your feedback about them.

* Regarding `#finish` method in Filesystem controller, I am not happy at all with nested conditionals. I tried to extract the creation of encryption to a separated method, but it looks a little weird to have the creation in a dedicated method and the deletion directly inside the conditional. Even more, this approach continues growing the length of controller, which has a `TODO` suggesting to split and reduce it.

  An alternative could be to move the logic inside `to_be_encrypted?` branch to a new `#set_encryption` method at [BlkDevice model](https://github.com/yast/yast-storage-ng/blob/7d909a43f606a257437fa34afbe24bb09b99d332/src/lib/y2storage/blk_device.rb#L32), but I am not sure if it should be within the scope of this PR or not. In addition, I must admit that I do not know if that could collide or overlap with the [`create_encryption` factory method](https://github.com/yast/yast-storage-ng/blob/7d909a43f606a257437fa34afbe24bb09b99d332/src/lib/y2storage/fake_device_factory.rb#L542) since there are few lines doing exactly the same. Thoughts?

* I simulated the API to add the path to the password file to the crypttab in `Y2Storage::Encryption` class, adding temporary `password_file` and `password_file=` methods. Here is where I think I need the @joseivanlopez's help, as was said in the IRC channel last Friday. Isn't it?

* I am not sure if breaking the help string in two different parts is a good idea or, instead, a problem for translators.

Remember that UI was discussed in the IRC channel last Friday, when I proposed both, one [using a checkbox](https://paste.opensuse.org/29804956) and [this with radio buttons]( https://paste.opensuse.org/14479832).

Thanks in advance :)

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/filesystem.rb
+++ b/src/lib/y2partitioner/actions/controllers/filesystem.rb
@@ -343,7 +343,7 @@ module Y2Partitioner
             enc = blk_device.create_encryption(name)
 
             if random_password
-              enc.password_file = '/dev/urandom' # FIXME
+              enc.password_file = "/dev/urandom" # FIXME
             else
               enc.password = encrypt_password
             end
--- a/src/lib/y2partitioner/actions/controllers/filesystem.rb
+++ b/src/lib/y2partitioner/actions/controllers/filesystem.rb
@@ -342,8 +342,11 @@ module Y2Partitioner
             name = Y2Storage::Encryption.dm_name_for(blk_device)
             enc = blk_device.create_encryption(name)
 
+            # FIXME: extract this to its own method. Also, pay attention to the hardcoded value for
+            # password_file, which for now is here only to test until a final decision about how to do it
+            # better will be taken.
             if random_password
-              enc.password_file = "/dev/urandom" # FIXME
+              enc.password_file = "/dev/urandom"
             else
               enc.password = encrypt_password
             end
--- a/src/lib/y2partitioner/actions/controllers/filesystem.rb
+++ b/src/lib/y2partitioner/actions/controllers/filesystem.rb
@@ -57,6 +57,9 @@ module Y2Partitioner
         # @return [String] Password for the encryption device
         attr_accessor :encrypt_password
 
+        # @return [Boolean] Whether the device should use a random password every boot
+        attr_accessor :random_password
+
         # Name of the plain device
         #
         # @see #blk_device
@@ -338,7 +341,12 @@ module Y2Partitioner
           if to_be_encrypted?
             name = Y2Storage::Encryption.dm_name_for(blk_device)
             enc = blk_device.create_encryption(name)
-            enc.password = encrypt_password
+
+            if random_password
+              enc.password_file = '/dev/urandom' # FIXME
+            else
+              enc.password = encrypt_password
+            end
           elsif blk_device.encrypted? && !encrypt
             blk_device.remove_encryption
           end
--- a/src/lib/y2partitioner/widgets/encrypt_password.rb
+++ b/src/lib/y2partitioner/widgets/encrypt_password.rb
@@ -99,9 +99,13 @@ module Y2Partitioner
           RadioButtonGroup(
             Id(:pwd_type),
             VBox(
-              Left(RadioButton(Id(:manual_pwd), _("Ask for password on boot to mount the swap"), true)),
+              Left(
+                RadioButton(Id(:manual_pwd), _("Ask for password on boot to mount the swap"), true)
+              ),
               password_fields,
-              Left(RadioButton(Id(:random_pwd), _("Auto-generate a new &random password on every boot"))),
+              Left(
+                RadioButton(Id(:random_pwd), _("Auto-generate a new &random password on every boot"))
+              )
             )
           )
         )
@@ -143,11 +147,11 @@ module Y2Partitioner
             "</p>\n"
           )
         else
-        _(
-          "<p>\n" \
-            "You will need to enter your encryption password.\n" \
-          "</p>\n"
-        )
+          _(
+            "<p>\n" \
+              "You will need to enter your encryption password.\n" \
+            "</p>\n"
+          )
         end
       end
 
--- a/src/lib/y2partitioner/widgets/encrypt_password.rb
+++ b/src/lib/y2partitioner/widgets/encrypt_password.rb
@@ -15,33 +15,25 @@ module Y2Partitioner
       end
 
       # @macro seeAbstractWidget
-      def help
-        # help text for cryptofs
-        _(
-          "<p>\n" \
-            "You will need to enter your encryption password.\n" \
-          "</p>\n" \
-          "<p>\n" \
-            "If the encrypted file system does not contain any system file and therefore is\n" \
-            "not needed for the update, you may select <b>Skip</b>. In this case, the\n" \
-            "file system is not accessed during update.\n" \
-          "</p>\n"
-        )
-      end
-
-      # @macro seeAbstractWidget
       def validate
-        msg = checker.error_msg(pw1, pw2)
-        return true unless msg
+        return true if random_password?
 
-        Yast::Report.Error(msg)
+        error_msg = checker.error_msg(pw1, pw2)
+        return true unless error_msg
+
+        Yast::Report.Error(error_msg)
         Yast::UI.SetFocus(Id(:pw1))
+
         false
       end
 
       # @macro seeAbstractWidget
       def store
-        @controller.encrypt_password = pw1
+        if random_password?
+          @controller.random_password = true
+        else
+          @controller.encrypt_password = pw1
+        end
       end
 
       # @macro seeAbstractWidget
@@ -56,27 +48,15 @@ module Y2Partitioner
           MarginBox(
             1.45,
             0.5,
-            VBox(
-              Password(
-                Id(:pw1),
-                Opt(:hstretch),
-                # Label: get password for user root
-                # Please use newline if label is longer than 40 characters
-                _("&Enter a Password for your File System:"),
-                ""
-              ),
-              Password(
-                Id(:pw2),
-                Opt(:hstretch),
-                # Label: get same password again for verification
-                # Please use newline if label is longer than 40 characters
-                _("Reenter the Password for &Verification:"),
-                ""
-              ),
-              VSpacing(0.5)
-            )
+            form_content
           )
         )
+      end
+
+      # @macro seeAbstractWidget
+      def help
+        # help text for cryptofs
+        help_encryption_password + help_skip
       end
 
     private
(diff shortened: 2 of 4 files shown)


src/lib/y2partitioner/actions/controllers/filesystem.rb
```diff
@@ -338,7 +341,12 @@ def finish
           if to_be_encrypted?
             name = Y2Storage::Encryption.dm_name_for(blk_device)
             enc = blk_device.create_encryption(name)
-            enc.password = encrypt_password
+
+            if random_password
+              enc.password_file = '/dev/urandom' # FIXME
```
jreidinger: FIXME what? without details it is not much useful

ancorgs:
> An alternative could be to move the logic inside to_be_encrypted? branch to a new #set_encryption method at BlkDevice model, but I am not sure if it should be within the scope of this PR or not.

That's up to you, but sounds like a good enhancement. :wink:

> In addition, I must admit that I do not know if that could collide or overlap with the create_encryption factory method since there are few lines doing exactly the same. Thoughts?

Which looks to me like another reason to add that method to `BlkDevice` or as a class method to `Encryption`. So that method can be used in your PR, in the mentioned `create_encryption` factory method and in `Planned::CanBeEncrypted`.

Let's say a `BlkDevice#encrypt` method that can receive either a password or a flag indicating we want a random password and additionally an optional dm_name (it will generate its own if omitted). Defining a good API for it would be quite challenging, but it could be then used in three different places.

src/lib/y2partitioner/widgets/encrypt_password.rb
```diff
@@ -14,34 +14,26 @@ def initialize(controller)
...
         false
       end
 
       # @macro seeAbstractWidget
       def store
-        @controller.encrypt_password = pw1
+        if random_password?
+          @controller.random_password = true
+        else
+          @controller.encrypt_password = pw1
```
jreidinger: do you test it if you are going back? I suspect that as random password and also encrypt_password is not reset, that it would still contain old value.

src/lib/y2partitioner/widgets/encrypt_password.rb
```diff
@@ -93,6 +73,94 @@ def pw1
       def pw2
         Yast::UI.QueryWidget(Id(:pw2), :Value)
       end
+
+      # @return [Boolean]
+      def random_password_allowed?
+        @random_password_allowed ||= @controller.filesystem_type.to_sym == :swap
```
jreidinger: please add explanation why random password is allowed only for swap. Ideally link to some documentation or short explanation with googable content.

src/lib/y2partitioner/widgets/encrypt_password.rb
```diff
@@ -93,6 +73,94 @@ def pw1
...
+      end
+
+      def extended_content
+        VBox(
+          RadioButtonGroup(
+            Id(:pwd_type),
+            VBox(
+              Left(RadioButton(Id(:manual_pwd), _("Ask for password on boot to mount the swap"), true)),
+              password_fields,
+              Left(RadioButton(Id(:random_pwd), _("Auto-generate a new &random password on every boot"))),
```
jreidinger: as I said above I expect this won't work well when going back as default is hardcoded here. Which I consider as future problem.

src/lib/y2partitioner/widgets/encrypt_password.rb
```diff
@@ -93,6 +73,94 @@ def pw1
...
+        end
+      end
+
+      def extended_content
+        VBox(
+          RadioButtonGroup(
+            Id(:pwd_type),
+            VBox(
+              Left(RadioButton(Id(:manual_pwd), _("Ask for password on boot to mount the swap"), true)),
+              password_fields,
```
jreidinger: well, I expect here can be more playing. How it does look in UI as those password fields is realted only to this radio button? I think that also enable/disable when you check/uncheck this radio button makes sense for me. And as this started to be more complicated I suggest to use here another object for this password fields that will be included in this widget as it starting to be more complicated and deserve some split.

I also expect that it more naturally allow better and more readable code in other parts which now heavilty mixture this password fields with other radio button selection.

src/lib/y2partitioner/widgets/encrypt_password.rb
```diff
@@ -93,6 +73,94 @@ def pw1
...
+          )
+        )
+      end
+
+      # @return [String]
+      def help_encryption_password
+        if random_password_allowed?
+          _(
+            "<p>\n" \
+              "You will need to enter your encryption password or allow to system\n" \
```
jreidinger: I probably write "allow system to auto-generate" ( see "to" position )

src/lib/y2partitioner/widgets/encrypt_password.rb
```diff
@@ -93,6 +73,94 @@ def pw1
...
+      end
+
+      # @return [String]
+      def help_skip
+        _(
+          "<p>\n" \
+            "If the encrypted file system does not contain any system file and therefore is\n" \
+            "not needed for the update, you may select <b>Skip</b>. In this case, the\n" \
+            "file system is not accessed during update.\n" \
+          "</p>\n"
```
jreidinger: to be honest now I am lost. This widget talk about setting encrypted password and now it mention some update. Relict of old help probably? But quite confusing for me.

jreidinger: I also do not see that skip option ( unless I overlook it ).

src/lib/y2storage/encryption.rb
```diff
@@ -79,6 +81,23 @@ def in_etc?
       in_etc_crypttab?
     end
 
+    # Returns the absolute path to a file containing the encryption password
+    #
+    # To be used as a third field in the crypttab file
+    #
+    # @return [String]
+    def password_file
+      @password_file
```
jreidinger: you mention that it returns string, but in fact it can also return nil. So allow to return nil in documentation or ensure that nil cannot be returned.

src/lib/y2partitioner/actions/controllers/filesystem.rb
dgdavid: Just a note to myself to not forget to review and ideally delete this literal from the code, because it is not clear to me yet that this will be the final way to set the random password for SWAP partition.

Anyway, I am going to remove it or add details, as appropriate, in next update.

Thanks.

src/lib/y2storage/encryption.rb
```diff
@@ -79,6 +81,23 @@ def in_etc?
...
+    end
+
+    # Sets the absolute path to a file containing the encryption password
+    #
+    # @param filename [String] absolute path to the file
```
jreidinger: can I somehow reset it if I do not want to use file? ( just hint, I know it, but you should document it ).

mvidner:
@dgdavid Thank you for the PR!
Other members of the team may already know this, but I'd like to know the context and references. Is this related to a Bugzilla ticket or a GH issue, or a mailing list post, or a forum post? If yes, please mention it in the PR description, and also in a commit message.

mvidner:
I can see how a random password makes sense for a swap partition: its contents are not needed after a shutdown, and can be reinitialized on boot-up.
But what if a clueless user chooses a random password for a data partition? Will we let them lose data?

src/lib/y2partitioner/widgets/encrypt_password.rb
mvidner: OK, the presence of this method answers my earlier question :)
@jreidinger : I don't think we need to link anything, the reason is simple: what happens when you let YaST use a random password for /home? (or any other non-swap FS). But yes, a short comment is needed.

mvidner: @dgdavid The `||=` method of memoization does not work well when the value is a Boolean. I think a `.nil?` is needed.

jreidinger:
@mvidner currently it is allowed only for swap partitions. But even for swap it can be problematic if suspending to that swap.

src/lib/y2partitioner/widgets/encrypt_password.rb
jreidinger: @mvidner well, still I would like to have written down this reason, as not everyone knows about this and there are other devices where it can make sense like e.g. bcache

src/lib/y2partitioner/widgets/encrypt_password.rb
dgdavid: To be honest, no, I did not test that scenario. But I just tested it now and what I see is that the described problem could not be reproduce, unless I am wrong.

 I guess that does not happen because this piece of code is located in the `#store` method, which seems to be the [second to last called in the widget's "workflow"](https://github.com/yast/yast-yast2/blob/e5231602858c6353d3802b09ff7a6b3d6e64255f/library/cwm/src/lib/cwm/abstract_widget.rb#L162). So, at the moment that user decides to go back, the `#store` method it has not been executed yet.

What do you think? Am I right?

jreidinger: @dgdavid well, store is called when you leaving this dialog. If there is more actions, then when going back from next dialog it will be loaded and stored again. So I am talking about scenario where you ahve this dialog lets call it A and then another following which have back button, lets call it B. 

And not try this scenario:
A -> B (back) -> A
and in this case I believe that 
A does not have previous values ( due to hardcoded button box )
A have problem in this store when going again to B as then controller will have both random password and encrypt password.

src/lib/y2partitioner/widgets/encrypt_password.rb
dgdavid: @mvidner,

Oops! Yes, you are right. _It's not good to be in a hurry_ :innocent: 

This is, probably, the first method that I wrote when started to play with the bug a few days ago. I think that since filesystem does not will change and right now we are considering the swap as the only filesystem that allows this kind of encryption method, it could be declared in the `#initialize` method. Isn't it?

My doubt is about the predicate method. Should I to add a method, use an alias_method or simply forget them and use directly the variable itself?

I mean

```ruby
...

def random_password_allowed?
  @random_password_allowed
end
```

or

```ruby

...

private

...
attr_reader :random_password_allowed
alias_method :random_password_allowed?, :random_password_allowed
...
```

or simply use the variable

```ruby
...

do_something if random_password_allowed
...
private

attr_reader :random_password_allowed
...

```

src/lib/y2partitioner/widgets/encrypt_password.rb
dgdavid: I agree. Password fields should be enabled/disabled according to the selected option. I forgot to add that question to the PR description since I did not investigated/discovered yet how you do that in YaST UI. Links are welcome :)

Regarding complexity, I feel the same but I need more time to figure out how to improve the code. That's the reason why I opened the PR as WIP, for early feedback / help.

So, let me a couple of days to improve it :)

src/lib/y2partitioner/widgets/encrypt_password.rb
dgdavid: > Relict of old help probably

I guess, because that already was there. I only added a clarification about random password option.

In fact, the subject of [commit which added](https://github.com/yast/yast-storage-ng/commit/49731c72f6ee5415ddbd095b1d517c397523137b) it is 

` Bulk copy of yast-partitioner files with no adaptation `

Shall I take the opportunity and remove it?

src/lib/y2storage/encryption.rb
dgdavid: Ok. I will pay greater attention to this when update it.

dgdavid:
Finally, it seems that this feature right now make no sense (@ancorgs, @aschnell  or @kobliha could bring more information about why). So, I am going to close the PR.

Thank you all for the review and comments.

kobliha:
Thank you @dgdavid for bringing this PR up. We are sorry that we can't do the library part now - and this PR doesn't make sense without the library. Anyway, it was a nice PR, there were good comments and good code. I've learned a lot myself. Sometimes we just don't finish what we start, so let's postpone this. Maybe you'll be able to reuse it in the future! :)

Thank you!

PR #656, opened 2018-06-05
Fix subvolume creation

joseivanlopez:
PBI: https://trello.com/c/KHSMmp19/392-sles15-p2p1-1067510-build3334storage-ng-partitioner-in-installed-system-add-snapshots-1-snapshot-prefix-to-the-subvolumes

### DIFF
--- a/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
+++ b/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
@@ -30,6 +30,8 @@ module Y2Partitioner
   module Dialogs
     # Popup dialog to create a btrfs subvolume
     class BtrfsSubvolume < Popup
+      include Yast::Logger
+
       attr_reader :filesystem
       attr_reader :form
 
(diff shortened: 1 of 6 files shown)


src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -178,18 +178,19 @@ def hierarchy_validation
...
+        # @see Y2Storage::Filesystems::Btrfs#subvolumes_prefix
         # @see Y2Storage::Filesystems::Btrfs#btrfs_subvolume_path
         def fix_path
           self.value = filesystem.canonical_subvolume_name(value)
           return if value.empty?
 
-          default_subvolume_path = filesystem.default_btrfs_subvolume.path
-          prefix = default_subvolume_path.empty? ? "" : default_subvolume_path + "/"
+          prefix = filesystem.subvolumes_prefix
+          prefix << "/" unless prefix.empty?
```
jreidinger: as there was problem, is possible to add logging like what is value and what is prefix?

PR #657, opened 2018-06-05
Added dumper mocking to not break parallel tests

lslezak:
- This is a preparation for the upcoming `parallel_tests` support, fixes a race condition when running several tests in parallel
- Actually fixed by Ancor during the YaST workshop
- The target is SLE15-GA to less diverge from master


ancorgs:
Maybe it's worth to merge this in SLE-15-GA, just for the sake of not diverging both code streams. This cannot break anything anyways.

PR #659, opened 2018-06-06
Allow to set crypttab names

joseivanlopez:
PBI: https://trello.com/c/OGssTRzO/393-osdistribution-p2p1-1094963-opensuse-150-installation-or-upgrade-fails-if-disk-contains-encrypted-partition-luks

* (done) It requires [this PR](https://github.com/openSUSE/libstorage-ng/pull/536) to pass tests.
* (done) It is also needed to update the dependency with libstorage-ng.


src/lib/y2storage/blk_device.rb
```diff
@@ -58,6 +58,10 @@ class BlkDevice < Device
     #   @raise [Storage::Exception] if the devicegraph provided as first
     #     argument is not the probed one
     #
+    #   @raise [DeviceNotFoundByName] if no device is found
```
ancorgs: I don't think this is true. The wrapped version of the method (the one that is described here) will never raise those two kinds of exceptions. It will return `nil`, as it was already documented.

src/lib/y2storage/crypttab.rb
```diff
@@ -0,0 +1,60 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2storage"
```
ancorgs: Why is "y2storage" required here? Specially weird since this file itself is required from `y2storage.rb`.

src/lib/y2storage/crypttab.rb
```diff
@@ -0,0 +1,60 @@
...
+    attr_reader :path
+
+    # Reads a crypttab file and returns its entries
+    #
+    # @return [Array<SimpleEtcCrypttabEntry>]
+    def read_entries
+      entries = Storage.read_simple_etc_crypttab(path)
+      entries.map { |e| SimpleEtcCrypttabEntry.new(e) }
+    rescue Storage::Exception
+      log.warn("Not possible to read the crypttab file: #{path}")
```
ancorgs: Just a warning?

src/lib/y2storage/crypttab.rb
ancorgs: I guess the correct thing here is `require "y2storage/simple_etc_crypttab_entry"`

src/lib/y2storage/simple_etc_crypttab_entry.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    # @!method password
+    #   @return [String]
+    storage_forward :password
+
+    # @!method crypt_options
+    #   @return [Array<String>]
+    storage_forward :crypt_options
+
+    # Device for the crypttab entry
```
ancorgs: I would be more explicit about the fact that this method returns the device on top of which the `Encryption` device is defined, not the `Encryption` block device itself. I understand it's a more useful approach (for example in cases in which the `Encryption` device is not there in the devicegraph), but it's kind of weird.

src/lib/y2storage/simple_etc_crypttab_entry.rb
```diff
@@ -0,0 +1,112 @@
...
+  class SimpleEtcCrypttabEntry
+    include StorageClassWrapper
+    wrap_class Storage::SimpleEtcCrypttabEntry
+
+    # @!method name
+    #   @return [String] name of the resulting encrypted block device
+    storage_forward :name
+
+    # @!method device
+    #   @return [String] path to the underlying block device
```
ancorgs: This is not necessarily the path of the underlying block device. This may also contain the UUID of the `Encryption` device itself.

PR #660, opened 2018-06-07
WIP PR to discuss DeviceMapper names and encrypt

ancorgs:
This PR fixes https://bugzilla.suse.com/show_bug.cgi?id=1094157 (although tests are not passing on Travis because it assumes that `Device#devicegraph` is public, something I adjusted locally).

In general, the bug was caused by the fact that we must select the dm_name of the `Encryption` device when creating it and YaST is expected to create DeviceMapper names that look nice and are related to the `Encryption#blk_device`. For example "cr_ata-XXX-part1" or "cr_system-home". But if the block device is a partition, its name can change several times, since partition numbers get reassigned during the partitioning process quite often and that may cause the name of the `BlkDevice` and the name of its `Encryption` to get out of sync (e.g. `/dev/mapper/cr_sda4` on top of `/dev/sda2`).

@aschnell @joseivanlopez So to improve this PR, I have several questions about management of DeviceMapper names for Encryption devices in the library.

This PR add some logic to adapt the name of the auto-generated names which looks like this (more or less):

```ruby
def update_dm_names(devicegraph)
  # Select only encryption devices with names auto-generated by YaST
  encryptions = Encryption.all(devicegraph).select(&:auto_dm_name?).sort_by(&:sid)

  # Reset all auto-generated names
  encryptions.each { |enc| enc.dm_table_name = "" }

  # And then reassign them according to the current names of the
  # block devices
  encryptions.each do |enc|
    enc.dm_table_name = Encryption.dm_name_for(enc.blk_device)
    # FIXME: synchronizing both values here feels wrong and hacky.
    # I would have expected #name to be updated based on #dm_table_name,
    # at least for non-active Luks
    enc.name = "/dev/mapper/#{enc.dm_table_name}"
end
```

The main current problem is explained in the FIXME. `#dm_table_name` and `#name` are initialized and synced in the constructor but only there.

There is a second aspect of the same problem. During installation libstorage-ng activates the existing LUKS devices with DeviceMapper names of the form "/dev/mapper/cr-auto-$NUM". That means that if the user decides to assign a mount point to an existing encrypted filesystem it will look like "cr-auto-1" in the final crypttab, instead of following the naming convention of the rest of YaST (described above).

I currently have this code to detect `Encryption` devices that can be adapted to the YaST convention:

```ruby
def auto_dm_name?
  result = userdata_value(:auto_dm_name)
  return result unless result.nil?

  # The name was not set via #encrypt or the Ruby setter,
  # so it comes most likely from probing. During a fresh installation
  # there is no /etc/crypttab so it's pretty safe to assume libstorage-ng
  # auto-proposed the name.
  Yast::Mode.installation && Yast::Stage.initial
end
```

Which means during installation we recognize those "cr-auto-X" names as auto-generated and thus we adapt them to the YaST convention whenever needed. But I can easily foresee problems with that because the current code of `.update_dm_names` modifies `Encryption#name` which I expect to be pretty dangerous for active LUKS devices.

So finally to the topic:

- Does it make sense to offer a method in the library to change `#dm_table_name` and `#name` at once (i.e. keeping them in sync)?
- How bad would that be (changing the device name) if the LUKS is active?
- What happens if `#dm_table_name` and `#name` are out of sync for an `Encryption` device? I can see that `#dm_table_name` is used to write the crypttab so changing that field looks like a possible way to get the desired final name even if the current `#name` in use is not synced, but I wonder about the side effects of getting both out of sync.


ancorgs:
Of course, one of the MANY possible solutions to the exposed problems would be:

- Make sure that `Encryption#set_dm_table_name` also syncs `Encryption#name`, so it can be easily and safely used to adjust the name of newly created (and thus, still not active) `Encryption` objects in staging.

- Offer some method to overwrite the name that will be written to crypttab (instead of always using `#dm_table_name`), so for objects that are probed and active we don't need to change the dangerous fields but just tell them to use a nicer name when adding themselves to crypttab. Let's say `Encryption#crypttab_name` that gets initialized by `#set_dm_table_name` but can be overridden afterwards if a probed encrypted partition is going to be reused.

aschnell:
Synchronise the name and dm_table_name for LUKSes should be possible. I do not see real problems here.

Changing the name for active LUKSes will likely cause trouble. After all the library has to be able to work with the device and thus needs the name. The kernel supports to rename active DM devices but that is not supported by libstorage-ng.

Using a different name in crypttab than the name of the active DM device can also cause trouble, see e.g. https://bugzilla.suse.com/show_bug.cgi?id=774247#c18 (maybe dracut behaves different than mkinitrd did but anyway it could cause trouble - needs more research).

ancorgs:
Ok, so let's go with point 1 of my suggested solution only (making sure that `#set_dm_table_name` keeps everything in sync) and let's leave point 2 out.

That would be enough to fix the reported bug and would be enough to remove the ugly code from #659.

The only problem is that reused encrypted filesystems (the second scenario in the PR description) will result in a not-so-nice crypttab entry. But that's already a very specific use-case and will not cause any real trouble (mainly cosmetic).

ancorgs:
We already reached a conclusion, so restarting the coding on #666.

PR #663, opened 2018-06-11
Run the unit tests in parallel

lslezak:
Speed up the unit tests by running them in parallel using the [parallel_tests](https://github.com/grosser/parallel_tests) gem.

- The local speed up is about `npcu`-times, i.e. with quad core CPU the speed up is about 4x (on my machine with HT enabled the speed up is ~4.4x)
- The Travis speed up is about 1.4x (4min15s -> 3min7s), the [Travis doc](https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System) says the worker has ~1.5CPU power. That's still quite a good improvement...
- This will speed up also build in OBS, the workers use `-smp 8` or even `-smp 16` KVM option, the speed up should be significant there
- This is SLE15-GA compatible change, it does not change anything in SLE15-GA, only the SLE15-SP1/Leap15.1 will be affected
- I have already submitted the required Ruby gems to Factory: [rubygem-parallel](https://build.opensuse.org/request/show/615280), [rubygem-parallel_tests](https://build.opensuse.org/request/show/615279), I'll do that for SLE15-SP1 as well...

## AI lslezak :smiley: 

- [x] Submit the gems to SP1
- [ ] Write a documentation how to use parallel_tests and parallel Travis

### DIFF
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -47,6 +47,10 @@ BuildRequires:	yast2 >= 4.0.73
 BuildRequires:	yast2-ruby-bindings >= 4.0.6
 BuildRequires:	rubygem(yast-rake)
 BuildRequires:	rubygem(rspec)
+# speed up the tests in SLE15-SP1+
+%if 0%{?sle_version} >= 150100
+BuildRequires:	rubygem(parallel_tests)
+%endif
 # communicate with udisks
 BuildRequires:	rubygem(ruby-dbus)
 PreReq:         %fillup_prereq
--- a/test/spec_helper.rb
+++ b/test/spec_helper.rb
@@ -53,8 +53,9 @@ if ENV["COVERAGE"]
   # track all ruby files under src
   SimpleCov.track_files("#{SRC_PATH}/lib/**/*.rb")
 
-  # use coveralls for on-line code coverage reporting at Travis CI
-  if ENV["TRAVIS"]
+  # use coveralls for on-line code coverage reporting at Travis CI,
+  # coverage in parallel tests is handled directly by the "test:unit" task
+  if ENV["TRAVIS"] && !ENV["PARALLEL_TEST_GROUPS"]
     require "coveralls"
     SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter[
       SimpleCov::Formatter::HTMLFormatter,
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,1 @@
+service_name: travis-ci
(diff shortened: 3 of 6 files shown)


package/yast2-storage-ng.spec
```diff
@@ -47,6 +47,7 @@ BuildRequires:	yast2 >= 4.0.73
 BuildRequires:	yast2-ruby-bindings >= 4.0.6
 BuildRequires:	rubygem(yast-rake)
 BuildRequires:	rubygem(rspec)
+BuildRequires:	rubygem(parallel_tests)
```
joseivanlopez: I see the only one reason to diverge here is the availability of this package. Would it be possible to avoid it? For example, adding some kind of check to ask whether the package exists or not. Because master and SLE-15-GA branches would be exact the same after this PR (and probably during some time), so it would be a pain to maintain the differences in version numbers when the code is the same.

test/spec_helper.rb
```diff
@@ -52,15 +52,6 @@ def require(path)
 
   # track all ruby files under src
   SimpleCov.track_files("#{SRC_PATH}/lib/**/*.rb")
-
```
joseivanlopez: Are we losing coveralls? Or is it still used?

lslezak: The coveralls needs to be moved to a different place when running parallel tests. The old approach sent the coverage at the end when collecting the coverage statistics. But that does not work properly when running the tests in parallel, each parallel process would submit a separate partial coverage only for the executed tests.

We need to send the coverage just once, after *all* parallel threads (processes) are finished.

The coverage is sent directly from `rake test:unit` after all parallel tests are finished. See
https://github.com/yast/yast-rake/blob/master/lib/tasks/test_unit.rake#L29-L35

For the backward compatibility when using the standard sequential RSpec the coverage is sent in the old way (from `spec/test_helper.rb`). It would be better to unify this, but then we would need to touch ~100 packages...

package/yast2-storage-ng.spec
lslezak: That's a good point! :+1: It should be possible to check the target version in the .spec file.

However, currently it seems to be a bit broken: [SLE15-SP1](https://build.suse.de/project/prjconf/SUSE:SLE-15-SP1:GA) defines `%sle_version 150100`, but [Leap 15.1](https://build.opensuse.org/project/prjconf/openSUSE:Leap:15.1) still has
```
%suse_version 1500
%is_opensuse 1
%sle_version 150000
```
which looks wrong to me...

lslezak: Reported in https://bugzilla.suse.com/show_bug.cgi?id=1097123...

PR #672, opened 2018-06-13
allow for numbers > 32 bit in region dialog (bsc#1065258)

wfeldt:
libyui's IntField widget works only with 32 bits. Since the major visible
advantage over a plain InputField are the controls allowing to
increase/decrease the value by one - which is not very helpful here - I
think it's fine to use an InputField here.

I've also changed the logic for the min/max values presented: it used to
show the first and last available blocks regardless whether these belong to
the same unused region.  This runs into the problem that in nearly all cases
you run into an error if you just confirm the defaults (for instance the min
value used to be _before_ the first partition).

Instead it now shows the default (largest) region (which is always a valid
range).

### DIFF
--- a/src/lib/y2partitioner/dialogs/partition_size.rb
+++ b/src/lib/y2partitioner/dialogs/partition_size.rb
@@ -267,36 +267,39 @@ module Y2Partitioner
 
         # @macro seeCustomWidget
         def contents
-          min_block = @regions.map(&:start).min
-          # FIXME: libyui widget overflow :-(
-          max_block = @regions.map(&:end).max
-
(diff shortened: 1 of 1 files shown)


ancorgs:
Version 4.0.190 means this should be submitted to the SLE-15-GA branch. So please, cherry-pick it there.

If your intention is to diverge and include this only in TW but not in the SLE-15 code stream. Then version should have been 4.1.0.

wfeldt:
The card ha an 'update' label.

wfeldt:
replaced by https://github.com/yast/yast-storage-ng/pull/673

PR #675, opened 2018-06-14
Fixed crash in the error callback when passing UTF-8 string from libstorage-ng (bsc#1096758)

lslezak:
The strings received from libstorage have ASCII-8BIT encoding which makes troubles when it contains UTF-8 characters. Processing such strings might result in the `Encoding::CompatibilityError` exception.

See https://bugzilla.suse.com/show_bug.cgi?id=1096758

### DIFF
--- a/test/y2storage/callbacks/callbacks_examples.rb
+++ b/test/y2storage/callbacks/callbacks_examples.rb
@@ -29,6 +29,19 @@ RSpec.shared_examples "general #error examples" do
     subject.error("the message", "the what")
   end
 
+  # SWIG returns ASCII-8BIT encoded strings even if they contain UTF-8 characters
+  # see https://sourceforge.net/p/swig/feature-requests/89/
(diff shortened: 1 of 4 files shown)


test/y2storage/callbacks/callbacks_examples.rb
```diff
@@ -29,6 +29,19 @@
...
 
+  # SWIG returns ASCII-8BIT encoded strings even if they contain UTF-8 characters
+  # see https://sourceforge.net/p/swig/feature-requests/89/
+  it "handles ASCII-8BIT encoded messages with UTF-8 characters" do
+    expect(Yast::Report).to receive(:yesno_popup) do |message, options|
+      expect(message).to include "🍺"
+      expect(options[:details]).to include "🍻"
+    end
+    subject.error(
+      "testing UTF-8 message: 🍺".force_encoding("ASCII-8BIT"),
```
ancorgs: Beers in the tests! I love it!

PR #677, opened 2018-06-14
Add format checkbox when importing mount points

joseivanlopez:
PBI: https://trello.com/c/S2PrGCQz/388-ostumbleweed-p5-1094924-partition-imported-during-installation-would-all-be-formatted-by-default

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/fstabs.rb
+++ b/src/lib/y2partitioner/actions/controllers/fstabs.rb
@@ -21,6 +21,7 @@
 
 require "yast"
 require "yast/i18n"
+require "y2storage/storage_manager"
 require "y2partitioner/device_graphs"
 require "y2storage"
 
@@ -35,6 +36,13 @@ module Y2Partitioner
 
         # @return [Y2Storage::Fstab] fstab file selected to import mount points
         attr_accessor :selected_fstab
+
+        # @return [Boolean] whether the system volumes must be formatted
+        #
+        # @note A volume is considered as "system volume" when it is mounted in certain
+        #   specific mount point like /, /usr, etc. See {#system_mount_points}.
+        attr_accessor :format_system_volumes
+        alias_method :format_system_volumes?, :format_system_volumes
 
         # Constructor
         def initialize
@@ -106,6 +114,11 @@ module Y2Partitioner
 
       private
 
+        # System mount points are taken from old code, see
+        # https://github.com/yast/yast-storage/blob/master_old/src/modules/FileSystems.rb#L438
+        SYSTEM_MOUNT_POINTS = ["/", "/usr", "/var", "/opt", "/boot"].freeze
+        private_constant :SYSTEM_MOUNT_POINTS
+
         # System devicegraph
         #
         # @return [Y2Storage::Devicegraph]
@@ -125,6 +138,13 @@ module Y2Partitioner
         # @return [Y2Storage::DiskAnalyzer]
         def disk_analyzer
           DeviceGraphs.instance.disk_analyzer
+        end
+
+        # Current system architecture
+        #
+        # @return [Storage::Arch]
+        def arch
+          @arch ||= Y2Storage::StorageManager.instance.arch
         end
 
         # Error when some entries in the selected fstab cannot be imported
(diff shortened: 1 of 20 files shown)


src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -104,6 +109,9 @@ def import_mount_points
 
       private
 
+        SYSTEM_VOLUMES_MOUNT_POINTS = ["/", "/usr", "/var", "/opt", "/boot", "/boot/zipl"].freeze
```
jreidinger: on which is this based? Please add comment explaining what it is.

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -173,12 +185,30 @@ def can_be_imported?(entry)
...
+          system_volume_mount_point?(mount_point) || !device.formatted?
+        end
+
+        def system_volume_mount_point?(mount_point)
+          system_volume_mount_points.include?(mount_point)
+        end
+
+        def system_volumes_mount_points
+          mount_points = SYSTEM_VOLUMES_MOUNT_POINTS.dup
+          mount_points.delete("/boot/zipl") unless arch.s390?
```
jreidinger: what about different approach and having arch specific array. which is concatenated for given architecture?

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -35,6 +36,10 @@ class Fstabs
         # @return [Y2Storage::Fstab] fstab file selected to import mount points
         attr_accessor :selected_fstab
 
+        # @return [Boolean] whether the system partition must be formatted
```
jreidinger: what is definition of system partition?

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -127,6 +140,13 @@ def disk_analyzer
           DeviceGraphs.instance.disk_analyzer
         end
 
+        # Current system architecture
+        #
+        # @return [Storage::Arch]
+        def arch
+          @arch ||= Y2Storage::StorageManager.instance.arch
+        end
```
ancorgs: Do we need something that `Yast::Arch` cannot provide? If not, I would use `Yast::Arch` (just like the rest of YaST and most places of y2-storage-ng). The StorageManager arch is only used in y2-storage-ng when `Arch#efiboot?` if needed. In all other places it uses `Yast::Arch`.

src/lib/y2storage/blk_device.rb
```diff
@@ -79,6 +79,12 @@ class BlkDevice < Device
     storage_forward :size, as: "DiskSize"
     storage_forward :size=
 
+    # @!method active?
+    #   Checks whether the device is active (LV, Encryption, etc)
```
ancorgs: It could be better documented what "active" means.

src/lib/y2storage/crypttab.rb
```diff
@@ -31,14 +31,19 @@ class Crypttab
     CRYPTTAB_PATH = "/etc/crypttab"
     private_constant :CRYPTTAB_PATH
 
+    # @return [Filesystems::Base]
```
ancorgs: NP: documentation. To know what `#filesystem` is really about I need to check the documentation of the constructor, which is not obvious from here.

I would document here what the attribute is about, not in the constructor.

test/y2storage/devicegraph_test.rb
```diff
@@ -391,19 +391,21 @@ def with_sda2_deleted(initial_graph)
 
       it "does not perform a system lookup" do
         expect(Y2Storage::BlkDevice).to_not receive(:find_by_any_name)
+        devicegraph.find_by_any_name("/dev/sda1")
```
ancorgs: :+1: Good catch about these broken tests!... which is kind of embarrassing, by the way

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -267,8 +330,56 @@ def setup_blk_filesystem(filesystem)
...
+          Y2Storage::Encryption.use_crypttab_names(fixed_devicegraph, crypttab)
+          fixed_devicegraph
+        end
+
+        # Selects the crypttab contained in the same filesystem than the selected fstab
+        #
+        # @return [Y2Storage::Crypttab, nil] nil if the filesystem does not contain
+        #   a crypttab file.
+        def crypttab
+          return @crypttab if @crypttab_found
```
ancorgs: Hmmm. Does this mean the crypttab is not refreshed even if `#selected_fstab` changes? Looks like a bug to me.

ancorgs: What about creating a unit test case for that corner case? Just in case some other developer feel tempted to cache that in the future. Only if it's easy enough, of course.

PR #682, opened 2018-06-22
- mask systemd mount and swap units (bsc#1073633)

aschnell:
For https://trello.com/c/kkmPuoxR/94-3-1073633-l3-system-auto-mounts-ext4-partition-on-xen-pv-device-after-fs-resize-causes-yast-error.


src/lib/y2storage/inhibitors/systemd_units.rb
```diff
@@ -0,0 +1,49 @@
...
+# find current contact information at www.suse.com.
+
+require "yast2/execute"
+
+module Y2Storage
+  # class to inhibit systemd mount and swap units
+  class SystemdUnits
+    include Yast::Logger
+
+    def inhibit
```
joseivanlopez: NP: This method (and `uninhibit`) could be written in a more compact way, but I am fine as it is now.

```
def inhibit
  log.info "mask systemd units"
  Yast::Execute.locally!("/usr/lib/YaST2/bin/mask-systemd-units", "--mask")
rescue Cheetah::ExecutionFailed => e
  log.error "masking systemd units failed #{e.message}"
end
```

aschnell: Fine, I will change it.

package/yast2-storage-ng.spec
```diff
@@ -88,6 +90,7 @@ rake install DESTDIR="%{buildroot}"
 %{yast_dir}/lib
 %{yast_desktopdir}/*.desktop
 %{yast_fillupdir}/*
+%{yast_ybindir}/mask-systemd-units
```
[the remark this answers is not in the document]
aschnell: The wildcards have to risk that file disappear without being noticed. But since this seems to be common I will change it.

PR #684, opened 2018-06-25
Bsc1073633

aschnell:
For https://trello.com/c/kkmPuoxR/94-3-1073633-l3-system-auto-mounts-ext4-partition-on-xen-pv-device-after-fs-resize-causes-yast-error.


src/bin/mask-systemd-units
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+
+# This script can mask or unmask systemd mount and swap units since systemd
+# and YaST mounting the same file system leads to race conditions (see bsc
+# #1073633). Unfortunately there is no single systemctl command to mask those
+# units - so each must be masks individually.
```
shundhammer: typo:  masked

PR #685, opened 2018-06-28
Added additional searchkeys to desktop file

schubi2:


src/desktop/disk.desktop
```diff
@@ -12,6 +12,7 @@ X-SuSE-YaST-AutoInst=
 X-SuSE-YaST-Geometry=
 X-SuSE-YaST-SortKey=
 X-SuSE-YaST-AutoInstResource=
+X-SuSE-YaST-Keywords=storage,disk,system
```
imobachgs: np: What about "partition"?

PR #688, opened 2018-07-18
AutoYaST: fix exporting of BIOS RAIDs (bsc#1098594)

ancorgs:
Second part of https://trello.com/c/BEBfAZTf/142-1-p2-1098594-support-for-bios-defined-md-in-autoyast

It also includes a commit to add a mock that was missing in the testsuite (it was trying to fiddle with udevadm).

### DIFF
--- a/src/lib/y2storage/autoinst_profile/drive_section.rb
+++ b/src/lib/y2storage/autoinst_profile/drive_section.rb
@@ -151,7 +151,7 @@ module Y2Storage
       #   <drive> section, like a disk, a DASD or an LVM volume group.
       # @return [Boolean] if attributes were successfully read; false otherwise.
       def init_from_device(device)
-        if device.is?(:md)
+        if device.is?(:software_raid)
           init_from_md(device)
         elsif device.is?(:lvm_vg)
           init_from_vg(device)
--- a/src/lib/y2storage/autoinst_profile/partitioning_section.rb
+++ b/src/lib/y2storage/autoinst_profile/partitioning_section.rb
@@ -83,7 +83,7 @@ module Y2Storage
       def self.new_from_storage(devicegraph)
         result = new
         # TODO: consider also NFS and TMPFS
-        devices = devicegraph.md_raids + devicegraph.lvm_vgs + devicegraph.disk_devices
+        devices = devicegraph.software_raids + devicegraph.lvm_vgs + devicegraph.disk_devices
         result.drives = devices.each_with_object([]) do |dev, array|
           drive = DriveSection.new_from_storage(dev)
           array << drive if drive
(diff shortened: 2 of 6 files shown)


ancorgs:
I just did a push -f for pleasing rubocop without adding a new commit. Nothing serious, just a couple of newlines.

src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -151,7 +151,7 @@ def self.new_from_storage(device)
       #   <drive> section, like a disk, a DASD or an LVM volume group.
       # @return [Boolean] if attributes were successfully read; false otherwise.
       def init_from_device(device)
-        if device.is?(:md)
+        if device.is?(:software_raid)
           init_from_md(device)
```
lslezak: NP: Is the `:md` symbol still valid actually? I mean was that a typo or the check was wrong?

What about renaming `init_from_md` to `init_from_software_raid` to be consistent? Or is the `init_from_md` name OK?

src/lib/y2storage/autoinst_profile/partitioning_section.rb
```diff
@@ -83,7 +83,7 @@ def self.new_from_hashes(drives_array)
       def self.new_from_storage(devicegraph)
         result = new
         # TODO: consider also NFS and TMPFS
-        devices = devicegraph.md_raids + devicegraph.lvm_vgs + devicegraph.disk_devices
+        devices = devicegraph.software_raids + devicegraph.lvm_vgs + devicegraph.disk_devices
```
lslezak: So the MD devices are not handled any more? Or are they covered by the `software_raids`?

shundhammer: This is covered by the class hierarchy AFAICS.

src/lib/y2storage/autoinst_profile/drive_section.rb
ancorgs: Yes, it's valid to check for `:md`. It will tell you if a device is an MD RAID. But many MD RAIDs are not defined by software. If an MD RAID is defined by BIOS, then we should export it as a regular disk, not in the format used in AutoYaST to define MDs.

Another alternative would have been:
```ruby
if device.is?(:md) && device.software_defined?
```

But I considered the other to be more concise.

ancorgs: And about `init_from_md`... I wanted to keep the changes to a minimum. After all, it will still generate an entry of type CT_MD in the profile. Unfortunately, the AutoYaST profile are far from being precise when using storage-related terms. In AutoYaST, MD RAID == software RAID. AutoYaST makes no difference. Even if in reality, the set of software RAIDS are a subset of all the MD RAIDs.

PR #693, opened 2018-07-23
document XEN guest setup for testing (bsc#1085134)

wfeldt:


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Jul 23 13:55:03 UTC 2018 - snwint@suse.com
+
+- document XEN guest setup for testing (bsc#1085134)
+- 4.0.199
```
jreidinger: does we need version bump when there is no functionality change, just documentation and just internal one?

wfeldt: It has even a bugzilla reference. This just keeps the SLE-15-GA structure; I see no need to deviate here.

PR #694, opened 2018-07-25
use "Partition Table" instead of "Disk Label" (bsc#1070570)

aschnell:
For https://trello.com/c/0ygAghdd/15-sles15-p3-1070570-yast-partitioner-names-the-partition-table-disk-label.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -3,7 +3,7 @@ Wed Jul 25 21:41:48 CEST 2018 - aschnell@suse.com
 
 - use "Partition Table" instead of "Disk Label" in expert
   partitioner (bsc#1070570)
-- 4.0.200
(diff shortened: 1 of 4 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jul 25 21:41:48 CEST 2018 - aschnell@suse.com
+
+- use "Partition Table" instead of "Disk Label" in expert
+  partitioner (bsc#1070570)
+- 4.0.200
```
ancorgs: If this change is only for master and not for SLE15SP0 (which makes sense, we should not change labels in MU with no good reason), then this is not following the versioning.
http://yastgithubio.readthedocs.io/en/latest/versioning/

PR #695, opened 2018-07-26
make bsc#1098594 regression test work on s390

wfeldt:
This fixes the bsc#1098594 regression unit test in [partitioning_section_test.rb](../tree/SLE-15-GA/test/y2storage/autoinst_profile/partitioning_section_test.rb#L105) as otherwise `drive.device` would be nil on s390.

### DIFF
--- a/src/lib/y2storage/autoinst_profile/drive_section.rb
+++ b/src/lib/y2storage/autoinst_profile/drive_section.rb
@@ -204,7 +204,7 @@ module Y2Storage
         return false if disk.partitions.empty?
 
         @type = :CT_DISK
-        # FIXME: could anyone with knowledge leave a comment why s390 is special here?
+        # s390 prefers udev by-path device names (bsc#591603)
         @device = Yast::Arch.s390 ? disk.udev_full_paths.first : disk.name
         # if disk.udev_full_paths.first is nil go for disk.name anyway
         @device ||= disk.name
--- a/src/lib/y2storage/autoinst_profile/drive_section.rb
+++ b/src/lib/y2storage/autoinst_profile/drive_section.rb
@@ -204,7 +204,10 @@ module Y2Storage
         return false if disk.partitions.empty?
 
         @type = :CT_DISK
+        # FIXME: could anyone with knowledge leave a comment why s390 is special here?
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -204,7 +204,10 @@ def init_from_disk(disk)
         return false if disk.partitions.empty?
 
         @type = :CT_DISK
+        # FIXME: could anyone with knowledge leave a comment why s390 is special here?
```
jreidinger: my wild guess is that it is caused that s390 disk names depends on order of activation so very very unstable.

src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -204,7 +204,10 @@ def init_from_disk(disk)
...
         @device = Yast::Arch.s390 ? disk.udev_full_paths.first : disk.name
+        # if disk.udev_full_paths.first is nil go for disk.name anyway
+        @device ||= disk.name
```
jreidinger: as it goes this complex maybe group it with:
```
@device = 
  if Yast::Arch.s390
    disk.udev_full_paths.first || disk.name
  else
    disk.name
  end
```

so fallback is used only in s390 case.

src/lib/y2storage/autoinst_profile/drive_section.rb
jreidinger: OK, found reason:
https://github.com/yast/yast-autoinstallation/commit/8d4fe6c6c8dc37b866d39f1f95b3156db7bbd8b3

so this bug:
https://bugzilla.suse.com/show_bug.cgi?id=591603

src/lib/y2storage/autoinst_profile/drive_section.rb
wfeldt: I've tried a few variants already and I had to go with one rubocop likes.

src/lib/y2storage/autoinst_profile/drive_section.rb
wfeldt: Yes, that's also what Arvin told me.

Still, I think not letting `@device` become `nil` is not a bad idea anyway.

PR #697, opened 2018-07-30
 Warning if overwriting manually edited settings (bsc#1055756)

schubi2:


test/y2storage/clients/inst_disk_proposal_test.rb
```diff
@@ -510,6 +510,10 @@
         it "aborts" do
           expect(client.run).to eq :abort
         end
+        it "recognize that the user has NOT changed settings" do
```
imobachgs: recognizes

test/y2storage/clients/inst_disk_proposal_test.rb
```diff
@@ -564,6 +568,15 @@
             .and_return(second_proposal_dialog)
           client.run
         end
+        it "recognize that the user has changed settings" do
```
imobachgs: recognizes

src/lib/y2storage/clients/inst_disk_proposal.rb
```diff
@@ -88,6 +91,19 @@ def run
       # @return [Integer]
       attr_reader :initial_staging_revision
 
+      # The user has changed partition settings.
+      # Asking if these changes can be overwritten.
+      def overwrite_manual_settings?
+        return true unless @manual_changed
```
imobachgs: If you are using `attr_reader :manual_changed`, I would use here the provided method (`manual_changed` instead of `@manual_changed`).

src/lib/y2storage/clients/inst_disk_proposal.rb
```diff
@@ -45,6 +45,8 @@ class InstDiskProposal
       include InstDialogMixin
       include PartitioningFeatures
 
+      attr_reader :manual_changed # Settings has been changed by user
```
ancorgs: There are already ways to ask the `StorageManager` if the staging devicegraph has been set manually. Adding another attribute for the same here doesn't sound like a good idea.

ancorgs: For example, in `Clients::PartitionProposal` we have this method which achieves the same without the need of introducing a new error-prone variable:

```ruby
def manual_partitioning?
  storage_manager.proposal.nil?
end
```

ancorgs: Moreover, using an instance variable in `Clients::InstDiskProposal` implies that if you go forward in the installer and back again to this screen the flag is lost. Isn't it?

I mean, if I understand the code, this will not raise the warning (although it should):

 * Reach the screen for the first time
 * Run the expert partitioner and setup the system
 * Click next to go to the next installer screen
 * Press back to return to the Storage Proposal screen
 * Press Guided Setup

ancorgs: The fix of the fix is here https://github.com/yast/yast-storage-ng/pull/705

Tested to work even after going back

PR #699, opened 2018-07-31
[WIP] Handle inactive RAIDs

mvidner:
- https://bugzilla.suse.com/show_bug.cgi?id=1090010
- https://trello.com/c/8kPdHfZR


ancorgs:
I know this is just a WIP, but changing how `StorageManager.probed` and `StorageManager.staging` behave definitely looks like a very deep change. Hopefully we don't need to do such a change just to add a new check.

mvidner:
The libstorage-ng part (WIP) https://github.com/yast/yast-storage-ng/pull/699

mvidner:
This should not be necessary given the fixes in libstorage-ng.

PR #704, opened 2018-08-08
Skip partition type dialog when going back if it needs to be skipped

teclator:
https://trello.com/c/ewWzf3Vk

Probably not the best solution, so just provided a quick fix in order to check the problem

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug 14 11:57:28 UTC 2018 - knut.anderssen@suse.com
+
+- Partitioner: Permit going back when the partition dialog is
+  skipped (bsc#1075443)
+- 4.1.5
+
+-------------------------------------------------------------------
 Mon Aug 13 14:44:32 UTC 2018 - ancor@suse.com
 
 - Partitioner: fixed some strings that contained mistakes about
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:	4.1.4
+Version:	4.1.5
 Release:	0
 
 BuildRoot:	%{_tmppath}/%{name}-%{version}-build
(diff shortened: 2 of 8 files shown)


joseivanlopez:
Ancor has refreshed my why this is failing. In the wizard there is a ["type" step](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/actions/add_partition.rb#L49) where the partition type (primary, logical, extended) is selected, but this step simply returns :next when there is no option to several partition types (e.g., only a logical can be created). So going back to this "type" step always returns :next in that case. What we have to do is simply detect if we are going back and in that case returns :back instead of :next.

I would try something like:

```
def type
  return :back if Yast::GetInstArgs.going_back
  Dialogs::PartitionType.run(part_controller)
end
```

I don't know if `GetInstArgs` is the correct way to detect if we are going back in that case, but at least that is the idea.

teclator:
Exactly, that is the reason of the failure, and that was exactly my first attempt but this is a "special wizard" or Sequencer 

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/actions/transaction_wizard.rb 

It defines a method for skipping methods when going back

https://github.com/yast/yast-yast2/blob/master/library/sequencer/src/lib/ui/sequence.rb#L151

So basically the knowledge about if it was skipped or not is known by the Dialog itself, and it doesn't know if we are going back or not.

That is why I fixed in this way, but maybe it is not the best.

teclator:
After been discussing it on Jangouts during a while, we finally agree that the problem is that the knowledge about the available partitions types is in the widget itself, and the widget is who knows if the dialog should be skipped or not. That logic should be moved away from the widget and the widget or the dualog should only need to be run in case that there are more than one option to choose.

jreidinger:
BTW just note, if there is no type selection, you can return `:auto` if it is for product control. Maybe something similar we can implement for wizard and sequencer:

https://github.com/yast/yast-yast2/blob/253f7f7435950eb33541722524a7b024c6c4ffae/library/control/src/modules/ProductControl.rb#L1497

PR #705, opened 2018-08-10
Fix the check about a manual staging (bsc#1055756)

ancorgs:
The old fix was not working if the user went back and forth in the installer because the status was tracked in an extra instance variable, instead of checking the already existing variables that are always properly initialized.

See https://github.com/yast/yast-storage-ng/pull/697#discussion_r207894154 for details.


ancorgs:
Manually tested. But after updating the PR with some minor stuff, Travis looks kind of broken.

PR #709, opened 2018-08-14
Set mount_by option when reusing devices

imobachgs:


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Aug 14 11:43:19 UTC 2018 - igonzalezsosa@suse.com
```
ancorgs: Changelog out of order?

imobachgs: Yes, fixed now. Thanks!

PR #710, opened 2018-08-16
Prevent nil:NilClass crash in Kubic if disk too small

shundhammer:
### Trello

https://trello.com/c/1K6xCnFa/

### Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1099762

# Problem

That product's workflow doesn't use the usual storage proposal as a separate workflow step before the installation summary proposal, it uses the older client which is called at the installation summary proposal together with all the other proposals (software, bootloader, ...).

That older proposal client has largely fell out of use; it is otherwise only used for AutoYaST which has a bit different preconditions.

When it could not make any storage proposal because the disk was simply too small, it caused a crash because it tried to assign a `nil` proposal to the storage manager which gladly accepted that, but crashed when it tried to deep-copy it.

# Fix

- Added unit test with a small (5 GiB) disk to check for this error scenario

- Safeguard that deep copy (`safe_copy()`) against nil and complain in the log, but don't crash

- Properly set the proposal result to _failed_ with the usual proposal mechanics and return a _blocker_ status to the calling proposal frame

- Correctly order the proposal steps in the `control.xml` file of that product (_separate PR!_) so the storage proposal is called first, not the bootloader proposal (which already caused a blocker because there was no root filesystem, of course, but the user could not continue from there)

# Screenshots

Error case: Only 5 GiB disk (10 GiB minimum required)

kubic-4gb-disk-no-proposal-01

kubic-4gb-disk-no-proposal-02-expert-partitioner

# Limitations

Since the cause of the problem could be many things, we can't simply tell the user _"insufficient disk space"_. This would be desirable, but this will probably be very confusing in many other scenarios where we also can't make a storage proposal, but the cause is not simply a too small disk.

### DIFF
--- a/src/lib/y2storage/clients/partitions_proposal.rb
+++ b/src/lib/y2storage/clients/partitions_proposal.rb
@@ -92,6 +92,10 @@ module Y2Storage
         }
       end
 
+      def failed?
+        @failed
+      end
+
     private
 
       attr_reader :failed
@@ -163,13 +167,22 @@ module Y2Storage
 
       def ensure_proposed
         if storage_manager.proposal.nil?
-          storage_manager.proposal = GuidedProposal.initial
+          guided_proposal
         elsif !storage_manager.proposal.proposed?
           storage_manager.proposal.propose
         end
       rescue Y2Storage::Error
         @failed = true
         log.error("generating proposal failed")
+      end
+
+      # Make a guided proposal.
+      # @raise [Y2Storage::Error] if proposal failed
+      def guided_proposal
+        proposal = GuidedProposal.initial
+        raise Y2Storage::Error, "Guided proposal failed" if proposal.failed?
+        @failed = false
+        storage_manager.proposal = proposal
       end
 
       def expert_partitioner
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Aug 20 15:26:02 UTC 2018 - shundhammer@suse.com
+
+- Fixed crash in the Kubic proposal when insufficient disk space
+  (bsc#1099762)
+- 4.0.207
+
+-------------------------------------------------------------------
 Fri Aug 17 08:47:53 UTC 2018 - ancor@suse.com
 
 - AutoYaST: recognize Xen virtual partitions in the profile when
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:        4.0.206
+Version:        4.0.207
 Release:	0
 
 BuildRoot:	%{_tmppath}/%{name}-%{version}-build
--- a/test/data/devicegraphs/empty_hard_disk_128MiB.yml
+++ b/test/data/devicegraphs/empty_hard_disk_128MiB.yml
@@ -1,0 +1,4 @@
+---
+- disk:
+    name: /dev/sda
+    size: 128 MiB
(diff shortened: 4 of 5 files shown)


src/lib/y2storage/storage_manager.rb
```diff
@@ -387,6 +389,10 @@ def apply_storage_defaults
     #
     # @param [Devicegraph] devicegraph to copy
     def copy_to_staging(devicegraph)
+      if devicegraph.nil?
+        log.error("Not copying nil devicegraph to staging")
+        return
```
ancorgs: I have not gone through the whole PR yet (tomorrow maybe), but this looks to me like a way to silent a rather serious error... so it can shoot us by the back when less expected.

If someones tries to assign a nil devicegraph we used to get an exception, that's how the bug was found... and the cause is very clear. We can easily know that some particular piece of code failed to calculate a devicegraph and it didn't handle that case.

But if we simply do nothing when nil is assigned to staging and we simply keep the previous outdated staging devicegraph, then we are producing a rather subtle bug. That can result in a completely different error at another future point of the installation... quite difficult to track and maybe even unnoticed for quite some time.

You can say "the show must go on"... but I honestly don't want the show to go in with the wrong staging devicegraph. :see_no_evil:

shundhammer: I can do that check also on the application level, but I can almost guarantee you that we will get a very similar backtrace very soon. This code is much too fragile IMHO. And for the user, this means the installation stops right here and now.

shundhammer:
Now explicitly setting staging to nil if receiving nil.

I thought this could be handled entirely on the application level, but the problem cascades over all the levels of the application. The storage manager can handle a nil staging devicegraph quite well, I found, since it is initially nil anyway.

shundhammer:
BTW that test suite issue was a problem of another test overwriting the proposal settings via the product features (control.xml), and that other product would accept a 5 GiB disk just fine. My pragmatic approach was to use a 128 MiB disk to force a failure here rather than loading another control.xml at greath lengths.

ancorgs:
Setting staging to nil in that case is not what I meant at all. I wanted to keep things as they were.

If doing any change there, I was thinking more in the direction of raising an `ArgumentError` if nil is passed to `#copy_to_staging`.

ancorgs:
Although it's true that staging can be nil before libstorage is initialized, is only expected to remain nil up to the first probing. I would say that as soon as a first probing happens (even a simulated one), staging is never expected to be nil. So allowing to set it to nil sounds like a recipe for disaster.

shundhammer:
Now using a much more conservative and less destructive approach.

shundhammer:
Tested again with the latest Kubic ISO and a 4 GB disk. Works as expected.

src/lib/y2storage/clients/partitions_proposal.rb
```diff
@@ -92,6 +92,10 @@ def description
         }
       end
 
+      def failed?
+        @failed
+      end
```
ancorgs: Is this new method used somewhere?

shundhammer: In the unit tests for sure; this is the only way that information is accessible.

ancorgs: Well, if you have to extend the API of a class just in order to be able to write its unit tests, that usually means the unit tests are not properly designed. They should test the already existing API, without the need to add extra methods just to lurk into the internal status of the objects.

But let's not start a useless discussion about testing theory. The code itself looks good. So let's move on. :+1:

shundhammer: Frankly, I would have exported that information from the very start; it might be of interest to the caller. So, for me this kind of thing definitely belongs to an API, no matter if the current caller(s) use it or not.

ancorgs: Well, this is an `Installation::ProposalClient`, so the API is strictly (pre)defined. All proposal clients must adhere to the corresponding API - a single `#run` method that responds to `MakeProposal`, `AskUser`, `Description` and `Write`.

No room for innovation or extra stuff there, since the caller is always the proposal mechanism, which iterates through all the proposal clients using that generic API and making no difference between this proposal client and any other.

PR #717, opened 2018-08-23
RAID attributes: include "Active: Yes/No"

mvidner:
Companion to https://github.com/openSUSE/libstorage-ng/pull/554
for https://trello.com/c/8kPdHfZR

### DIFF
--- a/src/lib/y2partitioner/widgets/md_description.rb
+++ b/src/lib/y2partitioner/widgets/md_description.rb
@@ -57,10 +57,20 @@ module Y2Partitioner
       # @return [Array<String>]
       def raid_attributes
         [
+          raid_active,
           raid_type,
           raid_chunk_size,
           raid_parity
         ]
+      end
+
+      # Information about MD RAID being (in)active
+      #
+      # @return [String]
+      def raid_active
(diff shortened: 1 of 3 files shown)


src/lib/y2partitioner/widgets/md_description.rb
```diff
@@ -57,12 +57,22 @@ def raid_description
...
         ]
       end
 
+      # Information about MD RAID being (in)active
+      #
+      # @return [String]
+      def raid_active
+        # TRANSLATORS: RAID being active (assembled), where %s is replaced by
+        # 'Yes' when the device is active or by 'No' otherwise
+        format(_("Active: %s"), blk_device.active? ? _("Yes") : _("No"))
```
ancorgs: Maybe it's more translation-friendly to have the strings `"Active: Yes"` and `"Active: No"`

mvidner: Well, mayybe, but this way it is consistent with "Encrypted: Yes/No" which we already have.

ancorgs: Ah. Ok.

PR #721, opened 2018-08-28
Merge two PRs to master (one of them fixing the build failure)

ancorgs:
Merge #717 and #720

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,15 @@
 -------------------------------------------------------------------
+Tue Aug 28 13:27:53 UTC 2018 - ancor@suse.com
+
+- Fixed a wrong unit test.
+- 4.1.10
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,14 @@
+-------------------------------------------------------------------
+Tue Aug 28 13:27:53 UTC 2018 - ancor@suse.com
+
+- Fixed a wrong unit test.
+- 4.1.10
```
joseivanlopez: NP: I would add the version number comment in the previous entry (to keep it as similar as possible to the original commit).

ancorgs: But we must be consistent with what we send to the build service. And the version 4.1.10 is the one that include both things. It's not true that there was a 4.1.10 version only with @mvidner's improvement and without the test suite fix.

joseivanlopez: @ancorgs ah yes, looking at it like that, I agree :)

PR #723, opened 2018-08-29
Irst

jreidinger:
how it looks like:

snimek obrazovky_2018-08-29_16-30-38

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Aug 29 14:30:59 UTC 2018 - jreidinger@suse.com
+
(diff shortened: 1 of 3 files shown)


jreidinger:
@ancorgs yes, there is a test for it - https://github.com/yast/yast-storage-ng/blob/master/test/y2storage/partition_id_test.rb#L111

but problem is that we now handle all unknown partitions, so it always return string, so this test basically is now useless - https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/partition_id.rb#L65

jreidinger:
@ancorgs so this commit basically makes test useless https://github.com/yast/yast-storage-ng/commit/4ed49be8816237b6c68d9375c8ed489ba5d605cf

PR #725, opened 2018-08-30
Try to unmount devices for some actions

joseivanlopez:
PBI: https://trello.com/c/e1TpNp2Y/213-5-partitioner-handle-on-the-fly-unmounting-in-installed-system

Requires https://github.com/yast/yast-yast2/pull/817

NOTE

This PR is intentionally not merged yet. These changes should not be included as Self Update, so we are going to merge [this L3](https://trello.com/c/1icYnFLW/266-l3-sles15-1107298-autoyast-disk-directly-as-pv-doesnt-work) before. Maybe, if we are lucky, we don't get more L3s after merging the code of this PR and so it would not be included in the Self Update.

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/blk_device.rb
+++ b/src/lib/y2partitioner/actions/controllers/blk_device.rb
@@ -28,6 +28,10 @@ module Y2Partitioner
     module Controllers
       # This class offers helper methods to perform actions over a block device,
       # for example, to check if its filesystem exists on disk, it is mounted, etc.
+      #
+      # @note Several methods in this class use the term "committed" to refer to the
+      #   device in the system. For example, committed_device would be the real device
+      #   as it is currently in the system (its version in memory could be different).
       class BlkDevice
         # Block device
         #
@@ -41,7 +45,8 @@ module Y2Partitioner
         # @param device [Y2Storage::BlkDevice] it has to be a block device,
         #   (i.e., device#is?(:blk_device) #=> true).
         def initialize(device)
-          raise(TypeError, "param device has to be a block device") unless device.is?(:blk_device)
+          wrong_param = !device.respond_to?(:is?) || !device.is?(:blk_device)
+          raise(TypeError, "param device has to be a block device") if wrong_param
 
           @device = device
         end
@@ -111,7 +116,7 @@ module Y2Partitioner
         #   it does not support mounted shrink.
         #
         # @return [Boolean]
-        def need_unmount_for_shrinking?
+        def unmount_for_shrinking?
           return false unless mounted_committed_filesystem?
 
           !committed_filesystem.supports_mounted_shrink?
@@ -123,7 +128,7 @@ module Y2Partitioner
         #   it does not support mounted grow.
         #
         # @return [Boolean]
-        def need_unmount_for_growing?
+        def unmount_for_growing?
           return false unless mounted_committed_filesystem?
 
           !committed_filesystem.supports_mounted_grow?
--- a/src/lib/y2partitioner/actions/controllers/blk_device.rb
+++ b/src/lib/y2partitioner/actions/controllers/blk_device.rb
@@ -1,0 +1,147 @@
+# encoding: utf-8
+
+# Copyright (c) [2018] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2storage"
+require "y2partitioner/device_graphs"
+
+module Y2Partitioner
+  module Actions
+    module Controllers
+      # This class offers helper methods to perform actions over a block device,
+      # for example, to check if its filesystem exists on disk, it is mounted, etc.
+      class BlkDevice
+        # Block device
+        #
+        # @return [Y2Storage::BlkDevice]
+        attr_reader :device
+
+        # Constructor
+        #
+        # @raise [TypeError] if device is not a block device.
+        #
+        # @param device [Y2Storage::BlkDevice] it has to be a block device,
+        #   (i.e., device#is?(:blk_device) #=> true).
+        def initialize(device)
+          raise(TypeError, "param device has to be a block device") unless device.is?(:blk_device)
+
+          @device = device
+        end
+
+        # Whether the current filesystem of the device exists on the system
+        #
+        # @note The device could have a new filesystem that does not match to the real filesystem
+        #   that actually exists on disk (e.g., after formatting the device).
+        #
+        # @return [Boolean]
+        def committed_current_filesystem?
+          return false unless device.formatted?
+
+          committed_filesystem? &&
+            committed_filesystem.sid == device.filesystem.sid
+        end
+
+        # Whether the filesystem of the device exists on the system and it is mounted
+        #
+        # @note The filesystem on system is checked, independently of it matches to the current
+        #   filesystem of the device.
+        #
+        # @return [Boolean]
+        def mounted_committed_filesystem?
+          return false unless committed_filesystem?
+
+          mount_point = committed_filesystem.mount_point
+
+          !mount_point.nil? && mount_point.active?
+        end
+
+        # Whether the device exists on the system
+        #
+        # @return [Boolean] true if the device exists on disk; false otherwise.
+        def committed_device?
+          !committed_device.nil?
+        end
+
+        # Device taken from the system devicegraph
+        #
+        # @return [Y2Storage::BlkDevice, nil] nil if the device does not exist on disk yet.
+        def committed_device
+          @committed_device ||= system_devicegraph.find_device(device.sid)
+        end
+
+        # Whether the committed device is formatted (see {#committed_device})
+        #
+        # @return [Boolean] true if the device exists on disk and it is formatted;
+        #   false otherwise.
+        def committed_filesystem?
+          !committed_filesystem.nil?
+        end
+
+        # Filesystem of the committed device (see {#committed_device})
+        #
+        # @return [Y2Storage::BlkFilesystem, nil] nil if the device does not exist on disk or
+        #   it is not formatted.
+        def committed_filesystem
+          return nil unless committed_device?
+
+          committed_device.filesystem
+        end
+
+        # Whether the device needs to be unmounted to resize it (shrink)
+        #
+        # @note The filesystem must be unmounted when it exists on disk, it is mounted and
+        #   it does not support mounted shrink.
+        #
+        # @return [Boolean]
+        def need_unmount_for_shrinking?
+          return false unless mounted_committed_filesystem?
+
+          !committed_filesystem.supports_mounted_shrink?
+        end
+
+        # Whether the device needs to be unmounted to resize it (grow)
+        #
+        # @note The filesystem must be unmounted when it exists on disk, it is mounted and
+        #   it does not support mounted grow.
+        #
+        # @return [Boolean]
+        def need_unmount_for_growing?
+          return false unless mounted_committed_filesystem?
+
+          !committed_filesystem.supports_mounted_grow?
+        end
+
+      private
+
+        # Devicegraph that represents the current version of the devices in the system
+        #
+        # @note To check whether a filesystem is currently mounted, it must be checked
+        #   in the system devicegraph. When a mount point is "immediate deactivated", the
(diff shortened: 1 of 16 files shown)


src/lib/y2partitioner/immediate_unmount.rb
```diff
@@ -0,0 +1,115 @@
...
+    # Sets textdomain
+    def included(_target)
+      textdomain "storage"
+    end
+
+    def immediate_unmount(device, full_message: nil, note: nil, allow_continue: true)
+      loop do
+        case confirm_unmount(device, full_message, note, allow_continue)
+        when :unmount
+          unmount(device) || next
```
jreidinger: why this `|| next` is needed? it looks strange for me as even unmount just continue with result.

jreidinger: or maybe you missing something like
```
unmount(device) ? break true : next
```

src/lib/y2partitioner/immediate_unmount.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      buttons = inmmediate_unmount_buttons(allow_continue)
+
+      Yast2::Popup.show(message, headline: headline, buttons: buttons, focus: :cancel)
+    end
+
+    def unmount(device)
+      device.mount_point.immediate_deactivate
+      true
+    rescue Storage::Exception
```
jreidinger: I would really like to see at least logging of exception if not passing it to show_unmount_error so user can see some details why unmount failed.

src/lib/y2partitioner/immediate_unmount.rb
```diff
@@ -0,0 +1,115 @@
...
+      show_unmount_error
+      false
+    end
+
+    def show_unmount_error
+      message = _("The file system could not be unmounted")
```
jreidinger: this is quite useless without any details why it could not be unmounted. It can be temporary issue, critical issue or whatever in between.

src/lib/y2partitioner/immediate_unmount.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+
+    def immediate_unmount_message(device, full_message, note, allow_continue)
+      return full_message unless full_message.nil?
+
+      path = device.mount_point.path
+
+      note = note.nil? ? "" : "\n" + note
+
+      # TRANSLATORS %s is the name of the device to be unmounted (e.g., /dev/sda1)
```
jreidinger: there is no `%s`

src/lib/y2storage/blk_device.rb
```diff
@@ -317,6 +317,12 @@ def delete_filesystem
       remove_descendants
     end
 
+    def mount_point
```
jreidinger: blk_device is so important that it really deserves documentation for any of its methods.

src/lib/y2partitioner/actions/delete_device.rb
```diff
@@ -147,6 +153,18 @@ def used_by_md?
...
+      def try_unmount
+        return true unless need_try_unmount?
+
+        note = _("It cannot be deleted while mounted.")
+
+        immediate_unmount(controller.committed_device, note: note)
+      end
+
+      def need_try_unmount?
+        device.is?(:blk_device) && controller.mounted_committed_filesystem?
```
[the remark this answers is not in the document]
joseivanlopez: You are right. This was a last minute change yesterday. Thanks to warn.

src/lib/y2partitioner/immediate_unmount.rb
joseivanlopez: Yes, it is wrong. Your proposal is what I wanted to do. Thanks.

src/lib/y2partitioner/immediate_unmount.rb
joseivanlopez: I fear that we don't have the reason when libstorage-ng cannot unmount, but I will check.

src/lib/y2partitioner/immediate_unmount.rb
joseivanlopez: Old comment, thanks.

src/lib/y2partitioner/immediate_unmount.rb
jreidinger: So exception does not contain anything useful?

joseivanlopez: BTW, the old partitioner simply showed nothing in that case :)

joseivanlopez: I have checked and it has the error reason. I have added it.

src/lib/y2partitioner/actions/controllers/blk_device.rb
```diff
@@ -0,0 +1,147 @@
...
+        attr_reader :device
+
+        # Constructor
+        #
+        # @raise [TypeError] if device is not a block device.
+        #
+        # @param device [Y2Storage::BlkDevice] it has to be a block device,
+        #   (i.e., device#is?(:blk_device) #=> true).
+        def initialize(device)
+          raise(TypeError, "param device has to be a block device") unless device.is?(:blk_device)
```
jreidinger: NP: but if you do type protection, then maybe check also against nil and potential other misusage like string, so maybe something like `if !device.respond_to?(:is?) || !device.is?(:blk_device)`

src/lib/y2partitioner/actions/controllers/blk_device.rb
```diff
@@ -0,0 +1,147 @@
...
+        end
+
+        # Whether the filesystem of the device exists on the system and it is mounted
+        #
+        # @note The filesystem on system is checked, independently of it matches to the current
+        #   filesystem of the device.
+        #
+        # @return [Boolean]
+        def mounted_committed_filesystem?
+          return false unless committed_filesystem?
```
jreidinger: I think this has one problem. What happen when user modify fs e.g. switch from ext4 to btrfs and this ext4 is mounted? We should offer to unmount it. Do we do?

jreidinger: no, now I get it. Name is a bit complex too me :)  Why not name it simple `mounted?` ?

jreidinger: meaning that device is somewhere mounted.

src/lib/y2partitioner/actions/controllers/blk_device.rb
```diff
@@ -0,0 +1,147 @@
...
+          !committed_filesystem.supports_mounted_shrink?
+        end
+
+        # Whether the device needs to be unmounted to resize it (grow)
+        #
+        # @note The filesystem must be unmounted when it exists on disk, it is mounted and
+        #   it does not support mounted grow.
+        #
+        # @return [Boolean]
+        def need_unmount_for_growing?
```
jreidinger: maybe name it simple `unmount_for_growing`? and same above.

src/lib/y2partitioner/actions/controllers/blk_device.rb
```diff
@@ -0,0 +1,147 @@
...
+
+require "yast"
+require "y2storage"
+require "y2partitioner/device_graphs"
+
+module Y2Partitioner
+  module Actions
+    module Controllers
+      # This class offers helper methods to perform actions over a block device,
+      # for example, to check if its filesystem exists on disk, it is mounted, etc.
```
jreidinger: this class use quite a lot term commited, so maybe also explain it in class description.

src/lib/y2partitioner/dialogs/blk_device_resize.rb
```diff
@@ -148,6 +162,105 @@ def used_size_info
...
+      end
+
+      # Whether the device is going to be growed
+      #
+      # @return [Boolean]
+      def growing?
+        selected_size > device.size
+      end
+
+      # Whether the device is going to be growed more than 1 GiB
```
jreidinger: Can you please add link why more then 1 GiB is big growing? So we can in future check if value is still reasonable.

src/lib/y2partitioner/dialogs/blk_device_resize.rb
```diff
@@ -148,6 +162,105 @@ def used_size_info
...
+          _("You are extending a mounted filesystem by %{gibs} Gigabyte.\n" \
+            "This may be quite slow and can take hours. You might possibly want \n" \
+            "to consider umounting the filesystem, which will increase speed of \n" \
+            "resize task a lot.\n\n" \
+            "You can try to unmount it now, continue without unmounting or cancel.\n" \
+            "Click Cancel unless you know exactly what you are doing."),
+          gibs: growing_size.to_i / Y2Storage::DiskSize.GiB(1).to_i
+        )
+
+        immediate_unmount(controller.committed_device, full_message: message, allow_continue: true)
```
jreidinger: can we show it twice for growing and then again for big growing, or it is always exclusive?

joseivanlopez: In case the user selects "continue" the first time, then it will be asked again in case of big growing (> 1 GiB).

src/lib/y2partitioner/immediate_unmount.rb
```diff
@@ -0,0 +1,173 @@
...
+  #
+  # @note This feature is intended to allow to unmount devices before the commit phase, for example,
+  #   when a device is deleted or resized. But there are serveral scenarios where the actions require
+  #   unmount the device and the user is not asked for it. For instance, when a new partition table
+  #   is created over a disk and some of its partitions are already mounted.
+  module ImmediateUnmount
+    include Yast::Logger
+    include Yast::I18n
+
+  private
```
jreidinger: private does not work in mixins

src/lib/y2partitioner/immediate_unmount.rb
```diff
@@ -0,0 +1,173 @@
...
+
+    # Shows a Popup dialog to try to unmount the device
+    #
+    # @param device [Y2Storage::BlkDevice]
+    # @param full_message [String, nil] message to show in the dialog (param note would be ignored)
+    # @param note [String, nil] note to add to the generic message (ignored if full_message is used)
+    # @param allow_continue [Boolean] if it should allow to continue without unmounting
+    #
+    # @return [Boolean]
+    def immediate_unmount(device, full_message: nil, note: nil, allow_continue: true)
```
jreidinger: To be honest, I do not like now design when I see it whole as only method that makes sense for users of this mixin is exactly this one. So what I propose? Having class ( e.g. inner class for unmount that gets params on instance and do action ). This way it allows to:

1. not pass allow_continue everywhere
2. not need that included trick for translations
3. easier test it
4. if makes sense ( now it is a bit on edge ) create subclasses ( e.g. for allow continue versus not one ).

src/lib/y2storage/blk_device.rb
```diff
@@ -317,6 +317,16 @@ def delete_filesystem
...
     end
 
+    # Mount point of the filesystem
+    #
+    # @return [Y2Storage::MountPoint, nil] nil if the device is not formatted or its
+    #   filesystem has not mount point.
+    def mount_point
+      return nil unless formatted?
+
+      filesystem.mount_point
```
jreidinger: nice shortcut :+1: . Maybe search rest of existing code and use it where it make sense?

src/lib/y2partitioner/actions/controllers/blk_device.rb
joseivanlopez: Probably the terminology is a bit confusing at first, but I prefer to explicitly see "committed" in the method, because it makes pretty clear that it refers to the filesystem in the real system and not the current version in memory.

src/lib/y2storage/blk_device.rb
joseivanlopez: Well, this PR already includes a lot changes. Due to this is for GA, I would avoid to add more changes here and there. I would prefer to do this "refac" in a separate PR.

src/lib/y2partitioner/dialogs/blk_device_resize.rb
joseivanlopez: BTW, the threshold is 50 GiB, see https://github.com/yast/yast-storage/blob/SLE-12-SP4/src/include/partitioning/ep-dialogs.rb#L1229

src/lib/y2partitioner/immediate_unmount.rb
```diff
@@ -51,123 +40,166 @@ def included(_target)
...
+
+      # Error popup showed when the device cannot be unmounted
+      #
+      # @param details [String] error details
+      def show_unmount_error(details)
+        # TRANSLATORS: Error message when the device could not be unmounted
+        message = _("The file system could not be unmounted")
+
+        Yast2::Popup.show(message, headline: :error, details: details, buttons: :ok)
+      end
```
jreidinger: code looks now much cleaner for me :+1:

src/lib/y2partitioner/dialogs/blk_device_resize.rb
```diff
@@ -248,11 +248,15 @@ def growing?
         selected_size > device.size
       end
 
-      # Whether the device is going to be growed more than 1 GiB
+      # Whether the device is going to be growed more than 50 GiB
+      #
+      # @note Threshold to consider a big growing is defined in the old code, but there is
```
jreidinger: this sentence missing probably some "not" like "there is not any kind of explanation" otherwise if there is some kind of explanation, I would like to see it also here.

joseivanlopez: Yes, I wanted to say "there is not any ..."

PR #719, opened 2018-08-28
[RFC] initial draaft of partitioner research

jreidinger:

PR #726, opened 2018-08-31
Partitioner: allow software RAIDs on top of full disks

ancorgs:
Main PR for https://trello.com/c/G8Or8z0z/211-3-full-disks-as-member-of-md

This is divided in three commits (so it can reviewed in a per-commit basis, although the third one modifies some code from the first two).

### Allow software RAIDs on top of full disks

This makes the partitioner offer full disks as "Available Devices" in the RAID screen. The criteria to decide which disks are considered available is explained at https://github.com/yast/yast-storage-ng/pull/719/.

In addition, it adapts all other parts of the Partitioner to correctly handle MD RAIDs on top of full disks (only very minor stuff was needed).

### Better action to create partition table

This adapts the workflow of "Create New Partition Table" to properly report about the consequences and, in general, to work in a consistent way with the rest of the Partitioner actions.

In SLE-15 the button immediately presents a form to select the partition table type in case the device supports more than one.

form

And **after** the user selects one type it **always** shows a warning about all kind of devices to be destroyed, no matter if some device is really affected or not.

waning

Even better, if only one partition table type is possible, it still shows the form but with no question. So creating a partition table in a completely empty DASD device result in a misleading warning (nothing is going to be destroyed) on top of an empty wizard.

stupid

So the whole action was reimplemented to display the warning only if some devices are indeed going to be affected (including the list of affected devices) and to display that warning as soon as the user clicks the button (as any other Partitioner action).

new_warning

As seen in the screenshot, the check handles correctly situations in which the disk as a whole (no partitions) is part of an MD RAID or an LVM setup. And, of course, there are not empty wizard steps in the case of DASD or nothing like that. Now the workflow works in the expected way, as proved by the unit tests.

* **For a disk with some partitions**
  * shows a confirmation dialog
  * if the user rejects the confirmation
    * leaves the disk untouched
    * returns :back
  * if the user confirms
    * displays a dialog to select the partition table type
      * if the user goes back or cancels in the type selection
        * leaves the disk untouched
        *  returns :finish
    * if the user selects a type and clicks "next"
      * replaces the previous partition table with a new one of the correct type
      * returns :finish
* **For an S/390 DASD with one partition**
  * shows a confirmation dialog
    * if the user rejects the confirmation
      * leaves the DASD untouched
      * returns :back
    * if the user confirms
      * does not display a dialog to select the partition table type
      * replaces the previous partition table with a new empty DASD one
      * returns :finish
* **for an unused disk with no partitions**
  * does not display any confirmation dialog
  * displays a dialog to select the partition table type
    * if the user goes back or cancels in the type selection
      * leaves the disk untouched
      *  returns :finish
    * if the user selects a type and clicks "next"
      * replaces the previous partition table with a new one of the correct type
      * returns :finish
* **for an unformatted ECKD DASD (it cannot have partition table)**
  * does not display any confirmation dialog
  * does not display any dialog to select the partition table type
  * displays an error popup
  * returns :back

And last but not least, the new implementation is more straightforward, without transactions, without a separate controller and with around 100 fewer lines of source code.

### More specific errors when a device is in use

In general, most of the checks already present in the Partitioner (with the exception of the mentioned "Create New Partition Table" action) were already able to correctly handle situations in which the disk was a direct member of an MD RAID or an LVM volume group. But the message about the device being in use was not informative enough.

in_use_old

Now the message includes the name of the device that is making the operation impossible (it's usually one, but there are corner cases in which it can be more than one), so the user has some clue about how to fix it.

in_use_new


ancorgs:
On the other hand, I just found this in one of the devicegraphs in our `/test/data/devicegraphs/` directory. 

md2


So it looks like we should add something else than disks (and multipath devices) to the list of available devices. MDs are also suitable and probably DM devices as well.

src/lib/y2partitioner/actions/controllers/md.rb
```diff
@@ -278,10 +283,23 @@ def available?(partition)
...
+        # A device is available if it is not already used by a RAID or
+        # used as physical volume or assigned to a mount point.
+        #
+        # @note This method does not check if the device is suitable (correct
+        # type, etc.), only if it's not already in use.
+        #
+        # @param device [Y2Storage::BlkDevice]
+        # @return [Boolean] true if can be used for a MD RAID; false otherwise.
+        def available?(device)
+          return false if device.lvm_pv || device.md
```
jreidinger: hmm, as in future here will be also bcache I start feeling that maybe this method suits better into device object?

test/y2partitioner/widgets/md_devices_selector_test.rb
```diff
@@ -281,7 +283,7 @@
             expect(controller.devices_in_md.map(&:name)).to eq [
               "/dev/sde3", "/dev/sda3", "/dev/sda2", "/dev/sda4"
             ]
-            expect(controller.available_devices).to be_empty
+            expect(controller.available_devices.map(&:name)).to eq ["/dev/sdb", "/dev/sdc"]
```
jreidinger: do we really want to be order dependent? For me `contain_exactly` is better unless we really want to check order.

src/lib/y2partitioner/actions/controllers/md.rb
ancorgs: I'm not sure. It's quite partitioner-specific.

jreidinger: @ancorgs maybe something like `used_in_logical_device?` can be that method? This is what we are checking here, if some other device use it.

src/lib/y2partitioner/actions/controllers/md.rb
```diff
@@ -278,10 +283,23 @@ def available?(partition)
...
+        #
+        # @note This method does not check if the device is suitable (correct
+        # type, etc.), only if it's not already in use.
+        #
+        # @param device [Y2Storage::BlkDevice]
+        # @return [Boolean] true if can be used for a MD RAID; false otherwise.
+        def available?(device)
+          return false unless device.component_of.empty?
+
+          device.filesystem.nil? || device.filesystem.mount_point.nil?
```
jreidinger: so if disk already has some filesystem, then it cannot used and at first it need somehow get rid of that filesystem, right?

src/lib/y2partitioner/actions/create_partition_table.rb
```diff
@@ -20,74 +20,72 @@
...
-      # Commit the action (creating a new partition table) to the staging
-      # devicegraph. This does not open a dialog.
-      def commit
-        log.info("Commit creating a new #{controller.type} partition table on #{disk_name}")
-        controller.create_partition_table
-        :finish
+      # Creates the new partition table
+      #
+      # @param type [Y2Storage::PartitionTables::Type]
+      def create_partition_table(type)
```
jreidinger: this method looks a bit like it should be in Partitionable itself. why original one does not remove descendants?

ancorgs: That's by design of libstorage-ng. Usually the methods there are kind of low-level with very little "implicit" or "automagical" behavior. The user of the library is expected to do the preparations explicitly if needed.

And the `Y2Storage` wrappers around the `Storage` classes try to respect as much as possible the original methods and ideas.

src/lib/y2partitioner/actions/create_partition_table.rb
```diff
@@ -96,10 +94,95 @@ def run?
...
+      # Whether performing the action can be considered non-destructive
+      #
+      # @return [Boolean] true if the disk is completely empty and unused or if
+      #   it only contains an empty partition table
+      def safe?
+        desc = disk.descendants.size
+        return true if desc.zero?
+        return false if desc > 1
+
+        disk.partition_table? || disk.encrypted?
```
jreidinger: I do not see from method documentation why encrypted disk is fine. I should kill that encrypted device, not?

src/lib/y2partitioner/actions/controllers/md.rb
ancorgs: Not exactly. If the disk already has a mount point, then it cannot be used. Having a filesystem is fine as long as is not assigned to any mount point.

src/lib/y2partitioner/actions/create_partition_table.rb
ancorgs: Yes, the definition of "non-destructive" needs to be improved. The idea is that killing an empty partition table or a completely unused encryption layer on the disk (that is, the disk is encrypted but not even formatted) is fine for someone clicking on "Create New Partition Table".

But strictly speaking, both things are "destructive" (they delete something, even if the deleted stuff is not used in any way or shape). So I will improve the documentation.

src/lib/y2partitioner/actions/create_partition_table.rb
```diff
@@ -96,10 +94,95 @@ def run?
...
+
+      # @see #confirm
+      def confirm_with_filesystem
+        fs = disk.filesystem.type.to_human
+        # TRANSLATORS: the first %s is replaced by the type of the filesystem
+        # (for example, "Btrfs" or "Ext2"), the second one by the disk device name
+        # ("/dev/sda" or similar).
+        msg = format(
+          _(
+            "This will delete the %s file system from the device\n\n" \
```
jreidinger: maybe mention that it destroy all data on this partition. as noone care about fs, but data is what matters.

src/lib/y2storage/blk_device.rb
```diff
@@ -351,13 +351,53 @@ def direct_md
...
+    def component_of
+      vg = lvm_pv ? lvm_pv.lvm_vg : nil
+      (dm_raids + [vg] + [md] + [multipath]).compact
+    end
+
+    # Equivalent of {#component_of} in which each device is represented by a
+    # string.
+    #
+    # @return [Array<String>]
+    def str_component_of
```
jreidinger: To be honest I really do not like name. When I see it first time, I have no idea what is that str...what about `component_names_of` ?

src/lib/y2partitioner/dialogs/partition_table_type.rb
```diff
@@ -47,7 +52,7 @@ def label
         # @macro seeAbstractWidget
         def help
           # helptext
-          _("<p>Choose the partition table type for the new partition.</p>")
+          _("<p>Choose the type for the new partition table.</p>")
```
jreidinger: is it "type for the new" or "type of the new"?

ancorgs: Both sounds equally valid to me. I just respected the previous one, but I can change it to "of" if you prefer so. :-)

src/lib/y2partitioner/actions/create_partition_table.rb
jreidinger: so what about adding something like "safe" variant or having convention that original ones that is rough ends up "!" and new safer ones will be without it? I just think that partitioner is not ideal place for it.

src/lib/y2partitioner/dialogs/partition_table_type.rb
jreidinger: @mvidner your fine tuned language sense here would be great :wink:

src/lib/y2storage/blk_device.rb
ancorgs: I would prefer `component_of_names`

src/lib/y2partitioner/actions/create_partition_table.rb
ancorgs: To avoid becoming inconsistent, we would need to do that for many libstorage-ng methods, not only `create_partition_table`. That is, renaming a bunch of libstorage-ng methods to `create_xxx!` or 'add_yyy!` and redefining the corresponding `create_xxx` and `add_yyy` to work in a non-libstorage way.

Sounds too aggressive to me (i.e. against the philosophy of keeping the wrapper thin). The whole libstorage-ng approach/API can be considered quite low-level not only in cases like this, but also in many other aspects. Once you change this in the wrapper, you will want to also change another behavior somewhere to make it also "smarter", and then another... where to stop?

package/yast2-storage-ng.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:	4.1.13
+Version:        4.1.14
```
jreidinger: indentation looks broken on github

src/lib/y2partitioner/actions/controllers/md.rb
```diff
@@ -96,6 +98,7 @@ def add_device(device)
             raise ArgumentError, "The device #{device} is already part of the Md #{md}"
           end
 
+          BlkDeviceRestorer.new(device).update_checkpoint if device.respond_to?(:partition_table)
```
jreidinger: btw why it is needed? comment here would be nice WHY it is needed to be here.

package/yast2-storage-ng.spec
ancorgs: it was done via `rake version:bump`

src/lib/y2partitioner/actions/create_partition_table.rb
```diff
@@ -96,10 +94,99 @@ def run?
...
+      #
+      # Destroying an unused partition table or an unused encryption layer is
+      # considered to be non-destructive in that regard.
+      #
+      # @return [Boolean] true if the disk (that may be encrypted) is completely
+      #   empty and unused or if it only contains an empty partition table
+      def safe?
+        desc = disk.descendants.size
+        return true if desc.zero?
+        return false if desc > 1
```
jreidinger: just matter of taste, but I do not like much this desc variable. Something like "descendants_count" is nicer for me.

src/lib/y2partitioner/dialogs/partition_table_type.rb
```diff
@@ -71,37 +76,37 @@ def validate
...
         textdomain "storage"
+
+        @disk = disk
+        @type_choice = TypeChoice.new(possible_types, default_type)
       end
 
       # @macro seeDialog
       def title
-        @controller.wizard_title
+        _("Create New Partition Table on %s") % @disk.name
```
jreidinger: rubocop did not complain? On other places `format` is used.

src/lib/y2partitioner/actions/create_partition_table.rb
ancorgs: It's just a shortcut for the comparison in the next two lines. Keeping it short was exactly the point. And it's a very local thing.

PR #735, opened 2018-09-11
bcache view

jreidinger:
snimek obrazovky_2018-09-11_15-55-15

depends on https://github.com/openSUSE/libstorage-ng/pull/572

then it can list also partitions:

snimek obrazovky_2018-09-12_16-43-53

bcache is also shown in system view:

bcache_all

and with subtree

snimek obrazovky_2018-09-17_09-42-23

and description of bcache device:

snimek obrazovky_2018-09-17_12-39-52

what is missing:
- [x] partitions support
- [ ] discuss collumns
- [x] subtree


src/lib/y2storage/bcache.rb
```diff
@@ -0,0 +1,77 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage/storage_class_wrapper"
+require "y2storage/partitionable"
+
+module Y2Storage
+  # A Bcachek device
```
ancorgs: Bcacheck :-)

src/lib/y2storage/bcache.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    def inspect
+      "<Bcache #{name} #{bcache_cset.inspect} -> #{blk_device}>"
+    end
+
+    # Default partition table type for newly created partition tables
+    # @see Partitionable#preferred_ptable_type
+    #
+    # @return [PartitionTables::Type]
+    def preferred_ptable_type
```
ancorgs: I wouldn't expect this here. First of all because this is not partitionable. Second, because there is apparently no need t overwrite the libstorage-ng default (if any, not even the case).

src/lib/y2storage/bcache_cset.rb
```diff
@@ -0,0 +1,61 @@
...
+module Y2Storage
+  # A Bcache caching set
+  #
+  # This is a wrapper for Storage::BcacheCset
+  class BcacheCset < Device
+    wrap_class Storage::BcacheCset
+
+    # @!method blk_device
+    #   @return [BlkDevice] returns a backing device for cache
+    storage_forward :blk_devices, as: "BlkDevice"
```
ancorgs: Is `:blk_devices` or `:blk_device`?

ancorgs: Probably the declaration is right but the documentation is a copy&paste from `Bcache`. Isn't it?

src/lib/y2storage/bcache.rb
jreidinger: well, I start this class as partitionable and then test reveals it is not, so it is there accidentally. I am not sure why it needs it, I took same approach as in disk. So I will remove it for now.

src/lib/y2storage/bcache_cset.rb
jreidinger: yes, will fix it.

src/lib/y2partitioner/widgets/bcache_device_description.rb
```diff
@@ -0,0 +1,65 @@
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
```
ancorgs: ~~2017~~ 2018

src/lib/y2partitioner/widgets/bcache_device_description.rb
```diff
@@ -0,0 +1,65 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2partitioner/widgets/disk_device_description"
+
+module Y2Partitioner
+  module Widgets
+    # Richtext filled with the description of a bcache devices
```
ancorgs: a bcache device~~s~~

src/lib/y2partitioner/widgets/bcache_device_description.rb
```diff
@@ -0,0 +1,65 @@
...
+      def device_description
+        super + bcache_description
+      end
+
+      # Specialized description for devices at backend of bcache
+      # @return [String]
+      def bcache_description
+        output = Yast::HTML.Heading(_("Bcache Devices:"))
+        output << Yast::HTML.List([
+                                    format(_("Backing device: %s"), device.blk_device.name.to_s),
```
ancorgs: NP: why do you chain `#to_s` to `#name`.

src/lib/y2partitioner/widgets/bcache_device_description.rb
```diff
@@ -0,0 +1,65 @@
...
+                                    format(_("Caching UUID: %s"), uuid),
+                                    format(_("Caching device: %s"), caching_device)
```
ancorgs: Following the usual YaST convention for upcasing, this should be "Caching Device", "Backing Device", etc.

src/lib/y2partitioner/widgets/disk_device_description.rb
```diff
@@ -23,7 +23,7 @@
 
 module Y2Partitioner
   module Widgets
-    # Richtext filled with the description of a disk device
+    # Richtext filled with the description of a disk devices
```
ancorgs: Why changing this to plural?

src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -221,6 +221,7 @@ def end_value(device)
       end
 
       DEVICE_ICONS = {
+        bcache:    Icons::HD, # TODO: have own icon for bcache
```
ancorgs: Please define `Icons::BCACHE`. The fact that we are reusing the HD png file so far should not be visible here. That's what we `y2partitioner/icons.rb` for.

src/lib/y2partitioner/widgets/overview.rb
```diff
@@ -234,6 +235,14 @@ def raid_items(md)
         CWM::PagerTreeItem.new(page)
       end
 
+      # @return [CWM::PagerTreeItem]
+      def bcaches_items
+        devices = device_graph.bcaches
+        page = Pages::Bcaches.new(devices, self)
+        children = devices.map { |v| disk_items(v) }
+        CWM::PagerTreeItem.new(page, children: children, icon: Icons::LVM) # TODO: new nice icon
```
ancorgs: Same than above, this should be `Icons::BCACHE`. No matter if we have not found a proper icon, we should define the constant already.

src/lib/y2partitioner/widgets/pages/bcaches.rb
```diff
@@ -0,0 +1,93 @@
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
```
ancorgs: ~~2017~~ 2018

src/lib/y2partitioner/widgets/pages/bcaches.rb
```diff
@@ -0,0 +1,93 @@
...
+        def initialize(bcaches, pager)
+          textdomain "storage"
+
+          @bcaches = bcaches
+          @pager = pager
+        end
+
+        # @macro seeAbstractWidget
+        def label
+          _("Bcaches")
```
ancorgs: As commented last week. "Bcaches" is not an English word. We should not make plurals out of it and so on. Please use just "Bcache" (as we do with the equivalent "RAID" section) or "Bcache Devices".

src/lib/y2partitioner/widgets/pages/bcaches.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+
+        # @macro seeCustomWidget
+        def contents
+          return @contents if @contents
+
+          table = ConfigurableBlkDevicesTable.new(devices, @pager)
+          icon = Icons.small_icon(Icons::HD) # TODO: icon for bcache
```
ancorgs: Just another example of why we should define the `Icons::BCACHE` constant.

src/lib/y2partitioner/widgets/pages/bcaches.rb
```diff
@@ -0,0 +1,93 @@
...
+        # @return [CWM::TreePager]
+        attr_reader :pager
+
+        # Returns all bcaches and their partitions
+        #
+        # @return [Array<Y2Storage::BlkDevice>]
+        def devices
+          bcaches.each_with_object([]) do |bcache, devices|
+            devices << bcache
+            devices.concat(bcache.partitions) if bcache.respond_to?(:partitions)
```
ancorgs: Why wouldn't `bcache` respond to `#partitions`?

src/lib/y2partitioner/widgets/pages/disk.rb
```diff
@@ -139,7 +140,18 @@ def label
         # @macro seeCustomWidget
         def contents
           # Page wants a WidgetTerm, not an AbstractWidget
-          @contents ||= VBox(DiskDeviceDescription.new(@disk))
+          @contents ||= VBox(description)
```
ancorgs: Please, mention in the general description of `Pages::Disk` that it's used also for bcache devices.

src/lib/y2partitioner/widgets/bcache_device_description.rb
jreidinger: hmm, that is in fact result of changing it later, as blk_device can be nil. I will fix it.

src/lib/y2partitioner/widgets/pages/system.rb
```diff
@@ -172,6 +172,15 @@ def nfs_devices
           device_graph.nfs_mounts
         end
 
+        # @return [Array<Y2Storage::Device>]
```
ancorgs: NP: you use `bcaches` in most method names, I find kind of inconsistent to use `bcache_devices` here. Not a big deal, just curious about the rationale (I guess just copy&pasting from `disk_devices`).

src/lib/y2partitioner/widgets/pages/system.rb
```diff
@@ -172,6 +172,15 @@ def nfs_devices
...
+        def bcache_devices
+          all = device_graph.bcaches
+          all.each_with_object([]) do |bcache, devices|
+            devices << bcache
+            devices.concat(bcache.partitions) if bcache.respond_to?(:partitions)
```
ancorgs: Again, why wouldn't `bcache` not respond to `#partitions`?

src/lib/y2partitioner/widgets/pages/bcaches.rb
jreidinger: well, I just copy same code as is for disks, which is also partitionable - https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/widgets/pages/disks.rb#L96

but maybe it is caused that not all disks are real disks that is partitionable?

src/lib/y2storage/devicegraph.rb
```diff
@@ -285,6 +286,10 @@ def nfs_mounts
       Filesystems::Nfs.all(self)
     end
 
+    def bcaches
```
ancorgs: Please document method (as obvious as it is, it's part of the public API).

test/y2partitioner/widgets/pages/bcaches_test.rb
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env rspec
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
```
ancorgs: ~~2017~~ 2018

src/lib/y2partitioner/widgets/pages/system.rb
jreidinger: ok, I will try to unify it to `bcache_devices`

test/y2partitioner/widgets/pages/bcaches_test.rb
```diff
@@ -0,0 +1,60 @@
...
+  let(:pager) { double("OverviewTreePager") }
+
+  include_examples "CWM::Page"
+
+  describe "#contents" do
+    let(:widgets) { Yast::CWM.widgets_in_contents([subject]) }
+    let(:table) { widgets.detect { |i| i.is_a?(Y2Partitioner::Widgets::BlkDevicesTable) } }
+    let(:bcaches_and_parts) do
+      device_graph.bcaches
+      # TODO: partitions does not work
```
ancorgs: Is this still true? As far as I understand, `Bcache` was fixed to descend from `Partitionable`

jreidinger: yes, no longer true

package/yast2-storage-ng.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:	4.1.15
+Version:	4.1.16
```
ancorgs: Please, don't increase version yet. I will do it in my pull request. I don't want to release what is in master right now without some corrections that are on #737

jreidinger: ok

src/lib/y2storage/blk_device.rb
```diff
@@ -217,6 +217,23 @@ def udev_full_ids
     storage_forward :create_blk_filesystem, as: "Filesystems::BlkFilesystem", raise_errors: true
     alias_method :create_filesystem, :create_blk_filesystem
 
+    # @!method create_bcache
+    #   Creates backing device build on current block device.
+    #
+    #   If the blk device has children, the children will become children of
```
ancorgs: This sentence looks like a wrong copy&paste

src/lib/y2storage/blk_device.rb
```diff
@@ -366,6 +383,20 @@ def multipath
...
+    #
+    # @return [Bcache, nil] nil if the device is not part of any bcache
+    def bcache
+      children.find { |dev| dev.is?(:bcache) }
+    end
+
+    # Bcache caching set device defined on top of the device
+    #
+    # @return [BcacheCset, nil] nil if the device is not used as bcache caching set
+    def direct_bcache_cset
```
ancorgs: I find the usage of `direct_` to be inconsistent between `#bcache` and `#direct_bcache_set`.

Can a bcache be defined with a LUKS as backing device? In that case, I would expect two methods:

* `#direct_bcache`
  * For `sda1 -> LUKS -> bcache0` it returns nil
  * For `sda1 -> bcache0` it returns bcache0

* `#bcache`
  * For `sda1 -> LUKS -> bcache0` it returns bcache0
  * For `sda1 -> bcache0` it returns bcache0

And the other way around. If a LUKS device can be a caching device, I would expect to have both `#direct_bcache_set` and `#bcache_set`.

Moreover, I find the usage of terms like "on top" potentially confusing in the descriptions, specially since it's used for both the bcache and the cset devices. What about something like "Bcache defined on top of the device, i.e. a Bcache device that uses this one as backing device".

src/lib/y2storage/blk_device.rb
```diff
@@ -382,11 +413,11 @@ def part_of_lvm_or_md?
...
     # one of its members.
     #
-    # @return [Array<Device>] a collection of MD RAIDs, DM RAIDs, volume groups
-    #   and multipath devices
+    # @return [Array<Device>] a collection of MD RAIDs, DM RAIDs, volume groups,
+    #   multipath, bcache and bcache_cset devices
     def component_of
       vg = lvm_pv ? lvm_pv.lvm_vg : nil
-      (dm_raids + [vg] + [md] + [multipath]).compact
+      (dm_raids + [vg] + [md] + [multipath] + [bcache] + [direct_bcache_cset]).compact
```
ancorgs: There is no `BcacheCset#name`, so make sure to handle that case in `#component_of_names`,

test/y2storage/bcache_cset_test.rb
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env rspec
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
```
ancorgs: Still 2017?

test/y2storage/bcache_test.rb
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env rspec
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
```
ancorgs: 2018

src/lib/y2storage/blk_device.rb
jreidinger: why you think so? It is also from libstorage, that it moves its children ( probably partitions or fs ) to bcache from blk_device.

src/lib/y2storage/blk_device.rb
jreidinger: Well, to be honest I do not play with LUKS yet on top of it. But I worry using luks on top of caching device kills performance.

src/lib/y2storage/bcache.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    # @!method self.all(devicegraph)
+    #   @param devicegraph [Devicegraph]
+    #   @return [Array<Bcache>] all the bcache devices in the given devicegraph,
+    #     in no particular order
+    storage_class_forward :all, as: "Bcache"
+
+    # @!method self.find_by_name(devicegraph, name)
+    #   @param devicegraph [Devicegraph]
+    #   @param name [String] kernel-style device name (e.g. "/dev/sda")
```
joseivanlopez: NP: the example is not a typical bcache name, better (e.g., "/dev/bcache0")

src/lib/y2storage/blk_device.rb
```diff
@@ -217,6 +217,23 @@ def udev_full_ids
...
+    #   the encryption device.
+    #
+    #   @return [Bcache]
+    storage_forward :create_bcache, as: "Bcache", raise_errors: true
+
+    # @!method create_bcache_cset
+    #   Creates caching set build on current block device.
+    #
+    #   Blk device have to not contain any children.
+    #
```
joseivanlopez: NP: I would try to document possible exceptions.

src/lib/y2storage/blk_device.rb
joseivanlopez: NP: I would try to document possible exceptions.

src/lib/y2storage/blk_device.rb
```diff
@@ -366,6 +383,20 @@ def multipath
       children.find { |dev| dev.is?(:multipath) }
     end
 
+    # Bcache device defined on top of the device
```
joseivanlopez: NP: What about rephrase it as: "Bcache in which the device is used as backing device."

src/lib/y2storage/blk_device.rb
ancorgs: I think so because it says "will become children of the encryption device" instead of "will become children of the bcache device".

src/lib/y2storage/blk_device.rb
```diff
@@ -395,7 +426,15 @@ def component_of
...
     def component_of_names
       # So far, all the possible elements on the array respond to #name
-      component_of.map(&:name)
+      component_of.map do |dev|
+        if dev.respond_to?(:name)
+          dev.name
+        elsif dev.respond_to?(:uuid)
+          dev.uuid
+        else
+          raise "Unknown tyoe if device #{dev.inspect}"
```
joseivanlopez: NP: The error message looks wrong.

src/lib/y2storage/blk_device.rb
ancorgs: Is not about performance or how correct it is, it's about using method names in a consistent way. I don't see the point in having a `#bcache` method that returns bcache devices placed directly (no LUKS in the middle) "on top" of this device and then a `#direct_bcache_cset` method that returns csets also placed directly on the device. It's inconsistent to each other and to the other methods in the class.

src/lib/y2storage/blk_device.rb
ancorgs: I don't know if displaying an UUID will help the users to know that they are removing the cset from a given bcache device. Something including the bcache name would probably be more informative.

On the other hand, raising an exception here may be a little bit radical. :wink:

src/lib/y2storage/blk_device.rb
```diff
@@ -395,7 +426,15 @@ def component_of
     # @return [Array<String>]
     def component_of_names
       # So far, all the possible elements on the array respond to #name
-      component_of.map(&:name)
+      component_of.map do |dev|
+        if dev.respond_to?(:name)
+          dev.name
+        elsif dev.respond_to?(:uuid)
```
ancorgs: I don't know if displaying an UUID will help the users to know that they are removing the cset from a given bcache device. Something including the bcache name would probably be more informative.

On the other hand, raising an exception here may be a little bit radical. :wink:

src/lib/y2storage/blk_device.rb
jreidinger: hmm, that scares me a bit that there is so many new methods needed. Would be possible to have only non-direct one and if needed define direct as something that takes a block?

like
```
def bcache
  children.find { |dev| dev.is?(:bcache) }
end

def direct
  encrypted? ? <NullObject returning nil> : self
end

# and then usage
dev.direct.bcache
dev.bcache
```

because otherwise we need for any new device device both methods, which looks quite bad for me as API for blk_device is already quite big.

jreidinger: btw this pull request also shows how many place I forgot in initial support for bcache. If there is new type of device it needs beside wrapper also add it to blk_device, devicegraph and also everywhere for downcast. I am not sure where it will lead with "read-only" LVMCache and other new potential types. API for some "god" objects is already quite big ( blk_device and devicegraph ).

src/lib/y2storage/blk_device.rb
jreidinger: Problem is that bcache caching device can be used in multiple bcaches and also for flash only volume, so the name can be really big. So I am not sure what would the best name for users.

src/lib/y2storage/blk_device.rb
jreidinger: @joseivanlopez libstorage-ng does not mention any exception for this call. Real testing will show reality. For create_bcache_cset it is written, so I will add it.

src/lib/y2storage/blk_device.rb
```diff
@@ -366,6 +385,21 @@ def multipath
       children.find { |dev| dev.is?(:multipath) }
     end
 
+    # Bcache device defined on top of the device, i.e. a Bcache device that uses
+    # this one as backing device
+    #
+    # @return [Bcache, nil] nil if the device is not part of any bcache
+    def bcache
```
ancorgs: I'm fine with no defining the corresponding `direct_xxx` methods. They are not needed so far so they would only bring noise.

But to keep consistency with the current API in which `#filesystem`, `#lvm_pv` and `#md` return the corresponding device, no matter if there is a LUKS in the middle or not, this should do the same. I don't care how often we will have a LUKS between the real block device and its bcache, but its theoretically doable.

The current API may be far from perfection, but the worst API is an inconsistent one. :-)

src/lib/y2storage/blk_device.rb
ancorgs: So what about something like this?

 * "cache set -bcache0-"
 * "cache set -bcache0, bcache3-"
 * "cache set" (*for flash only devices*)

That's still more informative about the devices to be affected than something like "0f517398-11ba-4c6c-9a5d-52b76c488303".

src/lib/y2storage/bcache_cset.rb
```diff
@@ -0,0 +1,70 @@
...
+  # This is a wrapper for Storage::BcacheCset
+  class BcacheCset < Device
+    include Yast::I18n
+    wrap_class Storage::BcacheCset
+
+    # @!method blk_devices
+    #   @return [Array<BlkDevice>] returns a list of devices used as caching ones in this set
+    storage_forward :blk_devices, as: "BlkDevice"
+
+    # @!method blk_devices
```
ancorgs: Copy&paste alarm!!! :wink:

ancorgs: Not only the name, also the description

src/lib/y2storage/bcache_cset.rb
```diff
@@ -0,0 +1,70 @@
...
+    #   @param devicegraph [Devicegraph]
+    #   @return [Array<BcacheCset>] all the bcache_csets in the given devicegraph,
+    #     in no particular order
+    storage_class_forward :all, as: "BcacheCset"
+
+    def inspect
+      "<BcacheCset #{uuid} #{blk_devices.inspect}>"
+    end
+
+    def display_name
```
ancorgs: Please document the method

src/lib/y2storage/bcache_cset.rb
```diff
@@ -0,0 +1,70 @@
...
+      textdomain "storage"
+      format(_("Cache Set for %s"), bcaches.map(&:name).sort.join(", "))
```
ancorgs: Shouldn't we cover the case in which bcaches is empty?

ancorgs: BTW, taking into account where it will be used, I would go for something shorter (maybe using `#basename` of devices) and with less uppercases.

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/actions/resize_blk_device.rb#L106
https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/actions/edit_blk_device.rb#L107

jreidinger: How it should look in that case? just "Cache Set" ?

ancorgs:
BTW, now that we merged the RAID changes before these ones, incrementing the version makes sense again. :-)

src/lib/y2storage/bcache_cset.rb
ancorgs: That should be enough, IMHO

jreidinger: hmm, but when I check its usage. I think it will be better to inform user what other devices will be affected. I will prepare something.

src/lib/y2storage/bcache_cset.rb
```diff
@@ -61,7 +61,8 @@ def display_name
       textdomain "storage"
       devices = bcaches.map(&:basename).sort.join(", ")
       # TRANSLATORS: status when cache set is not attached to any bcache
-      devices = _("not attached") if devices.empty?
+      return _("Cache set(not attached)") if devices.empty?
```
ancorgs: missing space before the parenthesis?

PR #736, opened 2018-09-11
Yaml support for importing/exporting Md devices

joseivanlopez:
Part of PBI https://trello.com/c/itizMqBG/274-5-partitions-in-raid

### DIFF
--- a/src/lib/y2storage/fake_device_factory.rb
+++ b/src/lib/y2storage/fake_device_factory.rb
@@ -35,7 +35,7 @@ module Y2Storage
   #
   class FakeDeviceFactory < AbstractDeviceFactory
     # Valid toplevel products of this factory
-    VALID_TOPLEVEL  = ["dasd", "disk", "lvm_vg"]
+    VALID_TOPLEVEL  = ["dasd", "disk", "md", "lvm_vg"]
 
     # Valid hierarchy within the products of this factory.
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/fake_device_factory.rb
```diff
@@ -250,6 +260,53 @@ def create_disk(_parent, args)
...
+      name = args["name"]
+      name ||= Md.find_free_numeric_name(devicegraph)
+
+      levels = {
+        "raid0"  => Storage::MdLevel_RAID0,
+        "raid1"  => Storage::MdLevel_RAID1,
+        "raid4"  => Storage::MdLevel_RAID4,
+        "raid5"  => Storage::MdLevel_RAID5,
+        "raid6"  => Storage::MdLevel_RAID6,
+        "raid10" => Storage::MdLevel_RAID10
```
jreidinger: why defining it here again? Why not simple to_s or to_sym ? I really do not like it as it e.g. miss container type and in general if we add another raid level it will miss here.

joseivanlopez: Yes, you are right. Basically, I was avoiding to use wrapper methods, because all this code was written before having the wrapper. But for this specific check it makes sense to start to use it.

PR #737, opened 2018-09-11
Display partitions in RAIDs list (and changing buttons)

ancorgs:
Part of https://trello.com/c/itizMqBG/274-5-partitions-in-raid

- Display partitions of RAIDs:
  - In the left tree
  - In the System view
  - In the RAID view
- In the RAID view, the buttons change with the selected row
- Do not allow to edit partitioned devices (will be soon also useful for whole disks)

partitionable_raids1

In addition, this also reverts some of the changes introduced in #738, which went too far affecting the "Hard Disks" section in very visible ways (even breaking staging).


joseivanlopez:
We should update here the version and changelog if this goes after #738.

package/yast2-storage-ng.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:	4.1.15
+Version:        4.1.16
```
jreidinger: bad indentation

PR #738, opened 2018-09-12
New view for partitionable Software RAID

joseivanlopez:
Part of PBI https://trello.com/c/itizMqBG/274-5-partitions-in-raid

This PR is only a part of the PBI. It will be completed by PR https://github.com/yast/yast-storage-ng/pull/737. For this reason, version number and changelog is not updated here, it will be done in PR https://github.com/yast/yast-storage-ng/pull/737.

Tests require https://github.com/openSUSE/libstorage-ng/pull/574


src/lib/y2partitioner/actions/delete_partitions.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      # Error when the device has no partition table
+      #
+      # @return [String, nil] nil if the device has a partition table
+      def missing_partition_table_error
+        return nil if device.partition_table?
+
+        # TRANSLATORS: Error when trying to delete partitions from a device that does not contain
+        # a partition table. %{name} is replaced by the name of the device (e.g. "/dev/sda").
+        format(_("The device %{name} does not contain a partition table."), name: device.name)
```
jreidinger: maybe add something like "does not contain a partition table, so no partition can be deleted."

src/lib/y2partitioner/widgets/partitions_delete_button.rb
```diff
@@ -0,0 +1,51 @@
...
+    class PartitionsDeleteButton < DeviceButton
+      def initialize(*args)
+        super
+        textdomain "storage"
+      end
+
+      # @macro seeAbstractWidget
+      def label
+        # TRANSLATORS: label for button to delete all partitions
+        _("Delete All...")
```
jreidinger: Hmm, why not "Delete partitions" ? this Delete All, is quite depending on context and in some context it can look strange.

joseivanlopez: Well, I have just implemented it as described in the document. But, anyway, this button only belongs to the "Partitions tab", so the context is always correct. @ancorgs what do you think? Would it better to rename the button as "Delete Partitions" ?

ancorgs: It's a list of partitions, so "Delete All" looks fine to me. In this first prototype, we do not longer have a button to delete all partitions in the table of disks, for example. If we reconsider that (I want to send a mail to Ken about it), then we should re-evaluate the label. But so far it looks correct to me.

PR #739, opened 2018-09-13
support full disk devices as lvm pv in autoyast profile (bsc#1107298)

wfeldt:
- unit tests missing

### DIFF
--- a/src/lib/y2storage/proposal/autoinst_devices_planner.rb
+++ b/src/lib/y2storage/proposal/autoinst_devices_planner.rb
@@ -123,9 +123,8 @@ module Y2Storage
       #
       # @note The part argument is used when we emulate the sle12 behavior to
       #   have partition 0 mean the full disk.
-      def planned_for_full_disk(drive, part = nil)
+      def planned_for_full_disk(drive, part)
         planned = Y2Storage::Planned::StrayBlkDevice.new
-        part ||= Y2Storage::AutoinstProfile::PartitionSection.new
         device_config(planned, part, drive)
         planned.lvm_volume_group_name = part.lvm_group
         add_device_reuse(planned, drive.device, part)
--- a/src/lib/y2storage/proposal/autoinst_devices_planner.rb
+++ b/src/lib/y2storage/proposal/autoinst_devices_planner.rb
@@ -89,14 +89,59 @@ module Y2Storage
       # @return [AutoinstIssues::List] List of AutoYaST issues to register them
       attr_reader :issues_list
 
+      # Returns an array of planned partitions for a given disk or the disk
+      # itself if there are no partitions planned
+      #
+      # @param disk [Disk,Dasd] Disk to place the partitions on
+      # @param drive [AutoinstProfile::DriveSection] drive section describing
+      #   the layout for the disk
+      # @return [Array<Planned::Partition, Planned::StrayBlkDevice>] List of planned partitions or disks
+      def planned_for_disk(disk, drive)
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -45,6 +45,7 @@ def self.attributes
           { name: :keep_unknown_lv },
           { name: :lvm2 },
           { name: :is_lvm_vg },
+          { name: :lvm_group },
```
ancorgs: According to AutoYaST documentation, there is no `lvm_group` attribute in a `<drive>` section. Only `<partition>` sections have that.

wfeldt: BTW, `pesize` is in `drive` and not in `partition` where it would logically belong to.

wfeldt: The autoyast profile looks not very consistent here and we should try to make it more logical going forward.

ancorgs: If it would be only `lvm_group` and `pesize`... There are quote some attributes in the current specification that are UTTERLY misplaced.

But in my opinion if we want to introduce another way of specifying the usage of full disks, we must first:

- decide/design how it's going to look like
- implement it
- then, document it

And that includes thinking about import, export, etc. and thinking about full disks as PV, full formatted disks, etc.

Just adding another attribute to the `<drive>` section "since we are here" sounds like the recipe to repeat the problem in the future (discovering via bug report than the code supported yet another way of doing things because it sounded like a good idea to whoever was touching the code). :wink:

src/lib/y2storage/proposal/autoinst_devices_planner.rb
```diff
@@ -89,14 +89,65 @@ def planned_devices(drives_map)
...
+      # @param disk [Disk,Dasd] Disk to place the partitions on
+      # @param drive [AutoinstProfile::DriveSection] drive section describing
+      #   the layout for the disk
+      # @return [Array<Planned::Partition, Planned::StrayBlkDevice>] List of planned partitions or disks
+      def planned_for_disk(disk, drive)
+        result = []
+
+        # if there are no partitions use the entire device
+        if drive.partitions.empty?
+          result = planned_for_full_disk(drive)
```
ancorgs: This does the same in the case in which we have just one partition with `partition_nr=0, create=false` (which was the scope of the PBI) and if we have an empty list of partitions (which is a different scenario that is reflected in the current AutoYaST documentation).

ancorgs:
IMO, this PR in its current shape goes a little bit too much beyond duty. I would focus on supporting the current format (just one partition with `partition_nr=0` and `create=false`) instead of expanding that format with new additional ways (that we should then export and so on).

Now that we are adding support to the partitioner to create more flexible layouts, we will need to find a way to specify all that in AutoYaST. But that's out of the scope of this PR/PBI and it's not something to be done just by thinking locally.

src/lib/y2storage/proposal/autoinst_devices_planner.rb
wfeldt: The created layout is the same as before in this case.

joseivanlopez:
@wfeldt, @ancorgs this PR should go to GA (now it is done to master).

wfeldt:
replaced by https://github.com/yast/yast-storage-ng/pull/740

PR #740, opened 2018-09-18
support full disk devices as lvm pv in autoyast profile (bsc#1107298)

wfeldt:
This allows partition 0 to stand as an alias for the full disk.


src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -211,12 +219,14 @@ def reuse_vgs(reused_vgs, devicegraph)
       # @param mds             [Array<Planned::Md>]        List of planned MD arrays to create
       # @param previous_result [Proposal::CreatorResult]   Starting point
       # @param parts_to_reuse  [Array<Planned::Partition>] List of partitions to reuse
```
wfeldt: That line should also go?

wfeldt: I see, it's removed in a later commit...

ancorgs:
@wfeldt Added a testcase with the minimal profile you attached to the card

wfeldt:
test case: lgtm

PR #741, opened 2018-09-19
Do not crash during the content info detection

dgdavid:
Related to https://bugzilla.suse.com/show_bug.cgi?id=1101979

> If there is any problem during the content info detection, it will not
be considered as windows partition.

### DIFF
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -210,6 +210,10 @@ module Y2Storage
 
       log.info("#{partition.name} is a windows partition") if is_win
       is_win
+
+    rescue Storage::Exception
+      log.warn("#{partition.name} content info cannot be detected")
+      false
     end
 
     # Obtain release names of installed systems in a disk.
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/disk_analyzer.rb
```diff
@@ -210,6 +210,10 @@ def windows_partition?(partition)
 
       log.info("#{partition.name} is a windows partition") if is_win
       is_win
+
+    rescue Storage::Exception
+      log.info("#{partition.name} content info cannot be detected")
```
ancorgs: Not sure if this should be `log.info` or something more serious like `log.warn`.

test/data/devicegraphs/mixed_disks.yml
```diff
@@ -15,6 +15,13 @@
     - free:
         size:         2 GiB
 
+    - partition:
```
ancorgs: Is it realistic to have a /dev/sda3 partition before an /dev/sda2 in an ms-dos partition table? Honest question, I would have to check.

dgdavid: Honestly, I don't know  :sweat_smile: but I am going to use the `windows-pc-gpt` devicegraph fixture as scenario of this _context_ instead, avoiding to modify the `mixed_disks`.

Thank you!

src/lib/y2storage/disk_analyzer.rb
dgdavid: Let me to change it from `.info` to `.warn`.

PR #746, opened 2018-09-25
add support for adding bcache

jreidinger:


src/lib/y2partitioner/actions/add_bcache.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+    private
+
+      def create_device(result)
+        backing = device_graph.blk_devices.find{ |d| d.name == result[:backing] }
+        raise "Invalid result #{result.inspect}. Backing not found." unless backing
+
+        caching = device_graph.blk_devices.find{ |d| d.name == result[:caching] }
+        raise "Invalid result #{result.inspect}. Backing not found." unless backing
```
ancorgs: Copy and paste: ~~backing~~caching

src/lib/y2partitioner/actions/add_bcache.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        bcache_nums = device_graph.bcaches.map(&:basename).map { |n| n[/bcache(\d+)/, 1].to_i }
+        new_num = (bcache_nums.max || -1) + 1
+        bcache = backing.create_bcache("/dev/bcache#{new_num}")
```
ancorgs: I would expect this numbering logic in the `Y2Storage::Bcache` class, not here.

src/lib/y2partitioner/actions/add_bcache.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        bcache_cset = device_graph.bcache_csets.find { |d| d.blk_devices.include?(caching) }
+        if !bcache_cset
+          if caching.respond_to?(:partition_table)
```
ancorgs: Why only if it responds to `#partition_table`?

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,172 @@
...
+module Y2Partitioner
+  module Dialogs
+    # Form to set the backing and caching device for bcache.
+    # Part of {Actions::AddBcache}.
+    class Bcache < Base
+
+      # @param device_graph [Y2Storage::DeviceGraph] graph on which it operates
+      # @param result [Hash] holder for result. It contains :caching and :backing keys that
+      #   have device name assigned. Content is undefined if dialog result is not next.
+      def initialize(device_graph, result)
```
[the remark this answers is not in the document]
jreidinger: we already do it, we just call it controller. So should I create object for it?

src/lib/y2partitioner/actions/add_bcache.rb
```diff
@@ -0,0 +1,69 @@
...
+            caching.remove_descendants
+          end
+          bcache_cset = caching.create_bcache_cset
+        end
+
+        bcache.attach_bcache_cset(bcache_cset)
+      end
```
ancorgs: Again, all this logic about reusing or creating cache sets is something I would expect in the `Y2Storage` classes (if not even in the library).

src/lib/y2partitioner/actions/add_bcache.rb
jreidinger: makes sense for me.

src/lib/y2partitioner/actions/add_bcache.rb
jreidinger: because we want to remove only partition table, nothing else. But I agree that it does not hurt if it is already there.

src/lib/y2partitioner/actions/add_bcache.rb
jreidinger: on which level of abstraction? Like method in Bcache like `bcache.add_caching(blk_device)` or in BcacheCset like `BcacheCset.find_or_create(blk_device)` ?

ancorgs:
Things that look strange to me:

### Passing the `result` hash back and forward as I/O argument

It feels so C++ish and so antiruby-ish... :wink:

I would pass a bcache device to the constructor of the dialog, that would be `nil` if we are creating a new bcache (so default values should be used in the form).

Then I would have methods like `Dialogs::Bcache#caching` or `Dialogs::Bcache#backing` to read the different aspects of the user selection.

### Logic about calculating the name or dealing with shared sets

That's general general logic that everyone dealing with bcaches will need, not specific to how the Partitioner works. I would expect that logic to be provided by the library. If not, I would expect it to live in the `Y2Storage` namespace. Definitely not in the action.

### Getting the `#available_devices` in the widget

I'm not so sure about this point. But maybe I would move that logic to the action to have more dumb dialogs. I would have to think about it twice.

src/lib/y2partitioner/dialogs/bcache.rb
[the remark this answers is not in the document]
ancorgs: No. That's not what the controller is about. Take a look to my summary comment below.

src/lib/y2partitioner/actions/add_bcache.rb
ancorgs: Then you should not be checking whether it responds to `#partition_table` (a.k.a. if it CAN have a partition table), but whether it HAS a partition table.

src/lib/y2partitioner/actions/add_bcache.rb
ancorgs: At first glance, both look ok to me

src/lib/y2partitioner/actions/add_bcache.rb
jreidinger: problem is that not all block devices can have it like partitions.

jreidinger: but lets make it generic a just call it always.

src/lib/y2partitioner/dialogs/bcache.rb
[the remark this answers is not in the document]
ancorgs: I will try to write a small document trying to clarify what the original intention behind the namespaces (`Dialogs`, `Widgets`, `Actions`, `Actions::Controllers`, etc.) was. Sure that with the hurry (and other circumstances) we twisted the original idea a bit and now it's not so clear.

src/lib/y2partitioner/actions/add_bcache.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    private
+
+      def create_device(dialog)
+        backing = dialog.backing_device
+        raise "Invalid result #{dialog.inspect}. Backing not found." unless backing
+
+        caching = dialog.caching_device
+        raise "Invalid result #{dialog.inspect}. Backing not found." unless caching
```
imobachgs: Backing -> Caching

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,182 @@
...
+
+        def label
+          _("Caching Device")
+        end
+
+        def items
+          available_devices.map do |dev|
+            case dev
+            when Y2Storage::BcacheCset
+              # if set have more then 1 device,
```
ancorgs: ~~have~~has

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,182 @@
...
+              # we still can identify it via this blk device as it cannot be shared
+              [dev.blk_devices.first.name, dev.display_name]
+            else
+              [dev.name, dev.basename]
```
ancorgs: All devices in the devicegraph have an internal numeric identifier called `sid`. We never display it to the user, but you could use it as the item id, instead of using names.

jreidinger: ok, so lets change it.

src/lib/y2storage/bcache.rb
```diff
@@ -59,6 +64,21 @@ def inspect
...
 
+    def self.usable_blk_devices(device_graph)
+      device_graph.blk_devices.select do |dev|
+        dev.component_of.empty? &&
+          (dev.filesystem.nil? || dev.filesystem.mount_point.nil?) &&
+          (!dev.respond_to?(:partitions) || dev.partitions.empty?)
+      end
+    end
+
+    # Device name suitable for newly created bcache
```
ancorgs: In other classes, this logic is built into the library itself. Like `Md.find_free_numeric_value` or the name of `PartitionTables::Base.unused_partition_slots`.

src/lib/y2storage/bcache.rb
```diff
@@ -59,6 +64,21 @@ def inspect
       "<Bcache #{name} #{bcache_cset.inspect} -> #{blk_device}>"
     end
 
+    def self.usable_blk_devices(device_graph)
```
ancorgs: I would call this partitioner-specific logic. It's the kind of things we have the actions for.

src/lib/y2storage/bcache.rb
jreidinger: so should I modify library? I do not see similar method in library for now: https://github.com/openSUSE/libstorage-ng/blob/master/storage/Devices/Bcache.h
I tried to avoid it if possible as ruby is faster to write something and easier for testing :)

src/lib/y2storage/bcache_cset.rb
```diff
@@ -67,6 +67,20 @@ def display_name
       format(_("Cache set (%s)"), devices)
     end
 
+    # Returns list of devices that can be used for caching. Like existing caching sets or
+    # suitable block devics.
+    def self.possible_devices(device_graph)
```
ancorgs: Same than above, I would expect this in the action. After all, it's about what devices to offer in the form.

src/lib/y2storage/bcache.rb
ancorgs: Take into account that bcache support in the library is in Hackweek state, so don't expect to see everything there.

Of course, I also prefer to read and test Ruby code, but for consistency with other parts in which available numbers and/or names are calculated, I honestly believe this should go to the library.

Sorry dude. :wink:

src/lib/y2storage/bcache.rb
jreidinger: ok

jreidinger: so action should pass it to dialog and dialog should provide it to widget, right?

src/lib/y2storage/bcache_cset.rb
jreidinger: well, one view is that it is devices for form, another view can be that it is devices where bcache can be created without destruction of data. But I can move it.

ancorgs:
About codeclimate. 

The duplication can surely be fixed/improved if that piece of code is moved to the action. Deciding which devices to offer in the form is partitioner-specific code. For example, the fact that we do not offer devices that are already part of an LVM and so on has nothing to do with the Y2Storage possibilities, it's a usability decision.

And of course, reporting than having different classes descending from `CWM::PushButton` is a code duplication is a completely stupid false positive.

src/lib/y2storage/bcache.rb
ancorgs: Yes, that's my idea. Pretty much like it's done here https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/actions/create_partition_table.rb#L167

src/lib/y2storage/bcache_cset.rb
ancorgs: Not really. For example, here we include devices that contain a filesystem as long as that filesystem has no mount point assigned currently. That can imply a very severe destruction of data, but it's ok to offer such devices from the Partitioner point of view.

src/lib/y2storage/bcache.rb
```diff
@@ -55,12 +60,16 @@ class Bcache < Partitionable
...
+    #   @param devicegraph [Devicegraph] in which search for free name
+    #   @return [String] full path to new bcache device like "/dev/bcache3"
+    storage_class_forward :find_free_name
+
     def inspect
       "<Bcache #{name} #{bcache_cset.inspect} -> #{blk_device}>"
     end
 
-  protected
-
```
ancorgs: Why removing this `protected`?

jreidinger: fat fingers in vim :) Will fix.

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,204 @@
...
+              [dev.sid.to_s, dev.name]
+            end
+          end
+        end
+
+        # @macro seeAbstractWidget
+        def help
+          # TRANSLATORS: %s stands for name of option
+          format(_(
+                   "<b>%s</b> is device that will be used as caching device for bcache." \
```
ancorgs: is **the** device that...

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,204 @@
...
+                   "It should be faster and usually is smaller then backing device, " \
```
ancorgs: smaller **than the**  backing device

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,204 @@
...
+                   "but it is not required. " \
+                   "Device will be reformatted so any previous content will be wiped out."
```
ancorgs: **The** device will be **formatted** so...

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,204 @@
...
+          # value can be empty string if there is no items
+          if value.nil? || value.empty?
+            Yast2::Popup.show(
+              _("Empty backing device is not yet supported"),
+              headline: _("Cannot Create Bcache")
+            )
+            return false
+          elsif value == @caching.value
+            Yast2::Popup.show(
+              _("Backing and Caching device cannot be identical."),
```
ancorgs: devices, in plural.

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,204 @@
...
+          @devices.map do |dev|
+            [dev.sid.to_s, dev.name]
+          end
+        end
+
+        # @macro seeAbstractWidget
+        def help
+          # TRANSLATORS: %s stands for name of option
+          format(_(
+                   "<b>%s</b> is device that will be used as backing device for bcache." \
```
ancorgs: is **the** device that will be used.

Or alternatively

is a device to be used

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,204 @@
...
+                   "It will define available space of bcache. " \
```
ancorgs: will define **the** available

or alternatively something more explicit like:

The resulting Bcache device will have the same size than this backing device.

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -0,0 +1,204 @@
...
+                   "Device will be reformatted so any previous content will be wiped out."
```
ancorgs: **The** device will be **formatted**

PR #750, opened 2018-09-27
Sle 15 ga bsc 1108831

schubi2:


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Sep 27 15:13:31 CEST 2018 - schubi@suse.de
+
+- AutoYaST proposal: Do not crash if existing boot partition
+  cannot be used without formating it. (bsc#1108831)
```
imobachgs: formatting

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -161,7 +161,15 @@ def boot_devices(devicegraph, devices)
       mountable_devices = find_mountable_devices(devices)
       return unless root?(mountable_devices)
       checker = BootRequirementsChecker.new(devicegraph, planned_devices: mountable_devices)
-      checker.needed_partitions
+      begin
+        result = checker.needed_partitions
+      rescue BootRequirementsChecker::Error
+        issues_list.add(:could_not_create_boot)
+        log.warn checker.warnings.to_s unless checker.warnings.empty?
```
imobachgs: `BootRequirementsChecker#warning` and `#error` methods will return an array of `SetupError` instances, so I am not sure about how those messages will look like in the log.

Could you please check it?

schubi2: I have tested. It looks like:
[#<Y2Storage::SetupError:0x0000000003bb5e60 @my_textdomain=["storage"], @message="MBR gap size is not enough to correctly install bootloader", @missing_volume=nil>, #<Y2Storage::SetupError:0x0000000003bb5a00 @my_textdomain=["storage"], @message="MBR gap size is not enough to correctly install bootloader", @missing_volume=nil>, #<Y2Storage::SetupError:0x0000000003bb5578 @my_textdomain=["storage"], @message="Boot disk has no partition table and it is not possible to boot from it.You can fix it by creating a partition table on the boot disk.", @missing_volume=nil>]
[#<Y2Storage::SetupError:0x0000000003bb4128 @my_textdomain=["storage"], @message="There is no device mounted at '/'", @missing_volume=nil>]

schubi2: I have tested it looks like e.g.:
warning:

[#<Y2Storage::SetupError:0x0000000003f1f1c8 @my_textdomain=["storage"], @message="MBR gap size is not enough to correctly install bootloader", @missing_volume=nil>, #<Y2Storage::SetupError:0x0000000003f1ecc8 @my_textdomain=["storage"], @message="Boot disk has no partition table and it is not possible to boot from it.You can fix it by creating a partition table on the boot disk.", @missing_volume=nil>]

error:

[#<Y2Storage::SetupError:0x0000000003f1d710 @my_textdomain=["storage"], @message="There is no device mounted at '/'", @missing_volume=nil>]

imobachgs: To be honest, I would prefer dumping the messages and not the full ruby objects. But we are already doing that in other storage-ng stuff.

PR #752, opened 2018-10-01
New Partitioner UI based on dynamic menu-buttons

ancorgs:
Main PR for https://trello.com/c/5rrmzwKh/276-3-partitioner-unleash-the-beast


src/lib/y2partitioner/widgets/device_buttons_set.rb
```diff
@@ -82,33 +86,64 @@ def refresh
       def calculate_buttons
         return [] if device.nil?
 
-        if device.is?(:partition)
-          partition_buttons
-        else
-          raid_buttons
+        types = [:partition, :software_raid, :lvm_vg, :lvm_lv, :stray_blk_device]
```
jreidinger: this looks like constant for list of supported device types?

src/lib/y2partitioner/widgets/device_menu_button.rb
```diff
@@ -0,0 +1,138 @@
...
+
+      # Device on which to act
+      #
+      # @return [Y2Storage::Device]
+      def device
+        return nil unless device_sid
+        working_graph.find_device(device_sid)
+      end
+
+      # @return [Array<[Symbol, String]>] list of menu options
```
jreidinger: well, this comment is a bit imprecise. I prefer here simple see for CWM method which defines how it exactly have to look.

src/lib/y2partitioner/widgets/device_menu_button.rb
```diff
@@ -0,0 +1,138 @@
...
+        actions.find { |action| widget_id_for_action(action) == id.to_sym }
+      end
+
+      # Checks whether there is a device on which to act
+      #
+      # @note An error popup is shown when there is no device.
+      #
+      # @return [Boolean]
+      def validate_presence
+        return true unless device.nil?
```
jreidinger: NP: but me is more nature `return true if device`

src/lib/y2partitioner/widgets/lvm_devices_table.rb
```diff
@@ -34,7 +34,8 @@ class LvmDevicesTable < ConfigurableBlkDevicesTable
       # @param devices [Array<Y2Storage::Lvm_vg, Y2Storage::Lvm_lv>] devices to display
       # @param pager [CWM::Pager] table have feature, that double click change content of pager
       #   if someone do not need this feature, make it only optional
-      def initialize(devices, pager)
+      # @param buttons_set [DeviceButtonsSet]
```
jreidinger: please document also why it is needed.

src/lib/y2partitioner/widgets/partition_add_button.rb
```diff
@@ -27,24 +27,16 @@ module Y2Partitioner
...
     class PartitionAddButton < DeviceButton
-      # Constructor
-      # @param pager [CWM::TreePager]
-      # @param table [Y2Partitioner::Widgets::ConfigurableBlkDevicesTable]
-      # @param device [Y2Storage::Device]
-      # @param short [Boolean] whether to display a short "Add" label
-      def initialize(pager: nil, table: nil, device: nil, short: false)
-        super(pager: pager, table: table, device: device)
-        @short_label = short
+      # Constructor. See parent class.
```
jreidinger: I worry this see won;t create automatic link in yardoc

src/lib/y2partitioner/dialogs/md.rb
```diff
@@ -33,6 +33,8 @@
 #   @see http://www.rubydoc.info/github/yast/yast-yast2/CWM%2FCustomWidget:${0}
 # @!macro [new] seeDialog
 #   @see http://www.rubydoc.info/github/yast/yast-yast2/CWM%2FDialog:${0}
+# @!macro [new] seeItemsSelection
+#   @see http://www.rubydoc.info/github/yast/yast-yast2/CWM%2FItemsSelection:${0}
```
jreidinger: it looks familiar for me :) https://github.com/yast/yast-storage-ng/pull/746/files#diff-e9f6c27fd34d282cdefbf6f09306cccdR36

src/lib/y2partitioner/widgets/lvm_devices_table.rb
ancorgs: Actually, that whole documentation should not be there, in `LvmDevicesTable`, which is not the base class. `#devices` and `#pager` are already (better) documented in the base class.

src/lib/y2partitioner/ui_state.rb
```diff
@@ -38,6 +36,26 @@ def initialize
...
 
+    # Title of the section listing the MD RAIDs
+    #
+    # @note This is defined in this class as the simplest way to avoid
+    #   dependency cycles in the Ruby requires. We might reconsider a more clean
+    #   approach in the future.
+    #
+    # @return [String]
+    def self.md_raids_label
+      N_("RAID")
```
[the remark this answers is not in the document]
ancorgs: I just moved stuff from the previous place in a hurry (kind of). With UIState being a singleton class, it makes indeed no sense to keep the old `N_` thingie.

src/lib/y2partitioner/widgets/device_menu_button.rb
```diff
@@ -129,6 +129,7 @@ def widget_id_for_action(action)
       # @param id [Symbol, String]
       # @return [Hash]
       def action_for_widget_id(id)
+        return nil if id.nil?
```
jreidinger: should be nil or exception? if so, then please document it that it can also return nil.

package/yast2-storage-ng.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:	4.1.20
+Version:        4.1.21
```
jreidinger: we should finally fix that indentation issue :)

src/lib/y2partitioner/widgets/device_menu_button.rb
ancorgs: Actually, I would never expect `#handle` to be called with no arguments (or with a nil argument) since `handle_all_events` is used.

Unfortunately, supporting that case is needed in order to be able to use `include_examples "CWM::AbstractWidget"` which always calls `#handle` with no arguments, so I have to adapt the class to ignore such wrong calls.

jreidinger: ah, yeah. That is problematic, because for calls it uses arity and rspec mock basically mocks call with *args, so -1 arity and it breaks this test. Please add comment for it, as it is quite tricky.

PR #753, opened 2018-10-01
Backport BlkDevice#component_of

imobachgs:
This coded has been backported from `master` because it is needed to find out when a drive should be included in an (exported) AutoYaST profile.


imobachgs:
Thanks!

PR #754, opened 2018-10-01
Complete support for using a disk as an LVM PV

imobachgs:
Recently we added some minimal support to use a whole disk as a physical volume on AutoYaST, a feature which is not even documented. The trick is that, if you set `partition_nr` to `0` for a partition, AutoYaST will not create any partition but the whole disk will be used.

After having the basic use case working, we have decided to improve/extend the support for this scenario:

* First of all, we have decided to deprecate the `partition_nr` hack. Now, to use a whole disk without partitioning it, you must set `disklabel` to `none` in the drive section. Of course, we still support the old way for backward compatibility reasons.
* Cloning support has been added.
* BTW, we have improved the AutoYaST proposal internals.

Currently, we support using the disk as an LVM PV and formatting the whole disk as a filesystem too. However, for this second use case, we should improve user's feedback (because if you do not set `format` to true and the filesystem does not exist yet, it will simply skip it silently).

Support for MD RAIDs will be implemented soon.

## Proposal

use-disk-as-pv

## pvscan

use-disk-as-pv-pvscan

## Exported profile differences

```diff
--- /root/autoinst.xml.backup	2018-10-01 10:56:20.136000000 -0400
+++ /root/autoinst.xml	2018-10-01 10:56:40.840000000 -0400
@@ -426,6 +426,22 @@
       <type config:type="symbol">CT_DISK</type>
       <use>all</use>
     </drive>
+    <drive>
+      <device>/dev/vdb</device>
+      <disklabel>none</disklabel>
+      <enable_snapshots config:type="boolean">false</enable_snapshots>
+      <initialize config:type="boolean">false</initialize>
+      <partitions config:type="list">
+        <partition>
+          <create config:type="boolean">false</create>
+          <format config:type="boolean">false</format>
+          <lvm_group>system</lvm_group>
+          <resize config:type="boolean">false</resize>
+        </partition>
+      </partitions>
+      <type config:type="symbol">CT_DISK</type>
+      <use>all</use>
+    </drive>
   </partitioning>
   <proxy>
     <enabled config:type="boolean">false</enabled>
```


src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -266,9 +290,19 @@ def partitions_from_hash(hash)
...
+      # Return the partition sections for the given disk
+      #
+      # @note If there is no partition table, an array containing a single section
+      #   (which represents the whole disk) will be returned.
+      #
+      # @return [Array<AutoinstProfile::PartitionSection>] List of partition sections
       def partitions_from_disk(disk)
-        collection = disk.partitions.reject { |p| skip_partition?(p) }
-        partitions_from_collection(collection.sort_by(&:number))
+        if disk.partition_table
```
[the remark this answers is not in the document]
jreidinger: ah, it is disk.partition_table and not local method. :duck:

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -252,7 +252,7 @@ def init_from_device(device)
         # unit (that equals to 8225280 bytes in my experiments).
         # According to the comments there, that was done due to bnc#415005 and
         # bnc#262535.
-        @size = device.size.to_i.to_s if create && !device.is?(:md)
+        @size = device.size.to_i.to_s if create && !(device.is?(:md) || device.is?(:disk))
```
jreidinger: what about e.g. multipath? why not use `:disk_device` ?

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -289,6 +289,8 @@ def init_fields_by_type(device)
           init_lv_fields(device)
         elsif device.is?(:md)
           init_md_fields(device)
+        elsif device.is?(:disk)
```
jreidinger: and same here.

src/lib/y2storage/planned/devices_collection.rb
```diff
@@ -0,0 +1,120 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "forwardable"
+
+module Y2Storage
+  module Planned
+    # This class holds a list of planned devices and offers an API to query the collection
+    #
+    # @example Add devices
```
jreidinger: example missing?

src/lib/y2storage/planned/devices_collection.rb
```diff
@@ -0,0 +1,120 @@
...
+      end
+
+      # Returns a new instance including the devices
+      #
+      # @param devices [Array<Y2Storage::Planned::Device>,Y2Storage::Planned::Device]
+      #   Devices to add
+      # @return [DevicesCollection]
+      def append(devices)
+        self.class.new(@devices + devices)
+      end
```
jreidinger: maybe document that class is immutable?

src/lib/y2storage/planned/devices_collection.rb
```diff
@@ -0,0 +1,120 @@
...
+      # @return [Array<Planned::StrayBlkDevice>]
+      def stray_blk_devices
+        devices.select { |d| d.is_a?(Planned::StrayBlkDevice) }
+      end
+
+      # Returns all devices, including nested ones
+      #
+      # @return [Array<Planned::Device>]
+      def all
+        partitions + disks + stray_blk_devices + vgs + lvs + mds
```
jreidinger: class is immutable. Can be element changed? if not, here it can be easily cached.

PR #755, opened 2018-10-01
[WIP] Cache downcasting map

jreidinger:
Measurement ( done by accumulating Time.now difference ) from simple start of module is that without cache it took

0.29208857400000154 seconds in this method with original implementation
and 0.07035962099999996 when cache is involved. Reason is huge number of
calls.

### DIFF
--- a/src/lib/y2storage/storage_class_wrapper.rb
+++ b/src/lib/y2storage/storage_class_wrapper.rb
@@ -231,15 +231,11 @@ module Y2Storage
       # @param object [Object] storage object to be wrapped
       def downcasted_new(object)
         @downcast_class_names.each do |class_name|
-          klass = StorageClassWrapper.class_for(class_name)
-          storage_class = klass.storage_class
+          map = StorageClassWrapper.downcast_map(class_name)
 
-          underscored = StorageClassWrapper.underscore(storage_class.name.split("::").last)
-          check_method = :"#{underscored}?"
-          cast_method = :"to_#{underscored}"
-          next unless Storage.public_send(check_method, object)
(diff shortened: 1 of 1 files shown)


mvidner:
Doing all this downcasting at the ruby level seems stupid, I think SWIG should do this automatically. Maybe this is helpful: https://stackoverflow.com/questions/27392602/swig-downcasting-from-base-to-derived `SWIG_TypeQuery`, `SWIG_NewPointerObj`

src/lib/y2storage/disk_device.rb
```diff
@@ -71,16 +71,18 @@ def types_for_is
     #
     # @return [Boolean]
     def disk_device?
+      return @disk_device unless @disk_device.nil?
+
       # If we cannot create partitions on it, this doesn't look like a disk
-      return false if respond_to?(:usable_as_partitionable?) && !usable_as_partitionable?
+      return @disk_device = false if respond_to?(:usable_as_partitionable?) && !usable_as_partitionable?
```
mvidner: The bicycle cop strikes:
> Cyclomatic complexity for disk_device? is too high. [7/6]

jreidinger: I know, trying to solve it

mvidner:
Yay, tests pass! How about this for .changes:
```
Improve UI responsiveness by caching
- StorageClassWrapper.downcasted_new
- StorageEnv#active?
- DiskDevice#types_for_is
```

ancorgs:
@jreidinger 's commit LGTM. So we only need to add the changelog+version and we are all set!

PR #756, opened 2018-10-02
document boot requirements with raid setups

wfeldt:


doc/raid.md
```diff
@@ -0,0 +1,73 @@
+# RAID setups - what works?
+
+## General considerations
+
+- How much stacking of multipath+raid+enc+lvm is allowed for /boot?
+- Avoid bcache for bootloader_partition and /boot?
```
jreidinger: I am a bit lost what do bcache here in raid document. What is relation?

doc/raid.md
```diff
@@ -0,0 +1,73 @@
...
+    - ! non-efi: grub2-install blocked in D-state for a very long time (3+ min)
+
+    - ! non-efi: pointing grub_installdevice to raid1 device directly results in error:
+      ```sh
+      grub2-install: error: diskfilter writes are not supported.
+      ```
+
+    - ! x86, msdos: wrong warning about missing bios_grub partition
+
+    - ! efi: (incorrect?) warning about esp on software raid
```
jreidinger: well, it was discussed in respective feature. Problem is that we depends on firmware what it supports and how much firmware checks ESP partition. In general we do not want to guarantee that it works, but you should be able to accept warning.

doc/raid.md
```diff
@@ -0,0 +1,73 @@
...
+      ```
+
+    - ! efi: no boot manager entry (in firmware) created
+
+- raidX over partitions, /boot on raidX, bootloader_partition not on raid
+    - works, grub_installdevice contains N physical disks
+    - efi: note there can be only one esp
+    - ! non-efi: need bootloader_partition on every disk constituting raidX or
+      change yast-bootloader to only install on those that have it (there should
+      be at least 1)
```
jreidinger: idea was that you need it on all disk to be able to boot from whatever you want. But we also discussed that we can relax installation failure if we can install it to at least some of the disks.

PR #757, opened 2018-10-02
fixed flickering testcase

schubi2:


ancorgs:
I wonder if fixing the testcase is good enough. It looks to me like that require is also missing in `src/lib/y2partitioner/widgets/fstab_options.rb` which is probably worse (and the root case).

PR #759, opened 2018-10-02
implement remove_bcache

jreidinger:


src/lib/y2partitioner/actions/delete_bcache.rb
```diff
@@ -0,0 +1,97 @@
...
+      def delete
+        log.info "deleting bcache raid #{device}"
+        device_graph.remove_bcache(device)
+      end
+
+      # Confirmation before performing the delete action
+      #
+      # @return [Boolean]
+      def confirm
+        if device.partitions.any?
```
jreidinger: I am not sure how to react if whole bcache device is used as source for something else like LVM or md raid. Should I check that there is no blk devices in children or predecessors of device?

ancorgs: Why would we need to check the predecessors?

Checking whether it has descendants and asking for confirmation about deleting those should be enough. By the way, that's something that I kind of expected to see in the base class `Actions::DeleteDevice`

jreidinger: yeah, me too. I am not sure if check for any descendants as it can be only empty partition_table or fs, which is probably fine, not?

PR #760, opened 2018-10-02
Do not crash when reusing a non-existent filesystem

imobachgs:
This PR extends AutoYaST to show an error when trying to reuse a non-existent filesystem. It is a fatal error because we do not know why the filesystem is not there (maybe it is an encrypted one and the user did not provide the password, or the probing was wrong, or whatever...). In such a case, we need the user to explicitly set `format=true` so we can create a new one in case it is missing.

reuse-missing-filesystem-error


src/lib/y2storage/autoinst_issues/missing_reusable_filesystem.rb
```diff
@@ -0,0 +1,56 @@
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
```
ancorgs: 2018

PR #762, opened 2018-10-04
Reorganize/modernize hierarchy of `Actions::DeleteDevice`

ancorgs:
This PR is a consequence of the conversation at https://github.com/yast/yast-storage-ng/pull/759#discussion_r222392912

We realized we were not sure about the correct responsibility and message from `Actions::DeleteDevice` and all its corresponding subclasses, so I decided to reorganize the code.

First of all, I found an action that is not longer used, so this contains a commit to remove it.

The second commit reorganizes the `Actions::DeleteDevice` hierarchy to share the checks using more generic messages, so we don't have to reinvent the  wheel in every subclass.

Last but not least, this reorganizes the "Overview" tab of a bcache device. It was using the same buttons than a disk, so there was not "Delete" button and it contained some actions that shouldn't be there (apart from containing "Hard Disk" in the title and similar inconsistencies).


ancorgs:
@jreidinger So this is the status. How do we move from there to something more sane?

jreidinger:
@ancorgs well, I suggest to follow delete_partitions and also use all descendants as base ( even if it means that we no longer use specialized messages ). For specific cases it can be extended ( like bcache with single cset that will be also removed, but it is not descendant). Errors makes sense for me to be specialized. I am just not sure if all error messages now makes sense.

src/lib/y2partitioner/widgets/pages/bcache.rb
```diff
@@ -0,0 +1,109 @@
...
+        # Constructor
+        #
+        # @param bcache [Y2Storage::Bcache]
+        # @param pager [CWM::TreePager]
+        def initialize(bcache, pager)
+          textdomain "storage"
+
+          @bcache = bcache
+          @pager = pager
+          self.widget_id = "bcache:" + bcache.name
```
jreidinger: I though that constructor of widget id is at some centralized place, as it is needed also later, not?

ancorgs: As commented on IRC, that's a different part of the code. This piece just looks like all its sibling pages.

PR #763, opened 2018-10-10
Add support for partitioned MD RAID devices to AutoYaST

imobachgs:
This PR adds support to use partitioned MD RAID devices to AutoYaST. By the way, it improves
the overall AutoYaST proposal code quality.

Basically, it allows to partition an MD RAID.

```xml
<drive>
  <device>/dev/md/0</device>
  <create config:type="boolean">false</create>
  <raid_options>
    <raid_type>raid0</raid_type>
  </raid_options>
  <partitions config:type="list">
    <partition>
      <filesystem config:type="symbol">ext4</filesystem>
      <mount>/</mount>
      <format config:type="boolean">true</format>
      <size>50%</size>
    </partition>
  </partitions>
  <type config:type="symbol">CT_MD</type>
  <use>all</use>
</drive>
```

Although the old format will be supported, the user can achieve the same behaviour
using setting `disklabel` to `none` (as you do to use a disk without creating partitions)

```xml
<drive>
  <device>/dev/md/0</device>
  <create config:type="boolean">false</create>
  <disklabel>none</disklabel>
  <raid_options>
    <raid_type>raid0</raid_type>
  </raid_options>
  <partitions config:type="list">
    <partition>
      <filesystem config:type="symbol">ext4</filesystem>
      <mount>/</mount>
      <format config:type="boolean">true</format>
      <size>50%</size>
    </partition>
  </partitions>
  <type config:type="symbol">CT_MD</type>
  <use>all</use>
</drive>
```

## Additional changes

* When a percentage is specified for a LV or a RAID partition, it is not possible to find out the
  final size until the VG or the MD RAID exists in the Devicegraph. This problem does not exist
  with disk partitions. However, in order to make the code easier to reason about, we are
  postponing the calculation in all cases.
* The logic to reuse VGs and MD RAIDs has been moved to `LvmCreator` and `MdCreator` classes.

## Next steps

* Support the old format again.
* When generating the new Devicegraph, AutoYaST should try to 1) reuse devices and 2) create
  devices. Currently it is done the other way around (which is wrong).
* Remove raid options from the partition section while exporting.

## See it in action!

In this example, AutoYaST is creating a partitioned RAID and using the partition as an LVM physical volume.

lvm-on-partitioned-raid


src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -49,7 +49,8 @@ def self.attributes
           { name: :pesize },
           { name: :type },
           { name: :use },
-          { name: :skip_list }
+          { name: :skip_list },
+          { name: :raid_options }
```
ancorgs: I'm just starting the review but, why has been this added to `<drive>`? I don't see the reason at first sight

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -305,6 +336,23 @@ def flexible_devices(devices)
...
+      # sizes adjusted.
+      #
+      # @param planned_partitions [Array<Planned::Partition>] List of planned partitions
+      # @return [Array<Planned::Partition>] New list of planned partitions with adjusted sizes
+      def sized_partitions(planned_partitions)
+        planned_partitions.map do |part|
+          new_part = part.clone
+          next new_part unless new_part.percent_size
+          disk = original_graph.find_by_name(part.disk)
+          new_part.max = new_part.min = new_part.size_in(disk)
```
ancorgs: I'm just wondering (I would really need to check)... since `#size_in` takes no alignment and such things into account... Will this work as expected? I mean, what happens if we set `min` and `max` to the same non-aligned value?

And in a similar fashion, can it happen that a drive with two partitions defined as 50% each ends up with unused space due to rounding?

I know these questions are probably not new, but something that was already present (and properly solved) in the algorithms since the beginning. But every time I see this I cannot prevent my brain to start spinning.

src/lib/y2storage/autoinst_profile/drive_section.rb
imobachgs: Because we have introduced partitioned RAIDs. In the past, each partition section defined a different RAID, so the RAID options were defined at that level.

However, now that each partition can define a partition itself, I think we should define these options at `drive` level.

src/lib/y2storage/proposal/autoinst_devices_creator.rb
ancorgs: As discussed in a call, @imobachgs will create a separate PBI about checking and/or unit-testing several situations regarding grain and alignment in AutoYaST like:

- Two partitions/LVs with size 50% when the capacity of the disk/VG is uneven (let's say 101 extents available).
- What happens if we set min and max to the same value which is not divisible by the grain
- What happens if we set min and max to the same value that is smaller than the grain
- Whether the final 16 KiB in a GPT can break some AutoYaST profile somehow

Our expectations is that all those situations are properly handled by the following steps (the ones that create the real devices in the devicegraph based on the Planned Devices generated by this part), but double checking and having unit tests will be nice.

PR #764, opened 2018-10-11
Better handling of the boot requirements for GPT and BIOS booting (legacy, no UEFI)

ancorgs:
This will be the main PR for https://trello.com/c/Obpi1nzI/363-3-better-support-for-installing-grub-into-protective-mbr

So far it only contains a commit to remove the misleading FIXME and improve the documentation.

More commits will be added to make that information reach the user.


src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -148,7 +148,15 @@ def errors_on_gpt
         errors = []
 
         if grub_part_needed_in_gpt? && missing_partition_for?(grub_volume)
-          errors << SetupError.new(missing_volume: grub_volume)
+          message = _(
+            "There is no partition of type BIOS Boot.\n" \
+            "Although other configurations would eventually make possible\n" \
```
wfeldt: missing `it`

wfeldt: Personally, I find the message is a bit too long.

wfeldt: There is no partition of type BIOS Boot.

It is strongly advised to use a BIOS Boot partition.

A setup without such a partition is not supported and may cause unnecessary problems later.

wfeldt: BTW, your mockup shows no effect of the line breaks. Maybe html markup is needed?

ancorgs: I guess I will just drop the line breaks, they would look weird in a `<li>` anyways.

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -149,12 +149,10 @@ def errors_on_gpt
...
           message = _(
-            "There is no partition of type BIOS Boot.\n" \
-            "Although other configurations would eventually make possible\n" \
-            "to boot the system, the usage of a BIOS Boot partition is\n" \
-            "the only encouraged and officially documented method for\n" \
-            "booting Linux from a disk with a GPT partition table in\n" \
-            "BIOS legacy mode (i.e. without EFI)."
+            "There is no partition of type BIOS Boot. " \
+            "For GPT devices it is strongly advised to use a BIOS Boot partition. " \
+            "A setup without such partition is not supported and may cause " \
```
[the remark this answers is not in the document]
ancorgs: argh. I will do a commit --amend for that

PR #765, opened 2018-10-11
add attributes to bcache

jreidinger:
Manual testing showed problems with reboot:

1. cache_mode works as expected after reboot
2. sequential cutoff and writeback_percent is reset.

Another problem is really rename in memory which result in failure as name is different then expected. Way to reproduce it is to create bcache0, then bcache1 and then delete bcache0.

## Conclusion

Rename was solved in libstorage-ng. For sequential cutoff and writeback percent as we do not get answer yet from Coly, we add only cache_mode, but commit to remove them can be easily reverted.

snimek obrazovky_2018-10-18_17-55-32


src/lib/y2storage/cache_mode.rb
```diff
@@ -0,0 +1,58 @@
...
+    }
+    private_constant :TRANSLATIONS
+
+    # Returns human readable representation of enum which is already translated.
+    # @return [String]
+    # @raise [RuntimeError] when called on enum value for which translation is not yet defined.
+    def to_human_string
+      textdomain "storage"
+
+      string = TRANSLATIONS[to_sym] or raise "Unhandled Cache mode value '#{inspect}'"
```
ancorgs: If we are going to raise an exception here, it would be good to have a testcase making Travis/Jenkins explode is a new constant appears in libstorage-ng, instead of letting the user find it.

src/lib/y2partitioner/dialogs/bcache.rb
```diff
@@ -199,6 +215,60 @@ def store
...
+                   "<i>Writethrough</i> is default mode which caches reading and do write parallel. " \
+                   "So in case of failure of caching device, no data is lost." \
+                   "<i>Writeback</i> is mode that caches also writes. This result in better " \
+                   "performance of write operations." \
+                   "<i>Writearound</i> is mode that use cache only for reads and writes new " \
+                   "content only to backing device." \
+                   "<i>None</i> is mode that do not use cache for read neither for write. " \
+                   "Useful mainly for temporary disabling cache before big sequential read/write " \
+                   "which otherwise result in intensive overwritting of cache device."
+          ), label)
```
ancorgs: Before entering into idiomatic suggestions, I miss some formatting here. The help of this dialog looks ugly as hell:

help-unformatted

ancorgs: Btw, while I'm reading the whole thing. It says "the caching device will be formatted so any previous content will be wiped out". Is that also true if the selected caching device is an existing bcache cset? Just curious

jreidinger: what version of CWM you use? It looks like bug of CWM itself that it does not separate individual entries. It should be separated by <p> each option.

jreidinger: I can try how ul will look for cache mode itself.

ancorgs: And now, my suggestions about wording. Nothing too big.

* _Writethrough_ reading operations are cached, writing is done in parallel to both devices. In case of failure of the caching device, no data is lost. This is the default mode.
* _Writeback_ both reading and writing operations are cached. This result in better performance when writing.
* _Writearound_ reading is cached, new content is written only to the backing device.
* _None_ disables all caching both for reading and writing. This is useful mainly for temporarily disabling the cache before any big sequential read or write, otherwise that would result in intensive overwriting on the cache device.

jreidinger: I tune a bit to something like, even list looks good:

snimek obrazovky_2018-10-19_12-56-53

ancorgs:
And another missing thing: adding `y2storage/cache_mode'  to `y2storage.rb` so it's included when we do `require "y2storage"`.

src/lib/y2partitioner/dialogs/bcache.rb
jreidinger: > Btw, while I'm reading the whole thing. It says "the caching device will be formatted so any previous content will be wiped out". Is that also true if the selected caching device is an existing bcache cset? Just curious

if it is existing bcache cset, then it is just attached and not created. So not wiped out.

PR #767, opened 2018-10-15
Limit bcache

jreidinger:


src/lib/y2partitioner/actions/add_bcache.rb
```diff
@@ -70,7 +70,9 @@ def usable_blk_devices
         device_graph.blk_devices.select do |dev|
           dev.component_of.empty? &&
             (dev.filesystem.nil? || dev.filesystem.mount_point.nil?) &&
-            (!dev.respond_to?(:partitions) || dev.partitions.empty?)
+            (!dev.respond_to?(:partitions) || dev.partitions.empty?) &&
+            # do not allow nested bcaches, see doc/bcache.md
+            !dev.ancestors.any? { |a| a.is?(:bcache) }
```
ancorgs: Wouldn't this be more readable using `#none?` instead of a negated `#any?`

Moreover, what about checking also for CSets to be even safer?

Something like:

```ruby
dev.ancestors.none? { |a| a.is?(:bcache, :bcache_cset) }
```

src/lib/y2partitioner/actions/delete_bcache.rb
```diff
@@ -35,6 +35,28 @@ def initialize(*args)
...
+
+      # Error when the bcache shares cset with any other bcache.
+      #
+      # @see doc/bcache.md
+      # @return [String, nil] nil if there is no error
+      def shared_cset
+        return nil unless device.bcache_cset
+        return nil if single_bcache_cset? || !device.exists_in_probed?
+
+        # TRANSLATORS: Error when trying to existing bcache that shares caches.
```
mvidner: trying to what?
"when trying to modify an existing bcache"?

src/lib/y2partitioner/actions/delete_bcache.rb
```diff
@@ -35,6 +35,28 @@ def initialize(*args)
...
+        errors.compact
+      end
+
+      # Error when the bcache shares cset with any other bcache.
+      #
+      # @see doc/bcache.md
+      # @return [String, nil] nil if there is no error
+      def shared_cset
+        return nil unless device.bcache_cset
+        return nil if single_bcache_cset? || !device.exists_in_probed?
```
ancorgs: In general it's preferred to use `Y2Partitioner::DeviceGraph` to decouple the code from the existence of the "holy and omnipresent" devicegraphs `probed` and `staging`.

Like here https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/actions/edit_md_devices.rb#L140

src/lib/y2partitioner/actions/delete_bcache.rb
ancorgs: The comments targeting the translators are useful to tell them what the placeholders mean or whether something is a title or a pop-up. But at some point we started to introduce comments explaining what is already said in the string they have to translate. Pretty useless. :wink:

src/lib/y2partitioner/actions/delete_bcache.rb
```diff
@@ -35,6 +35,28 @@ def initialize(*args)
...
+
+        # TRANSLATORS: Error when trying to existing bcache that shares caches.
+        _(
+          "The bcache shares cache set with other devices which can result in unreachable space " \
+            "if done without detach. But detach can take in some situation very long time."
```
mvidner: > The bcache shares a(?) cache set with other devices. This can result in unreachable space if done without detaching. Detaching can take a very long time in some situations.

Depending on the cardinality of the bcache-cset relation, it's "shares a cache set", "shares the cache set" or "shares its cache set".

ancorgs: Hard to read. Let me try to suggest an alternative:

This device shares its cache set with another bcache device. It's not safe to delete this device without detaching the cache set first. YaST cannot perform such operation, it may take a very long time.

doc/bcache.md
```diff
@@ -0,0 +1,17 @@
...
+
+bcache (abbreviated from block cache) is a cache in the Linux kernel's block layer,
+which is used for accessing secondary storage devices. It allows one or more
+fast storage devices, such as flash-based solid-state drives (SSDs), to act as
+a cache for one or more slower storage devices, such as hard disk drives (HDDs);
+this effectively creates hybrid volumes and provides performance improvements.
+
+### Limitations in YaST
+
+As name suggest in general it can be used on top of any block device. But YaST limits its usage.
```
mvidner: http://tinyurl.com/this-verb-form-are-wrong please fix

src/lib/y2partitioner/actions/delete_bcache.rb
ancorgs: BTW, what if the bcache is actually detached (because the user did it before running the Partitioner). How would that be seen by our probing mechanism? Would the bache device and its detached cset be part of the system devicegraph.

src/lib/y2partitioner/actions/add_bcache.rb
jreidinger: well, none looks good for me. About that CSet it is a bit tricky as Cset is not blocking device and cannot have any children. So ancestor cannot be Cset. But it does not hurt.

src/lib/y2partitioner/actions/delete_bcache.rb
jreidinger: @ancorgs detached bcache does not have caching set, so we do not see any caching device. I actually show it during initial review, as one of my testing bcache has detached cache.

jreidinger: @ancorgs so in this case, it works as expected as no caching set is there, so nothing blocks removal of bcache itself.

jreidinger: and about that cset, probably you will see it in device graph visualization, but nowhere else.

ancorgs: Ok. Visualization of the csets is probably another of the topics we have to think about. But not now.

jreidinger: you can of course use detached cset for another caching device. We can in fact even can allow to attach it to bcache that missing cache. see how it looks with detached cache set:

snimek obrazovky_2018-10-16_14-12-45

PR #768, opened 2018-10-16
Fix/export enable snapshots

imobachgs:


src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -439,6 +435,19 @@ def used?(disk)
...
         disk.partition_table ? disk.partition_table.type.to_s : NO_PARTITION_TABLE
       end
+
+      # Determines the value of the 'use' element for a disk/dasd device
+      #
+      # @param disk [Y2Storage::Disk, Y2Storage::Dasd] Disk
+      # @param partitions [Y2Storage::AutoinstProposal::PartitionSection] Set of partition sections
+      # @return [String] Value of the 'use' element for a disk.
+      def use_value_from_storage(disk, partitions)
+        if disk.partitions.any? { |i| windows?(i) }
```
ancorgs: I would document somewhere that this "logic" is inherited from the old times.

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -287,7 +287,7 @@ def partition_id_from(partition)
       def init_fields_by_type(device)
         if device.is?(:lvm_lv)
           init_lv_fields(device)
-        elsif device.is?(:disk_device)
+        elsif device.is?(:disk_device) || device.is?(:software_raid)
```
ancorgs: You could do `device.is?(:disk_device, :software_raid)`

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -420,6 +420,17 @@ def lvm_group_name(device)
...
+
+      # Determines whether the device has a fixed size (disk, RAID, etc.)
+      #
+      # It is used to find out whether the size specification should be included
+      # in the profile.
+      #
+      # @param device [Y2Storage::Device] Device
+      # @return [Boolean]
+      def fixed_size?(device)
+        device.is?(:disk_device) || device.is?(:software_raid)
```
ancorgs: Same than above

ancorgs:
As far as I can see, this introduces support for exporting that attribute in directly formatted disks and MDs. What about LVs?

imobachgs:
> As far as I can see, this introduces support for exporting that attribute in directly formatted disks and MDs. What about LVs?

Suprisingly, support for LVs was already implemented. I've tried it manually and it works.

PR #782, opened 2018-10-25
First version of temporary doc space_maker_improve_strategy.md

ancorgs:
Bootstraping the discussion about https://trello.com/c/Xiai2Fpa/369-3-define-a-better-spacemaker-strategy-brainstorm


joseivanlopez:
I like the idea of having a flexible way for defining `select_next_action`. This would give us quite a lot of room to adapt the algorithm to do "whatever" we want. The problem will be to find the priority between actions that most matches the user expectations. For example:

* Should we take into account all candidate disks when deciding next action? Or disk by disk?
* How to prioritize actions that leads to the same result, e.g., in this scenario [Linux][Free][Linux], and both Linux has the same size.

It is only a matter of deciding priorities, but I hope that priorities could be easily understandable for users.

ancorgs:
As a first step, I would just duplicate the current logic to a big extend. So we would have a SpaceMaker that acts in the same way than the current one but allow us to start tuning it step by step.

Something like:

```ruby
def select_next_action
  if linux_partitions.any?
    action_to_delete_the_end_linux_partition
  else
    if windows_partitions.any?(:not_resized_yet)
        action_to_resize_the_end_windows_partition
    else
        if other_partitions.any?
            action_to_delete_the_end_linux_partition
        ...
        end
    ...
    end
  ...
  end
end
```
  
The first step from there would be to introduce the exceptions requested in [fate#325885](https://fate.suse.com/325885) and [fate#323484](https://fate.suse.com/323484).

ancorgs:
I will make that idea more obvious by adding the content of my comment to the document

PR #783, opened 2018-10-26
Initial proposal tries to fit in only one disk first

dgdavid:
https://trello.com/c/hxkfVOnn/399-3-initial-proposal


src/lib/y2storage/guided_proposal.rb
```diff
@@ -227,11 +268,11 @@ def populated_settings
     # to smaller disk size.
     #
     # @return [Array<String>] names of the chosen devices
-    def candidate_roots
-      return [populated_settings.root_device] if populated_settings.root_device
+    def candidate_roots_for(devices)
```
joseivanlopez: What happens when the settings has one `root_device` and we are trying a proposal where devices has only one disk, but that disk is not the root device indicated in the settings? I guess no proposal is possible, isn't it?

joseivanlopez: I guess the old method `candidate_roots` would work perfectly if we use `populated_settings` as it was doing. So, maybe it is not necessary to re-implement it as `candidate_roots_for(devices)`, or am I missing something?

src/lib/y2storage/guided_proposal.rb
```diff
@@ -205,17 +205,58 @@ def candidate_devices_with_empty_partition_table(devicegraph)
...
+    # @return [Array<Symbol>]
+    def target_sizes
+      [:desired, :min]
+    end
+
     # Copy of the original settings including some calculated and necessary
     # values (mainly candidate_devices), in case they were not present
     #
     # @return [ProposalSettings]
     def populated_settings
```
joseivanlopez: NP: I would change the description of the method, because now, it does not add extra info to the settings by itself. I would say that "populated settings" are the settings used by each attempt of proposal, and some values could be "populated" during the process, for example, the candidate devices or the root device.

src/lib/y2storage/guided_proposal.rb
```diff
@@ -205,17 +205,58 @@ def candidate_devices_with_empty_partition_table(devicegraph)
...
+      return [settings.candidate_devices] if settings.candidate_devices
+      @candidate_devices ||= candidate_devices_from_settings || candidate_devices_from_analyzer
+    end
 
-      @populated_settings = populated
+    # List of candidate devices given in settings
+    #
+    # @return [<Array<Array<String>>]
+    def candidate_devices_from_settings
+      [settings.candidate_devices] if settings.candidate_devices
```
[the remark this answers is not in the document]
joseivanlopez: Umm, what happens when `candidate_devices` is given? Individual disks are not tried? Do we only want to use individual disks when candidate devices are not given?

src/lib/y2storage/guided_proposal.rb
```diff
@@ -205,17 +205,58 @@ def candidate_devices_with_empty_partition_table(devicegraph)
...
+    end
+
+    # List of candidate devices based on disks available in the analyzer
+    #
+    # By now, it consist on a sequence of available disk plus all of them at the end (unless
```
joseivanlopez: Typo: "it consists on"

src/lib/y2storage/guided_proposal.rb
```diff
@@ -205,17 +205,58 @@ def candidate_devices_with_empty_partition_table(devicegraph)
...
+    # there is only one, to avoid duplications). E.g, if available disks are
+    #
+    # ["/dev/sda", "/dev/sdb"]
+    #
+    # candidates list will be
```
joseivanlopez: s/list/devices

src/lib/y2storage/guided_proposal.rb
```diff
@@ -205,17 +205,58 @@ def candidate_devices_with_empty_partition_table(devicegraph)
...
+    #
+    # [["/dev/sda"], ["/dev/sdb"], ["/dev/sda", "/dev/sdb"]]
+    #
+    # @return [<Array<Array<String>>]
+    def candidate_devices_from_analyzer
+      disk_names = disk_analyzer.candidate_disks.map(&:name)
+      candidates = disk_names.zip
+      candidates << disk_names unless candidates.include?(disk_names)
+      candidates
```
joseivanlopez: NP: why not this version?

```suggestion
      disk_names = disk_analyzer.candidate_disks.map(&:name)
      disk_names.zip.append(disk_names).uniq
```

src/lib/y2storage/guided_proposal.rb
```diff
@@ -120,12 +120,12 @@ def calculate_proposal
       exception = nil
       saved_root_device = populated_settings.root_device
 
-      [:desired, :min].each do |target|
-
+      target_sizes.product(candidate_devices).each do |target, devices|
```
joseivanlopez: NP: it is a bit weird to me that `candidate_devices` actually is a list of groups of candidate devices. I would try to do it more explicit to avoid misleading.

```suggestion
      groups = group_candidate_devices
      target_sizes.product(groups).each do |target, devices|
```

And then,

```
      def group_candidate_devices
        candidate_devices.zip.append(candidate_devices).uniq
      end
```

So, `candidate_devices` does not need to group nothing by itself, it is only the plain list of candidate devices.

src/lib/y2storage/guided_proposal.rb
```diff
@@ -120,12 +120,12 @@ def calculate_proposal
...
         # reset root_device, else #candidate_roots will just use it
         populated_settings.root_device = saved_root_device
+        populated_settings.candidate_devices = devices
 
-        candidate_roots.each do |disk_name|
+        candidate_roots_for(devices).each do |disk_name|
```
joseivanlopez: See comment above about this method.

test/y2storage/guided_proposal_no_candidate_disks_test.rb
```diff
@@ -0,0 +1,303 @@
...
+        "fs_type"      => "xfs",
+        "mount_point"  => "/home",
+        "desired_size" => "20GiB",
+        "min_size"     => "10GiB",
+        "max_size"     => "40GiB"
+      }
+    ]
+  end
+
+  context "#propose with no candidate devices" do
```
joseivanlopez: Umm, I prefer to split this as:

```
describe "#propose" do
  context "with not candidate devices" do
    ...
  end
end

File: test/y2storage/guided_proposal_no_candidate_disks_test.rb [hunk 10]
```diff
@@ -0,0 +1,303 @@
...
+
+    context "in a system with a single disk" do
+      before do
+        create_empty_disk("/dev/sdc", disk_size)
+      end
+
+      context "having enough space for desired sizes" do
+        let(:disk_size) { needed_space_for_desired_sizes }
+
+        it "makes the proposal using those sizes" do
```
joseivanlopez: NP: s/those/desired/

test/y2storage/guided_proposal_no_candidate_disks_test.rb
```diff
@@ -0,0 +1,303 @@
...
+          expect(select_partition("swap", disk.partitions).size).to eq(1.GiB)
+          expect(select_partition("/", disk.partitions).size).to eq(40.GiB)
+          expect(select_partition("/home", disk.partitions).size).to eq(20.GiB)
+        end
+      end
+
+      context "having space only for minimum sizes" do
+        let(:disk_size) { needed_space_for_min_sizes }
+
+        it "makes the proposal using those sizes" do
```
joseivanlopez: NP: s/those/minimum/

test/y2storage/guided_proposal_no_candidate_disks_test.rb
```diff
@@ -0,0 +1,303 @@
...
+            expect(first_disk.partitions).to be_empty
+            expect(third_disk.partitions).to be_empty
+
+            expect(select_partition("swap", partitions).size).to eq(1.GiB)
+            expect(select_partition("/", partitions).size).to eq(40.GiB)
+            expect(select_partition("/home", partitions).size).to eq(20.GiB)
+          end
+        end
+
+        context "only if all devices are used" do
```
joseivanlopez: NP: "in none of them individually but all together"

test/y2storage/guided_proposal_no_candidate_disks_test.rb
```diff
@@ -0,0 +1,303 @@
...
+
+            expect(home_partition).to_not be_nil
+            expect(home_partition.size).to eq(10.GiB)
+
+            expect(swap_partition).to_not be_nil
+            expect(swap_partition.size).to eq(512.MiB)
+          end
+        end
+
+        context "using all of them" do
```
joseivanlopez: NP: "in none of them individually but all together"

src/lib/y2storage/guided_proposal.rb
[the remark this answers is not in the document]
dgdavid: > Do we only want to use individual disks when candidate devices are not given?

Exactly right!

Regarding to the code, I am realizing now that I get myself into a mess and I ended adding a weird `candidate_devices` method containing leftover. Anyway, I like more your `nil` approach :)

src/lib/y2storage/guided_proposal.rb
dgdavid: No, you are not missing anything :) Once again, you are right and that approach probably comes from my initial approach.

Thank you so much.

src/lib/y2storage/guided_proposal.rb
```diff
@@ -205,17 +205,58 @@ def candidate_devices_with_empty_partition_table(devicegraph)
...
+    # List of candidate devices
+    #
+    # If they are not given in the settings, disks found by the analyzer will be used.
+    #
+    # @see #candidate_devices_from_settings
+    # @see #candidate_devices_from_analyzer
+    #
+    # @return [<Array<Array<String>>]
+    def candidate_devices
+      return [settings.candidate_devices] if settings.candidate_devices
```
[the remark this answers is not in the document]
dgdavid: Yep! You are right!. Nice catch, thank you.

src/lib/y2storage/guided_proposal.rb
dgdavid: Because two main reasons:

* I did not think in this exact version :P
* I had a similar line of code but I though that could be a bit more complicate to understand.

However, I'll apply your suggestion.

Thanks!

src/lib/y2storage/guided_proposal.rb
dgdavid: This looks to me a good improvement. Believe or not, I could not find this more accurate name.

src/lib/y2storage/guided_proposal.rb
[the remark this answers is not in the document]
dgdavid: > Is this the use case?

I think so :)

src/lib/y2storage/guided_proposal.rb
```diff
@@ -205,17 +206,32 @@ def candidate_devices_with_empty_partition_table(devicegraph)
...
+      @populated_settings ||= settings.dup
+    end
 
-      populated = settings.dup
-      populated.candidate_devices ||= disk_analyzer.candidate_disks.map(&:name)
+    # List of candidate devices
+    #
+    # If they are not given in the settings, disks found by the analyzer will be used
+    #
+    # TODO: improve this documentation
```
dgdavid: Ey @joseivanlopez 

Could you please help me improving the documentation of this method?

Thanks in advance :)

src/lib/y2storage/guided_proposal.rb
```diff
@@ -212,8 +213,7 @@ def target_sizes
       [:desired, :min]
     end
 
-    # Copy of the original settings including some calculated and necessary
-    # values (mainly candidate_devices), in case they were not present
+    # Settings used by each attempt of proposal
```
joseivanlopez: I would extend the documentation here by saying that this "populated settings" are intended to be populated during the process of making a proposal, for example, by setting the candidate devices or the root device when it is not given in the original settings.

src/lib/y2storage/guided_proposal.rb
joseivanlopez: Sure, although I am not specially good in that area :)

```
# Candidate devices grouped for different proposal attempts
#
# When some candidate devices are indicated in the settings, the proposal is tried with all that candidate devices.
# But when no candidate devices are given, different attempts are done by using different sets of candidate devices. First, each available device is used lonely to make the proposal, and if no proposal was possible with any individual disk, a last attempt is done by using all available devices as candidate disks.
#
# @example
#
#   settings.candidate_devices #=> ["/dev/sda", "/dev/sdb"]
#   settings.group_candidate_devices #=> [["/dev/sda", "/dev/sdb"]]
#
#   settings.candidate_devices #=> nil
#   settings.group_candidate_devices #=> [["/dev/sda"], ["/dev/sdb"], ["/dev/sda", "/dev/sdb"]]

File: src/lib/y2storage/guided_proposal.rb [hunk 17]
```diff
@@ -212,51 +213,39 @@ def target_sizes
...
+    #
+    # A copy of original settings, which is intended to be populated during the process of make a
```
joseivanlopez: Typo: the process **of making** a

src/lib/y2storage/guided_proposal.rb
```diff
@@ -212,51 +213,39 @@ def target_sizes
...
+    # proposal. E.g, setting a value tha is was not given, such as the candidate devices or root
```
joseivanlopez: Fix text: a value that was not given,

PR #784, opened 2018-10-26
lower minimal mbr gap size to allow creating unaligned partitions

wfeldt:
For ms-dos partition tables the default minimal_mbr_gap setting of 1 MiB
prevents creating partitions in the expert partitioner starting at a lower
value (via the region dialog).

Note this is not an issue with gpt, there you can already do this.

### DIFF
--- a/src/lib/y2partitioner/actions/create_partition_table.rb
+++ b/src/lib/y2partitioner/actions/create_partition_table.rb
@@ -77,7 +77,10 @@ module Y2Partitioner
       # @param type [Y2Storage::PartitionTables::Type]
       def create_partition_table(type)
         disk.remove_descendants
-        disk.create_partition_table(type)
+        ptable = disk.create_partition_table(type)
+        if ptable.type.is?(:msdos)
+          ptable.minimal_mbr_gap = Y2Storage::DiskSize.B(512)
+        end
         Y2Storage::Filesystems::Btrfs.refresh_subvolumes_shadowing(device_graph)
       end
 
--- a/src/lib/y2partitioner/actions/create_partition_table.rb
+++ b/src/lib/y2partitioner/actions/create_partition_table.rb
@@ -79,6 +79,11 @@ module Y2Partitioner
         disk.remove_descendants
         ptable = disk.create_partition_table(type)
         if ptable.type.is?(:msdos)
+          # The default value is 1 MiB, preventing the creation of a
+          # partition starting at a lower value.
+          # 0.5 KiB is the space used by the partition table itself, you
+          # cannot go below that.
+          # Note that this does not affect partition alignment.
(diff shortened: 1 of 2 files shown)


ancorgs:
Well, if allowing to create misaligned partitions is important for the Partitioner. This looks like a safe place to put that code, although I miss a explanatory comment.

ancorgs:
And I guess we are sure than introducing this change only opens the door to specify misaligned partitions using the "custom region" option. "Maximum size" and "custom size" should keep producing aligned partitions, no matter what.

src/lib/y2partitioner/actions/create_partition_table.rb
```diff
@@ -77,7 +77,10 @@ def perform_action
       # @param type [Y2Storage::PartitionTables::Type]
       def create_partition_table(type)
         disk.remove_descendants
-        disk.create_partition_table(type)
+        ptable = disk.create_partition_table(type)
+        if ptable.type.is?(:msdos)
+          ptable.minimal_mbr_gap = Y2Storage::DiskSize.B(512)
```
joseivanlopez: After talking with Ancor now I understood why we want to create a non optimal aligned mbr gap, but, why 512 B?

joseivanlopez: When we are creating a partition without custom region option, does the resulting partition continue being optimal aligned at the beginning?

wfeldt: 0.5 KiB is the space used by the partition table itself. You cannot go below that.

As already said, this only brings us on par with gpt where this is already possible. This does not affect partition alignment in any way.

joseivanlopez: I would appreciate if you append this explanation to the comment in the code.

ancorgs: Our documentation of `PartitionTables::Msdos#mbr_gap` is not clear/precise enough, which makes those 512 B confusing. See https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/partition_tables/msdos.rb#L44

That states that mbr_gap is "_the space between the **end** of the MBR and the beginning of the first partition._". That is not true. Is the space between the START of the MBR (ie. the start of the disk) and the beginning of the first partition. Isn't it?

ancorgs: Note that documentation is only at `Y2Storage` level. There is no such method in libstorage-ng. In the library there is only `minimal_mbr_gap` (which is not documented).

wfeldt: > Is the space between the START of the MBR (ie. the start of the disk) and the beginning of the first partition. Isn't it?

Yes.

PR #788, opened 2018-10-30
fix boot requirements: legacy

wfeldt:
WIP
- some test cases are not yet adapted
- the 'whole disk' case gives the wrong warning (due to default = gpt): https://github.com/yast/yast-storage-ng/pull/788/commits/3637f8c8c8a9c21176f5ac4c216bfe9cfe435c79

The logic in the boot requirements check for legacy is a bit broken. So I start by putting things straight.

In the legacy (non-efi) case, using ms-dos partition table, in
https://github.com/yast/yast-storage-ng/blob/master/doc/boot-requirements.md
the entire 'with too small MBR gap' - section can go. It's basically the same as the 'mbr gap too small' part of the lvm/encryption case. I don't see why proposing /boot to embed grub is sometimes good and sometimes bad.

The code now consistently proposes a /boot partition if the mbr gap is too small. No error is raised. This should have near to no practical consequences but we have a number of test cases expecting this.

There's also no difference between 'mbr gap' and 'mbr gap + grub env block' anymore. This caused much of the test case adjustments as 256 KiB is now always ok.

Also, I'm pretty sure this

https://github.com/yast/yast-storage-ng/pull/788/commits/3566cdd1ffa4a6147ecf55718250fa6aed26ba04#diff-85b6d1f00cb4bd77ea48096f7b819477R29

is the wrong way. How to do it better?

### DIFF
--- a/doc/the-boot-partition-and-me.md
+++ b/doc/the-boot-partition-and-me.md
@@ -1,0 +1,83 @@
+## What is the boot partition used for and why would I need one?
+
+
+> #### Notes
+> - In this document only file systems supported by `storage-ng` are considered: btrfs, extX (ext2, ext3, ext4), xfs.
+> - With 'boot partition' a partition mounted at `/boot` (or containing the `/boot` directory directly) is meant.
+>   It is not to be confused with the BIOS GRUB partition, EFI System Partition, or PReP partition.
+> - While this document is basically valid for all architectures (points 1 and 3 in the below list) the focus here is on clarifying the situation
+>   on x86 with legacy BIOS and msdos partition table.
+
+A boot partition is used for three different purposes
+
+1. to hold kernel and initrd using a file system that can be read by grub
+2. to install grub into
+3. to hold the grub environment block (accessible as `/boot/grub2/grubenv` in the file system)
+
+There are different file system requirements for each use case
+
+1. extX, xfs, btrfs
+2. extX, btrfs
+3. readable in grub: extX, xfs, btrfs; writable in grub: extX, xfs
+
+So if you ever need a boot partition, extX would be the file system of choice.
+
+Let's get into detail.
+
+### 1. To hold kernel and initrd
+
+This would be needed in a complex setup where grub otherwise would not be
+able to read the kernel. As grub supports basically everything the only
+reason left would be to avoid entering the decryption key in grub for an
+encrypted boot partition.
+
+### 2. To install grub into
+
+On x86 with a legacy BIOS setup and using a msdos partition table grub
+would normally be installed into free space before the first partition ('mbr gap').
+
+But if the mbr gap is too small grub can alternatively be installed into the
+boot partition and a generic boot loader be put into the mbr to chainload
+grub from the boot partition.
+
+This is a setup discouraged by grub as this embedding of grub into a file
+system via block maps has the drawback that if the grub image files are
+inadvertently moved this would invalidate the block map and break
+grub until `grub2-install` is re-run.
+
+The mbr gap is considered large enough by `storage-ng` if it's at least
+256 KiB. This limit cannot be given exactly as it depends on the number of modules
+grub needs to boot (lvm, raid, encryption, file systems, ...).
+
+The chosen value is comfortably larger than the current real grub
+requirements (around 100 KiB) and much smaller than typical real mbr gap
+values of 1 MiB. Only older disks might have been partitioned with the obsolete
+chs (cylinder) layout in mind and only reserved one track (31.5 KiB) - which is too small for
+our purposes and is caught by our limit.
+
+### 3. To hold the grub environment block
+
+The [grub environment block](https://www.gnu.org/software/grub/manual/grub/grub.html#Environment-block)
+is a 1 KiB block reserved to store environment
+variables to be read and written by grub during the boot process. The block
+is visible as `/boot/grub2/grubenv` in the file system but should be
+accessed only via the `grub2-editenv` command.
+
+Entries can be read and written with `load_env` and `save_env` from grub. The catch
+here is that it can basically always be read but writing is limited to a few file systems
+as grub just does not want to get into trouble accidentally destroying data. Also, writing is
+not supported via lvm, raid, and encryption modules.
+
+The main usage of this block is `grub2-once` that sets a `next_entry`
+variable (boot entry to be used for the next boot). This variable is unset
+during the next boot by systemd in `grub2-once.service`.
+
+`save_env` is **not** used for this and we can live fine with a setup that
+does not allow writing the environment block from grub.
+
+### Conclusion
+
+A boot partition is not needed unless the mbr gap is too small. Because
+
+- we consciously live with the fact that you have to enter the decryption key in grub
+- write access in grub to the environment block is not needed in a typical setup
--- a/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
+++ b/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
@@ -146,6 +146,26 @@ module Y2Storage
         type ? type.is?(:btrfs) : false
       end
 
+      # Whether grub can be embedded into the boot (/boot) filesystem
+      #
+      # @return [Boolean] true if grub can be embedded into the boot filesystem.
+      #   False if the boot filesystem is unknown (not in the planned devices
+      #   or in the devicegraph) or can not embed grub.
+      def boot_fs_can_embed_grub?
+        type = boot_filesystem_type
+        type ? type.grub_ok? : false
+      end
+
+      # Whether grub can be embedded into the root (/) filesystem
+      #
+      # @return [Boolean] true if grub can be embedded into the root filesystem.
+      #   False if the root filesystem is unknown (not in the planned devices
+      #   or in the devicegraph) or can not embed grub.
+      def root_fs_can_embed_grub?
+        type = filesystem_type(device_for_root)
+        type ? type.grub_ok? : false
+      end
+
       # Type of the filesystem (planned or from the devicegraph) containing /boot
       #
       # @return [Filesystems::Type, nil] nil if there is no place for /boot either
@@ -157,14 +177,14 @@ module Y2Storage
       # Whether the partition table of the disk used for booting matches the
       # given type.
       #
-      # If the disk does not have a partition table, a GPT one will be assumed
-      # since it is the default type used in the proposal.
+      # It is possible to check for 'no partition table' by passing type nil.
       #
       # @return [Boolean] true if the partition table matches.
       #
       # @see #boot_disk
       def boot_ptable_type?(type)
-        return false if boot_ptable_type.nil?
+        return type.nil? if boot_ptable_type.nil?
+        return false if type.nil?
         boot_ptable_type.is?(type)
       end
 
@@ -264,12 +284,12 @@ module Y2Storage
         boot_planned_dev || boot_filesystem || root_planned_dev || root_filesystem || nil
       end
 
+      # Partition table type of boot disk
+      #
+      # @return [PartitionTables::Type, nil] partition table type of boot disk or nil
+      #   if it doesn't have a partition table
       def boot_ptable_type
-        return nil unless boot_disk
-        return boot_disk.partition_table.type unless boot_disk.partition_table.nil?
-
-        # If the disk end up being used, there will be a partition table on it
-        boot_disk.preferred_ptable_type
+        boot_disk.partition_table.type if boot_disk && !boot_disk.partition_table.nil?
       end
 
       # TODO: handle planned LV (not needed so far)
--- a/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
+++ b/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
@@ -65,6 +65,19 @@ module Y2Storage
       # If "/" is still not present in the devicegraph or the list of planned
       # devices, the first disk of the system is used as fallback.
       #
+      # FIXME: For RAID and LVM setups a list of disks would strictly be
+      #   correct (the disks that constitute the /boot file system).
+      #   The current approach is not that bad, though, as we don't have to
+      #   actually install the bootloader but just check if it will work.
+      #   Also, the extra partitions proposed are the *minimum* required to boot.
+      #
+      #   But particularly in asymmetric cases (like part of LVM on a GPT
+      #   disk, part on a MS-DOS disk) we have a problem: the requirements
+      #   will differ depending on which boot disk is picked (basically
+      #   random). For this to work properly we'd have to switch to tracking
+      #   all boot disks but this will also mean error messages like "BIOS
+      #   Boot on sda and (or?) MBR-GAP on sdb are missing".
+      #
       # @return [Disk]
       def boot_disk
         return @boot_disk if @boot_disk
@@ -76,6 +89,7 @@ module Y2Storage
         @boot_disk ||= boot_disk_from_planned_dev
         @boot_disk ||= boot_disk_from_devicegraph
         @boot_disk ||= devicegraph.disk_devices.first
+        @boot_disk = boot_disk_raid1(@boot_disk) || @boot_disk
 
         @boot_disk
       end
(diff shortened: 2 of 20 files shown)


src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -264,7 +282,18 @@ def device_for_boot
...
+      # Partition table type of boot disk
+      #
+      # @return [PartitionTables::Type, nil] partition table type of boot disk or nil
+      #   if it doesn't have a partition table
       def boot_ptable_type
+        # FIXME
+        #
+        # This needs to be sorted out to support plain disks (without partition table).
+        # The commented line below is possibly the right thing but it's implicitly assmued
+        # that an empty disk will get a gpt elsewhere.
```
ancorgs: Assuming this will get a GPT partition table was the right thing to do when `BootRequirementsChecker` was created, since it's original goal was to serve the Proposal. Reusing the checks for the Partitioner is something that came later.

In other words, in the scope of `BootRequirementsChecker#needed_partitions` (which answers the question "what additional partitions should the proposal create?") the historical behavior is right. If the boot/root disk has no partition table, it will get a GPT during the process for sure.

In the scope of `BootRequiremetsChecker#errors` or `BootRequirementsChecker#warnings` (which answer the question "what is wrong with this layout?") this is wrong.

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -148,11 +156,10 @@ def errors_on_gpt
...
         if grub_part_needed_in_gpt? && missing_partition_for?(grub_volume)
-          message = _(
-            "There is no partition of type BIOS Boot. " \
-            "For GPT devices it is strongly advised to use a BIOS Boot partition. " \
-            "A setup without such a partition is not supported and may cause " \
-            "problems now or in the future."
+          # TRANSLATORS: %s is a partition type, e.g. "BIOS Boot"
+          message = format(
+            _("A partition of type %s is needed to install the bootloader."),
+            PartitionId::BIOS_BOOT.to_human_string
```
ancorgs: We intentionally introduced this specific message (mentioning problems in the future) recently to address the complaints about enforcing a BIOS Boot partition (by users who managed to make their systems boot without it). Why do we want to revert the message now?

ancorgs:
_the 'whole disk' case gives the wrong warning (due to default = gpt)_

I commented on it. Assuming default == GPT is right for `#needed_partitions` (the proposal) but not for `#warnings` (the Partitioner). So we probably need to move the assumption to the right part of the code instead of having it in `#boot_ptable_type`.

_I don't see why proposing /boot to embed grub is sometimes good and sometimes bad._

As far as I understand the current implementation, it actually NEVER proposes /boot as a mechanism to embed Grub2. It proposes /boot for other purposes .

What follows is my understanding of the current implementation, basically based on the output of the tests. I'm not saying any of that is correct, just interpreting what I see in the tests.

Based on that, I would say the old code assumed that:

 * Grub2 itself (the core image) can only be installed in the MBR gap or in a Btrfs file-system over a plain partition[1].
 * When using any disk abstraction (LVM, RAID, encryption...) we have two options[2]
   * allocate Grub2 itself and, in addition, allocate the grubenv alongside Grub2.
   * a separate /boot
 * Of course, grubenv can be installed alongside Grub2 itself only if there is enough space for both.

[1] This is a huge difference with your PR, in which is clearly stated that ext2/3/4 also supports embedding Grub2. TBH, I am myself still confusing about whether ext2/3/4 supports embedding Grub2 or not and whether that is safe against future file-system operations. How is ext2 equal to (or different from) Btrfs in terms of embedding Grub2?

[2] Based on the information at [grub2-and-disk-abstractions](https://github.com/yast/yast-storage-ng/blob/master/doc/boot-partition.md#grub2-and-disk-abstractions).

Taking all that into account (that, of course, may be wrong) the current tests make some kind of sense. According to them:

- If the MBR gap is big enough for both Grub2 and its grubenv
  - Everything can be booted and we never need a separate /boot
- If the MBR gap is big enough for Grub2 itself but not for Grub2+grubenv
  - Using plain partitions we can still boot just fine
  - For LVM/RAID/encrypt we need the separate /boot
- If the MBR gap is too small to allocate Grub2
  - Plain btrfs works because is the only fs-type that supports embedding Grub2
  - Any other setup would fail because we cannot allocate Grub2 (although, in fact we could have proposed a separate Btrfs /boot instead of failing)

_The code now consistently proposes a /boot partition if the mbr gap is too small. No error is raised. [...] There's also no difference between 'mbr gap' and 'mbr gap + grub env block' anymore. This caused much of the test case adjustments as 256 KiB is now always ok._

I'm fine with that. I like the idea of breaking the separation between Grub2 core and grubenv. That was trying to fine-tune too much with no clear gain.

_How to do it better?_

I can take care of adapting the test-cases. It would be a good opportunity for me to ensure I understood everything (I think I did, but...).

doc/the-boot-partition-and-me.md
```diff
@@ -0,0 +1,82 @@
...
+would normally be installed into free space before the first partition ('mbr gap').
+
+But if the mbr gap is too small grub can alternatively be installed into the
+boot partition and a generic boot loader be put into the mbr to chainload
+grub from the boot partition.
+
+This is a setup discouraged by grub as this embedding of grub into a file
+system via block maps has the drawback that if the grub image files are
+inadvertently moved this would invalidate the block map and break
+grub until `grub2-install` is re-run.
```
ancorgs: Btrfs is safe in this regard. Isn't it?

At least, in the current code Btrfs is considered as the only safe file-system to embed Grub2, so I assume Btrfs includes mechanisms to never move the grub image file (or even a dedicated region for it).

doc/the-boot-partition-and-me.md
```diff
@@ -0,0 +1,82 @@
+## What is the boot partition used for and why would I need one?
+
+
+> #### Notes
+> - In this document only file systems supported by `storage-ng` are considered: btrfs, extX (ext2, ext3, ext4), xfs.
+> - With 'boot partition' a partition mounted at `/boot` (or containing the `/boot` directory directly) is meant.
+>   It is not to be confused with the BIOS GRUB partition, EFI System Partition, or PReP partition.
+
```
ancorgs: I would also add a note specifying this document describes only the situation with BIOS (aka legacy) booting and MS-Dos partition tables. Just to avoid confusion.

doc/the-boot-partition-and-me.md
wfeldt: I'm actually unsure here. The 64k gap in btrfs is not big enough for the whole grub image.

doc/the-boot-partition-and-me.md
```diff
@@ -0,0 +1,82 @@
...
+The [grub environment block](https://www.gnu.org/software/grub/manual/grub/grub.html#Environment-block)
+is a 1 KiB block reserved to store environment
+variables to be read and written by grub during the boot process. The block
+is visible as `/boot/grub2/grubenv` in the file system but should be
+accessed only via the `grub2-editenv` command.
+
+Entries can be read and written with `load_env` and `save_env` from grub. The catch
+here is that it can basically always be read but writing is limited to a few file systems
+as grub just does not want to get into trouble accidentally destroying data. Also, writing is
+not supported via lvm, raid, and encryption modules.
```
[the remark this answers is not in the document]
wfeldt: And that is not true (I checked with sle15-sp1 yesterday) - the grub env block is always the file.

doc/the-boot-partition-and-me.md
wfeldt: Anyway, despite this we propose (when we do) an **ext4** boot partition. It would look quite weird when we started to use btrfs for a separate boot partition. Even if the above statement were true and btrfs would be the only safe file system to be used for grub embedding. It just has an eerie feeling.

doc/the-boot-partition-and-me.md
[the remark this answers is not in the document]
ancorgs: Ok. We should ping Michael Chang about that. Just to clarify whether that's intentional. The preliminary document says "It will be for sure ready in time for SLE12-SP3". :wink:

doc/the-boot-partition-and-me.md
ancorgs: I agree we should keep proposing ext4 when we go for a separate `/boot` partition. I was just curious, since the information about block maps and its compatibility/fragility with different file-system types looks sometimes confusing to me.

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -26,48 +26,52 @@ module Y2Storage
...
+      # This function is intended to be used in #needed_partitions while
+      # creating a partition proposal. The catch is that when there is no
+      # partition table (yet) it is implicitly assumed that there will be a
+      # gpt created finally - so a grub partition is needed also in this
+      # case.
+      #
+      # @return [Boolean]
       def grub_partition_needed?
-        boot_ptable_type?(:gpt) && grub_part_needed_in_gpt?
+        (boot_ptable_type?(:gpt) || boot_ptable_type?(nil)) && grub_part_needed_in_gpt?
```
ancorgs: This should not directly assume the default type will be GPT. This should rely on `#preferred_ptable_type`. For example, it's quite likely that for Raspberry Pi  we will have to modify `#preferred_ptable_type` to return msdos (looks like Raspi doesn't support GPT in the booting device). Or any other circumstances, that's what we have `#preferred_ptable_type` for.

That is, something like this (just from the top of my head):
```ruby
gpt_used =
  if boot_ptable_type?(nil)
    boot_disk.preferred_ptable_type.is?(:gpt)
  else
    boot_ptable_type?(:gpt)
  end
gpt_used && grub_part_needed_in_gpt?
```

Although is probably nicer to simply have a flag in `#boot_ptable_type?` to tell it to keep the old behavior. So we could have something like

```ruby
boot_ptable_type?(:gpt, fallback_to_preferred_type: true) && grub_part_needed_in_gpt?
```

(ok, the argument could have a shorter name).

wfeldt: `boot_ptable_type` has cost me a lot of new gray hair already. And that even though it's only used withing the boot requirements code. I will try the first version if only to avoid the double semantics in `boot_ptable_type`.

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -26,48 +26,58 @@ module Y2Storage
...
+      # FIXME
+      #   It seems that a setup with xen virtual partitions (that are in
+      #   fact disks) also ends up here. In that case no grub partition will
+      #   be needed (below this case is indicated by boot_disk = nil).
+      #   This looks weird.
+      #
+      # @return [Boolean]
       def grub_partition_needed?
-        boot_ptable_type?(:gpt) && grub_part_needed_in_gpt?
+        (boot_ptable_type?(:gpt) || (boot_disk && boot_ptable_type?(nil))) && grub_part_needed_in_gpt?
```
ancorgs: This code directly assumes that GPT will be used as default type without asking to `#preferred_ptable_type`. I believe that's wrong. For example, according to fate#323484 is quite likely that we will need to tweak `#preferred_ptable_type` to not always blindly return GPT if running in a Rasberry Pi.

So I would add a flag for `#boot_ptable_type?` to tell it to keep using the old logic (fallback to the preferred type of there is none). Something like:

```ruby
boot_ptable_type?(:gpt, fallback_to_preferred: true) && grub_part_needed_in_gpt?
```

Which, roughly, would be kind of equivalent to:

```ruby
gpt_used =
  if boot_disk && boot_ptable_type?(nil)
    boot_disk.preferred_ptable_type.is?(:gpt)
  else
    boot_ptable_type?(:gpt)
  end
gpt_used && grub_part_needed_in_gpt?
```

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -166,22 +178,36 @@ def errors_on_gpt
...
         end
 
         errors
       end
 
-      # Specific error when the boot disk has not partition table
+      # Boot errors when there's no partition table
+      #
+      # @return [Array<SetupError>]
+      def errors_on_plain_disk
```
ancorgs: NitPicking: why not simply this?

```ruby
def errors_on_plain_disk
  [no_boot_partition_table_error]
end
```

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -166,22 +178,36 @@ def errors_on_gpt
...
-        elsif boot_partition_needed? && missing_partition_for?(boot_volume)
-          errors << SetupError.new(missing_volume: boot_volume)
+        if !mbr_gap_for_grub?
+          errors <<= mbr_gap_error
+          errors <<= if boot_can_embed_grub?
+            bad_config_warning
+          else
+            bad_config_error
+          end
         end
```
ancorgs: How are these errors presented, as different pop-ups or in the same dialog one below the other?

wfeldt: Presented as a bullet-list, looks ok.

src/lib/y2storage/boot_requirements_strategies/legacy.rb
wfeldt: Ok ok, I agree. I had it changed for this xen virtual partition setup thing anyway. See the reworked patch.

src/lib/y2storage/boot_requirements_strategies/legacy.rb
```diff
@@ -191,10 +233,49 @@ def unknown_boot_partition_table_error
...
       def mbr_gap_error
-        # TRANSLATORS: error message
-        error_message = _("MBR gap size is not enough to correctly install bootloader")
+        # TRANSLATORS: error message; %s is a human readable disk size like 256 KiB
+        error_message = format(
+          _(
+            "Not enough space before the first partition to install the bootloader. " \
+            "Leave at least %s."
+          ),
+          PartitionTables::Msdos::MBR_GAP_GRUB_LIMIT
```
ancorgs: This should be `PartitionTables::Msdos::MBR_GAP_GRUB_LIMIT.to_human_string` for consistency and proper i18n.

src/lib/y2storage/boot_requirements_strategies/legacy.rb
ancorgs: :+1:

wfeldt:
All tests pass, and the proposal is still wrong! :-((

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -403,8 +469,34 @@ def in_software_raid?(device)
...
+            dev.is?(:software_raid) && dev != boot_disk
+          end
+        end
+      end
+
+      # Check if device is a direct member of a RAID1 (RAID over entire disks).
+      #
+      # FIXME: The check is possibly overly strict: currently it enforces
+      #   that the disk is a member of a single RAID.
+      #   With BIOS RAIDs that might not be necessary. Also the limitation
```
ancorgs: I don't get why BIOS RAIDs are mentioned here. We just treat them always as regular disks (they return true for `is?(:disk_device)`) and I don't see anything about BIOS RAIDs in this method, just software RAIDs are considered. Isn't it?

wfeldt: I was thinking of:

https://github.com/openSUSE/libstorage-ng/blob/master/testsuite/probe/md-imsm1-devicegraph.xml

ancorgs: That's my point. In that case, `/dev/md/a` and `/dev/md/b` are both BIOS RAIDs. So they are not affected by this method. Basically in all regards, both the proposal and the partitioner consider them as disks (because they return true for `is?(:disk_device)`). And they are skipped in the first line of this function loop (because they return false for `is?(:software_raid)`).

Unless I'm overlooking something.

ancorgs: I find it less confusing now, without the reference to BIOS RAIDs.

PR #790, opened 2018-10-31
Find crypttab names

joseivanlopez:
PBI: https://trello.com/c/bY9jBdVR/424-find-devices-by-uuid

Requires libstorage-ng 4.1.54, see https://github.com/openSUSE/libstorage-ng/pull/589


src/lib/y2storage/devicegraph.rb
```diff
@@ -559,5 +557,78 @@ def udev_lookup_possible?
...
+    #
+    # @return [Device, nil] the found device, nil if no device matches the name
+    def deep_find(name)
+      log.info "Trying to find a device by name: #{name}"
+
+      # First check using the device name
+      device = find_by_name(name)
+      # If not found, check udev names directly handled by libstorage-ng
+      device ||= blk_devices.find { |dev| dev.udev_full_all.include?(name) }
+      log.info "Device #{device.inspect} found by its libstorage-ng name #{name}"
```
dgdavid: Looking the below lines of code, I think that the `blk_devices.find` call in the previous line could return `nil`. If so, then that info should be logged only if a devices was found, isn't it?

src/lib/y2storage/encryption.rb
```diff
@@ -102,6 +102,56 @@ def auto_dm_name=(value)
...
+    # /dev/disk/by-id/scsi-0ATA_Micron_1100_SATA_1652155452D8-part2
+    # /dev/disk/by-uuid/7a0c6309-7063-472b-8301-f52b0a92d8e9
+    # /dev/disk/by-path/pci-0000:00:17.0-ata-3-part2
+    #
+    # But it can also contain a value like UUID=111-222-333. In that case, the
+    # UUID value refers to the LUKS device instead of the underlying device.
+    # See {Luks#match_crypttab_spec?}.
+    #
+    # This method checks whether the underlying device of the encryption is the
+    # device indicated in the second column of an crypttab entry.
```
dgdavid: s/an/a

src/lib/y2storage/encryption.rb
```diff
@@ -102,6 +102,56 @@ def auto_dm_name=(value)
...
+    #
+    # Take into account that libstorage-ng discards during probing all the
+    # udev names not considered reliable or stable enough. This method only
+    # checks by the udev names recognized by libstorage-ng (not discarded).
```
dgdavid: NP question: is it "by" necessary?

src/lib/y2storage/encryption.rb
```diff
@@ -102,6 +102,56 @@ def auto_dm_name=(value)
...
+    # @return [Boolean]
+    def crypttab_name?
+      !crypttab_name.nil?
+    end
+
+    # Name specified in the crypttab file for this encryption device
+    #
+    # @note This relies on the userdata mechanism, see {#userdata_value}.
+    #
+    # @return [String, nil] nil if rypttab name is not known
```
dgdavid: s/rypttab/crypttab

src/lib/y2storage/devicegraph.rb
joseivanlopez: Good catch!

src/lib/y2storage/encryption.rb
joseivanlopez: I would say so.

doc/upgrade.md
```diff
@@ -0,0 +1,136 @@
+# Upgrade process
+
+This document summarizes some potential problems with storage-ng when trying to upgrade a system.
+
+During the product upgrade, that is, when selecting "Upgrade" option in the Linuxrc menu, all filesystems are mounted to discover the current installed Operating Systems. Then, the user selects what OS should be upgraded from the list of available Operating Systems.
+
+Once a OS is selected, the filesystem is mounted again and its /etc/fstab is parsed. Each entry in the fstab is also mounted, less the entry associated to the root mount point because root is already mounted.
```
ancorgs: Once **an** OS. Isn't it?

ancorgs: Less -> except

doc/upgrade.md
```diff
@@ -0,0 +1,136 @@
...
+```
+cr_home     UUID=000-111-222                            # default mount_by = uuid
+cr_home     /dev/sda2                                   # default mount_by = device name
+cr_home     /dev/sda2                                   # default mount_by = volume label
+cr_home     /dev/disk/by-id/scsi-11212-part2            # default mount_by = device id
+cr_home     /dev/disk/by-path/pci-0002-part2            # default mount_by = device path
+cr_home     /dev/disk/by-uuid/000-111-222
+```
+
+Several things to take into account here. When the UUID= format is used, that UUID refers to the LUKS device instead of the the underlying device. When the default mount_by is "volume label", the kernel name of the underlying device is used. When using "device id" or "device path" as default mount_by, the proper udev name of the underlying device is used. And finally, the user could indicate a device by using uuid udev name.
```
ancorgs: the the

doc/upgrade.md
```diff
@@ -0,0 +1,136 @@
...
+/dev/mapper/cr_home     /home   ext4    defaults 0 0
+```
+
+and a crypttab entry like:
+
+```
+cr_home     /dev/sda2
+```
+
+the device "/dev/sda2" is searched in the devicegraph and the crypttab name "cr_home" is stored in its associated encryption device. So, in case that there is an encryption device over "/dev/sda2", for example "/dev/mapper/cr-auto-1", that encryption device will save "cr_auto" as alternative name.
```
ancorgs: will save "cr_home" as alternative name

doc/upgrade.md
```diff
@@ -0,0 +1,136 @@
...
+
+Once this information is saved in the encryption devices, when an encryption name is searched in the devicegraph, we can check if the searched named is the alternative name of any encryption device. So, for this example, when we search in the devicegraph the device by "/dev/mapper/cr_home", we can return "/dev/mapper/cr-auto-1".
+
+This problem was fixed for SLE-15, see [related PBI](https://trello.com/c/OGssTRzO/393-osdistribution-p2p1-1094963-opensuse-150-installation-or-upgrade-fails-if-disk-contains-encrypted-partition-luks).
```
ancorgs: Not sure if I would link to Trello here. It's just a tool for team coordination (with restricted access). Better use something more stable (instead/in addition to) Trello, like the bugzilla entry or the pull request.

doc/upgrade.md
```diff
@@ -0,0 +1,136 @@
...
+mount /dev/mapper/cr_ata-111-part2 /home  => fail device /dev/mapper/cr_ata-111-part2 does not exist
+```
+
+The solution here is to avoid to use the fstab spec when that device name is not safe enough. So, instead of using the fstab spec, the udev uuid name of the encryption device will be used due to that name exists in the inst-sys:
+
+```
+mount /dev/disk/by-uuid/111-222-333 /home
+```
+
+Note that this problem could not appear when upgrading a system generated with old storage. The old storage created the fstab entries with the option "nofail" for some mount points, for example "/" or "/home":
```
ancorgs: I would say "this problem could not appear". I would say the symptom is different. Silently failing to mount the filesystem and performing the upgrade without it anyways is also a problem... or another side of the same problem.

doc/upgrade.md
```diff
@@ -0,0 +1,136 @@
...
+
+```
+/dev/mapper/cr_ata-111-part2                /home   ext4    nofail 0 0
+```
+
+So the upgrade process did not fail, but the partition was never mounted.
+
+Those problems are now fixed for SLE-15-SP1, see [related PBI](https://trello.com/c/bY9jBdVR/424-5-fix-upgrading-scenario-was-find-devices-by-uuid).
```
ancorgs: Again, a link to Trello should not be the only link.

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -332,35 +332,34 @@ def setup_blk_filesystem(filesystem)
...
-        # Duplicates the given devicegraph and modifies it by setting the encryption names
-        # from the crypttab file
+        # Duplicates the given devicegraph and saves the encryption names from the crypttab file
         #
         # @see #crypttab
         #
         # @param devicegraph [Y2Storage::Devicegraph]
         # @return [Y2Storage::Devicegraph]
-        def devicegraph_with_fixed_encryptions(devicegraph)
+        def devicegraph_with_crypttab_names(devicegraph)
```
ancorgs: Just thinking aloud: maybe now we don't need to duplicate the devicegraph now, since `save_crypttab_names` is less aggressive than `use_crypttab_names`.

src/lib/y2storage/devicegraph.rb
```diff
@@ -343,36 +350,27 @@ def find_by_name(name)
     # * It avoids system lookup in potentially risky scenarios (like an outdated
     #   {StorageManager#probed}).
     #
-    # @param name [String] can be a kernel name like "/dev/sda1" or any symbolic
+    # In case of LUKSes, the device might be found by using an alternative name,
+    # see {#alternative_names}.
+    #
+    # @see deep_search
```
ancorgs: deep_search? Isn't it deep_find?

src/lib/y2storage/devicegraph.rb
```diff
@@ -559,5 +557,79 @@ def udev_lookup_possible?
...
+    #
+    #   cr_home   /dev/sda1
+    #
+    # a possible alternative name for /dev/mapper/cr_home would be /dev/mapper/cr-auto-1, due to the
+    # encryption device cr-auto-1 has "cr_home" as crypttab_name (after parsing the crypttab file).
+    #
+    # @param device_name [String] a kernel name or udev name
+    # @return [Array<String>]
+    def alternative_names(device_name)
+      devices = encryptions.select { |e| e.crypttab_name? && device_name.match?(e.crypttab_name) }
```
ancorgs: Why is `#match?` used here to compare the strings?

src/lib/y2partitioner/actions/controllers/fstabs.rb
joseivanlopez: Umm, maybe, let me check if we have a new devicegraph each time that a fstab is imported.

src/lib/y2storage/devicegraph.rb
joseivanlopez: Yes :)

src/lib/y2storage/devicegraph.rb
joseivanlopez: Here strings are not exactly compared. It is checked that crypttab_name is included in the device_name, for example:

```
device_name #=> "/dev/mapper/cr_home"
device.crypttab_name #=> "cr_home"

device_name.match?(device.crypttab_name) #=> true
```

BTW, maybe `String#include?` is a better method for this.

src/lib/y2storage/luks.rb
```diff
@@ -0,0 +1,69 @@
...
+    storage_forward :uuid
+
+    # Whether the LUKS encryption device matches with a given crypttab spec
+    #
+    # The second column of /etc/crypttab contains a path to the underlying device
+    # of the encryption device (e.g., /dev/sda2). But it can also contain a value
+    # like UUID=111-222-333. In that case, the UUID value refers to the LUKS device
+    # instead of the underlying device.
+    #
+    # This method checks whether the underlying device of the encryption device is
```
ancorgs: I see a lot of text duplicated in this doc and in `Encryption#match_crypttab_spec?`

src/lib/y2storage/devicegraph.rb
```diff
@@ -559,5 +557,79 @@ def udev_lookup_possible?
...
+    #
+    # This is different from {BlkDevice.find_by_any_name} in several ways:
+    #
+    # * It will find any matching device, not only block devices (e.g. LVM VGs
+    #   also have a name but are not block devices).
+    # * It can be called on any devicegraph, not only probed.
+    # * It uses a system lookup only when necessary (i.e. all the cheaper
+    #   methods for finding the device have been unsuccessful).
+    # * It avoids system lookup in potentially risky scenarios (like an outdated
+    #   {StorageManager#probed}).
```
ancorgs: A lot of documentation repeated from the public method. A simple `@see #find_by_any_name` would be enough for this internal method. :-)

src/lib/y2partitioner/actions/controllers/fstabs.rb
joseivanlopez: After checking it, I am not sure if it is a good idea to populate the current devicegraph. When importing a fstab, the crypttab names are related to the the filesystem that contains the fstab being imported. But that crypttab names are not related at all with the current devicegraph in which the fstab is going to be imported. So, having those crypttab names in the current devicegraph sounds a little bit weird to me.

src/lib/y2storage/devicegraph.rb
dgdavid: For this case I would use `String#include?`

src/lib/y2partitioner/actions/controllers/fstabs.rb
ancorgs: Good point.

src/lib/y2storage/devicegraph.rb
ancorgs: Yes, with `include?` the intention is much more clear.

doc/upgrade.md
```diff
@@ -0,0 +1,138 @@
...
+
+```
+cr_home     /dev/sda2
+```
+
+the device "/dev/sda2" is searched in the devicegraph and the crypttab name "cr_home" is stored in its associated encryption device. So, in case that there is an encryption device over "/dev/sda2", for example "/dev/mapper/cr-auto-1", that encryption device will save "cr_home" as alternative name.
+
+Once this information is saved in the encryption devices, when an encryption name is searched in the devicegraph, we can check if the searched named is the alternative name of any encryption device. So, for this example, when we search in the devicegraph the device by "/dev/mapper/cr_home", we can return "/dev/mapper/cr-auto-1".
+
+This problem was fixed for SLE-15, see Pull Requests for [libstorage-ng](https://github.com/openSUSE/libstorage-ng/pull/535), [yast2-storage-ng](https://github.com/yast/yast-storage-ng/pull/659) and [yast2-update](https://github.com/yast/yast-update/pull/101) packages. And here the realated bug [bsc#1094963](https://bugzilla.suse.com/show_bug.cgi?id=1094963).
```
ancorgs: 1) Why "Pull Requests" with upper case?.
2) Realated

doc/upgrade.md
```diff
@@ -0,0 +1,138 @@
...
+mount /dev/mapper/cr_ata-111-part2 /home  => fail device /dev/mapper/cr_ata-111-part2 does not exist
+```
+
+The solution here is to avoid to use the fstab spec when that device name is not safe enough. So, instead of using the fstab spec, the udev uuid name of the encryption device will be used due to that name exists in the inst-sys:
+
+```
+mount /dev/disk/by-uuid/111-222-333 /home
+```
+
+Note that this problem could not appear when upgrading a system generated with old storage. The old storage created the fstab entries with the option "nofail" for some mount points, for example "/" or "/home":
```
ancorgs: I believe you forgot to delete this paragraph.

doc/upgrade.md
```diff
@@ -0,0 +1,138 @@
...
+
+Note that the old storage was creating fstab entries with the option "nofail" for some mount points, for example "/" or "/home". Therefore, a system generated with old storage could silently fail when trying to mount the filesystem, and the upgrade process could finish without that volume being mounted.
+
+```
+/dev/mapper/cr_ata-111-part2                /home   ext4    nofail 0 0
+```
+
+This could make the upgrade useless when some important mount point is not properly mounted, for example, when mounting an encrypted /usr silently fails.
+
+Those problems are now fixed for SLE-15-SP1, see Pull Requests for [libstorage-ng](https://github.com/openSUSE/libstorage-ng/pull/589), [yast2-storage-ng](https://github.com/yast/yast-storage-ng/pull/790) and [yast2-update](https://github.com/yast/yast-update/pull/115).
```
ancorgs: NP (grammar Nazi): same again here. Upper casing "Pull Requests" doesn't seem to make much sense. The fact that you can use it in singular and plural proves it's not a proper name.

PR #791, opened 2018-11-02
Make the proposal work on top of existing MD RAIDs

dgdavid:
Related to fate#326573 (see https://trello.com/c/atCIY5jT for detailed info) and [bug#1092417](https://bugzilla.novell.com/show_bug.cgi?id=1092417).

---
Initial proposal using the devicegraph from bsc#1092417 logs:

Initial proposal using the devicegraph from bsc#1092417 logs

:file_cabinet:  [bug_1092417_devicegraph.xml.zip](https://github.com/yast/yast-storage-ng/files/2543109/bug_1092417_devicegraph.xml.zip)

### DIFF
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -172,7 +172,7 @@ module Y2Storage
       disks.map { |d| @disks_data[data][d.name] }.flatten.compact
     end
 
-    # Obtains a list of disk devices, software RAIDs, and/or Bcaches
+    # Obtains a list of disk devices, software RAIDs, and Bcaches
     #
     # @see #default_disks_collection for default values when disks are not given
     #
@@ -295,18 +295,18 @@ module Y2Storage
       fs.release_name
     end
 
-    # Finds devices (disk devices or software raids) that are suitable for installing Linux
+    # Finds devices (disk devices and software RAIDs) that are suitable for installing Linux
     #
-    # From fate#326573 on, software raids with partition table or without children are also
+    # From fate#326573 on, software RAIDs with partition table or without children are also
     # considered as valid candidates.
     #
-    # @return [Array<BlkDevice>] candidate devices (disk devices and/or software RAIDs matching the
+    # @return [Array<BlkDevice>] candidate devices (disk devices and software RAIDs matching the
     #   conditions explained above)
     def find_candidate_disks
       find_candidate_software_raids + find_candidate_disk_devices
     end
 
-    # Finds software raids that are considered valid candidates for a Linux installation
+    # Finds software RAIDs that are considered valid candidates for a Linux installation
     #
     # Apart from matches conditions of #candidate_disk?, a valid software RAID candidate must
     # either, have a partition table or do not have children.
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -284,35 +284,14 @@ module Y2Storage
       fs.release_name
     end
 
-    # TODO: there is no reason to initally discard usb devices. The behavior is
-    # kept to avoid big changes in release phase.
-    #
     # Finds disk devices that are suitable for installing Linux.
-    #
-    # @note Usb devices are considered as candidate disk only if there are no
-    #   other candidate disks.
     #
     # @return [Array<BlkDevice>] candidate devices
     def find_candidate_disks
-      candidates = find_non_usb_candidate_disks
-      return candidates if candidates.any?
+      devicegraph.disk_devices.select { |d| candidate_disk?(d) }
 
-      find_usb_candidate_disks
     end
 
-    # Finds non usb devices that are suitable for installing Linux
-    #
-    # @return [Array<BlkDevice>] candidate devices
-    def find_non_usb_candidate_disks
-      devicegraph.disk_devices.select { |d| !d.usb? && candidate_disk?(d) }
-    end
-
-    # Finds usb devices that are suitable for installing Linux
-    #
-    # @return [Array<BlkDevice>] candidate devices
-    def find_usb_candidate_disks
-      devicegraph.disk_devices.select { |d| d.usb? && candidate_disk?(d) }
-    end
 
     # Checks whether a device can be used as candidate disk for installation
     #
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -172,19 +172,26 @@ module Y2Storage
       disks.map { |d| @disks_data[data][d.name] }.flatten.compact
     end
 
-    # Obtains a list of disk devices.
+    # Obtains a list of disk devices, software RAIDs, and/or Bcaches
     #
-    # @param disks [Array<Dasd, Disk, String>] disk device to analyze.
-    #   All disk devices by default.
-    # @return [Array<Dasd, Disk>]
+    # @see #default_disks_collection for default values when disks are not given
+    #
+    # @param disks [Array<BlkDevice, String>] blk device to analyze.
+    # @return [Array<BlkDevice>] a list of blk devices
     def disks_collection(disks)
-      return full_devices_list if disks.empty?
+      return default_disks_collection if disks.empty?
 
-      # Using BlkDevice because it is necessary to search in both, Dasd and Disk.
-      disks.map { |d| d.is_a?(String) ? BlkDevice.find_by_name(devicegraph, d) : d }.compact
+      disks.map! { |d| d.is_a?(String) ? BlkDevice.find_by_name(devicegraph, d) : d }
+      disks.compact
     end
 
-    def full_devices_list
+    # The default disks collection to be analyzed
+    #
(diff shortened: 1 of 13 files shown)


src/lib/y2storage/disk_analyzer.rb
```diff
@@ -178,10 +178,14 @@ def data_for(*disks, data)
     #   All disk devices by default.
     # @return [Array<Dasd, Disk>]
```
joseivanlopez: The return value could be any disk device, a software raid and a bcache, so I would document return as:

```
 # @return [Array<Dasd, Disk, Multipath, DmRaid, MdMember, Md, Bcache>] a disk device, software RAID or Bcahe
```

Or in a more general way:

```
 # @return [Array<BlkDevice>] a disk device, software RAID or Bcahe
```

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -178,10 +178,14 @@ def data_for(*disks, data)
...
-      disks = devicegraph.disk_devices if disks.empty?
+      return full_devices_list if disks.empty?
+
       # Using BlkDevice because it is necessary to search in both, Dasd and Disk.
-      disks = disks.map { |d| d.is_a?(String) ? BlkDevice.find_by_name(devicegraph, d) : d }
-      disks.compact
+      disks.map { |d| d.is_a?(String) ? BlkDevice.find_by_name(devicegraph, d) : d }.compact
+    end
+
+    def full_devices_list
```
joseivanlopez: NP: the name of the method does not sound too much accurate to me, and maybe it can lead to misleading. Why to not include that line in `disk_collection` method?

```
    def disks_collection(disks)
      disks ||= devicegraph.disk_devices + devicegraph.software_raids + devicegraph.bcaches
      disks = disks.map { |d| d.is_a?(String) ? BlkDevice.find_by_name(devicegraph, d) : d }
      disks.compact
    end
```

joseivanlopez: Please, also improve `disks_collection` documentation by saying that software raids and bcaches need to be analyzed since it is possible to find a Linux system installed on them (as you stated in the commit comment).

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -284,34 +288,35 @@ def release_name(filesystem)
...
     #
-    # Finds disk devices that are suitable for installing Linux.
+    # From fate#326573 on, software raids with partition table or without children are also
+    # considered as valid candidates.
     #
-    # @note Usb devices are considered as candidate disk only if there are no
-    #   other candidate disks.
-    #
-    # @return [Array<BlkDevice>] candidate devices
+    # @return [Array<DiskDevice,Md>] candidate devices
```
joseivanlopez: `DiskDevice` is not a class name, so I would use something like:

```
 # @return [Array<BlkDevice>] candidate devices (a disk devices or software RAIDs matching conditions explained above)
```

src/lib/y2storage/disk_analyzer.rb
dgdavid: > Why to not include that line in disk_collection method?

Because it made more sense to me having the "default_disks_collection" in a separate method, returning it directly when disks were not given, avoiding the not necessary mapping.

Furthermore, your example will not work because `[]` != `nil`. So, I prefer to have them split instead add more conditional logic to `disks_collection`. What do you think? Is it `default_disks_collection` more accurate for you?

Thank you so much.

test/y2storage/disk_analyzer_test.rb
```diff
@@ -210,28 +215,6 @@ def format_device(device)
       expect(candidate_disks).to contain_exactly("/dev/sda", "/dev/sdb", "/dev/sdc")
     end
 
-    context "when there are non-usb candidate devices" do
```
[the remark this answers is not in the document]
joseivanlopez: Sorry, that USB are included I mean.

src/lib/y2storage/guided_proposal.rb
```diff
@@ -244,10 +244,23 @@ def populated_settings
...
 
-      candidates = disk_analyzer.candidate_disks.map(&:name)
+      candidates = reorder_candidate_disks.map(&:name)
       candidates.zip.append(candidates).uniq
     end
 
+    # Sorts candidate disk based on their type: not USB first
+    #
+    # @return [Array<Disk>]
+    def reorder_candidate_disks
```
dgdavid: A better name is welcome :)

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -172,18 +172,29 @@ def data_for(*disks, data)
       disks.map { |d| @disks_data[data][d.name] }.flatten.compact
     end
 
-    # Obtains a list of disk devices.
+    # Obtains a list of disk devices, software RAIDs, and/or Bcaches
```
joseivanlopez: NP: why "and/or" ?

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -284,34 +295,39 @@ def release_name(filesystem)
       fs.release_name
     end
 
-    # TODO: there is no reason to initally discard usb devices. The behavior is
-    # kept to avoid big changes in release phase.
+    # Finds devices (disk devices or software raids) that are suitable for installing Linux
```
joseivanlopez: NP: I guess it should be "and" -> (disk devices and software raids)

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -284,34 +295,39 @@ def release_name(filesystem)
...
     #
-    # Finds disk devices that are suitable for installing Linux.
+    # From fate#326573 on, software raids with partition table or without children are also
+    # considered as valid candidates.
     #
-    # @note Usb devices are considered as candidate disk only if there are no
-    #   other candidate disks.
-    #
-    # @return [Array<BlkDevice>] candidate devices
+    # @return [Array<BlkDevice>] candidate devices (disk devices and/or software RAIDs matching the
```
joseivanlopez: NP: again, I guess "/or" is leftover.

src/lib/y2storage/guided_proposal.rb
joseivanlopez: "sorted_candidate_disks" ? But I am fine with current name :)

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -284,8 +284,8 @@ def free_spaces(graph, disk = nil)
       #
       # @return [Array<Dasd, Disk>]
       def disks_for(devicegraph, device_name = nil)
-        filter = device_name || candidate_disk_names
-        devicegraph.disk_devices.select { |d| filter.include?(d.name) }
+        filter = Array(device_name || candidate_disk_names)
```
joseivanlopez: NP: only a matter of taste:
```suggestion
        filter = device_name ? [device_name] : candidate_disk_names
```

src/lib/y2storage/disk_analyzer.rb
dgdavid: Hmmm... because the list could contain all of them (and) or whichever (or). Is it wrong? Do you prefer only "and"?

src/lib/y2storage/disk_analyzer.rb
dgdavid: Yep, you are right.

src/lib/y2storage/disk_analyzer.rb
dgdavid: Actually it is there intentionally, since it tried to say: "The returned candidate devices could be disk devices _and_ software RAIDs, _or_ only disk_devices _or_ only software RAIDs" .

However, if it is better expressed using only "and", I should change it :)

src/lib/y2storage/guided_proposal.rb
dgdavid: Curious, that was the first name that I chose, but sounds weird to me. So, if you are fine with the current one, I'll keep it as it.

src/lib/y2storage/disk_analyzer.rb
joseivanlopez: I would use "and".

src/lib/y2storage/disk_analyzer.rb
joseivanlopez: Yes, I prefer to see "and" there, because I only want to know what kind of devices it could return, not the conditions in which it returns this or the other.

src/lib/y2storage/guided_proposal.rb
joseivanlopez: As mentioned, I am fine with your proposal ;)

ancorgs:
@dgdavid I see several important changes here (proposing on top of MD RAIDs, fixed detection of Linux systems over MD RAIDs, USB disks are used as last resort instead of being always discarded if there is other disk) and no changelog entry.

dgdavid:
@ancorgs 

Sorry, I did not see your commit until now, a few hours after added the changelog in #792, where I only mentioned the main change. Shall I to update it again adding the all of them?

Once again, my apologies for this oversight :roll_eyes:

PR #794, opened 2018-11-08
SpaceMaker: refactor the code to be more flexible

ancorgs:
First step for https://trello.com/c/DVQH3p2n/397-8-bug1057436-improve-confusing-spacemaker-strategy-fate325885-preserve-irst-partitions-in-guided-proposal

This basically implements what is described at https://github.com/yast/yast-storage-ng/blob/master/doc/space_maker_improve_strategy.md

Please don't be too picky about name or level of documentation of internal methods. The next step would be to adapt them in order to support IRST and similar stuff.

It's much easier to review commit by commit.

### DIFF
--- a/src/lib/y2storage/proposal/space_maker.rb
+++ b/src/lib/y2storage/proposal/space_maker.rb
@@ -284,8 +284,9 @@ module Y2Storage
       # @return [Boolean] true if some operation was performed. False if nothing
       #   else could be done to reach the goal.
       def execute_next_action(planned_partitions, lvm_helper)
-        part = next_partition_to_resize
-        if part
+        part_sid = next_partition_to_resize
+        if part_sid
+          part = new_graph.find_device(part_sid)
           target_shrink_size = resizing_size(part, planned_partitions)
           shrink_partition(part, target_shrink_size)
         else
@@ -374,7 +375,7 @@ module Y2Storage
         settings.candidate_devices
       end
 
-      # Next partition to be deleted by {#resize_and_delete}
+      # Sid of the next partition to be resized by {#resize_and_delete}
       #
       # @return [Partition]
       def next_partition_to_resize
@@ -387,16 +388,27 @@ module Y2Storage
         log.info("Evaluating the following Windows partitions: #{part_names}")
 
         parts_by_disk = partitions_by_disk(part_names)
-        remove_linux_disks!(parts_by_disk) unless no_more_linux_or_other?
 
-        sorted_resizables(parts_by_disk.values.flatten).first
+        # Historical behavior: if there are still any Linux or "other" partition
+        # to delete, the Windows systems that are in the same disk than any Linux
+        # partition are not resized.
+        # Rationale: Users having a Windows and a Linux in the same disk have
+        # likely already resized Windows once (when installing that Linux).
+        # So they probably don't want to resize it again. But if there are no
+        # more Linux or "other" partitions to delete, then the next partition to
+        # delete would be a Windows one. In that case, all Windows partitions
+        # are candidates to be resized (it's better to resize than to delete).
+        remove_linux_disks!(parts_by_disk) if any_linux_or_other_left?
+
+        partition = sorted_resizables(parts_by_disk.values.flatten).first
+        partition.nil? ? nil : partition.sid
       end
 
       # @see #next_partition_to_resize
       #
       # @return [Boolean]
-      def no_more_linux_or_other?
-        (sorted_candidates[:linux] + sorted_candidates[:other] - new_graph_deleted_sids).empty?
+      def any_linux_or_other_left?
+        (sorted_candidates[:linux] + sorted_candidates[:other] - new_graph_deleted_sids).any?
       end
 
       # @return [Hash{String => Partition}]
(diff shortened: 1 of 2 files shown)


ancorgs:
Please note the 4 codeclimate issues are not really new issues, just moved ones (four issues fixed, four issues introduced).

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -237,57 +269,81 @@ def resize_and_delete(planned_partitions, keep, lvm_helper, disk: nil)
...
+      #
+      # @return [Boolean] true if some operation was performed. False if nothing
+      #   else could be done to reach the goal.
+      def execute_next_action(planned_partitions, lvm_helper)
+        part = next_partition_to_resize
+        if part
+          target_shrink_size = resizing_size(part, planned_partitions)
+          shrink_partition(part, target_shrink_size)
+        else
+          part_sid = next_partition_to_delete
```
joseivanlopez: I know that `nex_partiton_to_resize` and `next_partition_to_delete` are only internal methods, but for the sake of uniformity I would return the same in both cases, sid or Partition.

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -237,57 +269,81 @@ def resize_and_delete(planned_partitions, keep, lvm_helper, disk: nil)
...
+        sorted_candidates[:linux] + sorted_candidates[:other] + sorted_candidates[:windows]
+      end
+
+      # Sid of the next partition to be deleted by {#resize_and_delete}
+      #
+      # @return [Integer]
+      def next_partition_to_delete
+        result = (all_sorted_candidates - new_graph_deleted_sids).first
+
+        log.info "Next partition to delete: #{result}"
```
joseivanlopez: NP: it is a bit weird to see in log: "Next partition to delete: nil". I would print a more specific message when there is no partition to delete.

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -318,39 +374,38 @@ def candidate_disk_names
         settings.candidate_devices
       end
 
-      # Try to resize the existing windows partitions
+      # Next partition to be deleted by {#resize_and_delete}
```
joseivanlopez: "to be resized"

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -293,39 +374,38 @@ def candidate_disk_names
...
+        return nil unless settings.resize_windows
+        part_names = windows_part_names(disk_name)
+        return nil if part_names.empty?
 
-          success?(planned_partitions)
-        end
+        log.info("Evaluating the following Windows partitions: #{part_names}")
+
+        parts_by_disk = partitions_by_disk(part_names)
+        remove_linux_disks!(parts_by_disk) unless no_more_linux_or_other?
```
joseivanlopez: Quite difficult to read the double negation. Why not the following:
```suggestion
        remove_linux_disks!(parts_by_disk) if more_linux_or_other?
```

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -293,39 +374,38 @@ def candidate_disk_names
...
+
+        sorted_resizables(parts_by_disk.values.flatten).first
+      end
 
-        log.info "Didn't manage to free enough space by resizing Windows" unless success
+      # @see #next_partition_to_resize
+      #
+      # @return [Boolean]
+      def no_more_linux_or_other?
```
joseivanlopez: I said above, I would define it as:
```suggestion
      def more_linux_or_other?
        (sorted_candidates[:linux] + sorted_candidates[:other] - new_graph_deleted_sids).any?
      end
      
```

src/lib/y2storage/proposal/space_maker.rb
joseivanlopez: And another thing, why don't we resize windows when the disk contains a linux? I don't remember the reason.

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -451,29 +491,32 @@ def filter_partitions_by_type!(partitions, type, disk)
...
-      # @param planned_partitions [Array<Planned::Partition>] list of
-      #   partitions to allocate, used to know how much space is still missing
-      # @param disk [String] optional disk name to restrict operations to
       # @param lvm_helper [Proposal::LvmHelper] contains information about how
       #     to deal with the existing LVM volume groups
-      def delete_disk_content(planned_partitions, disk, lvm_helper)
-        log.info "BEGIN delete_disk_content with disk #{disk}"
+      # @return [BlkDevice]
+      def next_disk_to_wipe(lvm_helper)
+        log.info "Searching next disk to wipe (disk_name: #{disk_name})"
```
joseivanlopez: I would log this only if the action is actually restricted to one disk.

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -451,29 +491,32 @@ def filter_partitions_by_type!(partitions, type, disk)
...
       # like disks directly formatted or used as members of an LVM or software RAID.
       #
-      # @param planned_partitions [Array<Planned::Partition>] list of
-      #   partitions to allocate, used to know how much space is still missing
-      # @param disk [String] optional disk name to restrict operations to
       # @param lvm_helper [Proposal::LvmHelper] contains information about how
       #     to deal with the existing LVM volume groups
-      def delete_disk_content(planned_partitions, disk, lvm_helper)
-        log.info "BEGIN delete_disk_content with disk #{disk}"
+      # @return [BlkDevice]
```
joseivanlopez: Please, document that it could return nil.

src/lib/y2storage/proposal/space_maker.rb
ancorgs: That was already there in the very first prototype of the SpaceMaker. If I remember correctly the idea is that when you have a Windows and a Linux in the same disk, you have already resized Windows once (when you installed that Linux), so you likely don't want to resize it again.

src/lib/y2storage/proposal/space_maker.rb
ancorgs: Actually I find it kind of convenient to have a stable string for grepping.

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -374,7 +375,7 @@ def candidate_disk_names
         settings.candidate_devices
       end
 
-      # Next partition to be deleted by {#resize_and_delete}
+      # Sid of the next partition to be resized by {#resize_and_delete}
       #
       # @return [Partition]
```
joseivanlopez: NP: return type, now Integer or nil :)

PR #797, opened 2018-11-09
Improve initial proposal

joseivanlopez:
PBI: https://trello.com/c/GHmBJbGc/434-3-move-populatedsettings-to-guidedsetupinitial

Notes for reviewers:

* Codeclimate is complaining, but IMHO, in this case it does not deserve fixing it.
* Test coverage is decreasing a little bit, but those lines that have started to be uncovered were not directly tested before. They were tested just by luck.


src/lib/y2storage/proposal/settings_generator/base.rb
```diff
@@ -0,0 +1,112 @@
...
+        # @return [ProposalSettings, nil] nil if nothing else can be disabled in the current settings
+        def next_settings
+          initial_settings || calculate_next_settings
+        end
+
+      private
+
+        # Calculates the next settings to use after the initial settings
+        #
+        # Each derived class must implement this method.
```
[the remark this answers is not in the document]
joseivanlopez: You are right, I was very repetitive there.

doc/proposal.md
```diff
@@ -49,24 +49,26 @@ That second step is performed by an instance of
...
+instance of `Proposal::PlannedDevicesGenerator` is asked for the desired set of
+planned devices, so it will aim for the best possible size for each planned
+device. If the instance of `Proposal::DevicegraphGenerator` is not able to
+accommodate those planned devices, a second attempt will be performed. On that
+second attempt, the device generator will aim for a minimalistic set of planned
+devices, reducing the size expectations as much as possible. If the devicegraph
+generator also fails to allocate that smaller version of the planned devices, an
+exception is raised.
+
+During the installation a first proposal is automatically calculated when the
```
dgdavid: NP: there is an extra `the`

doc/proposal.md
```diff
@@ -49,24 +49,26 @@ That second step is performed by an instance of
...
+the installer reaches the proposal step. This initial proposal is performed
+without user interaction, and it is based on the settings defined for the current
+product. There is a specialized class `InititalGuidedProposal` to calculate this
+proposal. The initial proposal considers all available devices in the system as possible
```
dgdavid: Suggestion: s/The initial proposal/It

dgdavid: or s/this proposal. The initial proposal consider/this proposal, considering

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Fri Nov 16 09:32:05 UTC 2018 - jlopez@suse.com
+
+- Initial proposal tries all possible attempts over each candidate
+  device before switching to another device.
+- Related to bug#1115819 and fate#326573.
```
ancorgs: This is indeed related to those bugzilla/fate entries. But also (even more closely) to [bug#1102026](https://bugzilla.suse.com/show_bug.cgi?id=1102026) (which complains about the default proposal not sticking to the first disk) and [bug#1090383](https://bugzilla.suse.com/show_bug.cgi?id=1090383) (which also states the proposal should only try the first disk unless told to try something else).

src/lib/y2storage/guided_proposal.rb
```diff
@@ -39,117 +39,135 @@ module Y2Storage
...
-
-      exception = nil
-      saved_root_device = populated_settings.root_device
-
-      groups = group_candidate_devices
-      target_sizes.product(groups).each do |target, devices|
-        # reset root_device, else #candidate_roots will just use it
-        populated_settings.root_device = saved_root_device
-        populated_settings.candidate_devices = devices
+      try_with_each_target_size
```
ancorgs: Why this level of indirection? I mean having a method that is one-liner simply calling another method that, AFAICS, is not called anywhere else. :thinking:

src/lib/y2storage/guided_proposal.rb
```diff
@@ -39,117 +39,135 @@ module Y2Storage
...
-            exception = error
-          end
+      target_sizes.each do |target_size|
+        begin
+          log.info "Trying to make a proposal with target size: #{target_size}"
 
-          return true unless exception
+          @planned_devices = planned_devices_list(target_size)
+          @devices = devicegraph(@planned_devices)
+          # freezes settings only when proposal was correctly calculated
```
ancorgs: Maybe I will understand if I continue reading but... why only if we succeed? Shouldn't they be frozen always after the last attempt, no matter whether it succeeded or not?

Moreover, the fact that the passed settings are modified/completed during the execution should be very clearly documented.

src/lib/y2storage/guided_proposal.rb
joseivanlopez: That `try_with_each_target_size` is used in derived class `InitialGuidedProposal`. The idea is that `GuidedProposal` only has to try with different sizes to calculate a proposal, and `InitialGuidedProposal` has to do more things before trying with different sizes.

src/lib/y2storage/guided_proposal.rb
joseivanlopez: Umm, the problem is that `InitialGuidedProposal` cannot freeze the settings until it has done all possible attempts, which implies to call `try_with_each_target_size` several times. Maybe, we can freeze the settings in `#calculate_proposal` method.

src/lib/y2storage/guided_proposal.rb
```diff
@@ -39,117 +39,135 @@ module Y2Storage
...
+          settings.freeze
+          return true
+        rescue Error => error
```
ancorgs: I have to admit that I had to double check if the exception was assigned to the pre-existing variable `error` or whether a new local `error` variable got created (that are the kind of things that could change between major Ruby versions).

In that regard, I find the old code more readable... and maybe also more secure for an hypothetical change in an hypothetical Ruby 3.0.

joseivanlopez: I don't get what you mean. A block is a closure, so there is no way to use an internal `error` variable. You have to explicitly declare the variable as a local to the block to not use the external `error`.

```
error = nil

foo do |bar|
  error = "foo"
end

p error #=> foo

foo do |bar; error| # error declared as local
  error = "foo"
end

p error #=> nil
``` 

If ruby changes this, it will break all existing programs.

src/lib/y2storage/guided_proposal.rb
```diff
@@ -39,117 +39,135 @@ module Y2Storage
...
+      # will be calculated.
       #
-      # @see GuidedProposal#initialize
-      # @see Proposal::InitialStragegies::Legacy#initial_proposal
-      # @see Proposal::InitialStragegies::Ng#initial_proposal
+      # @see InitialGuidedProposal
+      # @see #initialize
       #
-      # @param settings [ProposalSettings] if nil, default settings will be used
-      # @param devicegraph [Devicegraph] starting point. If nil, the probed
```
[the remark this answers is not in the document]
joseivanlopez: Umm, it is defined in `InitialGuidedProposal#initialize`. I replaced it by a `@see InitialGuidedProposal`. Should I restore it?

src/lib/y2storage/guided_proposal.rb
ancorgs: Never mind, I found why in the subclass.

src/lib/y2storage/guided_proposal.rb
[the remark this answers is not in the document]
ancorgs: Oh, I found it in `GuidedProposal#initialize`. Never mind, I had somehow overlooked it there.

src/lib/y2storage/guided_proposal.rb
ancorgs: That would work better, IMO.

src/lib/y2storage/guided_proposal.rb
ancorgs: Your example about closures doesn't apply 1:1 to this case. I'm NOT discussing the clarity of this case:

```ruby
error = nil

foo do |bar|
  error = "foo"
  puts "#{bar}"
end
```

Anyway, I don't have a strong opinion. I'm fine with the current version after all.

But the clarity of this one:

```ruby

error = nil

foo do |bar|
  puts "#{bar}"
rescue Exception => error
  puts "catched"
end
```

PR #798, opened 2018-11-13
SpaceMaker: IRST support + extract prospect actions to dedicated classes

ancorgs:
Main PR for https://trello.com/c/DVQH3p2n/397-bug1057436-improve-confusing-spacemaker-strategy-fate325885-preserve-irst-partitions-in-guided-proposal

Review commit by commit.

Bonus: fixed 4 historical CodeClimate issues in SpaceMaker

### DIFF
--- a/src/lib/y2storage/proposal/space_maker.rb
+++ b/src/lib/y2storage/proposal/space_maker.rb
@@ -119,7 +119,7 @@ module Y2Storage
         prospects = SpaceMakerProspects::List.new(settings, disk_analyzer)
 
         disks_for(result).each do |disk|
-          prospects.unwanted_partition_entries(disk).each do |entry|
+          prospects.unwanted_partition_prospects(disk).each do |entry|
             sids = partition_killer.delete_by_sid(entry.sid)
             @all_deleted_sids.concat(sids)
           end
@@ -247,7 +247,7 @@ module Y2Storage
 
         prospects = SpaceMakerProspects::List.new(settings, disk_analyzer)
         disks_for(new_graph, disk_name).each do |disk|
-          prospects.add_entries(disk, lvm_helper, keep)
+          prospects.add_prospects(disk, lvm_helper, keep)
         end
 
         until success?(planned_partitions)
@@ -266,7 +266,7 @@ module Y2Storage
       # @return [Boolean] true if some operation was performed. False if nothing
       #   else could be done to reach the goal.
       def execute_next_action(planned_partitions, prospects, disk_name = nil)
-        prospect_action = prospects.next_available_entry
+        prospect_action = prospects.next_available_prospect
         if !prospect_action
           log.info "No more prospects for SpaceMaker (disk_name: #{disk_name})"
           return false
--- a/src/lib/y2storage/proposal/space_maker.rb
+++ b/src/lib/y2storage/proposal/space_maker.rb
@@ -1,5 +1,3 @@
-#!/usr/bin/env ruby
-#
 # encoding: utf-8
 
 # Copyright (c) [2015] SUSE LLC
--- a/src/lib/y2storage/proposal/space_maker.rb
+++ b/src/lib/y2storage/proposal/space_maker.rb
@@ -29,6 +29,7 @@ require "y2storage/disk_size"
 require "y2storage/free_disk_space"
 require "y2storage/proposal/partitions_distribution_calculator"
 require "y2storage/proposal/partition_killer"
+require "y2storage/proposal/space_maker_prospects"
 
 module Y2Storage
   module Proposal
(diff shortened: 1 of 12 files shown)


src/lib/y2storage/proposal/space_maker_prospects/partition_prospect.rb
```diff
@@ -0,0 +1,66 @@
...
+          @region_start = partition.region.start
+          @analyzer = disk_analyzer
+        end
+
+        # Whether there was a Linux partition in the same disk of the target
+        # partition (in the original devicegraph).
+        #
+        # @return [Boolean]
+        def linux_in_disk?
+          return @linux_in_disk unless @linux_in_disk.nil?
```
dgdavid: Why not 

```ruby
return @linux_in_disk if @linux_in_disk
```
?

dgdavid: Or directly use memoization

```ruby
 @linux_in_disk ||= analyzer.linux_partitions(disk_name).any?
```

src/lib/y2storage/proposal/space_maker_prospects/partition_prospect.rb
```diff
@@ -52,6 +52,16 @@ def linux_in_disk?
           @linux_in_disk = analyzer.linux_partitions(disk_name).any?
         end
 
+        # Whether there was a Windows system in the same disk of the target
+        # partition (in the original devicegraph).
+        #
+        # @return [Boolean]
+        def windows_in_disk?
```
dgdavid: The same than above, why not use `if` w/o `nil?` instead `unless` or directly a memoization?

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -114,14 +115,19 @@ def provide_space(original_graph, planned_partitions, lvm_helper)
       #   partitions
       def delete_unwanted_partitions(original_graph)
         result = original_graph.dup
-        partition_killer = PartitionKiller.new(result, candidate_disk_names)
+        types = [:windows, :linux, :other].select { |t| settings.delete_forced?(t) }
+        log.info "Types to forcely delete: #{types}"
```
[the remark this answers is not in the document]
ancorgs: Actually that line disappears in the next commit.

src/lib/y2storage/proposal/space_maker_prospects/partition_prospect.rb
ancorgs: Those tricks don't work with boolean variables. You really need to check with `#nil?` Otherwise `false` values are not cached. Isn't it?

src/lib/y2storage/proposal/space_maker_prospects/partition_prospect.rb
ancorgs: Same than above.

src/lib/y2storage/proposal/space_maker_prospects/partition_prospect.rb
dgdavid: Yes, you're right... I forgot this drawback memoizating `false` values.

src/lib/y2storage/proposal/space_maker_prospects/list.rb
```diff
@@ -0,0 +1,239 @@
...
+          else
+            @all_delete_partition_entries[type]
+          end
+        end
+
+        # Next prospect action that should be executed by SpaceMaker
+        #
+        # @return [SpaceMakerProspects::Base, nil] nil if there are no more
+        #   available prospects
+        def next_available_entry
```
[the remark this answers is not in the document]
joseivanlopez: Maybe, it would be clearer an implementation like:

```suggestion
        def next_available_entry
          next_resize_partition(allow_linux_in_disk: false) ||
            next_delete_partition(allow_windows: false) ||
            next_wipe_disk ||
            next_resize_partition(allow_linux_in_disk: true) ||
            next_delete_partition(allow_windows: true)
        end
```

src/lib/y2storage/proposal/space_maker_prospects/base.rb
```diff
@@ -0,0 +1,69 @@
...
+        #
+        # @return [Boolean] false if the action has already being performed or
+        #   if it has become impossible as side effect of another performed action
+        def available?
+          @available
+        end
+
+        # @return [String]
+        def to_s
+          "<#{sid} (#{device_name})>"
```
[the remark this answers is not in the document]
ancorgs: I though about it, but in all the places in which this is used (for logging purposes), it actually adds more noise/redundancy than useful information.

src/lib/y2storage/proposal/space_maker_prospects/list.rb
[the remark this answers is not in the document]
ancorgs: Hmmm, maybe. But the `:allow_windows` name is already imprecise. It actually excludes Windows systems AND partitions marked as `#last_resort?`...

So I would keep it as it is for the time being. Trying to shorten it too much can lead to a lot of new argument in the methods with not-so-precise names.

PR #808, opened 2018-12-04
Avoid unsafe characters for subvolumes path

joseivanlopez:
PBI: https://trello.com/c/m7eAH7Dw/525-storage-subvolumes

### DIFF
--- a/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
+++ b/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
@@ -1,6 +1,6 @@
 # encoding: utf-8
 
-# Copyright (c) [2017] SUSE LLC
+# Copyright (c) [2017-2018] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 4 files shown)


src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -140,44 +151,50 @@ def focus
...
-        # @return [Boolean] true if path does not exist
-        def uniqueness_validation
-          return true unless exist_path?
+        # @return [String, nil] nil if the path does not contain illegal characters
+        def content_error
+          regexp = Regexp.new(".*[#{ILLEGAL_CHARS}]+.*")
 
-          Yast::Popup.Error(format(_("Subvolume name %s already exists."), value))
-          false
+          return nil unless regexp.match?(value)
```
jreidinger: this is not needed to construct such complex regexp. Simply use
`return nil unless regexp.match?(/[#{ILLEGAL_CHARS}]/)`

src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -140,44 +151,50 @@ def focus
...
+
+          _("Subvolume path contains illegal characters. Be sure it does\n" \
+            "not include white spaces, tabs, line breaks or commas.")
```
jreidinger: hmm, this is not much robust when we extend list of illegal characters. Why not simply list them out? with maybe some escaping with inspect?

lslezak: Hm, displaying something like `\t` would be really confusing for the beginners as you have `Tab` written on the physical keyboard. Similar to `\n` and others, I'd definitely avoid that.

joseivanlopez: As suggested by @lsleza, I have extended the message with "or similar special characters".

PR #815, opened 2018-12-18
 Do not touch eMMC boot partitions (mmcblk*boot*) (bsc#1119316)

schubi2:


src/lib/y2storage/disk.rb
```diff
@@ -92,5 +92,16 @@ def preferred_ptable_type
...
+
+    # Whether this device can be in general treated like a disk for YaST
+    # purposes
+    #
+    # @see Devicegraph::disk_devices
+    #
+    # @return [Boolean]
+    def disk_device?
+      # Filter out all eMMC boot partitions (mmcblk*boot*)
+      basename =~ /mmcblk\d+boot\d+/ ? false : super
```
ancorgs: Just in case, I would use `^` and `$` here, just to make sure the expected sequence is not in the middle of a VERY UNFORTUNATE name.

On the other hand, according to (this link](https://github.com/endlessm/systemd/commit/0efe40e3bd0e85483e940314fa91adb036b7420d) provided in one of the comments of the bug, looks like we also need to filter out names line "/dev/mmcblk1rpmb"

PR #819, opened 2019-01-08
Find reusable MD devs also by sysfs_name (bsc#1120979)

pief:


pief:
Full disclosure: this is my first Ruby code whatsoever and also my first contribution to Yast sources. Compared to BSC#1120979 I had to move the extended find into a function of its own to satisfy the Rubocop line length check.

ancorgs:
I don't think this is the correct fix for the problem. The problem is that the name of the `Planned::Md` object is "/dev/md0" and the name of the `Y2Storage::Md` object is "/dev/md/0".

So what we are doing wrong is the generation of the `Planned::Md` name. In other words, I would say the real bug is here: https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/autoinst_profile/partition_section.rb#L231

ancorgs:
Closed in favor of #835

PR #820, opened 2019-01-09
Tests for resizing Windows

ancorgs:
See https://trello.com/c/gdXTlipJ/288-bug1057436-chapter-three-better-calculation-of-resize-windows

These tests proves that there are several situations in which the over-simplistic logic to calculate how much a partition must be reduced leads to Windows been reduced less than needed... which can have very bad consequences.

In fact, the current simplistic logic (labeled as temporary but never improved so far) only works reliably when there are no other spaces in the disk, apart from the one we are planning to create by resizing.


ancorgs:
This is included (and the proven bugs fixed) in #822

PR #821, opened 2019-01-09
Link to storage client

joseivanlopez:
Installation summary links to storage client instead of Expert Partitioner directly. Moreover, the Guided Setup wizard can be now canceled, see https://bugzilla.suse.com/show_bug.cgi?id=1121442.

PBI: https://trello.com/c/53IEUxrB/577-susecaas-platform-4-p2-1099485-caasp-specific-partitioning-warning-dialog-missing-in-storage-ng

~~Requires this be merged: https://github.com/yast/yast-yast2/pull/886~~

### DIFF
--- a/test/y2storage/clients/partitions_proposal_test.rb
+++ b/test/y2storage/clients/partitions_proposal_test.rb
@@ -222,22 +222,20 @@ describe Y2Storage::Clients::PartitionsProposal do
     end
 
     describe "#ask_user" do
-      RSpec.shared_context "run expert partitioner" do
+      RSpec.shared_context "run storage client" do
         before do
-          allow(Y2Partitioner::Dialogs::Main).to receive(:new).and_return(expert_dialog)
-          allow(expert_dialog).to receive(:run).and_return(result)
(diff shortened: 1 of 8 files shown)


test/y2storage/clients/partitions_proposal_test.rb
```diff
@@ -257,20 +255,18 @@
...
-        context "and the expert partitioner returns :next" do
-          let(:result) { :next }
+        context "and the storage client returns :cancel" do
+          let(:result) { :cancel }
 
-          before do
-            allow(expert_dialog).to receive(:device_graph).and_return(new_devicegraph)
+          it "returns a hash with :back for 'workflow_sequence' key" do
+            result = subject.ask_user(param)
+            expect(result).to be_a(Hash)
```
dgdavid: NP: I think that line is not really necessary. The above expectation should be enough.

joseivanlopez: Well, due to the nature of `#ask_user` it should return a hash, I would prefer to check it explicitly.

PR #822, opened 2019-01-10
Windows better resize

ancorgs:
This basically implements https://trello.com/c/gdXTlipJ/288-bug1057436-chapter-three-better-calculation-of-resize-windows

With this, the tests at #816 are all green, which means:

- The resizing calculation is not longer confused by other (probably useless) spaces in the disk. This used to have catastrophic consequences.
- When resizing a Windows with a misaligned end (quite common in GPT), we don keep the misalignment. Instead, we take the opportunity to fix it and avoid the gap.

In addition, the remaining test (most adaptations were done already in #825) is now adapted to the new (more correct) behavior because:

- Windows partitions are not always reduced 1 extra MiB just in case. Only when there is a real misalignment to fix.
- With the previous behavior, when there was a free space >1MiB at the end of a GPT partition, the partition being resized was reduced 16.5KiB more than actually needed (even introducing a misalignment). No matter if that partition was not at the end of the disk. That doesn't happen anymore.

### DIFF
--- a/src/lib/y2storage/planned/assigned_space.rb
+++ b/src/lib/y2storage/planned/assigned_space.rb
@@ -153,7 +153,11 @@ module Y2Storage
       #
       # @return [DiskSize]
       def total_missing_size
-        total_needed_size - disk_space.disk_size
+        if disk_space.disk_size >= total_needed_size
+          DiskSize.zero
+        else
+          total_needed_size - disk_space.disk_size
+        end
       end
 
       # Space consumed by the EBR of one logical partition in a given disk
--- a/src/lib/y2storage/planned/assigned_space.rb
+++ b/src/lib/y2storage/planned/assigned_space.rb
@@ -80,6 +80,7 @@ module Y2Storage
       def valid?
         return false if wrong_usage_of_reused_partition?
         return false unless primary_partitions_fit?
+        return true if disk_space.growing?
         return true if usable_size >= DiskSize.sum(partitions.map(&:min), rounding: align_grain)
         # At first sight, there is no enough space, but maybe enforcing some
         # order...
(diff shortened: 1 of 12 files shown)


ancorgs:
@joseivanlopez This is now ready for a first serious review.

src/lib/y2storage/proposal/partitions_distribution_calculator.rb
```diff
@@ -326,6 +331,124 @@ def add_unused_spaces(dist_hashes, spaces)
...
+      # @see #all_planned_partitions
+      #
+      # @return [Planned::Partition]
+      def planned_single_pv
+        res = Planned::Partition.new(nil)
+        res.min_size = lvm_space_to_make(1)
+        res
+      end
+
+      def missing_size_in_growing_space(dist_hashes, align_grain)
```
joseivanlopez: As discussed offline, please document this method.

src/lib/y2storage/proposal/partitions_distribution_calculator.rb
```diff
@@ -326,6 +331,124 @@ def add_unused_spaces(dist_hashes, spaces)
...
+      #
+      # @see resizing_size
+      # @see FreeDiskSpace#growing?
+      #
+      # @param free_spaces [Array<FreeDiskSpace>] initial list of free spaces in
+      #   the system (#growing? returns false for all of them)
+      # @param partition [Partition] partition to be resized
+      # @return [Array<FreeDiskSpace>] list of free spaces containing a growing one
+      #   (added to the list or replacing the original space affected by the resizing)
+      def add_or_mark_growing_space(free_spaces, partition)
```
joseivanlopez: As commented offline, maybe a comment the case where the next free space is the first region of an extended partition. And a test case?

src/lib/y2storage/planned/assigned_space.rb
```diff
@@ -134,6 +135,27 @@ def usable_size
...
       end
 
+      # Total size needed to actually allocate all the assigned planned partitions
+      # in the space, no matter what the real size of the space is
+      #
+      # Used when resizing existing partitions to make space.
+      #
+      # @return [DiskSize]
+      def total_needed_size
+        result = DiskSize.sum(partitions.map(&:min), rounding: align_grain)
```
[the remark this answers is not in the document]
ancorgs: Well, it's just a sum of sizes. Not sure if that's worth a test.

PR #824, opened 2019-01-17
Pop-up for exceptions on actiongraph calculation (bsc#1120070)

ancorgs:
See https://trello.com/c/GddxSfg1/599-3-sles15-sp1-1120070-sles15-sp1-beta1zeppelin-storage-exception-occurred-while-installation

The real fix is preventing libstorage-ng from raising the exception (https://github.com/openSUSE/libstorage-ng/pull/604), but this also helps by replacing the default YaST pop-up (basically a crash) with something more informative.

**Note:** the text of the message changed a bit. Screenshots need to be adjusted.

popup

When clicking in "Details...", the message from the exception is shown.

popup-details


ancorgs:
I created the following PBI for the mentioned follow-up changes. https://trello.com/c/iMoOGVxg/310-follow-up-of-bsc1120070-report-invalid-probed-devicegraph

I will now adapt the current message a bit and add a `TODO` to the code, so it doesn't get overlooked.

PR #825, opened 2019-01-22
SpaceMaker: avoid two passes if not needed (bug#1057436) 

ancorgs:
See https://trello.com/c/0ZARc9rP/602-1-bug1057436-chapter-two-avoid-two-passes-if-not-needed

Review by commit.

The **second commit** adapts the test in a way that is equivalent to the adaptations found in https://github.com/yast/yast-storage-ng/pull/822. That's because this pull request mitigates the effect of some problems of the original `PartitionsDistributionCalculator.#resizing_size` that are fixed in that other PR. To be precise, this "hides" the following problems described in the other pull request

- Windows partitions are reduced 1 extra MiB if there is some space in the disk. That is done just in case, even if there is no real misalignment to fix. With two passes, the first resizing introduced that space, so the second one added the extra MiB.
- When there is a free space >1MiB at the end of a GPT partition, the partition being resized is reduced 16.5KiB more than actually needed (even introducing a misalignment). Again, the first resizing introduced such space, so the second one was affected by the extra 16.5KiB.

As said, #822 really fixes those problems in `PartitionsDistributionCalculator.#resizing_size`, the current PR only hides then in the tests because it avoids calling the method more than strictly needed.

The **third and fourth commits** are there only to make debugging easier (based on experience).

### DIFF
--- a/src/lib/y2storage/proposal/space_maker.rb
+++ b/src/lib/y2storage/proposal/space_maker.rb
@@ -246,6 +246,9 @@ module Y2Storage
       # @param disk_name [String, nil] optional disk name to restrict operations to
       #
       def resize_and_delete(planned_partitions, keep, lvm_helper, disk_name: nil)
+        # Note that only the execution with disk_name == nil is the final one.
+        # In other words, if disk_name contains something, it means there will
+        # be at least a subsequent call to the method.
         log.info "Resize and delete. disk_name: #{disk_name}, planned partitions:"
         planned_partitions.each do |p|
           log.info "  mount: #{p.mount_point}, disk: #{p.disk}, min: #{p.min}, max: #{p.max}"
--- a/src/lib/y2storage/proposal/space_maker.rb
+++ b/src/lib/y2storage/proposal/space_maker.rb
@@ -176,16 +176,25 @@ module Y2Storage
         # anywhere but sometimes it is known in advance on which disk they
         # should be created.
         #
-        # Start by assigning space to them.
-        #
-        # The result (if successful) is kept in @distribution.
-        #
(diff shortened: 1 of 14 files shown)


src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -385,6 +397,26 @@ def disks_for(devicegraph, device_name = nil)
...
+        # wouldn't be executed for particular disks anyway
+        return false if parts_by_disk.empty?
+
+        return true if parts_by_disk.size > 1
+
+        # In theory, the specific disk mentioned in the planned partitions
+        # (note that, at this point, we are sure there is only one) should be
+        # included in the set of candidate disks. Return false if that's not the
+        # case or if there is more than one candidate disk.
+        parts_by_disk.keys != candidate_disk_names
```
joseivanlopez: Why cannot there be several candidates? I guess we can have several candidate devices but current planning is using only one. I guess I am missing something.

joseivanlopez: Yes, I am fine with your solution. But, thinking again on it, maybe the following is enough:

```
if candidate_disks.size > 1
  planned_partitions_by_disk(partitions).each do |disk, parts|
    resize_and_delete(parts, keep, lvm_helper, disk_name: disk)
  end
end

resize_and_delete(partitions, keep, lvm_h
```

Yes, this solution is only valid when the keys of `parts_by_disks` are always included in the `candidate_disk_names` (and this should be always the case), but due to  the code was not ensuring it, I see your proposal a better solution to ensure the same behavior as before.

PR #827, opened 2019-01-23
do not include nil in package list of used storage features for FC devices (bsc#1122781)

aschnell:
For https://trello.com/c/OkmGMBQG/632-sle15-sp1-1122781-p1-installer-throws-error-message-after-selection-of-multipath-disk-on-sles-15-sp1-beta2.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jan 23 10:11:58 CET 2019 - aschnell@suse.com
+
+- do not include nil in package list of used storage features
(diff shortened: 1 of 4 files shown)


ancorgs:
Replying to myself. SLE-15-GA is not affected because the module `PackageSystem` in SLE-15 does not call `#shellescape` on each package name. That was introduced here https://github.com/yast/yast-yast2/commit/e857e1766a9c36a6106e8e26e4db0167b857b6e0

aschnell:
AFAIS the problem was introduced by https://github.com/yast/yast-storage-ng/commit/ee14493e4c5a98dca959591900ab677f1149f0c9.

PR #828, opened 2019-01-24
Adapt to bcache changes

joseivanlopez:
Add support for Flash-only Bcache devices.

Requires https://github.com/openSUSE/libstorage-ng/pull/603.

Part of PBI: https://trello.com/c/T7taTEix/606-5-use-the-bcache-command-for-probing


src/lib/y2partitioner/widgets/bcache_device_description.rb
```diff
@@ -63,9 +63,18 @@ def caching_device
       end
 
       def backing_device
-        return "" unless device.blk_device
+        return "" if device.is?(:flash_bcache) || device.backing_device.nil?
```
jreidinger: what about marking somehow flash only volume? As bcache can have two possibilities. bcache_cset that is not associated and flash_only volume. So what about using something like `_("(flash only volume)")` for them?

joseivanlopez:
Closed in favor of https://github.com/yast/yast-storage-ng/pull/829.

See https://github.com/openSUSE/libstorage-ng/pull/607 for details.

PR #829, opened 2019-01-30
Adapt to bcache changes

joseivanlopez:
Add support for Flash-only Bcache devices.

Requires openSUSE/libstorage-ng#608.

Part of PBI: https://trello.com/c/T7taTEix/606-5-use-the-bcache-command-for-probing


src/lib/y2storage/bcache.rb
```diff
@@ -33,36 +37,56 @@ class Bcache < Partitionable
...
     # @!attribute sequential_cutoff
     #   Size for cache consider read as sequential and do not cache it.
+    #
+    #   This method does not make sense for Flash-only Bcache devices.
+    #
     #   @return [DiskSize]
     storage_forward :sequential_cutoff, as: "DiskSize"
     storage_forward :sequential_cutoff=
 
     # @!method self.create(devicegraph, name)
```
ancorgs: I guess this should be something like

```
# @!method self.create(devicegraph, name, type = BcacheType::BACKED)
```

src/lib/y2storage/bcache.rb
```diff
@@ -84,14 +108,33 @@ class Bcache < Partitionable
     #   @return [String] full path to new bcache device like "/dev/bcache3"
     storage_class_forward :find_free_name
 
+    # Whether the Bcache is flash-only
+    #
+    # @return [Boolean]
+    def flash_only?
+      type == BcacheType::FLASH_ONLY
```
ancorgs: NP: a possible (nicer?) alternative: `type.is?(:flash_only)`

src/lib/y2storage/blk_device.rb
```diff
@@ -218,19 +218,19 @@ def udev_full_ids
     alias_method :create_filesystem, :create_blk_filesystem
 
     # @!method create_bcache(name)
-    #   Creates backing device build on current block device with given name.
+    #   Creates Bcache device with given name over the current block device.
```
ancorgs: maybe this is more clear saying "Creates a bcache device with the given name, using this device as backing device"

src/lib/y2storage/bcache.rb
```diff
@@ -33,36 +37,56 @@ class Bcache < Partitionable
...
 
-    # @!method blk_device
-    #   @return [BlkDevice] returns a backing device for cache
-    storage_forward :blk_device, as: "BlkDevice"
-
     # @!attribute cache_mode
     #   Mode in which cache operates.
+    #
+    #   This method does not make sense for Flash-only Bcache devices.
+    #
```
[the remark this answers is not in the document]
joseivanlopez: As we talked in jangouts, it simply works, although internally the library does not use that value at all.

src/lib/y2storage/bcache.rb
```diff
@@ -33,36 +37,56 @@ class Bcache < Partitionable
...
     #   @return [CacheMode]
     storage_forward :cache_mode, as: "CacheMode"
     storage_forward :cache_mode=
 
     # @!attribute writeback_percent
     #   Target percent of dirty pages in writeback mode.
+    #
+    #   This method does not make sense for Flash-only Bcache devices.
```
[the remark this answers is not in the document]
joseivanlopez: Same answer.

src/lib/y2storage/bcache.rb
joseivanlopez: Sure, much nicer ;). Thanks.

PR #832, opened 2019-01-31
Limit ESP to min size on aarch64 (bsc#1119318)

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1119318

## Trello

https://trello.com/c/vjCvQuGp/650-1-sles15-sp1-p2-1119318-installer-suggest-way-too-big-uefi-esp

## Problem

The UEFI ESP partition was considered much too large on aarch64. Small flash storage (16 GiB SD cards etc.) are quite common for RasPI, and using 512 MiB on such a device only for the ESP consumes a considerable part of the available storage already.

## Fix

Limit the size of that partition to the minimum on that architecture.

### DIFF
--- a/doc/boot-requirements.md
+++ b/doc/boot-requirements.md
@@ -133,14 +133,13 @@
 			- **requires to mount the existing firmware partition at /boot/vc**
 		- and there is no firmware partition in the system
 			- **only requires to use the existing EFI partition**
-- when proposing an new EFI partition
+- when proposing a new EFI partition
 	- **requires /boot/efi to be a non-encrypted vfat partition**
 	- **requires /boot/efi to be close enough to the beginning of disk**
 	- when aiming for the recommended size
-		- **requires /boot/efi to be exactly 500 MiB large (enough for several operating systems)**
+		- **requires /boot/efi to be exactly 256 MiB large (always FAT32 min size)**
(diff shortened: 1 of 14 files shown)


doc/boot-requirements.md
```diff
@@ -270,8 +270,11 @@
 	- when proposing an new EFI partition
 		- **requires /boot/efi to be a non-encrypted vfat partition**
 		- **requires /boot/efi to be close enough to the beginning of disk**
-		- when aiming for the recommended size
-			- **requires /boot/efi to be exactly 500 MiB large (enough for several operating systems)**
-		- when aiming for the minimal size
-			- **requires it to be at least 256 MiB (min size for FAT32 in drives with 4-KiB-per-sector)**
-			- **requires it to be at most 500 MiB (enough space for several operating systems)**
+		- if aarch64
```
joseivanlopez: How do you generate this changes? I don't expect this output according to the unit tests. In fact, x86 section should not change, but only Raspberry Pi section.

This document can be updated with `rake doc:bootspecs`. Did you do that?

shundhammer: Argh. No, of course not; this thing is checked in. So why do we even check that in if this is auto-generated?

PR #835, opened 2019-02-01
Fix MD reusing (bsc#1120979 and bsc#1121720)

ancorgs:
See https://trello.com/c/lZCZ7gmP/652-3-1121720-and-1120979-problems-reusing-mds-via-autoyast-caused-by-dev-md0-vs-dev-md-0

This synchronizes the name used for MDs by libstorage-ng and AutoYaST (now both use strings like "/dev/md/0"). That fixes the reported bugs about reusing existing MDs.

On the other hand, that also broke some scenarios when creating a new MD. This also includes a change to ensure that case works with both "/dev/md0" and "/dev/md/0".

Last but not least, there are extra commits to ensure the whole thing keeps working in the future if libstorage-ng changes its behavior and starts returning strings like "/dev/md0" for `Md#name` (instead of the current "/dev/md/0").

Moreover many tests has been added to ensure everything works with all reasonable combinations of the following (that's more than 30 scenarios/combinations):

- /dev/md/0 vs /dev/md0 used in the profile
- /dev/md/0 vs /dev/md0 reported by libstorage-ng
- Old format (using "/dev/md" + partition_nr) vs new format (specifying the device name) for `:CT_MD`
- Reusing vs creating an MD and its members (even trying to reuse non-existent things).

Bonus tracks: added some missing mocks in other parts of the testsuite that were likely producing undesired accesses to /dev.


src/lib/y2storage/planned/md.rb
```diff
@@ -101,6 +101,24 @@ def add_devices(md_device, devices)
...
+      # Whether the given name matches the name of the planned MD
+      #
+      # Apart from directly comparing the strings, this method is also
+      # able to compare a string with a format like "/dev/md0" (by default,
+      # the planned MD uses the libstorage-ng format for MD names, which looks
+      # like "/dev/md/0").
+      #
+      # @param value [String] name been searched for
+      # @return [Boolean] true if the names match
+      def name?(value)
```
joseivanlopez: In the hypothetical case that `libstorage-ng` returns `/dev/md0` and `autoyast` indicates `dev/md/0`, this method returns `false`. Should it return `true` for such situation? Or that situation cannot happen?

ancorgs: It shouldn't happen right now, at least using the typical AutoYaST profile.

But to make sure, I will try to enforce that situation by defining an unusual (but valid) profile and simulating the "/dev/md0" behavior for libstorage-ng. Maybe there is an opportunity window for this to fail in the future.

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -295,8 +295,8 @@ def reuse_mds(reused_mds, previous_result)
       # @return                [Proposal::CreatorResult] Result containing the specified MD RAIDs
       def create_mds(mds, previous_result, devs_to_reuse)
         mds.reduce(previous_result) do |result, md|
-          devices = result.created_names { |d| d.respond_to?(:raid_name) && d.raid_name == md.name }
-          devices += devs_to_reuse.select { |d| d.raid_name == md.name }.map(&:reuse_name)
+          devices = result.created_names { |d| d.respond_to?(:raid_name) && md.name?(d.raid_name) }
```
joseivanlopez: NP: as we were discussing, please, document here that `d.raid_name` is always equal to `md.name` when using the new AY format, or it is `/dev/md/X` otherwise.

src/lib/y2storage/proposal/autoinst_md_planner.rb
```diff
@@ -156,6 +155,23 @@ def raid_level_from(raid_options)
         issues_list.add(:invalid_value, raid_options.raid_type, :raid_type, "raid1")
         Y2Storage::MdLevel::RAID1
       end
+
+      # MD device to be reused by the given planned MD
+      #
+      # @param md      [Planned::Md] Planned MD RAID
+      def find_md_to_reuse(md)
```
joseivanlopez: NP: only a reminder. In SP1 we have the "alternative_name" mechanism for `Devicegraph#find_by_any_name`. Maybe we could use it.

ancorgs: I have implemented this and just found it brings no value, at least with the current implementation of `Devicegraph#find_by_any_name`. For each alternative name all methods are tried before passing to the next one. That means the disk lookup is potentially used even more often than before.

I will try to adapt the implementation of `Devicegraph#find_by_any_name`

PR #837, opened 2019-02-06
Partitioner: abort gracefully in installed system (bsc#1123837)

ancorgs:
See https://trello.com/c/deksXijN/671-3-sles15-sp1-p1-1123837-error-popup-shown-twice-followed-by-an-unhandled-exception

### DIFF
--- a/src/lib/y2partitioner/clients/main.rb
+++ b/src/lib/y2partitioner/clients/main.rb
@@ -51,10 +51,7 @@ module Y2Partitioner
           inhibitors = Y2Storage::Inhibitors.new
           inhibitors.inhibit
 
-          # Quit immediately if probing failed
-          return nil unless storage_manager.probed
-
-          return nil if partitioner_dialog.run != :next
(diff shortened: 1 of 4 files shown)


src/lib/y2partitioner/clients/main.rb
```diff
@@ -93,9 +93,15 @@ def commit
...
       #
-      # @return [Dialogs::Main]
+      # @return [Dialogs::Main, nil] nil if it was not possible to get the
+      #   devicegraphs (probing failed)
       def partitioner_dialog
-        @partitioner_dialog ||= Dialogs::Main.new(storage_manager.probed, storage_manager.staging)
+        return @partitioner_dialog if @partitioner_dialog
+
+        # Force quitting if probing failed
+        return nil unless storage_manager.probed
```
joseivanlopez: NP: I would add a comment to explain that a popup is shown with the error during the probing, and the client is forced quitting only when the user selects abort option.

PR #842, opened 2019-02-12
Limit Bcache Support to x86_64 Arch

shundhammer:
## Trello

https://trello.com/c/f1pcGKbD/673-3-bcache-in-non-x8664-systems

## JIRA

https://jira.suse.de/browse/SLE-4329

## Problem

Bcache is only really supported (on the kernel and upstream tools side) on a limited range of platforms. Currently, this is only the x86_64 architecture.

### Installation / Upgrade and Expert Partitioner in the Installed System

Right after probing, check if bcache is in use (i.e. if there is an existing bcache device).

Probing is safe since this only uses external tools like the `blkid` command and checks for `/sys` etc. paths that the kernel generated.

If there is an existing bcache, but we don't fully support bcache on that architecture, post a warning, but let the user proceed anyway.

This is implemented via the DevicegraphSanitizer which is run over the probed devicegraph to check for problems. Now a new error class for bcache is added if any bcache device is found and the platform does not support bcache.

### Partitioner

Don't offer bcache operations. Don't add the bcache branch to the tree in the partitioner.

## Testing

- Slightly hacked up the code to make the warning appear on my system and tested with "yast2 disk" on my workstation (see screenshot below)

- Checked manually that there is no bcache branch in the partitioner's tree, i.e. no bcache devices or operations are show

- Unit tests

## Screenshots

bcache-not-supported-warning

Warning message

### DIFF
--- a/src/lib/y2storage/devicegraph_sanitizer.rb
+++ b/src/lib/y2storage/devicegraph_sanitizer.rb
@@ -22,6 +22,8 @@
 require "yast"
 require "yast/i18n"
 require "y2storage/devicegraph"
+require "y2storage/bcache"
+require "y2storage/bcache_cset"
 
 Yast.import "Mode"
 
@@ -29,7 +31,7 @@ module Y2Storage
   # Class to sanitize a devicegraph
   #
   # A devicegraph can contain certain errors, for example, an LVM VG with missing PVs.
-  # This class fix wrong devices (typically by removing them).
+  # This class fixes wrong devices (typically by removing them).
   #
   # @example
   #   sanitizer = DevicegraphSanitizer.new(devicegraph)
@@ -88,6 +90,7 @@ module Y2Storage
       device = error.device
 
       fix_error_for_lvm_vg(devicegraph, device) if device.is?(:lvm_vg)
+      fix_error_for_bcache(devicegraph, device) if device.is?(:bcache) || device.is?(:bcache_cset)
       devicegraph
     end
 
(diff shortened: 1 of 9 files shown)


src/lib/y2storage/devicegraph_sanitizer.rb
```diff
@@ -162,6 +166,45 @@ def missing_pvs?(vg)
       vg.lvm_pvs.any? { |p| p.blk_device.nil? }
     end
 
+    # Errors related to LVM VGs in the given devicegraph
```
ancorgs: Errors related to Bcache (copy&paste error)

test/y2storage/devicegraph_sanitizer_test.rb
```diff
@@ -69,6 +69,29 @@
...
+          fake_scenario(scenario)
+          expect(subject.errors).to be_empty
+        end
+      end
+
+      context "on an architecture that does not support bcache (ppc)" do
+        let(:architecture) { :ppc }
+
+        it "raises an error" do
+          expect { fake_scenario(scenario) }.to raise_error(Y2Storage::Error)
```
ancorgs: I don't get why an exception is raised here.

I was expecting subject.errors to contain the error about unsupported bcache, but not an exception

ancorgs: Ok. I took a closer look and I found the exception is raised in the StorageManager if the user decides to abort (i.e. if the callback to inform about the error returns false).
https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/storage_manager.rb#L473

During normal execution that exception is caught. That's not the case of the `fake_scenario` helper (intentionally or not, likely nobody thought about it).

In any case, I don't think we should rely on the un-captured StorageManager exception in order to test `DevicegraphSanitizer#errors`. We should really check that `#error` contains what we want, not that our testing infrastructure explodes because we didn't provide a valid callback to report the error to the user (that's what we are testing by expecting an exception).

shundhammer: OK. I prefer that approach a lot, but I didn't find out **why** I kept getting that exception. The missing callback is what I hadn't found, so I'll have another look at that.

shundhammer: Now using the same technique as the other tests for that class.

PR #844, opened 2019-02-13
use realistic device name for DASD

aschnell:
## Problem

Unrealistic device name in unit tests make unit tests more difficult to understand.

## Solution

Use realistic device names.

### DIFF
--- a/test/y2storage/autoinst_proposal_test.rb
+++ b/test/y2storage/autoinst_proposal_test.rb
@@ -265,6 +265,10 @@ describe Y2Storage::AutoinstProposal do
           { "mount" => "/", "partition_nr" => 1, "create" => false }
         end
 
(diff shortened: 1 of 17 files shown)


test/y2storage/autoinst_proposal_test.rb
```diff
@@ -265,6 +265,10 @@
           { "mount" => "/", "partition_nr" => 1, "create" => false }
         end
 
+        let(:partitioning) do
+          [{ "device" => "/dev/dasda", "use" => "all", "partitions" => [root, home] }]
+        end
```
joseivanlopez: Why was this new `let` needed?

aschnell: Otherwise sda was handled by AutoYaST.

PR #845, opened 2019-02-13
do not recreate DASD partition table

aschnell:
## Problem

Recreating a DASD partition table on DASDs using virtio-blk could be dangerous. See https://bugzilla.suse.com/show_bug.cgi?id=1112037.

## Solution

Do not recreate DASD partition tables in the proposal.


aschnell:
Not desired, see https://bugzilla.suse.com/show_bug.cgi?id=1112037#c40.

PR #846, opened 2019-02-13
Allow to modify Bcache devices

joseivanlopez:
## Problem

Now, the Expert Partitioner allows to create Bcache devices without a caching set, but there is no way to edit the Bcache later to add a caching set.

- https://jira.suse.de/browse/SLE-4329
- https://trello.com/c/cQo16PQ5/672-editing-an-in-memory-bcache-device-devices-and-cache-mode

## Solution

Add option in the Expert Partitioner for editing a Bcache device. This should allow to change its caching set and its cache mode. Moreover, it should allow to change the caching set of a Bcache only when the Bcache is not using a caching set already (on the real system).

Requires these changes in the library: https://github.com/openSUSE/libstorage-ng/pull/613.

## Testing

- Added unit tests
- Tested manually

## Screenshots

virtualbox_opensuse tumbleweed_15_02_2019_16_45_35

virtualbox_opensuse tumbleweed_15_02_2019_16_45_46

virtualbox_opensuse tumbleweed_19_02_2019_16_36_48

virtualbox_opensuse tumbleweed_19_02_2019_16_37_31

virtualbox_opensuse tumbleweed_18_02_2019_15_26_18

### DIFF
--- a/src/lib/y2partitioner/actions/delete_bcache.rb
+++ b/src/lib/y2partitioner/actions/delete_bcache.rb
@@ -22,6 +22,7 @@
 require "yast"
 require "y2partitioner/actions/delete_device"
 require "y2partitioner/actions/controllers/bcache"
+require "y2partitioner/ui_state"
 
 module Y2Partitioner
   module Actions
@@ -36,6 +37,7 @@ module Y2Partitioner
         textdomain "storage"
 
         @bcache_controller = Controllers::Bcache.new(bcache)
+        UIState.instance.select_row(bcache)
       end
 
     private
--- a/src/lib/y2partitioner/actions/delete_bcache.rb
+++ b/src/lib/y2partitioner/actions/delete_bcache.rb
@@ -1,6 +1,6 @@
 # encoding: utf-8
 
-# Copyright (c) [2018] SUSE LLC
+# Copyright (c) [2018-2019] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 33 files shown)


src/lib/y2storage/bcache.rb
```diff
@@ -53,15 +53,17 @@ class Bcache < Partitionable
...
     #
     #   @raise [storage::Exception] if attaching failed
     #   @raise [storage::LogicException] for a Flash-only Bcache device or when
     #     the Bcache device already has a caching set.
     #
     #   @param set [BcacheCset] set to attach
-    storage_forward :attach_bcache_cset
+    storage_forward :add_bcache_cset
+
+    storage_forward :remove_bcache_cset
```
ancorgs: No documentation for this?

src/lib/y2storage/devicegraph.rb
```diff
@@ -381,6 +381,15 @@ def remove_bcache(bcache)
       remove_with_dependants(bcache_cset) if bcache_cset && bcache_cset.bcaches.empty?
     end
 
+    def remove_bcache_cset(bcache_cset)
```
ancorgs: Documentation?

src/lib/y2partitioner/actions/delete_bcache.rb
```diff
@@ -21,48 +21,66 @@
...
+        return nil if safe_detach_bcache?
 
-        # TRANSLATORS: Error when trying to modify an existing bcache that shares caches.
+        # TRANSLATORS: Error message when detach is not a safe action
         _(
-          "The bcache shares its cache set with other devices.\nThis can result in unreachable space " \
-            "if done without detaching.\nDetaching can take a very long time in some situations."
+          "The bcache cannot be deleted because it shares its cache set with other devices.\n" \
+          "Detaching is required to avoid possible unreachable space, but detaching action\n" \
+          "could take a very long time in some situations."
```
ancorgs: The sentence, specially the part  "but detaching action could take..." sounds a little bit strange to me.

I would merge the old and the new sentence in something like this.
```
The bcache cannot be deleted because it shares its cache set with other devices.\n
Deleting it without detaching the device first can result in unreachable space.\n
Unfortunately detaching can take a very long time in some situations.
```

src/lib/y2partitioner/actions/edit_bcache.rb
```diff
@@ -0,0 +1,112 @@
...
+      #
+      # @return [String, nil] nil if the caching device can be modified.
+      def non_editable_error
+        return nil if editable_bcache_cset?
+
+        # TRANSLATORS: error message, %{name} is replaced by a bcache name (e.g., /dev/bcache0)
+        format(
+          _("The Bcache %{name} is already created on disk and its caching\n" \
+            "cannot be modified. To modify the caching device, remove the Bcache\n" \
+            "and create it again."),
```
ancorgs: I had some concerns about this message. But I just checked and saw is almost identical to the message we have always shown for RAIDs, so I guess the same wording is fine for the bcache case. Anyways, leaving my original message below for reference (in fact, the capitalization is still a real issue IMO).

_It's not fully true that it "cannot be modified", the point is that we refuse to do it in the Partitioner. That aspect (possible but not supported by the tool) is more clear in other messages, but here we make it sound like a limitation of the technology._

_Moreover, asking the user to delete the bcache (by the way, we inconsistently capitalize it as "Bcache" and/or "bcache") makes it sound like a safe or even recommended operation._

PR #849, opened 2019-02-18
Partitioner: button to activate crypt devices

ancorgs:
## Problem

When executed on an already installed system (i.e. not in installation), the Partitioner of storage-ng does not contain any way to activate pre-existing but inactive LUKs device.

The old partitioner (pre-ng) didn't executed an activation by default either, but it offered two ways of activating such device so it could be mounted or used in any other way.

1) The user can click on "Edit" for an encrypted partition and then, in the first screen of the Edit wizard, select to mount the existing file-system. In that case, the next step of the Edit wizard consists on asking the password and activating the existing LUKS.

2) There is also a general option "Provide Crypt Passwords" that is available in  "Configure..." just alongside the similar "Configure Multipath" also used to activate inactive devices (Multipath devices instead of LUKS ones).

See more in the following links 

- [bsc#1113515](https://bugzilla.suse.com/show_bug.cgi?id=1113515)
- [Trello card](https://trello.com/c/HVqpFjMJ)

## Solution

Due to fundamental differences in how activation of devices works in storage-ng, solution 1 cannot be re-implemented.

This pull request re-implements solution 2. But due to the mentioned differences in approach, the "Provide Crypt Passwords" button added in this pull request:

- activates all kind of devices, not only the crypt ones
- refreshes everything (i.e. reprobes the whole devicegraph)

## Testing

- Tested manually
- Unit test added

## Screenshots

activate-crypt

None of the options of "Configure" and "Rescan Devices" buttons has never been explained in the help (not even in the old storage). Let's fix it.

activate-help

### DIFF
--- a/src/lib/y2partitioner/widgets/configure.rb
+++ b/src/lib/y2partitioner/widgets/configure.rb
@@ -71,6 +71,20 @@ module Y2Partitioner
         :redraw
       end
 
+      # @macro seeAbstractWidget
+      def help
+        # TRANSLATORS: help text for the Partitioner
+        _(
+          "<p>The <b>Configure</b> button offers several options to activate devices " \
+          "that were not detected by the initial system analysis. After activating the " \
+          "devices of the choosen technology, the system will be rescanned and all the " \
+          "information about storage devices will be refreshed. " \
+          "Thus, the <b>Provide Crypt Passwords</b> option is also useful when no " \
+          "encryption is involved, to activate devices of other technologies like LVM " \
+          "or Multipath.</p>"
+        )
+      end
+
     private
 
       # @return [Array<Yast::WidgetTerm>]
--- a/src/lib/y2partitioner/widgets/configure.rb
+++ b/src/lib/y2partitioner/widgets/configure.rb
@@ -61,7 +61,7 @@ module Y2Partitioner
       # @return [:redraw, nil] :redraw when some configuration client was
       #   executed; nil otherwise.
       def handle(event)
-        action = Action.find(event["ID"])
+        action = find_action(event["ID"])
 
         return nil unless action
         return nil unless warning_accepted?(action) && availability_ensured?(action)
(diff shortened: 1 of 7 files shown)


src/lib/y2partitioner/widgets/configure.rb
```diff
@@ -202,6 +204,10 @@ def initialize(id, label, icon, client, pkgs)
         # are indexed in this constant in order to reuse the existing
         # translations from yast2-storage
         WARNING_TEXTS = {
+          crypt: N_(
+            "Rescanning crypt devices cancels all current changes.\n" \
```
jreidinger: NP: "discards" sounds better for me then "cancels"

src/lib/y2partitioner/widgets/configure.rb
```diff
@@ -275,6 +281,16 @@ def warning_text
...
         end
+
+        # Value for the 'activate' argument of {Reprobe#reprobe}
+        #
+        # For most cases this returns nil, which implies simply honoring the
+        # default behavior.
+        #
+        # @return [Boolean, nil]
+        def activate
+          id == :crypt ? true : nil
```
jreidinger: to be honest i start thinking that here it is a bit abusing generic class where it more fits by common class hierarchy. I see there many "child" specific actions and this is one of it ( others are e.g. label defined elsewhere, warning text and others). why not create child class and have factory that create child based on id? If there is reason please document it in Action class description. Because as it is now, reuse single or subset of actions elsewhere is problematic.

src/lib/y2partitioner/widgets/configure.rb
ancorgs: Intentionally done for symmetry with all the other texts (that were probably inherited from the old storage).

src/lib/y2partitioner/widgets/configure.rb
ancorgs: I don't think there is a reason to avoid a hierarchy other than keeping things simple. These "actions" are not supposed to be reused elsewhere, they are just some very simplistic objects to represent the options of the "Configure" menu-button. They will never be used out of the scope of such button.

Anyway, I will take a look to the possibility of introducing a hierarchy and a factory class.

joseivanlopez:
Remember changelog, etc.

src/lib/y2partitioner/widgets/configure.rb
ancorgs: @jreidinger I added a commit which changes this to use a hierarchy of actions. I don't think the factory class helps in this case, so I just reorganized the code a bit to use the new classes directly.

Now is more correct from the OOP point of view, but to be honest I'm not sure if the code has become more readable or simply bigger. :-)

PR #851, opened 2019-02-18
handle device_order for MD RAIDs during installation

aschnell:
## Problem

The device order for MD RAIDs was not handled by AutoYaST during installation.

https://bugzilla.suse.com/show_bug.cgi?id=1083542

## Solution

Handle the device order during installation.

## Testing

- Extended unit test
- Tested manually

### DIFF
--- a/src/lib/y2storage/proposal/autoinst_md_planner.rb
+++ b/src/lib/y2storage/proposal/autoinst_md_planner.rb
@@ -119,6 +119,7 @@ module Y2Storage
         md.name = raid_options.raid_name if raid_options.raid_name
         md.chunk_size = chunk_size_from_string(raid_options.chunk_size) if raid_options.chunk_size
         md.md_parity = MdParity.find(raid_options.parity_algorithm) if raid_options.parity_algorithm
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/proposal/autoinst_md_planner.rb
```diff
@@ -119,6 +119,7 @@ def add_raid_options(md, raid_options)
         md.name = raid_options.raid_name if raid_options.raid_name
         md.chunk_size = chunk_size_from_string(raid_options.chunk_size) if raid_options.chunk_size
         md.md_parity = MdParity.find(raid_options.parity_algorithm) if raid_options.parity_algorithm
+        md.devices_order = raid_options.device_order if !raid_options.device_order.empty?
```
ancorgs: Just a note for the future.

Take into account you have `#any?` as some kind of inverse for `#empty?`. Just mentioning it because I find

```ruby
whatever if device_order.any?
```

way more understandable than

```ruby
whatever if !device_order.empty?
```

And if negating the condition if unavoidable, I would have used `unless`

```ruby
whatever unless device_order.empty?
```

PR #852, opened 2019-02-18
Help Texts for Guided Setup

shundhammer:
## Problem

There were no help texts yet in the _guided setup_.

- Trello: https://trello.com/c/iCjvewN0/

- Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1121801

## Testing

Manually tested with the proposal test client:

```
/sbin/yast2 ./proposal_testing.rb ../../test/data/devicegraphs/windows-pc.yml
```

(or any other suitable YAML or XML devicegraph)

## Screenshots

help-text-01

If there are Windows partitions that can be resized, this can contain two more lines that refer to the resize options.

help-text-02

help-text-03

This is a bit lame; in particular, it does not explain the pros and cons of the different filesystem types. It also does not explain why and when a user would want a separate /home partition and the pros and cons of Btrfs snapshots.

The problem with that is that those widges are created quite dynamically, depending on the settings on control.xml. It would be pretty awkward to explain things that are not available to the user if they are not configured for the product; and we'd have to do it for both _legacy_ and _ng_ settings.

But at least this is conistent with the help text for the initial proposal dialog:

help-text-proposal

## Unit Tests

`rake test:unit` runs without problems locally; it's only in Travis where some recent changes to libstorage-ng about Bcache have not arrived yet.

### DIFF
--- a/src/lib/y2storage/dialogs/guided_setup/select_root_disk.rb
+++ b/src/lib/y2storage/dialogs/guided_setup/select_root_disk.rb
@@ -161,6 +161,43 @@ module Y2Storage
           update_windows_settings if activate_windows_actions?
         end
 
+        def help_text
+          # TRANSLATORS: Help text for root disk selection
+          msg = _(
+            "<p>" \
+            "Select the disk where to create the root filesystem. " \
+            "</p><p>" \
+            "This is also the disk where any boot-related partitions " \
+            "will be created as necessary: /boot, ESP (EFI System " \
+            "Partition), BIOS-Grub. " \
+            "That means that this disk should be usable by the machine's " \
+            "BIOS / firmware." \
+            "</p><p>" \
+            "In this dialog you can also choose what to do with existing partitions:" \
+            "</p><p>" \
+            "<ul>" \
+            "<li>Do not modify (keep them as they are)</li>" \
+            "<li>Remove if needed</li>" \
+            "<li>Remove even if not needed (always remove)</li>" \
+            "</ul>"
+          )
+
+          # TRANSLATORS: Help text for root disk selection, continued
+          msg << _(
+            "<ul>" \
+            "<li>Resize if needed (Windows partitions only)</li>" \
+            "<li>Resize or remove if needed (Windows partitions only)</li>" \
+            "</ul>" \
+            "<p>" \
+            "That last option means to try to resize the Windows partition(s) to " \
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
```diff
@@ -107,6 +107,44 @@ def update_settings!
           settings.encryption_password = password
         end
 
+        # rubocop:disable Metrics/MethodLength
+        def help_text
+          # TRANSLATORS: Help text for the partitioning scheme (LVM / encryption)
+          _(
+            "<p>" \
+            "Select the parititioning scheme:" \
```
ancorgs: parititioning

shundhammer: Oops...

shundhammer: I'll add that fix to my next PR.

src/lib/y2storage/dialogs/guided_setup/select_root_disk.rb
```diff
@@ -161,6 +161,43 @@ def update_settings!
...
 
+        def help_text
+          # TRANSLATORS: Help text for root disk selection
+          msg = _(
+            "<p>" \
+            "Select the disk where to create the root filesystem. " \
+            "</p><p>" \
+            "This is also the disk where any boot-related partitions " \
+            "will be created as necessary: /boot, ESP (EFI System " \
+            "Partition), BIOS-Grub. " \
```
ancorgs: Not sure if we can say that any boot-related partition will be created there.

The point is that we consider that the BIOS / firmware will use that disk for booting, which doesn't always imply we will create the boot-related partitions there (at least not in all booting systems).

shundhammer: I fear using a 100% correct wording here would pretty much defeat the purpose of being a help text because it would quickly become too complicated.

I think I'll cheat my way around that by adding _typically_.

shundhammer: See also PR#853

ancorgs: Yes, much better with "typically". :-)

PR #855, opened 2019-02-20
Preserve old filesystem label (bsc##1087229)

jreidinger:
## Problem

Insonsistency with old partitioner which keeps old label if just refformating fs.

- https://bugzilla.suse.com/show_bug.cgi?id=1087229

## Solution

lets keep label even when reformat filesystem.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Feb 20 11:55:30 UTC 2019 - jreidinger@suse.com
+
+- filesystem label is kept when using existing partitioning
+  (bsc#108722)
+- 4.1.64
+
+-------------------------------------------------------------------
 Tue Feb 19 14:40:11 UTC 2019 - ancor@suse.com
 
 - Partitioner: new option "Provide Crypt Passwords" (bsc#1113515).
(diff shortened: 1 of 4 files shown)


src/lib/y2partitioner/actions/controllers/filesystem.rb
```diff
@@ -502,8 +502,7 @@ def current_value_for(attribute)
           when :mount_point
             filesystem.mount_path
           when :label
-            # Copying the label from the filesystem in the disk looks unexpected
-            new?(filesystem) ? filesystem.label : nil
+            filesystem.label
```
ancorgs: Please, reference the bug report here. We re-use the label from the previous file-system but not other attributes. That's surprising to me, so I would tend to consider it a bug and "fix" it, unless it's stated to be intentional.

test/y2partitioner/actions/controllers/filesystem_test.rb
```diff
@@ -595,13 +595,6 @@ def create_partition(size)
         expect(subject.partition_id).to eq(type.default_partition_id)
       end
 
-      context "when the previous filesystem exists in the disk" do
```
ancorgs: For the same reason than above, I would not simply delete this test, but I would state here than the label from the previous file-system is reused on purpose.

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Feb 20 11:55:30 UTC 2019 - jreidinger@suse.com
+
+- filesystem label is kept when using existing partitioning
+  (bsc#108722)
```
ancorgs: Sorry, I LGTMed this too fast. The bug number is wrong, missing 9 at the end

PR #858, opened 2019-02-20
Improve import mount points

joseivanlopez:
## Problem

When using the "Import Mount Points" feature of the Expert Partitioner, it crashes when the selected fstab file contains an entry with an unsupported filesystem type (e.g., ntfs-3g).

- https://bugzilla.suse.com/show_bug.cgi?id=1103391
- https://trello.com/c/QGZTOJ9T/692-1-osdistribution-p3-1103391-handle-unsupported-fstype-when-importing-etc-fstab

## Solution

Avoid to import a fstab entry when it contains an unsupported filesystem type. Now it shows an error message instead of crashing.

Moreover, when we were importing mount points, the "mount by" type of each entry was not imported, but the default "mount by" type was used. And the label of the previous filesystem was also removed. This was fixed too, so now the "mount by" indicated in the fstab is respected and also the label is preserved when the device was already formatted. 

## Testing

- Added unit tests
- Tested manually

## Screenshots

virtualbox_opensuse tumbleweed_22_02_2019_14_53_02

virtualbox_opensuse tumbleweed_22_02_2019_14_53_15


src/lib/y2partitioner/dialogs/import_mount_points.rb
```diff
@@ -96,12 +96,24 @@ def label
...
         end
 
-        def init
-          self.value = true
+        def opt
+          [:notify]
+        end
+
+        def handle
+          store
```
jreidinger: why it is needed? it will redraw table? please mention it in comment as it is not much obvious why it is called.

src/lib/y2partitioner/filesystems.rb
```diff
@@ -0,0 +1,44 @@
...
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage/filesystems/type"
+
+module Y2Partitioner
```
jreidinger: I really do not like that decision about supported FS lives in partitioner and not in storage. Is there reason for it?

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -440,15 +441,7 @@ def label
...
       def items
-        Y2Storage::Filesystems::Type.all.select { |fs| supported?(fs) }.map do |fs|
-          [fs.to_sym, fs.to_human_string]
-        end
-      end
-
-    private
-
-      def supported?(fs)
-        SUPPORTED_FILESYSTEMS.include?(fs.to_sym)
```
jreidinger: SUPPORTED_FILESYSTEM still exists in this file.

src/lib/y2partitioner/filesystems.rb
joseivanlopez: Partitioner is offering a limited subset of filesystems. That list of supported filesystems is something only related to the Partitioner.

src/lib/y2partitioner/widgets/format_and_mount.rb
joseivanlopez: I forgot to remove it. Thanks.

src/lib/y2partitioner/dialogs/import_mount_points.rb
joseivanlopez: Ok, I added the comment only in the commit description. I am going to copy it here too.

src/lib/y2partitioner/filesystems.rb
ancorgs: That's something we discussed offline. The list of possible filesystems used in the Partitioner dropdowns is there basically for historic reasons. The library supports quite some file-systems that are not present in the Partitioner.

So this is not really about the libstorage-ng or `Y2Storage` capabilities, but about what has been decided to offer in the partitioner (which is a subset of all the types actually supported in `Y2Storage`).

But that needs to be better explained.

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -247,6 +258,16 @@ def known_fs_type?(entry)
           !entry.fs_type.is?(:auto) && !entry.fs_type.is?(:unknown)
         end
 
+        # Whether a fstab entry has a supported filesystem type
+        #
+        # Only some filesystem types are supported by YaST, see {Filesystems.all}.
```
ancorgs: This comment, for exampled helps to the confusion that is being discussed in other comments. This is not a list of file-system supported by YaST, but a lit of file-systems offered by the Partitioner forms (which is a subset of the former).

joseivanlopez: Right! I wanted to say Partitoner instead of YaST.

src/lib/y2partitioner/dialogs/import_mount_points.rb
jreidinger: Thanks, comment in commit is good, but doing this archeology to find reason is annoying. My simple rule is comment is code why we are doing it and comment in commit why we are changing it. So when you read in code you immediately see rationale for code and when you are interested why we change something e.g. between SPs or major releases, you will see it in blame.

PR #860, opened 2019-02-27
Do not export the partition_type attribute when belongs to GPT

dgdavid:
## Problem

When cloning a system with GPT, AutoYaST still generating
 
> <partition_type>primary</partition_type>

However, this attribute does not make sense when dealing with GPT partition tables.

- https://bugzilla.suse.com/show_bug.cgi?id=1115751
- https://trello.com/c/9ZIgDgWF

## Solution

Do not export it when the partition belongs to a GPT partition table.

## Testing

- Added a new unit test
- Also tested manually

### DIFF
--- a/src/lib/y2storage/autoinst_profile/partition_section.rb
+++ b/src/lib/y2storage/autoinst_profile/partition_section.rb
@@ -124,7 +124,7 @@ module Y2Storage
       #   @return [Integer] the partition number of this partition
 
       # @!attribute partition_type
-      #   @return [String] undocumented attribute that can only contain "primary"
+      #   @return [String, nil] the partition type of this partition (only can be "primary")
 
       # @!attribute subvolumes
       #   @return [Array<SubvolSpecification>,nil] list of subvolumes or nil if not
@@ -297,7 +297,7 @@ module Y2Storage
       def init_partition_fields(partition)
         @create = !NO_CREATE_IDS.include?(partition.id)
         @partition_nr = partition.number
-        @partition_type = "primary" if partition.type.is?(:primary)
+        @partition_type = "primary" if primary_partition?(partition)
         @partition_id = partition_id_from(partition)
         @lvm_group = lvm_group_name(partition)
         @raid_name = partition.md.name if partition.md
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -124,6 +124,7 @@ def self.attributes
       #   @return [Integer] the partition number of this partition
 
       # @!attribute partition_type
+      #   TODO: change documentation
```
dgdavid: @imobachgs, 

Can I remove the "undocumented attribute" text here?

imobachgs: Yes, please. Actually it is not undocumented anymore.

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -431,6 +432,14 @@ def lvm_group_name(device)
       def fixed_size?(device)
         device.is?(:disk_device, :software_raid)
       end
+
+    private
+
+      def include_partition_type?(partition)
```
imobachgs: np: The name is confusing to me. I read that I am setting the `@partition_type` to `primary` if the section includes a partition type (which is not exactly what you mean). I would prefer something like `primary_partition?` or something like that.

imobachgs: Please, document the method even if it is private. That documentation will help your teammates if they need to check your code :) In this case it may seems obvious, but we have `Planned::Partition` and `Partition` classes (and `PartitionSection` too) so it is nice the have the arguments annotated.

imobachgs:
Just for the records, this minor fix does not deserve a version bump.

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -431,6 +431,18 @@ def lvm_group_name(device)
...
       end
+
+      # Determines whether given partition is primary or not
+      #
+      # Always false when the partition belongs to a GPT partition table.
+      #
+      # @param partition [Y2Storgae::Partition] the partition to check
+      # @return [Boolean] true when is a primary partition; false otherwise
+      def primary_partition?(partition)
+        return false if partition.partition_table.type.is?(:gpt)
```
ancorgs: It would be better to use something more generic like `PartitionTable#:extended_possible?`

dgdavid: Thank you so much for the explanation, @ancorgs 

I am going to do a new PR fixing this :)

PR #863, opened 2019-03-04
Fix missing translations in partitioner

dgdavid:
## Problem

The filesystem roles in partitioner were not being translated due a wrongly placed `textdomain`.

- https://bugzilla.suse.com/show_bug.cgi?id=1127756

## Solution

Move the `textdomain` to the `initialize` method.

## Testing

- *Tested manually*

## Screenshots

| Before | After |
|--------|-------|
| unfixed_filesytem_role_selection | fixed_filesystem_role_selection |

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Mar  4 16:31:25 UTC 2019 - David Díaz <dgonzalez@suse.com>
+
+- Translates the filesystem roles properly (bsc#1127756).
+- 4.1.69
+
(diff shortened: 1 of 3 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Mar  4 16:31:25 UTC 2019 - David Díaz <dgonzalez@suse.com>
+
+- Translates the filesystem roles properly (bsc#1124456).
```
ancorgs: Wrong bug number?

dgdavid: Oops! :no_mouth:

PR #864, opened 2019-03-06
Avoid to crash when frozen String and fix missing translation

dgdavid:
## Problems

There were a missing translation an a "can't modify frozen String" crash.

- https://bugzilla.suse.com/show_bug.cgi?id=1125006
- https://bugzilla.suse.com/show_bug.cgi?id=1127181

## Solution

Simply fix those errors.

### DIFF
--- a/src/lib/y2partitioner/dialogs/partition_table_type.rb
+++ b/src/lib/y2partitioner/dialogs/partition_table_type.rb
@@ -94,8 +94,8 @@ module Y2Partitioner
 
       # @macro seeDialog
       def title
-        # TRANSLATORS: %s is a device name like /dev/sda
-        format("Create New Partition Table on %s", @disk.name)
+        # TRANSLATORS: %{device_name} is a device name like /dev/sda
+        format(_("Create New Partition Table on %{device_name}"), device_name: @disk.name)
       end
 
       # @macro seeDialog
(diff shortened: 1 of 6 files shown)


src/lib/y2partitioner/dialogs/partition_table_type.rb
```diff
@@ -94,8 +94,8 @@ def initialize(disk, possible_types, default_type)
 
       # @macro seeDialog
       def title
-        # TRANSLATORS: %s is a device name like /dev/sda
-        format("Create New Partition Table on %s", @disk.name)
+        # TRANSLATORS: %{device_name} is a device name like /dev/sda
+        format(_("Create New Partition Table on %{device_name}"), device_name: @disk.name)
```
ancorgs: NP: When there is only one placeholder in the string, I tend to avoid the ruby-specific way (i.e. `%{variable_name}`), because I believe is more error-prone for translators that are used to the more widely used placeholders like `%s`.

lslezak: On the other hand I find "Create New Partition Table on %{device_name}" more readable than "Create New Partition Table on %s".

In the latter case I need to read the TRANSLATORS comment and think about it more. In the first case I just read it.

So it depends on your taste...

dgdavid: Thank you @ancorgs and @lslezak 

Actually, I was discussing it with @joseivanlopez before doing the PR :)

> is more error-prone for translators that are used to the more widely used placeholders like %s

but

>  I find "Create New Partition Table on %{device_name}" more readable

So, lets merge it :) :crossed_fingers:  :crossed_fingers:

PR #865, opened 2019-03-06
Prevent crash if resizing a newly formatted filesystem

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1124146

## Trello

https://trello.com/c/OUo0x3S3/

## Problem

In the partitioner, when the user changed an existing partition to be formatted and then immediately clicked on "resize", there would be a crash with a libstorage exception.

## Cause

The code tried to call `detect_size_info` on the filesystem, but that filesystem did not yet exist in the probed/system devicegraph.

## Fix

Now also checking if the planned filesystem is the same as the existing one or if it is new.

## Test

Manual test on my workstation with "yast2 disk" and a USB stick.
Changed the existing VFAT filesystem to ext4 and then hit _Resize_.

### DIFF
--- a/src/lib/y2partitioner/dialogs/blk_device_resize.rb
+++ b/src/lib/y2partitioner/dialogs/blk_device_resize.rb
@@ -93,8 +93,7 @@ module Y2Partitioner
       attr_reader :space_info
 
       def detect_space_info
-        return unless formatted? && controller.committed_device? && !swap?
-        return unless controller.committed_current_filesystem?
+        return unless controller.committed_current_filesystem? && !swap?
         @space_info = device.filesystem.detect_space_info
       end
 
--- a/src/lib/y2partitioner/dialogs/blk_device_resize.rb
+++ b/src/lib/y2partitioner/dialogs/blk_device_resize.rb
@@ -94,6 +94,7 @@ module Y2Partitioner
 
       def detect_space_info
         return unless formatted? && controller.committed_device? && !swap?
+        return unless controller.committed_current_filesystem?
         @space_info = device.filesystem.detect_space_info
       end
 
--- a/src/lib/y2partitioner/dialogs/blk_device_resize.rb
+++ b/src/lib/y2partitioner/dialogs/blk_device_resize.rb
@@ -94,7 +94,25 @@ module Y2Partitioner
 
       def detect_space_info
         return unless controller.committed_current_filesystem? && !swap?
-        @space_info = device.filesystem.detect_space_info
+        begin
+          @space_info = device.filesystem.detect_space_info
+        rescue Storage::Exception => e
+          detect_space_info_failed_warning(e)
+        end
+      end
+
+      # Show a warning popup about an error during detect_space_info.
+      #
+      # @param err [Storage::Exception] libstorage-ng exception for details
+      def detect_space_info_failed_warning(err)
+        log.warn "detect_space_info for #{device.name} failed: #{err.what}"
+
+        # TRANSLATORS: Warning message when the user wanted to resize a filesystem
+        # and there was a problem getting information about that filesystem.
+        msg = _("Obtaining information about free space on this filesystem failed.\n" \
+          "Resizing it might or might not work. If you continue, there is a risk\n" \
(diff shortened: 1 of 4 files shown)


src/lib/y2partitioner/dialogs/blk_device_resize.rb
```diff
@@ -94,6 +94,7 @@ def next_handler
 
       def detect_space_info
         return unless formatted? && controller.committed_device? && !swap?
+        return unless controller.committed_current_filesystem?
```
ancorgs: Why did you remove the `!swap?` part? As far as I understand the previous code, `Filesystem#detect_space_info` should not be called for swap filesystems. But that check is gone now.

src/lib/y2partitioner/dialogs/blk_device_resize.rb
```diff
@@ -94,6 +94,7 @@ def next_handler
...
         @space_info = device.filesystem.detect_space_info
```
ancorgs: Moreover, to avoid similar crashes in the future, maybe we should do something like

```ruby
begin
  @space_info = device.filesystem.detect_space_info
rescue Storage::Exception
  log.error "Storage exception calling detect_space_info"
  # Should we alert the user in addition to registering the error?
  @space_info = nil
end
```

Uncatched libstorage-ng exceptions look uglier than exceptions thrown in Ruby. Moreover, displaying the widget without the "currently used" info is acceptable, so an error detecting the available space is actually not severe enough to interrupt the execution.

Maybe we could even inform the user that something weird happened checking that filesystem, in addition to just logging the error. Then is up to the user whether to try the resizing or not.

src/lib/y2partitioner/dialogs/blk_device_resize.rb
shundhammer: I didn't. That's an extra line of code. Otherwise the entire expression would have been too long for Rubocop's liking.

src/lib/y2partitioner/dialogs/blk_device_resize.rb
lslezak: If we cannot determine the space info then the FS is probably broken or in a bad shape. In that case resizing it might be very dangerous, it might fail or the data can be lost. In this case I'd really appreciate a warning message with "resizing it may fail or cause data loss" info.

src/lib/y2partitioner/dialogs/blk_device_resize.rb
ancorgs: Right. I got confused because then the `formatted? && controller.committed_device?` part is absolutely redundant with the condition you added. Isn't it? This should be enough and more clear:

```ruby
def detect_space_info
  return unless controller.comitted_current_filesystem? && !swap?
  # the rest
end
```

shundhammer: I am not sure. I did the defensive approach. Can't there be other situations where only the old condition is true?

ancorgs: I'm pretty sure this is totally sure.

First, semantically:

- If `formatted?` is false, then there is no filesystem on the device. Thus `committed_current_filesystem?` MUST be false as well.
- If `controller.committed_device?` is false, then the device was not in the original devicegraph. Thus, its filesystem was not there either.

And now looking at the code:

- If `formatted?` is false, then `committed_current_filesystem?` is also false because that method starts with `return false unless device.formatted?`
- If `controller.committed_device?` is false, then `committed_current_filesystem?` is also false because it directly relies on `committed_filesystem` whose first line is `return nil unless committed_device?`.

shundhammer: OK then. Removing the extra conditions.

shundhammer:
Warning popup:

detect-space-info-warning

ancorgs:
Please merge https://github.com/shundhammer/yast-storage-ng/pull/1, everything else looks good.

shundhammer:
Had to rebase from master.

PR #867, opened 2019-03-08
Add Bcache support to AutoYaST

imobachgs:
## Problem

Adds support for Bcache to AutoYaST (fate#325346).

## Solution

Now it is possible to write a `<drive/>` section that specifies a *bcache device* setting the `type` element to `CT_BCACHE`.

In order to specify which devices are used as backing or caching devices for a given Bcache, two new elements are now supported in the `<partition/>` section: `bcache_caching_for` and `bcache_backing_for`. The example below contains just the relevant parts.

Additionally, Bcache options can be specified by using a `bcache_options` section.

```xml
    <drive>
      <type config:type="symbol">CT_DISK</type>
      <device>/dev/vdb</device>
      <disklabel>msdos</disklabel>
      <use>all</use>
      <partitions config:type="list">
        <partition>
          <!-- It can serve as caching device for several bcaches -->
          <bcache_caching_for config:type="list">
            <listentry>/dev/bcache0</listentry>
          </bcache_caching_for>
          <size>max</size>
        </partition>
      </partitions>
    </drive>

    <drive>
      <type config:type="symbol">CT_DISK</type>
      <device>/dev/vdc</device>
      <use>all</use>
      <!-- <disklabel>none</disklabel> -->
      <disklabel>msdos</disklabel>
      <partitions config:type="list">
        <partition>
          <!-- It can serve as backing device just for one bcache -->
          <bcache_backing_for>/dev/bcache0</bcache_backing_for>
        </partition>
      </partitions>
    </drive>

    <drive>
      <type config:type="symbol">CT_BCACHE</type>
      <device>/dev/bcache0</device>
      <bcache_options>
        <cache_mode>writethrough</cache_mode>
      </bcache_options>
      <use>all</use>
      <partitions config:type="list">
        <partition>
          <mount>/data</mount>
          <size>20GiB</size>
        </partition>
        <partition>
          <mount>swap</mount>
          <filesystem config:type="symbol">swap</filesystem>
          <size>1GiB</size>
        </partition>
      </partitions>
    </drive>
  </partitioning>
```

## Screenshots

bcache-non-partitioned

bcache-partitioned


src/lib/y2storage/autoinst_profile/bcache_options_section.rb
```diff
@@ -0,0 +1,62 @@
...
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Y2Storage
+  module AutoinstProfile
+    # Thin object oriented layer on top of a <raid_options> section of the
```
ancorgs: ~~raid_options~~ bcache_options

src/lib/y2storage/autoinst_profile/bcache_options_section.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      # @!attribute cache_mode
+      #   @return [String] Cache mode
+
+      # @param parent [#parent,#section_name] parent section
+      def initialize(parent = nil)
+        @parent = parent
+      end
+
+      # Clones RAID device options into an AutoYaST <raid_options> profile section
```
ancorgs: ~~raid_options~~ bcache_options

src/lib/y2storage/autoinst_profile/bcache_options_section.rb
```diff
@@ -0,0 +1,62 @@
...
+      # @param device [Bcache] Bcache device
+      # @return [BcacheOptionsSection] Bcache options section
+      def self.new_from_storage(device)
+        result = new
+        result.init_from_bcache(device)
+        result
+      end
+
+      # Method used by {.new_from_storage} to populate the attributes when
+      # cloning RAID options
```
ancorgs: ~~RAID~~ bcache

src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -309,6 +320,26 @@ def init_from_md(md)
...
+        @type = :CT_BCACHE
+        @device = bcache.name
+        collection =
+          if bcache.filesystem
+            [bcache]
+          else
+            bcache.partitions
+          end
+        @partitions = partitions_from_collection(collection)
+        @enable_snapshots = enabled_snapshots?(collection.map(&:filesystem).compact)
```
ancorgs: Not exactly related to this code review, but this makes me wonder whether it would make sense to use snapshots for a root filesystem on top of a bcache device. I mean, such setup will likely need a separate "/boot", since I don't believe Grub can boot from the bcache device. And with a separate "/boot", grub will likely not be able to run from a previous snapshot.

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -48,6 +48,7 @@ class PartitionSection < SectionWithAttributes
       NO_CREATE_IDS = []
       private_constant :NO_CREATE_IDS
 
+      # rubocop:disable Metrics/MethodLength
```
ancorgs: Maybe moving the definition to a constant can save us from disabling the MethodLength check for `self.attributes`.

ancorgs: Btw, I don't see the check being restored for the rest of the file.

src/lib/y2storage/planned/bcache.rb
```diff
@@ -0,0 +1,78 @@
...
+    # Specification for a Y2Storage::Bcache object to be created during the proposal
+    #
+    # @see Device
+    class Bcache < Device
+      include Planned::CanBeFormatted
+      include Planned::CanBeMounted
+      include Planned::CanBeEncrypted
+      include MatchVolumeSpec
+
+      # @return [name] device name of the MD RAID
```
ancorgs: First of all, ~~MD RAID~~~ bcache device.

On the other hand, is it realistic to expect the device actually gets this name? If not, I would describe this as the tentative name, at most.

src/lib/y2storage/planned/can_be_bcache_member.rb
```diff
@@ -0,0 +1,74 @@
...
+
+module Y2Storage
+  module Planned
+    # Mixin for planned devices that can act as Bcache caching/backing devices
+    # @see Planned::Device
+    module CanBeBcacheMember
+      # @return [Array<String>] name of the Bcache devices to which this device will serve as
+      #   caching
+      attr_accessor :bcache_caching_for
+      # @return [String] name of the Bcache device to which this device will serve as backing
```
ancorgs: or nil if this is not a backing device

src/lib/y2storage/planned/device.rb
```diff
@@ -116,7 +116,9 @@ def reuse?
       #
       # @return [Boolean]
       def component?
-        (respond_to?(:lvm_pv?) && lvm_pv?) || (respond_to?(:md_member?) && md_member?)
+        (respond_to?(:lvm_pv?) && lvm_pv?) ||
```
ancorgs: Please update documentation to also mention bcache

src/lib/y2storage/proposal/autoinst_bcache_planner.rb
```diff
@@ -0,0 +1,128 @@
...
+      #
+      # @param bcache [Planned::Bcache] Planned Bcache
+      def find_bcache_to_reuse(bcache)
+        dev_by_name = devicegraph.find_by_any_name(bcache.name, alternative_names: true)
+
+        return dev_by_name if dev_by_name && dev_by_name.is?(:bcache)
+        nil
+      end
+
+      # Given a user specified RAID type, it returns the RAID level
```
ancorgs: Again, RAID comment instead of bcache one

ancorgs:
I don't see any handling of the case in which two different block devices are used as backing device for the same bcache

ancorgs:
And same question for several devices used as caching of the same bcache. That's something that in theory could work with bcache, but the Partitioner does not allow it... which is a safe bet in my opinion.

src/lib/y2storage/autoinst_issues/multiple_bcache_members.rb
```diff
@@ -0,0 +1,67 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage/autoinst_issues/issue"
+
+module Y2Storage
+  module AutoinstIssues
+    # The proposal several backing or caching devices for the same bcache
```
ancorgs: I guess there is a missing word in this sentence. Something like "includes" or "contains"

src/lib/y2storage/autoinst_issues/multiple_bcache_members.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      # Return the error message to be displayed
+      #
+      # @return [String] Error message
+      # @see Issue#message
+      def message
+        # TRANSLATORS: 'bcache_name is the bcache device name (e.g., '/dev/bcache0');
+        # 'role' is the kind of device (e.g., 'caching' or 'backing').
+        _(
+          "%{bcache_name}: only one %{role} device can be specified per bcache."
```
ancorgs: Not sure how nice is to have the untranslated word "backing" or "caching" in all messages.

E.g.: "/dev/bcache0: sólo puede especificarse un dispositivo backing para cada bcache"

Well, the profile of course use English names, so I guess that's acceptable.

src/lib/y2storage/autoinst_profile/partition_section.rb
```diff
@@ -48,40 +48,41 @@ class PartitionSection < SectionWithAttributes
...
+        { name: :create_subvolumes },
+        { name: :resize },
+        { name: :pool },
+        { name: :used_pool },
+        { name: :stripes },
+        { name: :stripe_size, xml_name: :stripesize },
+        { name: :bcache_backing_for },
+        { name: :bcache_caching_for }
+      ].freeze
+
```
ancorgs: I usually add a line like this right after the definition of the constant... to avoid temptations of programmers ignoring the `self.attributes` method and accessing the constant directly (which breaks the uniform access principle).

```ruby
private_constant :ATTRIBUTES
```

src/lib/y2storage/planned/bcache.rb
```diff
@@ -0,0 +1,78 @@
...
+    # Specification for a Y2Storage::Bcache object to be created during the proposal
+    #
+    # @see Device
+    class Bcache < Device
+      include Planned::CanBeFormatted
+      include Planned::CanBeMounted
+      include Planned::CanBeEncrypted
+      include MatchVolumeSpec
+
+      # @return [String] tentative device name of the bcache
```
ancorgs: BTW, is clearly stated in the corresponding AY documentation that the name used in the profile for bcache devices is just to identify them in the profile itself and that users shoudn't expect those names to be final?

imobachgs: No, I will add a note.

src/lib/y2storage/autoinst_issues/multiple_bcache_members.rb
imobachgs: You are right. I've decided to make "backing device" and "caching device" as the strings to translate.

src/lib/y2storage/planned/bcache.rb
imobachgs: Now it is :)

PR #869, opened 2019-03-12
Add missing icons

joseivanlopez:
## Problem

Some icons used by the Expert Partitioner are not provided (neither by [yast-theme](https://github.com/yast/yast-theme/tree/master/icons/hicolor/scalable/apps) nor by [libyui-qt](https://github.com/libyui/libyui-qt/tree/master/src/icons)): *fcoe*, *xpram*, *dasd* and *zfcp* icons are missing. As consequence, the *Configure* menu button does not show the icon for some of its options, see the following screenshot:

* Related to https://bugzilla.suse.com/show_bug.cgi?id=1122174
* https://trello.com/c/IcPWJBUW/817-3-missing-icons-in-partitioner-configure
* [old_icons.tar.gz](https://github.com/yast/yast-storage-ng/files/2960648/old_icons.tar.gz)
* https://github.com/libyui/libyui-qt/pull/101

## Solution

@hellcp has provided [svg version](https://github.com/libyui/libyui-qt/pull/101) for the missing icons.

After some discussion in IRC, it seems that [these icons in the old partitioner](https://github.com/yast/yast-storage-ng/files/2960648/old_icons.tar.gz) are quite pointless because they are quite similar between them. But we have now new versions, so let's use them. 

## Testing

* Tested manually

## Screenshots

VirtualBox_openSUSE Tumbleweed_13_03_2019_16_27_15

### DIFF
--- a/src/lib/y2partitioner/widgets/configure.rb
+++ b/src/lib/y2partitioner/widgets/configure.rb
@@ -22,6 +22,7 @@
 require "yast"
 require "cwm"
 require "y2partitioner/widgets/reprobe"
+require "y2partitioner/icons"
 
 Yast.import "Stage"
 Yast.import "Popup"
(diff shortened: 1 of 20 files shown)


ancorgs:
After some conversation in IRC I would say the general feeling was: if we can get good SVG replacements for these icons, then use those new ones. If we don't get them, just remove any icon in that menu-button, they are not really needed.

On the other hand, I would not change the current strings. We are too close to the Leap 15.1 release and the gain doesn't justify the risk of having missing translations.

src/lib/y2partitioner/widgets/configure.rb
```diff
@@ -89,9 +89,17 @@ def help
...
+      end
+
+      # Generates an item for the given action
+      #
+      # @param action [Action]
+      # @return [Yast::WidgetTerm]
+      def item(action)
+        return Item(Id(action.id), action.label) unless action.icon?
+
+        Item(Id(action.id), Yast::Term.new(:icon, action.icon), action.label)
```
dgdavid: Just a doubt, why to keep this?

In the PR description you wrote

> For now, **all icons have been dropped** from the Configure menu button.

joseivanlopez: Well, maybe in the future we find good icons for those actions, so I have extended the `Action` class to allow both: with and without icon. This will make easier to add icons then.

joseivanlopez:
The margin is always there, it does not matter if there is icon or not. You can see it in other menu buttons of the expert partitioner.

PR #872, opened 2019-03-14
Sort Partition IDs Logically

shundhammer:
## Trello

https://trello.com/c/uJDzzpFV/

## Bugzilla

_No corresponding bug_

## Problem

Partition IDs (types) were not sorted in any meaningful way in the partitioner.

## Fix

Introduced a sort order on the Ruby level of that type (enum which is imported and mapped from libstorage-ng).

We now have a table in the Ruby code for the desired sort order and a spaceship operator (`<=>`) that uses that table so the normal Ruby `sort` function can be used:

https://github.com/yast/yast-storage-ng/pull/872/files#diff-ecc44dd40509958fa302edf2358c7f8fR60

Of course the partitioner still only shows those types that are actually applicable, not just every value of that enum.

Values that are not in that sort order table are sorted at the end.

part-sort-order

## Notice

This is for SLE-15 SP2. We need to wait until we have a branch for SLE-15 SP1, so for now there is no change log entry and no version bump to avoid complaints of the OBS change log check scripts for entries being out of chronological order.

### DIFF
--- a/test/y2storage/partition_id_test.rb
+++ b/test/y2storage/partition_id_test.rb
@@ -128,7 +128,7 @@ describe Y2Storage::PartitionId do
     it "sorts Linux Native before Linux Swap" do
       linux = Y2Storage::PartitionId::LINUX
       swap = Y2Storage::PartitionId::SWAP
-      expect(linux.sort_order < swap.sort_order).to eq true
+      expect(linux.sort_order).to be < swap.sort_order
     end
 
     it "sorts Linux partition IDs naturally" do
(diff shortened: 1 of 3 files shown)


test/y2storage/partition_id_test.rb
```diff
@@ -123,4 +123,68 @@
       end
     end
   end
+
+  describe "#sort_order" do
+    it "sorts Linux Native before Linux Swap" do
+      linux = Y2Storage::PartitionId::LINUX
+      swap = Y2Storage::PartitionId::SWAP
+      expect(linux.sort_order < swap.sort_order).to eq true
```
dgdavid: NP: You also can use 

```ruby
expect(linux.sort_order).to be < swap.sort_order
```

which seems more readable to me.

shundhammer: OK, good idea.

PR #873, opened 2019-03-14
Display pathnames correctly in a Right-to-Left context (bsc#1128091)

mvidner:
## Problem

In Arabic, the device table displays `/dev/sda1` as `dev/sda1/`

- https://bugzilla.suse.com/show_bug.cgi?id=1128091

## Solution

Use [Unicode BiDi formatting characters](https://www.unicode.org/reports/tr9/#Directional_Formatting_Codes) in the source (creating a machine "translation" of the string).

Disabled for TUI: libyui-ncurses does not handle the BiDi characters well, truncating the strings. TUI does not have the general RTL direction anyway.

If you start off with a JEOS and only install the bare minimum of packages needed to test this, you may end up seeing dotted box characters in place of the BiDi controls. Installing *noto-sans-fonts.rpm* helps.

## Testing

- [x] Added a new unit test
- [x] Tested manually
- [x] Fix and test handling `/`

## Screenshots

<table>
<tr><td>before</td><td>after</td></tr>
<tr><td>

</td><td>

</td>
</tr></table>

### DIFF
--- a/src/lib/bidi_markup.rb
+++ b/src/lib/bidi_markup.rb
@@ -1,7 +1,7 @@
 # Bidirectional Text: Left-to-right (Latin) and right-to-left (Arabic).
 #
 # See https://en.wikipedia.org/wiki/Bidirectional_text
-module Bidi
+module BidiMarkup
   LRE = "\u{202A}".freeze
   RLE = "\u{202B}".freeze
   PDF = "\u{202C}".freeze
--- a/src/lib/bidi_markup.rb
+++ b/src/lib/bidi_markup.rb
@@ -1,6 +1,46 @@
+# Copyright (c) [2020-2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
 # Bidirectional Text: Left-to-right (Latin) and right-to-left (Arabic).
 #
 # See https://en.wikipedia.org/wiki/Bidirectional_text
+#
+# If you feel lost, try wrapping a piece of text that belongs together with
+# {.first_strong_isolate First Strong Isolate}.
+#
+# Quoting from the standard, Unicode TR 9, Section 6.3 Formatting,
+# where UPPER CASE simulates a RTL script:
+#
+# > The most common problematic case is that of neutrals on the
+# > boundary of an embedded language. This can be addressed by
+# > setting the level of the embedded text correctly. For example,
+# > with all the text at level 0 the following occurs:
+# >
+# >     Memory:  he said "I NEED WATER!", and expired.
+# >     Display: he said "RETAW DEEN I!", and expired.
+# >
+# > If the exclamation mark is to be part of the Arabic quotation,
+# > then the user can select the text I NEED WATER! and explicitly
+# > mark it as embedded Arabic, which produces the following result:
+# >
+# >     Memory:  he said "(RLI)I NEED WATER!(PDI)", and expired.
+# >     Display: he said "!RETAW DEEN I", and expired.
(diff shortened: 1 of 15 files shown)


src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -148,8 +148,25 @@ def end_title
...
+
+      def left_to_right(s)
+        # ncurses would get confused and truncates the text instead
+        return s unless bidi_supported?
+
+        # If we used a plain device.name
+        # then it would display as dev/sda1/ in a RTL context.
+        # Unicode BiDi Left-to-right embedding (LRE),
+        # then Pop Directional Formatting (PDF)
+        format("\u{202A}%s\u{202C}", s)
```
mvidner: This is the actual fix. The rest of the PR 1) works around a libyui-ncurses bug 2) adjusts the tests.

mvidner:
The PR rules demand screenshots... and that's a good thing because comparing the screenshots made me realize I have a bug in handling the root `/` path.

src/lib/y2partitioner/widgets/blk_devices_table.rb
shundhammer: In the "after" screenshot I see `boot/efi/`. Is that the expected outcome? Whenever a user tries to use that path for anything, there will probably be an error since the system expects `/boot/efi`.

I understand this must be difficult for users of such an RtoL language (Arabic or Hebrew); if your natural reading direction is right-to-left, how is a user supposed to read such a technical path specification that is really still entirely Western-style? They might still start reading that part from the right, only then to realize that this is really left-to-right; but `boot/efi/` seems wrong in both worlds.

shundhammer (approved):
LGTM, and AFAICS it fixes this particular bug.

But I fear we will need this in more places, so in the long run, we will probably need those Bidi routines in a more central place so they are accessible from other packages as well.

Also, we will probably need this in C++ as well; libstorage-ng comes to mind. We have places where libstorage-ng generates user-visible messages that include both translated text and paths: The action graph and the subvolume actions. They will have the same problem; unless the Arabic and the Hebrew translator inserted those bidi markers manually in each message.

mvidner:
>  we will probably need those Bidi routines in a more central place 

Yup, that's likely, so I named the library so that it is easily movable without needing to change the usage places. I've even checked for collisions with [other bidi gems](https://rubygems.org/search?query=bidi) (which are solving different problems, so merging is unlikely).

mvidner:
Actually, *most* of the other places do not need special handling like this, because they can be solved with the *usual special handling*, which is using BiDi control characters in the translated messages. The storage table is special in having cells containing just filesystem paths, which noone has thought would need translation.

src/lib/bidi_markup.rb
```diff
@@ -0,0 +1,138 @@
+# Bidirectional Text: Left-to-right (Latin) and right-to-left (Arabic).
```
joseivanlopez, line 1: Missing license.

test/y2partitioner/widgets/columns/mount_point_test.rb
```diff
@@ -105,8 +105,35 @@
...
+        expect(subject.value_for(device)).to eq("\u2066" \
+                                                "/\u2068\u0641\u064A\u062F\u064A\u0648\u2069" \
+                                                "/\u2068\u0642\u062F\u064A\u0645\u0629\u2069" \
+                                                "\u2069")
+      end
+
+      context "when the path is root /" do
+        let(:device_name) { "/dev/sda" }
+
+        it "returns the mount path with wrapped with LTR Isolate" do
```
joseivanlopez, line 134: Leftover *with*.

src/lib/bidi_markup.rb
```diff
@@ -0,0 +1,138 @@
...
+# >     Memory:  he said "I NEED WATER!", and expired.
+# >     Display: he said "RETAW DEEN I!", and expired.
+# >
+# > If the exclamation mark is to be part of the Arabic quotation,
+# > then the user can select the text I NEED WATER! and explicitly
+# > mark it as embedded Arabic, which produces the following result:
+# >
+# >     Memory:  he said "(RLI)I NEED WATER!(PDI)", and expired.
+# >     Display: he said "!RETAW DEEN I", and expired.
+module BidiMarkup
```
lslezak, line 25: Is this `yast2-storage-ng` specific problem? I think this should be moved to `yast2`, just in case we need the same solution elsewhere....

src/lib/bidi_markup.rb
mvidner: Ah, right. Added.

src/lib/bidi_markup.rb
mvidner: In the end it looks like it might indeed be specific to storage, but the file is designed to be easily movable once we find out we need it.

PR #874, opened 2019-03-18
Fix bcache tests

imobachgs:


test/y2storage/autoinst_profile/drive_section_test.rb
```diff
@@ -409,7 +409,10 @@ def lvm_vg(name)
     end
 
     context "given a bcache" do
-      before { fake_scenario("btrfs_bcache.xml") }
+      before do
+        allow(Yast::Arch).to receive(:x86_64).and_return("x86_64")
```
joseivanlopez: [spec_helper.rb](https://github.com/yast/yast-storage-ng/blob/master/test/spec_helper.rb#L82) is already mocking this. You only have to `let(:architecture) { :x86_64 }`

joseivanlopez: The same comment for the rest of test.

imobachgs: Thanks. I did not know about that `:architecture` :) On the other hand, it looks pretty tricky to me. I would go for tags or something like that. But, for the team being, I guess it is good enough.

PR #877, opened 2019-03-21
document boot partition strategy

wfeldt:
## Problem

The boot loader should be installed on the same disk that gets the operating system (root partition).

- Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1094927
- Scrum board: https://trello.com/c/zgu7Iq9V

## Solution

Documentation adjusted and implemented according to doc.

## Testing

- updated and added unit tests
- manual integration tests

### DIFF
--- a/doc/boot-partition-location.md
+++ b/doc/boot-partition-location.md
@@ -1,0 +1,62 @@
+## Boot loader partitions in the Guided Proposal
+
+This document describes the locations of the separate `/boot` partition (if needed) and
+the boot loader partition (ESP, grub2 or zipl) the Guided Proposal assumes.
+
+This does not imply that it is technically required to follow the Guided
+Proposal. Users can in fact use the Expert Partitioner to do as they please
+(but may get warnings).
+
+### The `/boot` partition
+
+The proposal will always create the `/boot` partition (if one is needed) on
+the disk containing the root ('/') file system.
+
+It will never use an existing one.
+
+### Boot loader partition (ESP, grub2, zipl)
+
+The boot loader is usually installed into its own partition. Except on
+x86-legacy with msdos partition table that uses the mbr gap.
+
+This means
+
+- BIOS GRUB partition (x86-legacy, GPT partition table)
+- EFI System Partition (EFI systems: x86, aarch64)
+- PReP (ppc)
+- ZIPL partition (s390)
+
+These partitions are expected to be on the boot disk (disk containing the
+`/boot` file system). Note that this is identical to the disk containing the
+root file system (see previous chapter) as far as the Guided Proposal is
+concerned.
+
+An existing suitable partition will be reused. If none is found a new one
+is proposed on the boot disk.
+
+### Rationale
+
+The goal here is to keep the core operating system on a single disk if at
+all possible. Having the boot loader elsewhere would be technically possible
+but is impractical.
+
+#### Pro
+
+Experience shows that setups with operating system and boot loader on
+different disks tend to run into trouble as identifying the correct disk is
+a weak spot in boot loaders.
+
+Also, there might be unwanted interaction with other operating systems
+installed on the same machine.
+
+And, finally, the user might take 'the operating system disk' to another
+machine and have good chances it will simply 'just work'.
+
+ESPs could be shared but as they typically are of limited size having you
+'own' ESP looks better.
+
+#### Con
+
+There is the rumor that Windows expects a single ESP for the whole system.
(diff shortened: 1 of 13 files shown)


test/y2storage/boot_requirements_checker_ppc_test.rb
```diff
@@ -64,8 +65,22 @@
       context "if there is already a suitable PReP partition in the disk" do
         let(:prep_partitions) { [prep_partition] }
 
-        it "does not require any partition (PReP will be reused and Grub2 can handle this setup)" do
-          expect(checker.needed_partitions).to be_empty
+        context "and it is on the boot disk" do
+          let(:boot_disk) { dev_sda }
+
+          it "does not require any partition (PReP will be reused and Grub2 can handle this setup)" do
+            expect(checker.needed_partitions).to be_empty
```
joseivanlopez: One thing came to my mind. What happens if the existing PReP partition is formatted? Checking the code I see we would try to reuse it, what I think it is wrong. If that is true, maybe we should add some kind of control to avoid to reuse formatted partitions (and the same applies to zipl and bios_boot).

test/y2storage/initial_guided_proposal_test.rb
```diff
@@ -162,22 +164,51 @@
...
+        it "allocates the root device in the biggest device" do
           proposal.propose
 
-          expect(used_devices).to contain_exactly("/dev/sda", "/dev/sdb", "/dev/sdc")
+          expect(disk_for("/").name).to eq "/dev/sdb"
         end
 
-        it "allocates the root device in the biggest device" do
-          proposal.propose
+        # FIXME: this is currently not working, see bsc#1130392
```
joseivanlopez: Ups, this is a quite important bug. After disabling a volume feature (snapshots, adjust_by_ram, or the volume itself) the disabled feature is never restored. Good catch!

Definitely, [`Yast.deep_copy`](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/initial_guided_proposal.rb#L41) does not work as I expected.

jreidinger: @joseivanlopez can you get more details how and why deep code does not work as expected?

doc/boot-partition-location.md
```diff
@@ -0,0 +1,62 @@
+## Boot loader partitions in the Guided Proposal
+
+This document describes the locations of the separate `/boot` partition (if needed) and
+the boot loader partition (ESP, grub2 or zipl) the Guided Proposal assumes.
```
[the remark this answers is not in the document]
wfeldt: What I meant was 'for' grub2 etc; and PReP is for grub2.

test/y2storage/boot_requirements_checker_ppc_test.rb
wfeldt: It's not formatted. Reuse means 'install grub there'. It's the same as with bios grub. If the partition already exists you have to assume it is (has been) used. There's no way to distinguish use/non-use.

test/y2storage/initial_guided_proposal_test.rb
joseivanlopez: Well, somehow I supposed a different behavior (I don't know why because Yast.deep_copy is well documented). What we need here it is something like the following (a recursive version):

```
class Foo 
    attr_accessor :bar

    def initialize
        @foobar = [2, 2]
    end 

    def deep_copy
        f = dup 

        instance_variables.each do |v| 
            value = instance_variable_get(v)
            f.instance_variable_set(v, value.dup)
        end 

        f   
    end 

private

    attr_accessor :foobar
end

f = Foo.new
f.bar = [1, 1]

f2 = f.deep_copy
f2.bar << 1
f2.instance_variable_get(:@foobar) << 2

f.send(:bar) #=> [1, 1]
f2.send(:bar) #=> [1, 1, 1]

f.send(:foobar) #=> [2, 2]
f2.send(:foobar) #=> [2, 2, 2]
```

Do you think something like this Foo#deep_copy is a good idea?

jreidinger: Well, problem is that deep_copy is intended only for old component system. If you have own class I suggest to use common ruby approach ( whatever it is ). Maybe just overwrite dup?

joseivanlopez: Yes, I guess we can add `ProposalSettings#dup` with a logic similar as my example.

PR #879, opened 2019-03-22
Fix boot disk detection

joseivanlopez:
## Problem

The boot disk was not being properly detected when a separate /boot was created.

The device containing the root filesystem was always considered as boot disk, even when /boot was explicitly created as a separate partition. This confused to the Expert Partitioner, and the user was wrongly informed about some possible problems with the current devices configuration.

For example, in a non-UEFI x86 system, if the user creates a LVM PV over a complete disk (e.g., /dev/sda) and then creates /boot in a partition from another disk (e.g., /dev/sdb1), the disk used by LVM was considered the boot disk (/dev/sda instead of /dev/sdb) and the Expert Partitioner complains because the boot disk has no partition table.  

Something similar can happens in other scenarios, for example, when using Bcache. In fact, this bug was detected when Bcache was being tested by QA.

* https://bugzilla.suse.com/show_bug.cgi?id=1129787
* PBI: https://trello.com/c/WYoMDTMZ/852-bsc1129787-bootrequirementschecker-fails-to-determine-the-boot-disk

## Solution

Fix boot disk detection. Now, if a separate /boot exists, such device is considered the boot disk.

## Testing

- Added unit tests
- Tested manually (wrong warning is gone)


jreidinger:
I do not think so for now, as bcache does not have grub2 module, so probably grub2 will have problem to assembly it and load initrd from it. But maybe in future it will be added.

joseivanlopez:
@wfeldt @jreidinger but I guess it works with a separate boot, right?

wfeldt:
lgtm

wfeldt:
> but I guess it works with a separate boot, right?

Sure; I was just wondering.

jreidinger:
@joseivanlopez yes, as on separate boot is initrd which is capable of assembly bcache.

PR #880, opened 2019-03-26
Fix bcache tests

joseivanlopez:
## Problem

The `DevicegraphSanitizer` complains when Bcache is used with an architecture different to *x86_64*. In unit tests, architecture has to be mocked to avoid test failures when the tests are executed in another architecture. This has demonstrated to be quite error prone. With new tests involving Bcache, OBS will fail when generating the package for other architectures if you forget mocking the architecture.

Errors for Bcache should be mocked in general, and only unmocked when needed.

* Follow-up of https://trello.com/c/WYoMDTMZ/852-bsc1129787-bootrequirementschecker-fails-to-determine-the-boot-disk

* https://bugzilla.suse.com/show_bug.cgi?id=1129787

## Solution

Errors for Bcache are generally mocked. Architecture mocking is also removed from unit tests that don't require it anymore.

### DIFF
--- a/test/spec_helper.rb
+++ b/test/spec_helper.rb
@@ -87,6 +87,16 @@ RSpec.configure do |c|
       allow(Yast::Arch).to receive(:s390).and_return(architecture == :s390)
       allow(Yast::Arch).to receive(:aarch64).and_return(architecture == :aarch64)
     end
+
+    # Bcache is only supported for x86_64 architecture. The sanitizer complains when
+    # Bcache is used with another architecture. Bcache errors are mocked here to avoid
+    # to have to indicate x86_84 architecture in every test using Bcache (which has
+    # demonstrated to be quite error prone).
+    #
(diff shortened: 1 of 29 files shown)


ancorgs:
Maybe it also makes sense to remove all the `Arch` mocks that were introduced only to avoid the sanitizer error.

test/spec_helper.rb
```diff
@@ -87,6 +87,16 @@ def require(path)
       allow(Yast::Arch).to receive(:s390).and_return(architecture == :s390)
       allow(Yast::Arch).to receive(:aarch64).and_return(architecture == :aarch64)
     end
+
+    # Bcache is only supported for x86_64 architecture. The sanitizer complains when
+    # Bcache is used with another architecture. Bcache errors are mocked here to avoid
+    # to have to indicate x86_84 architecture is every test using Bcache (which has
```
jreidinger: typo `is`. Should be `in`

PR #884, opened 2019-03-28
Add support for NFS in AutoYaST

joseivanlopez:
## Problem

In SLE-12, AutoYaST supported installation over NFS. The way to indicate NFS drives in the AutoYaST profile is not documented at all, but it can be achieved by giving a *\<device\>* inside the *\<partition\>* section. This invalidates the profile because *\<partition\>* cannot contain a *\<device\>*, but it works in SLE-12.

Storage-ng has no support for installing over NFS with AutoYaST, so the *\<device\>* hack should be brought back again.

* https://bugzilla.suse.com/show_bug.cgi?id=1130256
* https://trello.com/c/x4LYDnwe/865-3-l3-important-bug-1130256-autoyast-for-a-nfs-root-fails-at-partitioning

## Solution

Add support for installing over NFS as before, that is, by specifying a device inside the partition section. Such device contains the NFS share in which to install.

NOTE: exporting NFS is not implemented in this PR. The idea is to create a follow-up PBI to implement both, import and export, but using a new style for defining NFS drives (current style invalidates the profile).

## Testing

- Added unit tests
- Tested manually

### DIFF
--- a/src/lib/y2storage/planned/devices_collection.rb
+++ b/src/lib/y2storage/planned/devices_collection.rb
@@ -163,14 +163,8 @@ module Y2Storage
       #
       # @return [Array<Planned::Device>]
       def all
-        @all ||= partitions +
-          disks +
-          stray_blk_devices +
-          vgs +
-          lvs +
-          mds +
-          bcaches +
-          nfs_filesystems
+        @all ||= [].concat(partitions, disks, stray_blk_devices, vgs, lvs, mds,
+          bcaches, nfs_filesystems)
       end
 
       # Returns the list of devices that can be mounted
--- a/src/lib/y2storage/planned/devices_collection.rb
+++ b/src/lib/y2storage/planned/devices_collection.rb
@@ -1,6 +1,6 @@
 # encoding: utf-8
 
-# Copyright (c) [2018] SUSE LLC
+# Copyright (c) [2018-2019] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -138,6 +138,13 @@ module Y2Storage
         @bcaches ||= devices.select { |d| d.is_a?(Planned::Bcache) }
       end
 
+      # Returns the list of planned NFS filesystems
+      #
+      # @return [Array<Planned::Nfs>]
+      def nfs_filesystems
+        @nfs_filesystems ||= devices.select { |d| d.is_a?(Planned::Nfs) }
+      end
+
       # Returns the list of planned LVM logical volumes
       #
       # @return [Array<Planned::LvmLv>]
@@ -156,7 +163,14 @@ module Y2Storage
       #
       # @return [Array<Planned::Device>]
       def all
-        @all ||= partitions + disks + stray_blk_devices + vgs + lvs + mds + bcaches
+        @all ||= partitions +
+          disks +
+          stray_blk_devices +
+          vgs +
+          lvs +
+          mds +
+          bcaches +
+          nfs_filesystems
       end
 
       # Returns the list of devices that can be mounted
--- a/src/lib/y2storage/proposal/autoinst_devices_creator.rb
+++ b/src/lib/y2storage/proposal/autoinst_devices_creator.rb
@@ -351,9 +351,9 @@ module Y2Storage
       # @return [Proposal::CreatorResult] Result containing the specified volume groups
       def set_up_lvm(vgs, devs_to_reuse)
         # log separately to be more readable
-        log.info "BEGIN: set_up_lvm: vgs=#{vgs.inspect}"
-        log.info "BEGIN: set_up_lvm: previous_result=#{creator_result.inspect}"
-        log.info "BEGIN: set_up_lvm: devs_to_reuse=#{devs_to_reuse.inspect}"
+        log.info "set_up_lvm: vgs=#{vgs.inspect}"
+        log.info "set_up_lvm: previous_result=#{creator_result.inspect}"
+        log.info "set_up_lvm: devs_to_reuse=#{devs_to_reuse.inspect}"
 
         vgs.reduce(creator_result) do |result, vg|
           pvs = creator_result.created_names { |d| d.pv_for?(vg.volume_group_name) }
@@ -417,7 +417,8 @@ module Y2Storage
       def create_logical_volumes(devicegraph, vg, pvs)
         lvm_creator = Proposal::LvmCreator.new(devicegraph)
         lvm_creator.create_volumes(vg, pvs)
-      rescue RuntimeError
+      rescue RuntimeError => error
+        log.error error.message
         lvm_creator = Proposal::LvmCreator.new(devicegraph)
         new_vg = vg.clone
         new_vg.lvs = flexible_devices(vg.lvs)
--- a/src/lib/y2storage/proposal/autoinst_devices_creator.rb
+++ b/src/lib/y2storage/proposal/autoinst_devices_creator.rb
@@ -81,25 +81,12 @@ module Y2Storage
         log.info "planned devices = #{planned_devices.to_a.inspect}"
         log.info "disk names = #{disk_names.inspect}"
 
-        # Process planned partitions
-        parts_to_create, parts_to_reuse, creator_result =
-          process_partitions(planned_devices, disk_names, original_graph.duplicate)
+        reset
 
-        # Process planned disk like devices (Xen virtual partitions and full disks)
-        planned_disk_like_devs = process_disk_like_devs(planned_devices, creator_result.devicegraph)
+        @planned_devices = planned_devices
+        @disk_names = disk_names
 
-        devs_to_create = parts_to_create
-
-        # Add planned disk like devices to reuse list so they can be considered for lvm and raids
-        # later on.
-        devs_to_reuse = parts_to_reuse + planned_disk_like_devs
-
-        creator_result = add_mds(planned_devices, devs_to_create, devs_to_reuse, creator_result)
-        creator_result = add_bcaches(planned_devices, devs_to_create, devs_to_reuse, creator_result)
-        creator_result = add_vgs(planned_devices, devs_to_create, devs_to_reuse, creator_result)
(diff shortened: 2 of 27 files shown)


src/lib/y2storage/planned/devices_collection.rb
```diff
@@ -156,7 +163,14 @@ def stray_blk_devices
...
       def all
-        @all ||= partitions + disks + stray_blk_devices + vgs + lvs + mds + bcaches
+        @all ||= partitions +
+          disks +
+          stray_blk_devices +
+          vgs +
+          lvs +
+          mds +
+          bcaches +
+          nfs_filesystems
```
jreidinger: NP: when I see this huge sum and also know that in past some customers have system with thousands of disks I start worrying about performance in that case. Maybe concat helps here to avoid recreation of resulting array multiple times?

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -89,27 +88,18 @@ def populated_devicegraph(planned_devices, disk_names)
...
-        devs_to_reuse.concat(mds_to_reuse.flat_map(&:partitions))
-
-        # Process planned bcaches
-        bcaches_to_create, bcaches_to_reuse, creator_result =
-          process_bcaches(planned_devices, devs_to_reuse, creator_result)
-        devs_to_reuse.concat(bcaches_to_reuse.flat_map(&:partitions))
+        creator_result = add_mds(planned_devices, devs_to_create, devs_to_reuse, creator_result)
+        creator_result = add_bcaches(planned_devices, devs_to_create, devs_to_reuse, creator_result)
+        creator_result = add_vgs(planned_devices, devs_to_create, devs_to_reuse, creator_result)
+        creator_result = add_nfs_filesystems(planned_devices, devs_to_create, creator_result)
```
jreidinger: what is reason to also return that creator_result and not having it as rw parameter only? I am not sure if this is more readable.

jreidinger: And also a bit confuse me as for me it looks like creator result is not modified. Is it modified or not ( I mean third parameter).

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
```diff
@@ -0,0 +1,212 @@
...
+      # @return [Array<Planned::Nfs>] Planned NFS filesystems
+      def planned_devices(drive)
+        # TODO: planned device with new format
+
+        planned_devices_old_format(drive)
+      end
+
+    private
+
+      NEW_FORMAT_MANDATORY_VALUES = { drive: [:device], partition: [:mount] }.freeze
```
jreidinger: can you document what this hash means? If someone will modify it in future, so he knows. ( like what is key and what is value ).

src/lib/y2storage/planned/devices_collection.rb
joseivanlopez: Ok, makes sense to me.

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
```diff
@@ -0,0 +1,212 @@
...
+      # @param format [:new, :old] whether the section is using the new or old AutoYaST style
+      # @return [Array<Symbol>]
+      def mandatory_partition_values(format)
+        mandatory_values(format)[:partition]
+      end
+
+      # Mandatory values depending on the AutoYaST style
+      #
+      # @param format [:new, :old] new or old AutoYaST style
+      # @return [Hash<Symbol, Array<Symbol>>]
```
jreidinger: here link documentation for constant or document it here and link constants to this method. But knowing meaning of symbols in hash is important to be written.

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
```diff
@@ -0,0 +1,212 @@
...
+      def path(share)
+        server_and_path(share).last || ""
+      end
+
+      # Name of the server and the shared directory from a NFS share
+      #
+      # @param share [String] e.g., "192.168.56.1:/root_fs"
+      # @return [Array<String>]
+      def server_and_path(share)
+        share.split(":")
```
jreidinger: can happen that there is multiple collong? and is it correct in that case? what about  nfs with password like `jreidinger:ubercoolpassword@192.168.56.1:/root_fs` can value look like this? Maybe is possible to simple prepend `nfs://` and use url parsing there? or can you link to documentation how nfs share can look like?

src/lib/y2storage/proposal/autoinst_devices_creator.rb
joseivanlopez: Well, first of all, I did this change only because rubocop was complaining because too many assignments. I agree this solution is quite confusing. The last parameter (creator_result) is not modified inside the method, for this reason a new creator_result is always returned. I will check if it is possible to modify it inside each method.

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
jreidinger: ah, username and password looks like not possible. Please link to `https://tools.ietf.org/html/rfc2224` so we know in future how this url can look like.

jreidinger: ah, but I found that path can be omitted, so I worry in that case it won't work.

joseivanlopez: Ok, fixed.

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -428,36 +326,54 @@ def create_mds(mds, previous_result, devs_to_reuse)
...
+      # @param vgs [Array<Planned::LvmVg>] List of planned volume groups to add
+      # @param devs_to_reuse [Array<Planned::Partition, Planned::StrayBlkDevice>] List of devices
+      #   to reuse as Physical Volumes
+      #
+      # @return [Proposal::CreatorResult] Result containing the specified volume groups
+      def set_up_lvm(vgs, devs_to_reuse)
+        # log separately to be more readable
+        log.info "BEGIN: set_up_lvm: vgs=#{vgs.inspect}"
+        log.info "BEGIN: set_up_lvm: previous_result=#{creator_result.inspect}"
+        log.info "BEGIN: set_up_lvm: devs_to_reuse=#{devs_to_reuse.inspect}"
```
jreidinger: hmm, if you want to use some keywords for easier filtering I suggest something more distinguish then `BEGIN:` I think that `set_up_lvm:` is enough for quick grepping of logs.

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -492,6 +408,22 @@ def create_bcache(devicegraph, bcache, backing_devname, caching_devname)
...
+      # Creates a volume group in the given devicegraph
+      #
+      # @param devicegraph [Devicegraph]                    Starting devicegraph
+      # @param vg          [Planned::LvmVg]                 Volume group
+      # @param pvs         [Planned::Partition,Planned::Md] List of physical volumes
+      # @return            [Proposal::CreatorResult] Result containing the specified volume group
+      def create_logical_volumes(devicegraph, vg, pvs)
+        lvm_creator = Proposal::LvmCreator.new(devicegraph)
+        lvm_creator.create_volumes(vg, pvs)
+      rescue RuntimeError
```
jreidinger: that a bit scary me that it just catch any runtime error and do not even log it. What if something else happens then what you expect? I think we should log at least exception message.

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
```diff
@@ -187,25 +206,29 @@ def mandatory_values(format)
       # Name of the server from a NFS share
       #
       # @param share [String] e.g., "192.168.56.1:/root_fs"
-      # @return [String]
+      # @return [String] e.g., "192.168.56.1"
       def server(share)
         server_and_path(share).first || ""
```
jreidinger: does "" makes sense here? I think that nfs without server is not possible.

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
```diff
@@ -187,25 +206,29 @@ def mandatory_values(format)
...
       end
 
       # Name of the shared directory from a NFS share
       #
       # @param share [String] e.g., "192.168.56.1:/root_fs"
-      # @return [String]
+      # @return [String] e.g., "/root_fs"
       def path(share)
         server_and_path(share).last || ""
```
jreidinger: hmm, if path is not defined I think path is according to that RFC ".", not?

src/lib/y2storage/proposal/autoinst_devices_creator.rb
joseivanlopez: This code was not added as part of this PR, I have only moved it. But ok, I will remove that "BEGIN".

src/lib/y2storage/proposal/autoinst_devices_creator.rb
joseivanlopez: Again, this was not added in this PR. I will log the error message.

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
joseivanlopez: Well, the idea is to create the NFS with the value given in the profile. If no server was given, then we will create it without a server name. libstorage-ng will complain later if the NFS share cannot be mounted.

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
joseivanlopez: I am not sure, but anyway, libstorage-ng does not support NFS without a path, so maybe we need another card for that.

jreidinger: yep, please create card for it.

PR #888, opened 2019-04-02
Added .cirrus.yml

lslezak:
## Problem

Travis is sometimes too slow and does not support Docker builds directly (only running docker commands inside an Ubuntu VM).

## Solution

Experimentally try another CI: https://cirrus-ci.com/


jreidinger:
@lslezak do you check also https://github.com/features/actions/ ?

lslezak:
@jreidinger not yet, it's currently in public beta testing and you have to register for the beta wait list... so maybe later.

PR #889, opened 2019-04-03
Fix resetting proposal settings

dgdavid:
## Problem

Initial settings are not being restored before each proposal attempt because

> `Yast.deep_copy` is quite a fail that only copies one level, leaving the second level as references instead of copies. — [extracted from Ancor's comment](https://github.com/yast/yast-storage-ng/pull/889#discussion_r271681321)

See also, 

- https://bugzilla.suse.com/show_bug.cgi?id=1130392
- https://trello.com/c/KJpkiH4P/889-1130392-storage-disabled-volumes-stay-disabled-during-proposal-process

## Solution

~~Overwrite the [`ProposalSettings#dup`](https://github.com/yast/yast-storage-ng/compare/SLE-15-SP1...feature/bsc-1130392-fix-reset-proposal-settings?expand=1#diff-30d4c4c7ff63d32d4f2fa075e77d3290R209)~~ Added a [`ProposalSettings#deep_copy`](https://github.com/yast/yast-storage-ng/pull/889/files#diff-30d4c4c7ff63d32d4f2fa075e77d3290R208) method to make a copy based on a `Marshal` serialization  instead of use `Yast.deep_copy` because

>  It did its job with the YCP -> Ruby translation, but it should never be used in new code because it does not do what it promises. We should use Ruby mechanisms for performing deep copies, such as Marshall. —  [extracted from Ancor's comment](https://github.com/yast/yast-storage-ng/pull/889#discussion_r271681321)

## Testing

- Added a new unit tests
- Fixed some existing tests

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Apr  3 07:12:34 UTC 2019 - David Diaz <dgonzalez@suse.com>
+
+- Fix initial proposal resetting settings properly before a new
+  attempt in other candidates (bsc#1130392).
+- 4.1.79
+
+-------------------------------------------------------------------
 Tue Apr  2 15:37:22 UTC 2019 - José Iván López González <jlopez@suse.com>
 
 - Add support for installing over NFS with AutoYaST (bsc#1130256).
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,8 +1,9 @@
 -------------------------------------------------------------------
 Wed Apr  3 07:12:34 UTC 2019 - David Diaz <dgonzalez@suse.com>
 
-- Fix initial proposal resetting settings properly before a new
-  attempt in other candidates (bsc#1130392).
+- Fix initial proposal to make a clean copy of initial settings
+  before switching to another candidate device (bsc#1130392).
+- Related to bsc#1102026 and bsc#1090383.
 - 4.1.79
 
 -------------------------------------------------------------------
--- a/src/lib/y2storage/initial_guided_proposal.rb
+++ b/src/lib/y2storage/initial_guided_proposal.rb
@@ -38,7 +38,7 @@ module Y2Storage
     def initialize(settings: nil, devicegraph: nil, disk_analyzer: nil)
       super
 
-      @initial_settings = self.settings.deep_copy
+      @initial_settings = self.settings
     end
 
   private
@@ -171,7 +171,7 @@ module Y2Storage
     #   devices have changed.
     def reset_settings
       self.space_maker = nil
-      self.settings = initial_settings.deep_copy
+      self.settings = initial_settings
     end
 
     # Creates a generator of settings
--- a/src/lib/y2storage/initial_guided_proposal.rb
+++ b/src/lib/y2storage/initial_guided_proposal.rb
@@ -38,7 +38,7 @@ module Y2Storage
     def initialize(settings: nil, devicegraph: nil, disk_analyzer: nil)
       super
 
-      @initial_settings = self.settings.dup
+      @initial_settings = self.settings.deep_copy
     end
 
   private
@@ -171,7 +171,7 @@ module Y2Storage
     #   devices have changed.
     def reset_settings
       self.space_maker = nil
-      self.settings = initial_settings.dup
+      self.settings = initial_settings.deep_copy
     end
 
     # Creates a generator of settings
--- a/src/lib/y2storage/initial_guided_proposal.rb
+++ b/src/lib/y2storage/initial_guided_proposal.rb
@@ -38,7 +38,7 @@ module Y2Storage
     def initialize(settings: nil, devicegraph: nil, disk_analyzer: nil)
       super
 
-      @initial_settings = Yast.deep_copy(self.settings)
+      @initial_settings = self.settings.dup
     end
 
   private
@@ -171,7 +171,7 @@ module Y2Storage
     #   devices have changed.
     def reset_settings
       self.space_maker = nil
-      self.settings = Yast.deep_copy(initial_settings)
+      self.settings = initial_settings.dup
     end
 
     # Creates a generator of settings
--- a/src/lib/y2storage/proposal_settings.rb
+++ b/src/lib/y2storage/proposal_settings.rb
@@ -206,7 +206,7 @@ module Y2Storage
     # Produces a deep copy of settings
     #
     # @return [ProposalSettings]
-    def dup
+    def deep_copy
       Marshal.load(Marshal.dump(self))
     end
 
--- a/src/lib/y2storage/proposal_settings.rb
+++ b/src/lib/y2storage/proposal_settings.rb
@@ -203,6 +203,13 @@ module Y2Storage
       load_features
     end
 
+    # Produces a deep copy of settings
+    #
+    # @return [ProposalSettings]
+    def dup
+      Marshal.load(Marshal.dump(self))
+    end
+
     # Whether encryption must be used
     # @return [Boolean]
     def use_encryption
--- a/test/y2storage/initial_guided_proposal_test.rb
+++ b/test/y2storage/initial_guided_proposal_test.rb
@@ -237,7 +237,7 @@ describe Y2Storage::InitialGuidedProposal do
     # Test, at hight level, that settings are reset between candidates
     #
     # Related to bsc#113092, settings must be **correctly** reset after moving to another (group
-    # of) candidate device(s). To check that, the first candidate will be enought small to make not
+    # of) candidate device(s). To check that, the first candidate will be small enough to make not
     # possible the proposal on it even **after adjust the initial settings**, expecting to have a
     # valid proposal **with original settings** in the second candidate.
     context "when a proposal is not possible for a candidate even after adjust the settings" do
@@ -259,7 +259,7 @@ describe Y2Storage::InitialGuidedProposal do
 
         expect(used_devices).to contain_exactly("/dev/sdb")
 
-        # having expeted mount points means that settings were reset properly, since in the
+        # having expected mount points means that settings were reset properly, since in the
         # previous attempts swap and separated home should be deleted
         expect(mount_points).to include("swap", "/home", "/")
       end
--- a/test/y2storage/initial_guided_proposal_test.rb
+++ b/test/y2storage/initial_guided_proposal_test.rb
@@ -234,6 +234,37 @@ describe Y2Storage::InitialGuidedProposal do
       end
     end
 
+    # Test, at hight level, that settings are reset between candidates
+    #
+    # Related to bsc#113092, settings must be **correctly** reset after moving to another (group
+    # of) candidate device(s). To check that, the first candidate will be enought small to make not
+    # possible the proposal on it even **after adjust the initial settings**, expecting to have a
+    # valid proposal **with original settings** in the second candidate.
+    context "when a proposal is not possible for a candidate even after adjust the settings" do
+      include_context "candidate devices"
+
+      let(:candidate_devices) { ["/dev/sda", "/dev/sdb"] }
+
+      let(:control_file_content) { ng_partitioning_section }
+
+      before do
+        sda.size = 2.GiB
+      end
+
+      it "resets the settings before attempting a new proposal with next candidate" do
+        proposal.propose
+
+        partitions = proposal.devices.partitions
+        mount_points = partitions.map(&:filesystem_mountpoint).compact
+
+        expect(used_devices).to contain_exactly("/dev/sdb")
+
+        # having expeted mount points means that settings were reset properly, since in the
+        # previous attempts swap and separated home should be deleted
+        expect(mount_points).to include("swap", "/home", "/")
+      end
+    end
+
     context "when a proposal is not possible" do
       include_context "candidate devices"
 
--- a/test/y2storage/proposal/settings_generator/ng_test.rb
+++ b/test/y2storage/proposal/settings_generator/ng_test.rb
@@ -30,6 +30,10 @@ describe Y2Storage::Proposal::SettingsGenerator::Ng do
     settings.volumes.find { |v| v.mount_point == mount_point }
   end
 
+  def remove_object_ids(input)
+    input.gsub(/:0x\w+ /, "")
+  end
+
   describe "#next_settings" do
     before do
       stub_product_features("partitioning" => partitioning_features)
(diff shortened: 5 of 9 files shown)


test/y2storage/initial_guided_proposal_test.rb
```diff
@@ -234,6 +234,37 @@
       end
     end
 
+    # Test, at hight level, that settings are reset between candidates
+    #
+    # Related to bsc#113092, settings must be **correctly** reset after moving to another (group
+    # of) candidate device(s). To check that, the first candidate will be enought small to make not
```
ancorgs: small enough (surplus "t")

test/y2storage/initial_guided_proposal_test.rb
```diff
@@ -234,6 +234,37 @@
...
+
+      it "resets the settings before attempting a new proposal with next candidate" do
+        proposal.propose
+
+        partitions = proposal.devices.partitions
+        mount_points = partitions.map(&:filesystem_mountpoint).compact
+
+        expect(used_devices).to contain_exactly("/dev/sdb")
+
+        # having expeted mount points means that settings were reset properly, since in the
```
ancorgs: expe**c**ted

test/y2storage/proposal_settings_test.rb
```diff
@@ -39,6 +39,51 @@ def stub_partitioning_features(features = {})
...
+
+    before do
+      stub_partitioning_features(partitioning)
+    end
+
+    it "returns a deep copy of settings" do
+      dup = settings.dup
+
+      # Let's simply check two nested levels: it's expected to found the same amount of objects with
+      # a different identity. In other words, object must looks equal but being different.
```
ancorgs: Nitipicking english correction: "object must look equal but be different"

test/y2storage/proposal_settings_test.rb
```diff
@@ -39,6 +39,51 @@ def stub_partitioning_features(features = {})
...
+    end
+
+    before do
+      stub_partitioning_features(partitioning)
+    end
+
+    it "returns a deep copy of settings" do
+      dup = settings.dup
+
+      # Let's simply check two nested levels: it's expected to found the same amount of objects with
```
ancorgs: ~~to found~~ to find

test/y2storage/proposal_settings_test.rb
```diff
@@ -39,6 +39,51 @@ def stub_partitioning_features(features = {})
...
+      dup_volumes_desired_sizes_ids = dup_volumes_desired_sizes.map(&:object_id)
+
+      settings_volumes = settings.volumes
+      settings_volumes_ids = settings_volumes.map(&:object_id)
+      settings_volumes_desired_sizes = settings_volumes.map(&:desired_size)
+      settings_volumes_desired_sizes_ids = settings_volumes_desired_sizes.map(&:object_id)
+
+      expect(dup.object_id).to_not eq(settings.object_id)
+      expect(dup_volumes_ids).to_not eq(settings_volumes_ids)
+      expect(dup_volumes_desired_sizes_ids).to_not eq(settings_volumes_desired_sizes_ids)
```
ancorgs: I don't see any check about the objects "looking the same". I just see checks about them "not being the same".

dgdavid: :thinking: :thinking: :thinking:  Do you know what? You're right. Let me fix it :)

src/lib/y2storage/initial_guided_proposal.rb
```diff
@@ -38,7 +38,7 @@ class InitialGuidedProposal < GuidedProposal
     def initialize(settings: nil, devicegraph: nil, disk_analyzer: nil)
       super
 
-      @initial_settings = Yast.deep_copy(self.settings)
+      @initial_settings = self.settings.dup
```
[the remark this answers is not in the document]
dgdavid: @lslezak,

Exactly, it "does not work". I guess you mean that we only need to use `initial_settings.dup` inside the `#reset_settings` method, since it is there where the copy is being made. But for the same reason we do not need to use none of them at this point, `Yast.deep_copy` nor `ProposalSettings#dup`. However, I think we want to hold a _real_ deep copy here.

@joseivanlopez, thoughts?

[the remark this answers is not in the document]
dgdavid: >  However, I think we want to hold a real deep copy **of the initial/original settings** here.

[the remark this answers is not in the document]
lslezak: OK, then we should just add a comment why we are using own `deep_copy` implementation here as it might be confusing later.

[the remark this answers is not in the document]
dgdavid: Just curiosity,  

Why do you think it is more confusing to see `self.settings.dup` than `Yast.deep_copy(self.settings)`?

In both case, what comes to my mind when reading the code is

> Ok, let's store a deep copy of current settings as `initial_settings`.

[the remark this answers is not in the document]
lslezak: Ah, I was not talking about using deep copy, that's fine regardless which implementation is used. 

I meant why we need implementing own version of deep copy and not using `Yast.deep_copy` directly. Something like

```ruby
# we cannot use Yast.deep_copy here because it keeps references to nested objects
```

(or something similar, I do not know the exact reason).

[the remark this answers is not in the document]
ancorgs: @lslezak Well, that was exactly the root of the problem. @joseivanlopez  (who wrote the original code) assumed `Yast.deep_copy` would perform a deep copy of the passed object. But guess what: it does not!

`Yast.deep_copy` is quite a fail that only copies one level, leaving the second level as references instead of copies. It did its job with the YCP -> Ruby translation, but it should never be used in new code because it does not do what it promises. We should use Ruby mechanisms for performing deep copies, such as `Marshall`.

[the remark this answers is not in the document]
joseivanlopez: Indeed, this copy could be optional because `#reset_settings` is call at the beginning of each try.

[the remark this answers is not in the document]
dgdavid: > Indeed, this copy could be optional because #reset_settings is call at the beginning of each try.

So, what do you think it is better? Keep it as it is, or remove this copy?

src/lib/y2storage/proposal_settings.rb
```diff
@@ -203,6 +203,13 @@ def for_current_product
       load_features
     end
 
+    # Produces a deep copy of settings
+    #
+    # @return [ProposalSettings]
+    def dup
```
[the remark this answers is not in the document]
dgdavid: Hmmmm, I think you are right. I agree.

@ancorgs, @joseivanlopez if you do not mind, I'll change it.

src/lib/y2storage/initial_guided_proposal.rb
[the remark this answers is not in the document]
dgdavid: BTW, thanks to this discussion I realized that there is another place where we must to replace `Yast.deep_copy` by `ProposalSettings#(dup|deep_copy)`

`SettingsGenerator` has also an `initial_settings`

https://github.com/yast/yast-storage-ng/blob/06c72586a788e209bb04dc350a0a2ec66ba542b2/src/lib/y2storage/proposal/settings_generator/base.rb#L58-L60

Which is private method

https://github.com/yast/yast-storage-ng/blob/06c72586a788e209bb04dc350a0a2ec66ba542b2/src/lib/y2storage/proposal/settings_generator/base.rb#L90-L101

That make a "deep_copy" of the settings (ProposalSettings)

https://github.com/yast/yast-storage-ng/blob/06c72586a788e209bb04dc350a0a2ec66ba542b2/src/lib/y2storage/proposal/settings_generator/base.rb#L106-L109

So, we could change the `#initial_settings` to 

```ruby
def initial_settings
   return nil if @used_initial_settings
   @used_initial_settings = true

   settings.dup # or settings.deep_copy
end
```

Right?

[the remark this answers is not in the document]
dgdavid: Something similar for `SettingsGenerator#initialize_settings`.

We could change it from 

https://github.com/yast/yast-storage-ng/blob/06c72586a788e209bb04dc350a0a2ec66ba542b2/src/lib/y2storage/proposal/settings_generator/base.rb#L79-L81

to 

```ruby
def initialize_settings(settings)
  settings.dup # or sttings.deep_copy
end
```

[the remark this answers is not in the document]
jreidinger: yep, `Yast.deep_copy` goal is to only provide way how old ycp do its object copy. So in general I really do not recommend to do it in new code. It e.g. even do not dup strings as it cause serious slow down of whole yast code and old yast use immutable string, so to fix it Yast.deep_copy do not dup strings and do not handle at all types that is not allowed by yast component system.

src/lib/y2storage/proposal_settings.rb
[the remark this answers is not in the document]
dgdavid: Done in commit https://github.com/yast/yast-storage-ng/pull/889/commits/f3746c8a7b849c10facee636339868176e30130c

test/y2storage/proposal/settings_generator/ng_test.rb
```diff
@@ -75,14 +79,17 @@ def volume_from_settings(settings, mount_point)
...
 
     context "when called for first time" do
       it "returns the same values as the initial settings" do
         next_settings = subject.next_settings
 
-        expect(next_settings.to_s).to eq(settings.to_s)
+        settings_content = remove_object_ids(settings.to_s)
+        next_settings_content = remove_object_ids(next_settings.to_s)
+
+        expect(next_settings_content).to eq(settings_content)
```
dgdavid: If this approach looks good to you, I could do the same with the added example to test the initial guided proposal at  https://github.com/yast/yast-storage-ng/pull/889/files#diff-024fcd1dc73eb2f382db64fe6cb7529fR264. But if it smells bad... I'll undo it :)

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Apr  3 07:12:34 UTC 2019 - David Diaz <dgonzalez@suse.com>
+
+- Fix initial proposal resetting settings properly before a new
```
lslezak: This is hardly readable/understandable, could you please rephrase it a bit? Keep in mind that this is also read by the customers...

test/y2storage/proposal_settings_test.rb
```diff
@@ -39,6 +39,55 @@ def stub_partitioning_features(features = {})
...
+
+    before do
+      stub_partitioning_features(partitioning)
+    end
+
+    it "returns a deep copy of settings" do
+      copy = settings.deep_copy
+
+      # Let's simply check two nested levels: it's expected to find the same amount of objects with
+      # a different identity. In other words, object must look equal but be different.
```
lslezak:
```suggestion
      # a different identity. In other words, the objects must be equal but must be different instances.
```

test/y2storage/proposal_settings_test.rb
```diff
@@ -39,6 +39,55 @@ def stub_partitioning_features(features = {})
...
+    subject(:settings) { described_class.new_for_current_product }
+
+    let(:volumes) do
+      [
+        {
+          "mount_point" => "/", "fs_type" => "xfs", "weight" => 60,
+          "desired_size" => "20GiB", "max_size" => "40GiB"
+        },
+        { "mount_point" => "/home", "fs_type" => "xfs", "weight" => 40, "desired_size" => "10GiB" },
+        # This should reuse the existing logical swap
```
ancorgs: I don't get this comment (sorry for not mentioning it before). Looks like a copy&paste which doesn't make sense in the scope of this test.

test/y2storage/proposal/settings_generator/ng_test.rb
lslezak: Comparing string representations of objects does not seem to be the right way to me, I'd avoid that if possible. IMHO too fragile...

test/y2storage/proposal_settings_test.rb
dgdavid: A left over from the copy&paster, :man_facepalming:

src/lib/y2storage/proposal_settings.rb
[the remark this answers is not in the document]
joseivanlopez: Well, to be honest, I haven't thought about this before, but I agree with @lslezak. IIRC, none class from the ruby Standard Library implements its own `#dup`. So, I would keep `#dup` with its original meaning.

test/y2storage/proposal/settings_generator/ng_test.rb
dgdavid: The test was already using the string representation (I guess because of `ProposalSettings` is redefining the [`#to_s`](https://github.com/yast/yast-storage-ng/blob/06c72586a788e209bb04dc350a0a2ec66ba542b2/src/lib/y2storage/proposal_settings.rb#L257))

So, shall we change this?

dgdavid: Changed. See https://github.com/yast/yast-storage-ng/pull/889/commits/7b3d778a20bef749c9bf7226dee40f3c77a445f4 and https://github.com/yast/yast-storage-ng/pull/889/commits/a75490ebf2f42e5ec565e888f440a8ca9aeda8a7.

lslezak:
Yes, looks better to me.

test/y2storage/proposal_settings_test.rb
```diff
@@ -39,6 +39,54 @@ def stub_partitioning_features(features = {})
...
+
+    let(:partitioning) do
+      { "proposal" => {}, "volumes" => volumes }
+    end
+
+    before do
+      stub_partitioning_features(partitioning)
+    end
+
+    it "returns a deep copy of settings" do
```
joseivanlopez: NP: Only an idea, why not do something like this:

```
    it "creates a new object" do
      expect(settings.deep_copy).to_not eql(settings)
    end
    
    it "creates an object with the same values" do
      expect(settings.deep_copy).to eq(settings)
    end

   it "creates an object with different references" do
     ids = settings.volumes.map(&:object_id).sort
     new_ids = settings.deep_copy.volumes.map(&:object_id).sort
     
     expect(new_ids).to_not eq(ids)
   end
```

joseivanlopez: I would also add some test for `#==` method.

joseivanlopez:
I see some tests are starting to fail because "double" objects cannot be marshaled. Maybe we can fix it by replacing `double` by `instance_double`, for example here https://github.com/yast/yast-storage-ng/blob/master/test/y2storage/dialogs/guided_setup/select_filesystem/ng_test.rb#L38

dgdavid:
@joseivanlopez 

Thank you for your feedback.

I already updated the PR with your suggestions, but I fix those tests that are failing. Using `instance_double` do not fix them.

joseivanlopez:
@dgdavid I was not sure about that solution (I was only guessing). First of all, we have to be sure that the problem is the combination of Marshall + mocking. This can be easily checked. And if we confirm that, then we have 3 options:

a) try to use real VolumeSpecification objects in tests, although I don't like it because we force to always use real settings.
b) do not redefine `ProposalSettings#==`. The problem I see with this is that `settings == settings.dup` but `settings != settings.deep_copy`.
c) do not use Marshal to redefine `ProposalSettings#==`, but maybe a good solution would require too much effort.

So, if finally the problem is Marshal + mocking, IMHO I would go with option b) iff  `settings != settings.deep_copy` is not a problem (not sure if this could affect in some parts of the code).

@ancorgs How do you see it?

ancorgs:
I wouldn't redefine `ProposalSettings#==` just to have nicer tests. That's a change with potentially many implications, definitely not something to do in the SLE-15-SP1 branch at this point in time.

Fixing the deep dup is the further I would go.

dgdavid:
@joseivanlopez, @ancorgs 

So, I'll undo the latest changes and keep the PR as it was [20 hours ago](https://github.com/yast/yast-storage-ng/pull/889#pullrequestreview-222256777).

Thank you!

PR #890, opened 2019-04-04
Backport: Add support for NFS in AutoYaST

joseivanlopez:
Backporting PR https://github.com/yast/yast-storage-ng/pull/884.

### DIFF
--- a/src/lib/y2storage/proposal/autoinst_nfs_planner.rb
+++ b/src/lib/y2storage/proposal/autoinst_nfs_planner.rb
@@ -38,11 +38,29 @@ module Y2Storage
 
     private
 
-      NEW_FORMAT_MANDATORY_VALUES = { drive: [:device], partition: [:mount] }.freeze
-
+      # This hash defines mandatory profile values for a NFS share using the old AutoYaST style.
+      #
+      # Keys are the name of the section, and values are the mandatory values for such section.
+      #
+      # With old format, the partition section must contain a device and a mount value, e.g.:
+      #
+      # <drive>
+      #   <device>/dev/nfs</device>
+      #   <partitions>
+      #     <partition>
+      #       <device>192.168.56.1:/root_fs</device>
+      #       <mount>/</mount>
+      #     </partition>
+      #   </partitions>
+      # </drive>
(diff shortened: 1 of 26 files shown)


src/lib/y2storage/proposal/autoinst_nfs_planner.rb
```diff
@@ -0,0 +1,235 @@
...
+      #       <device>192.168.56.1:/root_fs</device>
+      #       <mount>/</mount>
+      #     </partition>
+      #   </partitions>
+      # </drive>
+      OLD_FORMAT_MANDATORY_VALUES = { drive: [], partition: [:device, :mount] }.freeze
+
+      # Similar to {OLD_FORMAT_MANDATORY_VALUES}, but for the new AutoYaST style.
+      #
+      # TODO
```
imobachgs: I would add a note pointing out that it is available in SLE 15 SP1. Otherwise, I would think it is just missing, and we need to implement it.

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
```diff
@@ -0,0 +1,235 @@
...
+
+      # Whether any value is missing for the drive section
+      #
+      # Errors are registered when values are missing.
+      #
+      # @param drive [AutoinstProfile::DriveSection]
+      # @param format [:new, :old] whether the section is using the new or old AutoYaST style
+      #
+      # @return [Boolean]
+      def missing_drive_values?(drive, format)
```
imobachgs: As "format" is the name of a method, I would use something like "fmt".

src/lib/y2storage/proposal/autoinst_nfs_planner.rb
imobachgs: OK, after an offline discussion, it is clear that we will implement the "export" feature later.

imobachgs:
> In general, it looks good, but I did not see any modification to `AutoinstProfile` classes to export the NFS shares. And if it works out-of-the-box, a test would be nice.

We will implement this feature later. So all good :)

PR #896, opened 2019-04-09
Add support for NFS in Autoyast (new format)

joseivanlopez:
## Problem

NFS drives are specified in the AutoYaST profile by an "undocumented" option. That is, by defining a drive with a "/dev/nfs" device, and then, each partition defines a NFS share, e.g.:

```
<drive>
  <device>/dev/nfs</device>
  <partitions>
    <partition>
      <mount>/</mount>
      <device>server:path</device>
    </partition>
  </partitions>
</drive>
```

The <partition> section contains a <device> attribute. This is quite unexpected, because a <partition> section cannot contain a <device>. In fact, the AutoYaST profile does not validate.

The support for importing this kind of drives was already implemented in [here](https://github.com/yast/yast-storage-ng/pull/890). But now we have to find a better way for defining NFS drives and be able to import/export this new format.

Note: currently, old format cannot be exported at all because we don't want to continue supporting this format for new products (from SLE-15 on).  The new format will be available for SP0 (GA), SP1 and master.

* https://bugzilla.suse.com/show_bug.cgi?id=1130256
* https://trello.com/c/wATMQ3Mo/917-5-import-export-nfs-shares-with-new-profile-format

## Solution

Added support for importing/exporting new NFS format:

```
<drive>
  <device>server:path</device>
  <type>CT_NFS</drive>
  <partitions>
    <partition>
      <mount>/</mount>
    </partition>
  </partitions>
</drive>
```

## Testing

* Added unit tests
* Tested manually

### DIFF
--- a/src/lib/y2storage/proposal/autoinst_nfs_planner.rb
+++ b/src/lib/y2storage/proposal/autoinst_nfs_planner.rb
@@ -31,9 +31,9 @@ module Y2Storage
       # @param drive [AutoinstProfile::DriveSection] drive section describing NFS filesystems
       # @return [Array<Planned::Nfs>] Planned NFS filesystems
       def planned_devices(drive)
-        # TODO: planned device with new format
+        return planned_devices_old_format(drive) if old_format?(drive)
 
-        planned_devices_old_format(drive)
+        planned_devices_new_format(drive)
       end
 
     private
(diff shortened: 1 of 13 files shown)


src/lib/y2storage/proposal/autoinst_nfs_planner.rb
```diff
@@ -57,11 +57,25 @@ def planned_devices(drive)
...
+      #       <mount>/</mount>
+      #     </partition>
+      #   </partitions>
+      # </drive>
+      NEW_FORMAT_MANDATORY_VALUES = { drive: [:device], partition: [:mount] }.freeze
 
       private_constant :OLD_FORMAT_MANDATORY_VALUES, :NEW_FORMAT_MANDATORY_VALUES
 
+      def old_format?(drive)
+        drive.device == "/dev/nfs" || drive.partitions.any? { |p| !p.device.nil? }
```
imobachgs: np: You could write `drive.partitions.any?(&:device)` or even `drive.partitions.any? { |p| p.device }`.

PR #903, opened 2019-04-11
First version of btrfs_in_partitioner.md

ancorgs:
Better use the "view file" button for the `.md` file. That will render the Markdown file with the right format and the inline images.

See https://trello.com/c/1aS03eJZ/922-3-plan-the-partitioner-ui-change


doc/btrfs_in_partitioner.md
```diff
@@ -0,0 +1,326 @@
...
+## The target Btrfs features at a glance
+
+First of all, let's take a sligthly deeper look to these two Btrfs features in
+order to understand their basis.
+
+### Multi-device filesystem
+
+As already mentioned, this is not so different from the capabilities already
+offered by LVM or software RAIDs. It basically refers to the ability of Btrfs of
+distributing its data and metadata over several block filesystems. That can be
```
joseivanlopez: several block **devices**.

doc/btrfs_in_partitioner.md
```diff
@@ -0,0 +1,326 @@
...
+The first two screens remain identical. One for defining the size of the
+partition.
+
+![Partition creation step 1](btrfs_in_partitioner/new_part_1.png)
+
+And the role selection which allows to predefine some of the options for the
+following screen(s).
+
+![Partition creation step 2](btrfs_in_partitioner/new_part_2.png)
+
```
joseivanlopez: I expected the "Role dialog" to be the third stop. Once I have defined my partition (size, encryption, id, etc) then I decide which role to apply (i.e., how I want to use my partition). In fact, if I decide to not format the partition, the role dialog is not needed at all, so we can finish the wizard just in the second step.

joseivanlopez: This would also make the role "Raw partition" leftover.

ancorgs: But the point of the "Role" screen is to prefill the options in the wizard before making any other decision. It's kind of weird to have questions about format/dontformat and about the Partition ID before selecting the role.

ancorgs: I mean:

- Operating system: preselects "format", partition ID linux and the next available mount point
- Data: basically the same but with a different set of mount points
- Swap: preselects "format", the partition ID, the filesystem and the mount point
- EFI: preselects "format", the partition ID, the filesystem and the mount point
- Raw: preselect "don't format" and the LVM Partition ID.

If I select the role after having selected already whether to format and the partition ID... I don't know. It feels wrong.

joseivanlopez: Well, if the partition role affects to the partition id, then I agree.

PR #897, opened 2019-04-10
Remove unused LvmPv when editing a partition

dgdavid:
## Problem

Editing a partition with LvmPv not associated with any LvmVg could generate not doable operations in the real system. More specifically, 

> If a partition contains an LvmPv not associated to any LvmVg, it's possible to edit the partition and mark the "encrypt device" check-box. The result is a devicegraph in which a new Encrypt device in inserted between the partition and the existing LvmPv

- https://bugzilla.suse.com/show_bug.cgi?id=1129663
- https://trello.com/c/RMJphyX3/914-tumbleweed-1129663-setting-up-encrypted-partition-with-yast-errors-stub-doreallottext-called-for-lvmpv

## Solution

Remove unused LvmPv when a partition is edited.

## Testing

- Added a new unit test
- Also Tested manually

## Screenshots

Given the [following devicegraph](https://github.com/yast/yast-storage-ng/files/3062824/empty_lvm_pv.zip)

original_device_graph

### :bug:  Results with current code base

| After going through the /dev/sda2 edit dialog | After add encryption to /dev/sda2 |
|-----------------------------------|----------------------------------------|
| after_going_through_edit_dialog_sda3_wo_changes_before_fix png_sda2_wo_fix | after_donotformat_and_encrypt_sda2_wo_fix |

| After keep the /dev/sda3 encryption | After remove the /dev/sda3 encryption |
|-----------------------------------|----------------------------------------|
| after_going_through_edit_dialog_sda3_wo_changes_before_fix png_sda2_wo_fix | after_remove_encryption_sda3_before_fix |

### :heart: Results after merge proposed changes
| After going through the /dev/sda2 edit dialog | After add encryption to /dev/sda2 |
|-----------------------------------|----------------------------------------|
| after_going_through_edit_dialog_sda2_wo_changes |  after_donotformat_and-encrypt_sda2 |

| After keep the /dev/sda3 encryption | After remove the /dev/sda3 encryption |
|-----------------------------------|----------------------------------------|
| after_going_through_edit_dialog_sda3_wo_changes | after_remove_encryption_sda3_wo_format |

---

[empty_lvm_pv.zip](https://github.com/yast/yast-storage-ng/files/3062824/empty_lvm_pv.zip)

PR #904, opened 2019-04-12
Remove dead code

dgdavid:
This PR does not solve a specific problem. Instead, simply removes

* A [quite confusing comment](https://github.com/yast/yast-storage-ng/commit/ac181fb3516bcdbea57c9eaf140d781ed1a730af) that we found while working on  https://github.com/yast/yast-storage-ng/pull/897. It could be interpreted in two ways

  - It is describing an old behavior and clarifying that it is no longer available in the new Partitioner, or 
  - It is a forgotten note which should have been removed [when the `BlkDeviceRestorer` was included](https://github.com/yast/yast-storage-ng/commit/6b22232d69100a6b8428774399a85198047af3b6).

  In case of the first interpretation, it seems to be not necessary nowadays. For the second one, it is simply a comment that must be removed.

* A [leftover method](https://github.com/yast/yast-storage-ng/commit/a69c815544a0588c1c6b9e3a0f250c5c59c19d9a). It is private and is not being used.

---

:warning: Version bump not required :warning:

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/filesystem.rb
+++ b/src/lib/y2partitioner/actions/controllers/filesystem.rb
@@ -213,13 +213,6 @@ module Y2Partitioner
 
         # Makes the changes related to the option "do not format" in the UI, which
         # implies removing any new filesystem or respecting the preexisting one.
-        #
-        # @note With the current implementation there is a corner case that
-        # doesn't work like the traditional expert partitioner. If a partition
-        # preexisting in the disk is edited (e.g. replacing the filesystem with a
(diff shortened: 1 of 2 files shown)


joseivanlopez:
Wait! Why SP1?

dgdavid:
Thanks @joseivanlopez 

> Wait! Why SP1?

Maybe I am a bit lost. To keep `SLE-15-SP1` and `master` synchronized? I thought apply changes in that branch and then "port" them to `master`.

Is it wrong?

dgdavid:
Changed to `master`.

PR #907, opened 2019-04-16
New BTRFS section in Expert Partitoner

joseivanlopez:
NOTE FOR REVIEWERS: codeclimate is stupidly complaining. Feel free either to ignore it or to mark it as wontfix (for some reason, I am not allowed to mark it).

## Problem

This is the first step in the Expert Partitioner side for supporting multidevice BTRFS filesystems. There is already a BTRFS section in the Expert Partitioner, but it needs to be improved to show both: multidevice and non-multidevice BTRFS.

Moreover, each BTRFS should have its own page with two tabs: "Overview" and "Used Devices".

* https://trello.com/c/a4rSzGQk/931-3-partitioner-filesystems-entry-in-the-tree

## Solution

Adapt current BTRFS section and create specific page for each BTRFS.

BONUS: refactoring of code to generate the description of the devices (generally, the "Overview" tab content). 

## Testing

- Added units test
- Tested manually

## Screenshots

VirtualBox_openSUSE Tumbleweed_24_04_2019_14_58_44

VirtualBox_openSUSE Tumbleweed_24_04_2019_14_59_26

VirtualBox_openSUSE Tumbleweed_24_04_2019_14_36_05

VirtualBox_openSUSE Tumbleweed_24_04_2019_14_36_19

VirtualBox_openSUSE Tumbleweed_24_04_2019_14_36_31


src/lib/y2partitioner/actions/edit_btrfs.rb
```diff
@@ -0,0 +1,72 @@
...
+
+module Y2Partitioner
+  module Actions
+    # Action for editing a BTRFS filesystem, see {Actions::Base}
+    class EditBtrfs < Base
+      # Constructor
+      #
+      # @param filesystem [Y2Storage::Filesystems::Btrfs]
+      def initialize(filesystem)
+        super()
```
dgdavid: In submitted changes, you are using both, `super()` and `super`.

Better to choose, and use, only one. Isn't it? :)

src/lib/y2partitioner/widgets/description_section/bcache.rb
```diff
@@ -0,0 +1,122 @@
...
+          # TRANSLATORS: title for section about Bcache details
+          _("Bcache Devices:")
+        end
+
+        # @see DescriptionSection::Base#entries
+        def entries
+          ENTRIES
+        end
+
+        # Enty data about the backing device
```
dgdavid: Typo: "Ent**r**y".

The same for the next 3 methods.

src/lib/y2partitioner/widgets/description_section/blk_device.rb
```diff
@@ -0,0 +1,89 @@
...
+          # TRANSLATORS: title for section about block device details
+          _("Device:")
+        end
+
+        # @see DescriptionSection::Base#entries
+        def entries
+          ENTRIES
+        end
+
+        # Enty data about the udev by_path values
```
dgdavid: The same typo here, "Ent**r**y".

src/lib/y2partitioner/widgets/description_section/base.rb
```diff
@@ -0,0 +1,121 @@
...
+        #   @example
+        #
+        #     section.entries #=> [{ value: :device_name, help: :name }, { value: :device_size }]
+        #
+        #   The above section is composed by two entries. The value of the first one is generated by the
+        #   method :device_name and it uses the help field :name. And for the second entry, its value is
+        #   generated by calling to :device_size method. In this case there is no help field for the
+        #   entry.
+        #
+        #   @return [Array<Hash>]
```
ancorgs: Not sure if I fully understand this. For the value part, I don't see why a method name is used instead of the real content. And for the help part I'm not sure if I get the concept of the "help field".

In this example below (simplified on purpose):
```ruby
class Bcache < Base

  def entries
   [ { value: :bcache_cache_mode,     help: :cache_mod } ]
  end
end
```

I understand `bcache_cache_mode` is a method of `Widgets::DescriptionSection::Bcache` (although it's a little bit weird) but... where `cache_mod` comes from? Is a method? Defined where?

src/lib/y2partitioner/widgets/overview.rb
```diff
@@ -285,8 +285,18 @@ def nfs_section
...
       # @return [CWM::PagerTreeItem]
       def btrfs_section
-        page = Pages::Btrfs.new(self)
-        section_item(page, Icons::BTRFS)
+        filesystems = device_graph.btrfs_filesystems
+
+        page = Pages::BtrfsFilesystems.new(filesystems, self)
+        children = filesystems.map { |f| btrfs_item(f) }
+
+        CWM::PagerTreeItem.new(page, children: children, icon: Icons::BTRFS)
```
dgdavid: Now we're using the [`#section_item`](https://github.com/yast/yast-storage-ng/blob/721ea612f96651d377936836800cc701378985df/src/lib/y2partitioner/widgets/overview.rb#L336) method.

src/lib/y2partitioner/widgets/overview.rb
```diff
@@ -285,8 +285,18 @@ def nfs_section
...
+      end
+
+      # @return [CWM::PagerTreeItem]
+      def btrfs_item(filesystem)
+        page = Pages::Btrfs.new(filesystem, self)
+        CWM::PagerTreeItem.new(page)
```
dgdavid: Now we're using the [`#device_item`](https://github.com/yast/yast-storage-ng/blob/721ea612f96651d377936836800cc701378985df/src/lib/y2partitioner/widgets/overview.rb#L348) method

src/lib/y2partitioner/widgets/description_section/base.rb
ancorgs: Now I believe I understand `:help` points to one of the keys accepted as argument by `Widgets::Help::helptext_for`. That part is definitely not the most object oriented part, it looks a lot like old YaST.

Anyway, that leaded me to realize we are not checking whether it's a legitimate key. Take the already pointed `cache_mod` as example.

joseivanlopez: Yes, as commented in our call, I am going to improve this part. I will add a test to be sure we are providing the required helps.

src/lib/y2partitioner/actions/edit_btrfs.rb
joseivanlopez: In this case we need to use `super()` because the parent constructor must be called without parameters.

PR #912, opened 2019-04-26
Edit Btrfs filesystems

joseivanlopez:
## Problem

Currently, the button for editing a Btrfs filesystem opens a dialog to manage its subvolumes. The button should open a dialog for editing the filesystem options, similar to the dialog for editing a block device.

* https://trello.com/c/Icti2ZBT/947-3-partitioner-editing-btrfs-filesystems

## Solution

Now, the edit button opens a dialog to edit the Btrfs filesystem. The dialog allows to set the mount point, mount options, configure subvolumes, and enable/disable snapshots (this last only for root during installation).

## Testing

- Added unit tests
- Tested manually

## Screenshots

VirtualBox_openSUSE Tumbleweed_30_04_2019_11_32_54

### DIFF
--- a/src/lib/y2partitioner/filesystem_errors.rb
+++ b/src/lib/y2partitioner/filesystem_errors.rb
@@ -94,7 +94,9 @@ module Y2Partitioner
     def small_size_for_snapshots?(filesystem, new_size: nil)
       return false unless filesystem && filesystem_with_snapshots?(filesystem)
 
-      # FIXME: Does this check make sense for multidevice Btrfs?
+      # TODO: check size for multidevice Btrfs
+      return false if filesystem.multidevice?
+
       size = new_size || filesystem.blk_devices.first.size
       min_size = min_size_for_snapshots(filesystem)
 
--- a/src/lib/y2partitioner/filesystem_errors.rb
+++ b/src/lib/y2partitioner/filesystem_errors.rb
@@ -94,6 +94,7 @@ module Y2Partitioner
     def small_size_for_snapshots?(filesystem, new_size: nil)
       return false unless filesystem && filesystem_with_snapshots?(filesystem)
 
+      # FIXME: Does this check make sense for multidevice Btrfs?
       size = new_size || filesystem.blk_devices.first.size
       min_size = min_size_for_snapshots(filesystem)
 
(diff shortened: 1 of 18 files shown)


src/lib/y2partitioner/filesystem_errors.rb
```diff
@@ -94,6 +94,7 @@ def installing?
     def small_size_for_snapshots?(filesystem, new_size: nil)
       return false unless filesystem && filesystem_with_snapshots?(filesystem)
 
+      # FIXME: Does this check make sense for multidevice Btrfs?
```
ancorgs: I don't think so. So I would probably disable it.

src/lib/y2partitioner/widgets/pages/btrfs_filesystems.rb
```diff
@@ -45,7 +45,7 @@ def initialize(filesystems, pager)
 
         # @macro seeAbstractWidget
         def label
-          Y2Storage::Filesystems::Type::BTRFS.to_human
+          _("Btrfs")
```
ancorgs: Why this change?

src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -114,6 +114,21 @@ def plain_blk_devices
...
+      # When the filesystem is a non-multidevice, this method simply returns the base
+      # name of the block device (e.g., "sda1"). And for multidevice ones, it only returns
+      # the base name of the first block device plus a "+" symbol to indicate that the
+      # filesystem is multidevice (e.g., "sda1+").
+      #
+      # @return [String]
+      def blk_device_basename
+        info = blk_devices.first.basename
+        info << "+" if multidevice?
+
```
ancorgs: Can you check the value of blk_devices.first.basename right after this `<<`?

joseivanlopez: Are you worried by the reference? Take into account that `BlkDevice#basename` always returns a new String object, so no problem.

ancorgs:
I know it's done for the sake of reusing existing widgets, but that grouping of "Mounting options" looks a little bit weird. Specially when the "Enable Snapshots" is not there. It would look more natural if the widgets are just there, without the frame.

src/lib/y2partitioner/widgets/pages/btrfs_filesystems.rb
joseivanlopez: Well, originally it was as it is now, but I changed it in my previous PR. But now, I have realized that maybe we need such Btrfs label in different places, so maybe it is better to have a translated string for it, instead of calling the long `Y2Storage::Filesystems::Type::BTRFS.to_human` everywhere.

PR #913, opened 2019-04-30
Fix "Type" column

dgdavid:
## Problem

After merge #911, we realized that it contained below undesired behaviors

* It was displaying a wrong icon for a Btrfs multidevice filesystem _(because it was using the default one)_.
* It was using a wrong device name for a multidevice filesystem.
* It was displaying the label for a device which is part of a multidevice.
* It was displaying the mount point for a device which is part of a multidevice.
* It was displaying the unfinished "PV of" string in certain situation, when the VG does not have a basename (yet).

## Solution

Fix all of them.

## Testing

- Tested manually

## Screenshots

<details>

---

| Before |
|-------|
| Screenshot_opensusetumbleweed_2019-04-30_14:58:55 |

| After |
| -------- |
| Screenshot_opensusetumbleweed_2019-04-30_16:51:55 |

---

| Before |
| ------ |
| Screenshot_opensusetumbleweed_2019-04-30_15:28:51 |

| After |
| -------- |
| Screenshot_opensusetumbleweed_2019-04-30_15:30:05 |

### DIFF
--- a/src/lib/y2partitioner/widgets/blk_devices_table.rb
+++ b/src/lib/y2partitioner/widgets/blk_devices_table.rb
@@ -197,6 +197,7 @@ module Y2Partitioner
       def mount_point_value(device)
         fs = filesystem(device)
         return "" if fs.nil?
+        return "" if part_of_multidevice?(device)
 
         res = fs.mount_path
         res += " *" if fs.mount_point && !fs.mount_point.active?
--- a/src/lib/y2partitioner/widgets/blk_devices_table.rb
+++ b/src/lib/y2partitioner/widgets/blk_devices_table.rb
@@ -220,7 +220,8 @@ module Y2Partitioner
         partition: Icons::HD_PART,
         raid:      Icons::RAID,
         lvm_vg:    Icons::LVM,
-        lvm_lv:    Icons::LVM_LV
+        lvm_lv:    Icons::LVM_LV,
+        btrfs:     Icons::BTRFS
       }
 
       # Table icon for the device
--- a/src/lib/y2partitioner/widgets/blk_devices_table.rb
+++ b/src/lib/y2partitioner/widgets/blk_devices_table.rb
@@ -145,7 +145,10 @@ module Y2Partitioner
 
       def device_value(device)
         if device.is?(:blk_filesystem)
-          device.type.to_human_string
+          device_name =  [device.type.to_human_string]
+          device_name << device.blk_device_basename if device.multidevice?
+
+          device_name.join(" ")
         else
           device.name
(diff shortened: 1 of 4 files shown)


src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -145,7 +145,10 @@ def end_title
 
       def device_value(device)
         if device.is?(:blk_filesystem)
-          device.type.to_human_string
+          device_name =  [device.type.to_human_string]
+          device_name << device.blk_device_basename if device.multidevice?
+
+          device_name.join(" ")
```
joseivanlopez: The order with translations is always tricky. I would protect us against it:

```suggestion
          if device.multidevice?
            format(_("%{type} %{device}"),
              type: device.type.to_human_string,
              device: device.blk_device_basename)
            else
              device.type.to_human_string
            end
```

src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -187,7 +187,9 @@ def type_value(device)
 
       def filesystem_label_value(device)
         fs = filesystem(device)
-        # fs may be nil or a file system not supporting labels, like NFS
+        return "" if fs.nil?
+        return "" if fs.respond_to?(:multidevice?) && fs.multidevice?
```
joseivanlopez: As commented in IRC, we have to show the label when the device is a multidevice filesystem.

src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -197,6 +197,7 @@ def filesystem_label_value(device)
       def mount_point_value(device)
         fs = filesystem(device)
         return "" if fs.nil?
+        return "" if fs.respond_to?(:multidevice?) && fs.multidevice?
```
joseivanlopez: The same here. Show it If the device is a multidevice Btrfs.

src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -398,7 +403,6 @@ def device_label_for(device)
       # Whether the device belongs to a multidevice filesystem
       #
       # @param device [Device]
-      # @param fs [Y2Storage::Filesystems::Base]
       # @return [Booelean]
```
joseivanlopez: Typo: Boo**e**lean

dgdavid: :man_facepalming:

PR #914, opened 2019-04-30
Partitioner: fix behavior of tree sections that were initially empty

ancorgs:
## Problem

In theory, the Partitioner should start with all the sections (like "Hard Disks", "Volume Management", etc.) open by default, so the first level of devices is visible.

But when a section was initially empty and the first device was added, the section remained closed. That's wrong. When the very first device is added, the default behavior (open displaying the first level of devices) should be applied.

The wrong behavior can be seen at
https://openqa.opensuse.org/tests/920618#step/partitioning_full_lvm/51

Reported as [bsc#1133686](https://bugzilla.suse.com/show_bug.cgi?id=1133686)

Trello: https://trello.com/c/XYcpfC8F/957-ostumbleweed-p5-1133686-partitioner-misbehavior-of-tree-sections-that-were-initially-empty

In addition, while fixing this we noticed the status of the tree was not saved in several places in which it should (like when creating a new RAID or a new VG) and that it was saved in places where it was not needed (like rendering a DeviceMenuButton).

## Solution

To fix the problem with the empty sections, now we only store the status of only those tree entries that have children. So we do not longer confuse a closed entry with an empty one.

To fix the calls to `#save_open_items` we have introduced a new mixin that is used by all the widgets that produce an UI redrawing, so they all share the code for saving the status.

## Testing

- Tested manually so far
- Added one more unit test (+0.01 coverage)

### DIFF
--- a/test/y2partitioner/ui_state_test.rb
+++ b/test/y2partitioner/ui_state_test.rb
@@ -352,7 +352,7 @@ describe Y2Partitioner::UIState do
     context "if the open items has not been saved" do
       before { described_class.create_instance }
 
-      it "returns and empty hasg" do
+      it "returns and empty hash" do
         expect(ui_state.open_items).to eq({})
       end
     end
--- a/test/y2partitioner/ui_state_test.rb
+++ b/test/y2partitioner/ui_state_test.rb
@@ -348,12 +348,12 @@ describe Y2Partitioner::UIState do
     end
   end
 
-  describe "#open_items_ids" do
+  describe "#open_items" do
     context "if the open items has not been saved" do
       before { described_class.create_instance }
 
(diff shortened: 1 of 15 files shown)


test/y2partitioner/ui_state_test.rb
```diff
@@ -348,29 +348,29 @@
...
   end
 
-  describe "#open_items_ids" do
+  describe "#open_items" do
     context "if the open items has not been saved" do
       before { described_class.create_instance }
 
-      it "returns nil" do
-        expect(ui_state.open_items_ids).to be_nil
+      it "returns and empty hasg" do
```
joseivanlopez: Typo: hasg

PR #915, opened 2019-05-02
Allow to create multi-device Btrfs

joseivanlopez:
NOTE FOR REVIEWERS: codeclimate strikes again. Please, mark it as wontfix (for some reason, I am not allowed to do it).

## Problem

The Expert Partitioner shows existing multi-device Btrfs, but there is no way to create a new one.

* https://trello.com/c/C7o44w3f/948-5-partitioner-adding-and-modifying-multi-device-btrfs

## Solution

Added button to the Expert Partitioner to create a new Btrfs. That button opens a wizard with two steps:

* step 1: allows to select metadata/data RAID levels, and the devices to use for the new Btrfs.
* step2: allows to select mount point, fstab options, subvolumes, etc.

Moreover, a button for editing the used devices was added. Right now, the devices can only be edited when the filesystem does no exist on disk yet. 

## Testing

- Added unit tests
- Tested manually

## Screenshots

<details>
<summary>Show screenshots</summary>

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_48_40

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_48_59

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_49_39

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_50_11

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_50_44

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_51_17

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_51_40

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_52_22

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_54_17

VirtualBox_openSUSE Tumbleweed_10_05_2019_08_54_59

</details>

### DIFF
--- a/src/lib/y2partitioner/actions/resize_blk_device.rb
+++ b/src/lib/y2partitioner/actions/resize_blk_device.rb
@@ -109,13 +109,13 @@ module Y2Partitioner
         return nil if using_devs.empty?
 
         format(
-          # TRANSLATORS: %{name} is replaced by a device name (e.g., /dev/sda1).
-          # and %{users} is replaced by a comma-separated list of name devices
-          # (devices using the first one).
-          _("The device %{name} is in use (%{users}).\n" \
+          # TRANSLATORS: %{name} is replaced by a device name (e.g., /dev/sda1) and %{users} is replaced
+          # by a list of name devices (devices using the first one).
+          _("The device %{name} is being used by:\n" \
+            "%{users}\n\n" \
             "It cannot be resized.\n" \
             "To resize %{name}, make sure it is not used."),
-          name: device.name, users: using_devs.join(", ")
+          name: device.name, users: using_devs.join("\n")
         )
       end
 
--- a/src/lib/y2partitioner/widgets/btrfs_devices.rb
+++ b/src/lib/y2partitioner/widgets/btrfs_devices.rb
@@ -48,9 +48,9 @@ module Y2Partitioner
           Left(
             HVSquash(
               HBox(
-                metadata_raid_level_widget,
+                data_raid_level_widget,
                 HSpacing(1),
-                data_raid_level_widget
+                metadata_raid_level_widget
               )
             )
           ),
@@ -72,11 +72,10 @@ module Y2Partitioner
       #
       # @return [Boolean]
       def validate
-        current_errors = errors
-        return true if current_errors.none?
+        error = error_message
+        return true unless error
 
-        message = current_errors.join("\n\n")
-        Yast2::Popup.show(message, headline: :error)
+        Yast2::Popup.show(error, headline: :error)
 
         false
       end
@@ -111,8 +110,9 @@ module Y2Partitioner
       #
       # @return [String]
       def help_for_raid_levels
+        # TRANSLATORS: help text
         _("<h4><b>Btrfs RAID levels</b></h4>" \
-          "<p> Btrfs supports the following RAID levels for both, metadata and data:" \
+          "<p>Btrfs supports the following RAID levels for both, data and metadata:" \
             "<ul>" \
               "<li>" \
                 "<b>DUP:</b> stores two copies of each piece of data on the same device. " \
@@ -147,30 +147,43 @@ module Y2Partitioner
       #
       # @return [String]
       def help_for_default_raid_level
+        # TRANSLATORS: help text
         _("<p>" \
-            "When <b>DEFAULT</b> RAID level is used, Btrfs will select a RAID level depending on " \
+            "When the default RAID level is used, Btrfs will select a RAID level depending on " \
             "whether the filesystem is being created on top of multiple devices or using only one " \
             "device. For a single-device Btrfs, the tool also will distinguish between rotational " \
             "or not-rotational devices to choose the default value." \
           "</p>")
       end
 
-      # Errors for the selected values
+      # Message to show when there are errors
       #
-      # @return [Array<String>]
-      def errors
-        [
-          metadata_devices_error,
-          data_devices_error
-        ].compact
-      end
+      # @return [String, nil] nil if there is no error.
+      def error_message
+        data_error = data_devices_error
+        metadata_error = metadata_devices_error
 
-      # Error when the selected metadata RAID level cannot be used (according to the number of selected
-      # devices)
-      #
-      # @return [String, nil] nil if there is no error
-      def metadata_devices_error
-        raid_level_devices_error(:metadata)
+        return nil unless data_error || metadata_error
+
+        error_by =
+          if data_error && metadata_error
+            # TRANSLATORS: error message
+            _("RAID levels do not match the amount of selected devices.")
+          elsif data_error
+            # TRANSLATORS: error message
+            _("The RAID level does not match the amount of selected devices.")
+          else
+            # TRANSLATORS: error message
+            _("The metadata RAID level does not match the amount of selected devices.")
+          end
+
+        # TRANSLATORS: error message, where %{error_by} is replaced by an initial sentence.
+        intro = format(
+          _("%{error_by}\nModify the list of devices or adjust the RAID levels."),
+          error_by: error_by
+        )
+
+        [intro, data_error, metadata_error].compact.join("\n\n")
       end
 
       # Error when the selected data RAID level cannot be used (according to the number of selected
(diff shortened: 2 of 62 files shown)


src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -75,6 +76,68 @@ class Btrfs < BlkFilesystem
...
       storage_forward :configure_snapper=
 
+      # @!method metadata_raid_level
+      #
+      #   Gets the metadata RAID level
+      #
+      #   @return [BtrfsRaidLevel]
+      storage_forward :metadata_raid_level, as: "BtrfsRaidLevel"
+
+      # @!method metadata_raid_level=(level)
```
ancorgs: I'm not sure if this will look right in the rendered yardoc. Take a look at how these combinations of setter+getter are documented in other places using `@!attribute`. For example in `blk_filesystem.rb`.

src/lib/y2partitioner/widgets/btrfs_devices.rb
```diff
@@ -0,0 +1,228 @@
...
+
+      # @macro seeCustomWidget
+      def contents
+        VBox(
+          Left(
+            HVSquash(
+              HBox(
+                metadata_raid_level_widget,
+                HSpacing(1),
+                data_raid_level_widget
```
ancorgs: Maybe it's just me, but I find this to be in the wrong order. I would expect:

```
Data RAID Level   -  Metadata RAID Level
```

Instead of the other way around.

ancorgs: I would even consider going further, with something like:

```
RAID Level  -  RAID Level for Metadata
```

src/lib/y2partitioner/widgets/btrfs_devices.rb
```diff
@@ -0,0 +1,228 @@
...
+      def btrfs_devices_selector_widget
+        @btrfs_devices_widget ||= Widgets::BtrfsDevicesSelector.new(controller)
+      end
+
+      # Btrfs RAID levels help
+      #
+      # @return [String]
+      def help_for_raid_levels
+        _("<h4><b>Btrfs RAID levels</b></h4>" \
+          "<p> Btrfs supports the following RAID levels for both, metadata and data:" \
```
ancorgs: Any reason for the white space right after `<p>`?

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -491,6 +494,24 @@ def setup_default_btrfs_subvolumes
...
+      #
+      # @return [String, nil]
+      def display_name
+        return nil unless multidevice?
+
+        textdomain "storage"
+
+        # TRANSLATORS: display name when the Btrfs is multidevice
+        format(
+          _("Btrfs %{name} over %{num_devices} devices"),
```
ancorgs: The "+" sign makes this sentence hard to read

```
Btrfs sda1+ over 2 devices
```

I would prefer an alternative which doesn't imply that plus sign being in the middle. For example:

```
Btrfs over 2 devices (sda1+)
```

src/lib/y2storage/lvm_pv.rb
```diff
@@ -55,6 +58,27 @@ def plain_blk_device
...
+    # Only orphan PV has its own representation.
+    #
+    # @return [String, nil]
+    def display_name
+      return nil unless orphan?
+
+      textdomain "storage"
+
+      # TRANSLATORS: display name when the PV has no associated VG
+      format(_("Orphan PV on %{device}"), device: plain_blk_device.name)
```
ancorgs: Not sure if "Orphan PV" would mean something to most people out there. What about "Unused LVM PV on %s"?

src/lib/y2partitioner/widgets/btrfs_devices.rb
joseivanlopez: I haven't a strong preference. I would adapt it according to your last suggestion.

src/lib/y2partitioner/widgets/btrfs_devices.rb
joseivanlopez: Leftover.

src/lib/y2storage/filesystems/btrfs.rb
joseivanlopez: And something like:

```
Btrfs over 2 devices (sda1 ...)
```

src/lib/y2storage/lvm_pv.rb
joseivanlopez: New column type is using "Orphan PV". Should we "Unused LVM PV" instead?

src/lib/y2partitioner/widgets/btrfs_devices.rb
joseivanlopez: Then, should we reorder it in the Overview description? Right now, metadata comes first.

src/lib/y2storage/lvm_pv.rb
ancorgs: I think so

src/lib/y2storage/filesystems/btrfs.rb
ancorgs: ok

src/lib/y2partitioner/widgets/btrfs_devices.rb
ancorgs: Yes, it's also counterintuitive there IMO.

src/lib/y2partitioner/actions/resize_blk_device.rb
```diff
@@ -112,10 +112,11 @@ def used_device_error
           # TRANSLATORS: %{name} is replaced by a device name (e.g., /dev/sda1).
           # and %{users} is replaced by a comma-separated list of name devices
```
ancorgs: Outdated comment for the translators, this is not longer comma-separated

src/lib/y2partitioner/widgets/btrfs_devices.rb
```diff
@@ -147,6 +147,7 @@ def help_for_raid_levels
       #
       # @return [String]
       def help_for_default_raid_level
+        # TRANSLATORS: help text
         _("<p>" \
             "When <b>DEFAULT</b> RAID level is used, Btrfs will select a RAID level depending on " \
```
ancorgs: This is not longer DEFAULT in upcase. This would be now more like "When the default RAID level is used,..."

src/lib/y2partitioner/widgets/btrfs_devices.rb
```diff
@@ -155,30 +156,50 @@ def help_for_default_raid_level
...
+
+        return nil unless data_error || metadata_error
+
+        error_by =
+          if data_error && metadata_error
+            # TRANSLATORS: error message
+            _("RAID levels do not match the amount of selected devices.")
+          elsif data_error
+            # TRANSLATORS: error message
+            _("The data RAID level does not match the amount of selected devices.")
```
ancorgs: Since we removed the word "data" from the labels, I would do the same here.

src/lib/y2storage/btrfs_raid_level.rb
```diff
@@ -27,15 +28,37 @@ module Y2Storage
...
+    # Returns human readable representation of enum which is already translated.
+    #
+    # @raise [RuntimeError] when called on enum value for which translation is not defined yet.
     #
     # @return [String]
     def to_human_string
-      Storage.btrfs_raid_level_name(to_storage_value)
+      textdomain "storage"
+
+      value = TRANSLATIONS[to_sym] or raise "Unhandled Btrfs RAID level value '#{inspect}'"
```
ancorgs: I kind of fear that `raise`. Do we have a unit test or some other mechanism to detect unhandled RAID levels before they hit the user?

src/lib/y2storage/btrfs_raid_level.rb
```diff
@@ -27,15 +28,37 @@ module Y2Storage
...
+    extend Yast::I18n
+
     include StorageEnumWrapper
 
     wrap_enum "BtrfsRaidLevel"
 
-    # Returns human readable representation of enum which is already translated
+    TRANSLATIONS = {
+      default: N_("Default"),
+      single:  N_("SINGLE"),
```
ancorgs: Likely `SINGLE` and `DUP` deserve a comment for the translators. Something like "This is one of the RAID modes used by Btrfs, likely to be left as-is".

src/lib/y2storage/btrfs_raid_level.rb
joseivanlopez: No, we don't. And the same for other things like MdParity.

joseivanlopez: I have unit test to ensure everything is translated, and eureka! we have no translation for "UNKNOWN" RAID level :P

ancorgs: Will you create a separate pull request for that?

ancorgs: I see the fix is part of this PR already. Never mind.

PR #919, opened 2019-05-09
Regression unit test for bsc#1134330

ancorgs:
## Problem

AutoYaST fails when trying to reuse a non-formatted partition as-is (i.e. without formatting it). By definition, it means AutoYaST can hardly be used to reuse BIOS BOOT partitions.

- See https://bugzilla.suse.com/show_bug.cgi?id=1134330

## Solution

This pull request includes only a unit test showcasing the problem. Solution to be implemented.

## Testing

- Added regression unit test.


ancorgs:
Closed in favor of https://github.com/yast/yast-storage-ng/pull/922

PR #921, opened 2019-05-10
proposal frees space from existing multidevice btrfs properly (bsc#1096760)

wfeldt:
### Problem

Proposal does not delete all partitions constituting a multidevice btrfs but only those that are actually needed (step by step) to fulfill the requirements. That's in contrast to LVM where all partitions are deleted.

This PR unifies the handling across LVM, RAID, and btrfs.

- Scrum board: https://trello.com/c/Dv1iuLnb

### Solution

If a partition is deleted that is part of some common multidevice storage object (lvm, raid, btrfs), all partitions in the candidate disks constituting that object are deleted.

### Notes

- This handles only partitions. If whole disks are part of a multidevice storage object, these are not automatically reused - see Scrum board for an example where it makes a difference.
- As the implementation of these 3 seemingly similar cases shows, our classes lack a bit of uniformity.

### Testing

- unit tests
- manual integrations tests


ancorgs:
I remember the original reason to not unify RAID and LVM back then was the way in which both technologies (and some more) could be nested to produce situations that were tricky. I even had some examples and counterexamples and some kind of recursive prototype... but I cannot find any of that anymore (is likely in my office's computer).

wfeldt:
Well, you specifically asked for a uniform treatment of these 3 cases. In any case this is about raid composed of partitions which I hoped would rule out the tricky cases.

src/lib/y2storage/proposal/partition_killer.rb
```diff
@@ -137,33 +133,78 @@ def last_logical?(partition)
...
+
+        log.info "These partitions will be deleted: #{target_partitions.map(&:name)}"
         target_partitions.map { |p| delete_partition(p) }.flatten
       end
 
+      # Checks whether the partition is part of a multidevice storage object
+      #
+      # @param partition [Partition]
+      # @return [Boolean]
+      def related_partitions?(partition)
```
ancorgs: Where is this method used? Looks like a leftover to me.

wfeldt: Yep.

ancorgs:
Although I do not remember the details, I guess I abandoned my previous attempt to unify RAID and LVM because I was going too far, supporting scenarios like:

Part1 is a PV of VG1
Part2 and Part3 form MD1
MD1 is another PV of VG1

When deleting Part2 I was trying to remove Part3 (because it's part of the same MD) and also Part1 (because killing the MD did kill the PV, which made Part1 useless).

I can tell you that can end-up in all kind of corner cases so I gave up.

The simpler (and more realistic) approach implemented on this PR is indeed good enough for 99.9% cases, so LGTM (except the leftover method).

wfeldt:
I had discussed briefly with Arvin about a doing this purely abstract on the graph deleting a node and all nodes above it that then no longer have any descendants.

You'd first have to introduce a `deletable?` flag to not accidentally delete physical things (like disks) - but the real problem is that there's no uniform `delete` method for every object. Instead you have to use some specialized methods some times - which is a pity.

ancorgs:
While reviewing this PR I noticed that PartitionKiller is also used by the AutoInst proposal here:

https://github.com/yast/yast-storage-ng/blob/05ad85f9ed458ebc2d4528737d57daf8930ed3d2/src/lib/y2storage/proposal/autoinst_space_maker.rb#L142

Which can be dangerous if any of the related partitions are marked to be reused (create = false) in the profile. It looks to me like it would be very easy to write a testcase that makes the AutoInst proposal go nuts by just deleting one of the partitions in a LVM (or RAID) and reformatting the other. Isn't it?

PR #922, opened 2019-05-10
Fix bug#1134330 in SLE-15-GA

ancorgs:
## Problem

AutoYaST fails when trying to reuse a non-formatted partition as-is (i.e. without formatting it). By definition, it means AutoYaST can hardly be used to reuse BIOS BOOT partitions.

- See https://bugzilla.suse.com/show_bug.cgi?id=1134330

## Solution

This fixes two issues:

- First of all, the "missing reusable filesystem" issue was always reported twice due to a code duplication.
- More important, this avoids false positives when reporting such error. If the partition is not intended to be mounted and no filesystem type is specified, is probably ok if the partition to be reused is not currently formatted.

## Testing

- Regression unit test added

## Note for reviewer

Review commit by commit, it makes more sense.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri May 10 14:59:50 UTC 2019 - ancor@suse.com
+
+- AutoYaST: do not ask for a reusable filesystem when it's not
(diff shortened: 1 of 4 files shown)


kobliha:
@ancorgs Approved here and also tested by the customer. You can merge now, I believe :) Thx

PR #923, opened 2019-05-13
Unit test for reusing partitions of a former LVM

ancorgs:
## Problem

While discussing #921 we realized that using `Proposal::PartitionKiller` as-is in the `AutoInstProposal` code was not a good idea. `PartitionKiller` deletes some extra partitions to make more space. That can be wrong for some AutoYaST profiles.

## Solution

This PR provides a failing unit test that showcases the problem, which is relatively easy to trigger.

### DIFF
--- a/test/data/devicegraphs/lvm-two-disks.yml
+++ b/test/data/devicegraphs/lvm-two-disks.yml
@@ -1,0 +1,37 @@
+---
+- disk:
+    name: /dev/sda
+    size: 200 GiB
+    partition_table:  gpt
+    partitions:
+
+    - partition:
+        size:         unlimited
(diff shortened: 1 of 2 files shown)


wfeldt:
Also, LDL DASDs with their implicit partitions might need special treatment.

ancorgs:
Just for the record, implicit partition already have their own treatment (they are emptied instead of deleted). See https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/proposal/partition_killer.rb#L83

joseivanlopez:
Closing in favor of https://github.com/yast/yast-storage-ng/pull/937

PR #932, opened 2019-06-19
Separate vg fixes

ancorgs:
## Problem

The test coverage for #931 was intentionally simplistic since it makes little sense to test all the small details before getting feedback. Unfortunately that leaded to a bug been overlooked. The proposal managed to succeed when combining the "system" VG with some dedicated ones just by a matter of luck (and producing a clearly sub-optimal solution). It was in fact getting confused about some PVs.

## Solution

Fix the PV detection to have into account the case of many planned VGs.

### DIFF
--- a/src/lib/y2storage/proposal/phys_vol_strategies/base.rb
+++ b/src/lib/y2storage/proposal/phys_vol_strategies/base.rb
@@ -127,8 +127,7 @@ module Y2Storage
           @potential_part_sizes[partition] ||= space.usable_extra_size + partition.min_size
         end
 
-        # Planned partition that will hold an LVM PV of the planned physical
-        # volume
+        # Planned partition that will hold an LVM PV of the planned volume group
         #
         # @param space [Planned::AssignedSpace]
         # @return [Planned::Partition]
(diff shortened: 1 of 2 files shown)


src/lib/y2storage/proposal/phys_vol_strategies/base.rb
```diff
@@ -127,12 +127,15 @@ def potential_partition_size(partition, space)
           @potential_part_sizes[partition] ||= space.usable_extra_size + partition.min_size
         end
 
-        # Planned partition that will hold an LVM PV
+        # Planned partition that will hold an LVM PV of the planned physical
+        # volume
```
joseivanlopez: "of the planned volume group", isn't it?

ancorgs:
This is not going to be used until we build the UI, so I don't think a version bump is really needed.

src/lib/y2storage/proposal/phys_vol_strategies/base.rb
ancorgs: Ups, fixed.

PR #934, opened 2019-06-26
New rubocop

jreidinger:
## Problem

*Short description of the original problem.*

- https://bugzilla.opensuse.org/show_bug.cgi?id=1139270


src/lib/y2partitioner/actions/controllers/clone_partition_table.rb
```diff
@@ -71,7 +71,8 @@ def suitable_devices_for_cloning?
         #
         # @return [Array<Y2Storage::Partitionable>]
         def suitable_devices_for_cloning
-          @suitable_devices ||= working_graph.blk_devices.select { |d| suitable_for_cloning?(d) }
+          @suitable_devices_for_cloning ||= working_graph.blk_devices
```
joseivanlopez: In general, in this commit, why are all those renames needed?

src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -245,10 +245,11 @@ def reuse_device!(device)
 
       # @param filesystem [Filesystems::Base]
       def assign_mount_point(filesystem)
-        if mount_point && !mount_point.empty?
-          filesystem.mount_path = mount_point
-          filesystem.mount_point.mount_by = mount_by if mount_by
-        end
+        return unless mount_point
```
joseivanlopez: NP: why not this?

```
return if !mount_point || mount_point.empty?
```

.rubocop.yml
```diff
@@ -16,9 +16,10 @@ SingleLineBlockParams:
 Metrics/LineLength:
   Max: 105
 
-# Enforce if/unless at the end only for really short lines
-Style/IfUnlessModifier:
-  MaxLineLength: 60
+Metrics/BlockLength:
+  # rspec is know for big blocks as DSL
```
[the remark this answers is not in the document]
jreidinger: well, this metric check length of block. so anything starting with do/end how long it is. and in rspec whole you code is describe xxx do/end, so basically whole test. I think rubocop by default skip this cop for files ending with `.spec` but we use `.rb` suffix so I have to disable it manually.

src/lib/y2partitioner/actions/controllers/clone_partition_table.rb
jreidinger: it is new cop, that check for consistent caching of method call. This cop check that cache is kept in variable that has same name as method which result is cached. ( I hope I describe it clear ).

jreidinger: and link to cop https://rubocop.readthedocs.io/en/stable/cops_naming/#namingmemoizedinstancevariablename

src/lib/y2storage/proposal/lvm_helper.rb
```diff
@@ -105,11 +105,12 @@ def reusable_volume_groups(devicegraph)
       #   nil when no volume group will be reused.
       #
       # @see volume_group
+      # FIXME: return value of writer is really strange. It should be same as argument.
```
joseivanlopez: I agree. Maybe we can rename the method as `reuse_volume_group(vg)`

src/lib/y2storage/autoinst_issues/multiple_bcache_members.rb
```diff
@@ -67,9 +65,9 @@ def severity
       def message
         # TRANSLATORS: 'bcache_name is the bcache device name (e.g., '/dev/bcache0');
         # 'role' is the kind of device (e.g., 'caching device').
-        _(
-          "%{bcache_name}: only one %{role} can be specified per bcache."
-        ) % { bcache_name: bcache_name, role: _(ROLE[role]) }
+        format(_(
```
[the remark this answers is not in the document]
jreidinger: you mean white spaces? and new lines? rubocop is not against, this is just automatic change of original code that it tries to keep as it was.

src/lib/y2storage/planned/can_be_formatted.rb
jreidinger: well, both is possible and rubocop does not prevent it. I ( so personal taste ) just usually prefer having simple exit criteria. So return if nil and return if empty.

src/lib/y2storage/proposal/lvm_helper.rb
jreidinger: in fact see what I write to email to yast-devel. return value is ignored, so in the end, I simply remove return value completely.

src/lib/y2storage/autoinst_issues/multiple_bcache_members.rb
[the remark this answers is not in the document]
joseivanlopez: Yes, I know, but I would write each param in a new line for the sake of readability, and also the closing bracket in a new line.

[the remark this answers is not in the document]
jreidinger: you can(!) write it, just currently I do not know way how to force it in rubocop.

src/lib/y2partitioner/actions/controllers/clone_partition_table.rb
joseivanlopez: Ok, now I got it. Thanks!

src/lib/y2storage/planned/can_be_formatted.rb
joseivanlopez: Ok, yes, a matter of taste. In my case, I prefer to have only one line breaking the method workflow. And when the condition is too complex (IMHO in this case is easy enough) I tend to create a question method:

```
return if empty_mount_point?(mount_point)
```

But I am also fine with your proposal.

.rubocop.yml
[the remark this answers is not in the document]
joseivanlopez: Ok, thanks for the explanation. What do you thing about adding a TODO to rename test files as spec files? And there is a typo: "rspec is known .."

src/lib/y2storage/autoinst_issues/multiple_bcache_members.rb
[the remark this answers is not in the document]
joseivanlopez: Yes, I mean to modify it manually, because the rubocop proposal is not the most readable (IMHO). But not a big deal, you leave it as it is ;)

.rubocop.yml
```diff
@@ -28,3 +25,21 @@ Style/ZeroLengthPredicate:
 # the ".freeze" attribute for the constants is not nice
 Style/MutableConstant:
   Enabled: false
+
+# this failing in storage wrapper, will need more time to adapt
```
joseivanlopez: NP: typo "this is failing ..., it will need ..."

joseivanlopez:
I only miss a new version number and the changelog entry mentioning the related bug number. Otherwise, LGTM.

Many thanks for adapting it! Quite a good improvement.

PR #935, opened 2019-07-03
Add control file option to delete and resize partitions

joseivanlopez:
NOTE FOR REVIEWERS: please, mark codeclimate complaints as wontfix. For some reason, I cannot do it by myself.

## Problem

SUMA uses its particular wizard for the storage proposal. During the installation, SUMA always removes all existing partitions from the target devices, so it does not require to ask the user what to do with the existing partitions.

In the future, more products could have similar expectations about allowing the user to decide what to do with existing partitions. So, the control file should have a specific setting to indicate whether the user can modify the setting about deleting and resizing partitions.  

* https://trello.com/c/9T83F9me/1096-3-suma-proposal-non-configurable-delete-modes-and-resizewindows
* https://jira.suse.com/browse/SLE-7238
* Related card: https://trello.com/c/XnJnPdw6/1092-5-suma-proposal-new-wizard-implies-change-in-proposalsettings

## Solution

Added new option `delete_resize_configurable` in the control file. The user cannot modify the delete and resize options when `delete_resize_configurable` is set to false.

And now, if there is only one candidate disk for the installation and the value of this new option is `false`, the second step of the Guided Setup is skipped.

Bonus: widgets to select the delete/resize modes have been extracted to separated classes to make easy to reuse them in other dialogs (e.g., SUMA dialogs).

## Testing

* Added unit test
* Tested manually

### DIFF
--- a/src/lib/y2storage/dialogs/guided_setup/helpers/disk.rb
+++ b/src/lib/y2storage/dialogs/guided_setup/helpers/disk.rb
@@ -12,10 +12,10 @@
 # more details.
 #
 # You should have received a copy of the GNU General Public License along
-# with this program; if not, contact Novell, Inc.
+# with this program; if not, contact SUSE LLC.
 #
-# To contact Novell about this file by physical or electronic mail, you may
-# find current contact information at www.novell.com.
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
 
 require "yast"
 require "y2storage"
--- a/src/lib/y2storage/dialogs/guided_setup/helpers/disk.rb
+++ b/src/lib/y2storage/dialogs/guided_setup/helpers/disk.rb
@@ -1,0 +1,94 @@
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact Novell, Inc.
+#
+# To contact Novell about this file by physical or electronic mail, you may
+# find current contact information at www.novell.com.
+
+require "yast"
+require "y2storage"
+
+module Y2Storage
+  module Dialogs
+    class GuidedSetup
+      module Helpers
+        # Helper class to generate the label of a disk
+        class Disk
+          include Yast::I18n
+
+          # Constructor
+          #
+          # @param analyzer [Y2Storage::DiskAnalayzer]
+          def initialize(analyzer)
+            textdomain "storage"
+
+            @analyzer = analyzer
+          end
+
+          # Disk label used by dialogs
+          #
+          # The label has the form: "NAME, SIZE, [USB], INSTALLED_SYSTEMS".
+          #
+          # Examples:
+          #
+          #   "/dev/sda, 10.00 GiB, Windows, OpenSUSE"
+          #   "/dev/sdb, 8.00 GiB, USB"
+          #
+          # @return [String]
+          def label(disk)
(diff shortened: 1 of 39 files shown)


src/lib/y2storage/dialogs/guided_setup/helpers/disk.rb
```diff
@@ -0,0 +1,94 @@
...
+          #
+          # @return [String]
+          def label(disk)
+            data = [disk.name, disk.size.to_human_string]
+            data += type_labels(disk)
+            data += analyzer.installed_systems(disk)
+            data.join(", ")
+          end
+
+          private
```
imobachgs: np: Are not we "out-denting" this `private` word?

src/lib/y2storage/dialogs/guided_setup/widgets/base.rb
```diff
@@ -0,0 +1,114 @@
...
+          end
+
+          # Whether the widget is currently enabled
+          #
+          # @return [Boolean]
+          def enabled?
+            Yast::UI.QueryWidget(Id(widget_id), :Enabled)
+          end
+
+          private
```
imobachgs: np: outdent?

src/lib/y2storage/dialogs/guided_setup/widgets/linux_partition_actions.rb
```diff
@@ -0,0 +1,87 @@
...
+          alias_method :enable_on_init?, :enable_on_init
+
+          # Constructor
+          #
+          # @param widget_id [String]
+          # @param settings [Y2Storage::ProposalSettings]
+          # @param enabled [Boolean]
+          def initialize(widget_id, settings, enabled: true)
+            super(widget_id, settings)
+
```
imobachgs: np: why these blank lines?

src/lib/y2storage/dialogs/guided_setup/widgets/partition_actions.rb
```diff
@@ -0,0 +1,176 @@
...
+            @other_actions_widget ||=
+              OtherPartitionActions.new("#{widget_id}_other", settings, enabled: other_actions?)
+          end
+
+          # Adds vertical spacing between the widgets
+          #
+          # @param widgets [Array<Yast::Term>]
+          # @return [Array<Yast::Term>]
+          def add_spacing(widgets)
+            widgets.flat_map { |w| [w, VSpacing(1)] }.tap(&:pop)
```
imobachgs: nice trick :+1:

src/lib/y2storage/dialogs/guided_setup/widgets/root_disk_selector.rb
```diff
@@ -0,0 +1,177 @@
...
+                "This is also the disk where boot-related partitions " \
+                "will typically be created as necessary: /boot, ESP (EFI System " \
+                "Partition), BIOS-Grub. " \
+                "That means that this disk should be usable by the machine's " \
+                "BIOS / firmware." \
+              "</p>"
+            )
+          end
+
+          private
```
imobachgs: outdent?

src/lib/y2storage/dialogs/guided_setup/helpers/disk.rb
joseivanlopez: Yes, new version of rubocop prefers this way.

src/lib/y2storage/dialogs/guided_setup/widgets/linux_partition_actions.rb
joseivanlopez: Only a matter of taste. I tend to separate lines that are not related at all.

src/lib/y2storage/dialogs/guided_setup/helpers/disk.rb
jreidinger: well, in fact it is not that new version of rubocop prefers it, but new version has a bug that cause endless loop. For more details see comment in .rubocop.yml - https://github.com/yast/yast-storage-ng/blob/master/.rubocop.yml#L33

BTW no other module need it so far, as other modules does not have access modifier like 

```
  class A
  private
    ...
  end
```
which cause bug.

joseivanlopez: Ah, ok. Thanks for the info. BTW, I personally prefer this version without out-denting.

PR #937, opened 2019-07-08
AutoYaST: only delete indicated partitions

joseivanlopez:
## Problem

The Guided Proposal and AutoYaST use the `PartitionKiller` to remove partitions. When the `PartitionKiller` is told to delete a partition, it removes such partition and all the related partitions. For example, when the specified partition is a LVM PV, the rest of PVs are removed too. And something similar happens when the partition belongs to a MD RAID or to a multi-device Btrfs.

Such behavior may be useful for the Guided Proposal, but it could be a problem for AutoYaST, see https://github.com/yast/yast-storage-ng/pull/923. So, for the case of AutoYaST, the `PartitionKiller` 
should remove only the indicated partitions and nothing else.

* https://trello.com/c/9AZD9sn6/1054-2-autoyast-storage-ng-avoid-collateral-casualties-of-partitionkiller
* Related PR: https://github.com/yast/yast-storage-ng/pull/921
* Related bug: https://bugzilla.suse.com/show_bug.cgi?id=1096760

## Solution

Now, an extra parameter can be passed to the `PartitionKiller` to indicate if the related partitions should be removed. For AutoYaST, this parameter is used, and now the rest of LVM PVs, MD devices or multi-device Btrfs devices are not removed.

## Testing

* Added unit tests

### DIFF
--- a/src/lib/y2storage/proposal/partition_killer.rb
+++ b/src/lib/y2storage/proposal/partition_killer.rb
@@ -49,11 +49,15 @@ module Y2Storage
       #
       # @param device_sid [Integer] device sid of the partition
       # @return [Array<Integer>] device sids of all the deleted partitions
-      def delete_by_sid(device_sid)
+      def delete_by_sid(device_sid, delete_related_partitions: true)
         partition = find_partition(device_sid)
         return [] unless partition
 
-        delete_with_related_partitions(partition)
+        if delete_related_partitions
+          delete_with_related_partitions(partition)
+        else
+          delete_partitions([partition])
+        end
       end
 
       protected
--- a/src/lib/y2storage/proposal/partition_killer.rb
+++ b/src/lib/y2storage/proposal/partition_killer.rb
@@ -1,8 +1,4 @@
-#!/usr/bin/env ruby
-#
-# encoding: utf-8
-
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/proposal/partition_killer.rb
```diff
@@ -49,11 +49,15 @@ def initialize(devicegraph, disks = nil)
       #
       # @param device_sid [Integer] device sid of the partition
       # @return [Array<Integer>] device sids of all the deleted partitions
-      def delete_by_sid(device_sid)
+      def delete_by_sid(device_sid, delete_related_partitions: true)
```
jreidinger: please document new parameter

jreidinger: and also missing test for that new parameter. ( see only autoinst related tests )

joseivanlopez: Ok, I have improved the documentation.

Regarding the test, we don't have specific tests for `PartitionKiller` class. This is a very internal component, so it is not directly tested. Its behavior is indirectly tested with the proposal tests. And the same happens with the new parameter. The AutoYaST proposal tests are checking that the parameter is actually used.

PR #938, opened 2019-07-08
Disable enlarge swap for suspend in s390 by default

joseivanlopez:
## Problem

The control file allows to set the setting `adjust_by_ram` for each volume. This setting ensures that the volume size is, at least, as bigger as the RAM size. This setting is specially useful for the swap volume to suspend to disk.

The control file of some products (e.g., [SLE](https://github.com/yast/skelcd-control-SLES/blob/master/control/installation.SLES.xml#L153)) sets this setting to `true` for the swap volume. But there are some architectures (Z Systems) where it is not possible to resume from swap. In this case, it does not make sense to enlarge swap by default.
 
* https://trello.com/c/xNInsz6m/1088-1-storage-enlarge-to-ram-set-to-false-by-default-for-s390
* Related to: https://jira.suse.com/browse/SLE-6926

## Solution

Now, the setting `adjust_by_ram` is forced to `false` for the swap volume of Z Systems, even when the control file sets the value to `true`.

## Testing

* Added unit tests


src/lib/y2storage/storage_manager.rb
```diff
@@ -40,14 +41,13 @@ module Y2Storage
   # there are quite some code that could be extracted to a new place, mainly all
   # stuff related to testing (e.g., {#probe_from_yaml}).
   #
-  class StorageManager
+  class StorageManager # rubocop:disable ClassLength
```
jreidinger: hmm, this show some design problems with this class. It kind of becoming god object, which is not good. Also file is already 700 lines long, which is quite lot.

Quickly going over this class few hints what need not to be in this class:

1. logger and its setting. It should be separate file like e.g. logger for cheetah - https://github.com/yast/yast-yast2/blob/66bc64d08c6ff017b88e3cff63fd02d78bbcf7f1/library/system/src/lib/yast2/execute.rb#L43 so it lives in its own file and you just need to require it.

2. create_test_instance. It can live outside of production code. Ideally in tests in separate file.

3. configure_snapper? feels a bit meh in StorageManager. I expect it in some kind of proposal or installation specific class.

4. probe_from_* feels a bit strange as I expect to be done outside and just assign probe graph to manager

In general issue of this class is that it has too big scope, from low level operations like commit, access to probed, staging, probe to quite high level operations.

So I prefer to ideally address it or at least add there TODO with plan how to handle this situation.

test/spec_helper.rb
```diff
@@ -87,6 +87,17 @@ def require(path)
...
+
+      # If :storage_arch is defined, the Storage::Arch object is used instead of Yast::Arch.
+      if respond_to?(:storage_arch)
+        arch = Y2Storage::Arch.new(storage_arch)
+
+        allow(Y2Storage::Arch).to receive(:new).and_return(arch)
+
+        allow(storage_arch).to receive(:x86?).and_return(architecture == :x86)
+        allow(storage_arch).to receive(:ppc?).and_return(architecture == :ppc)
+        allow(storage_arch).to receive(:s390?).and_return(architecture == :s390)
```
jreidinger: why is not there all storage forwarded methods?

src/lib/y2storage/storage_manager.rb
joseivanlopez: Totally agree. In fact, this class and the filesystem controller deserve a refactoring card. @ancorgs is there any plan for that?

Anyway, I would address it with a separate card.

test/spec_helper.rb
joseivanlopez: These are the "architecture" methods. The other two (:efiboot? and :ppc_power_nv?) do not define the architecture itself, but properties related to the architecture. For those two properties, each tests mock them if it is needed, for example:

```
let(:architecture) { :x86 }

before do 
  allow(:storage_arch).to receive(:efiboot?).and_return(true)
end
```

jreidinger: so it does not make sense to have some default like false here? to avoid call to storage and potential arch dependent test?

src/lib/y2storage/storage_manager.rb
ancorgs: No concrete card for `StorageManager` so far, but feel free to create one.

joseivanlopez: Ok, I have created this card: https://trello.com/c/Tg1UvmmZ/385-storagemanager-class-needs-refactoring

PR #944, opened 2019-07-10
Complain if /boot is in a Thinly Provisioned LVM or a BCache

shundhammer:
## Trello

https://trello.com/c/DKVCXCdz/

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1134130

## Problem

Grub2 cannot boot from a thinly provisioned LVM or from bcache.

## Fix

Check if /boot is on a thinly provisioned LVM or a bcache and complain, i.e. add an error to the error list of the boot requirements analyzer.

That will lead to an error popup in the partitioner when the user tries to set up such a scenario (the proposal will not try that).

## Background: Thin LVM

An LVM LV can be a _thin pool_; then it can pretend to supply more disk space than is actually available (i.e. disk space can be _overcommitted_). In that setup, this _thin pool_ LV can have one or more LVM LV children that can actually have filesystems (they are called _thin_; don't confuse this with the _thin pool_):

- LVM VG (volume group)
  -  _thin pool_ LVM LV (logical volume)
      - _thin_ LVM LV (logical volume)
        - filesystem
      - _thin_ LVM LV (logical volume)
        - filesystem

A normal LVM LV does _not_ have LVM LV children; it has only a filesystem.


src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -468,6 +488,46 @@ def in_lvm?(device)
...
+      # @param device [Filesystems::Base, Planned::Device, nil]
+      # @return [Boolean] false if device is nil
+      def in_thin_lvm?(device)
+        return false if device.nil?
+
+        if device.is_a?(Planned::Device)
+          device.is_a?(Planned::LvmLv) && device.lv_type == LvType::THIN
+        else
+          # If this is not a BlkFilesystem (e.g. NFS), it can't be on thin LVM
+          return false unless device.respond_to?(:plain_blk_devices)
```
joseivanlopez: NP: why not this? `return false unless device.is?(:blk_filesystem)`

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -468,6 +488,46 @@ def in_lvm?(device)
...
+      # @param device [Filesystems::Base, Planned::Device, nil]
+      # @return [Boolean] false if device is nil
+      def in_bcache?(device)
+        return false if device.nil?
+
+        if device.is_a?(Planned::Device)
+          device.is_a?(Planned::Bcache)
+        else
+          # If this is not a BlkFilesystem (e.g. NFS), it can't be in a BCache
+          return false unless device.respond_to?(:plain_blk_devices)
```
joseivanlopez: Same comment here.

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
shundhammer: Because Ruby fans insist on "duck typing"; just because right now we only have that method for BlkFilesystem that doesn't necessarily mean that one day we won't extend the concept for other classes. That `unless` clause really only defends against a _no method_ error. If the method is there, by all means let's use it. And that's something different than explicitly checking a specific class.

I realize that this is not exactly what the comment in the code says; that comment was meant more as a practical example just _when_ the check is necessary. I myself learned that the hard way when the NFS test cases started failing. ;-)

PR #948, opened 2019-07-30
fixed exception when removing devices from btrfs

aschnell:
## Problem

The BlkDeviceRestorer tries to add a device that already exists.

- https://bugzilla.suse.com/show_bug.cgi?id=1142669

## Solution

Do not try to add a device that already exists.

## Bonus

The selected devices in the dialog for LVM VGs and btrfses are sorted by name now.

### DIFF
--- a/test/y2partitioner/blk_device_restorer_test.rb
+++ b/test/y2partitioner/blk_device_restorer_test.rb
@@ -1,0 +1,44 @@
+#!/usr/bin/env rspec
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require_relative "test_helper"
+
+require "y2partitioner/blk_device_restorer"
+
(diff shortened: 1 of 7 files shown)


test/y2partitioner/blk_device_restorer_test.rb
```diff
@@ -0,0 +1,44 @@
...
+
+describe Y2Partitioner::BlkDeviceRestorer do
+
+  let(:devicegraph) { "trivial_btrfs" }
+
+  before do
+    devicegraph_stub(devicegraph)
+  end
+
+  describe "#restore_from_checkpoint" do
```
joseivanlopez: I would suggest to write this test in a more rspec way. Maybe something like:

```

  subject { described_class.new(device) }

  let(:device) { Y2Partitioner::DeviceGraphs.instance.current.find_by_name(device_name) }

  describe "#restore_from_checkpoint" do
    let(:device_name) { "/dev/sda1" }

    context "when restoring a device used by filesystem itself" do
      let(:filesystem) { device.direct_blk_filesystem }

      before do
        # same sequence as in Controllers::BtrfsDevices.remove_device
        filesystem.remove_device(device)
      end

      it "does not raise an exception" do
        expect(subject.restore_from_checkpoint).to_not raise_error
      end

      it "does not remove the filesystem" do
        sid = filesystem.sid

        subject.restore_from_checkpoint

        expect(Y2Partitioner::DeviceGraphs.instance.current.find_device(sid)).to_not be_nil
      end
    end
  end
```

test/y2partitioner/blk_device_restorer_test.rb
```diff
@@ -0,0 +1,44 @@
...
+    it "does not crash when restoring device used by filesystem itself" do
+      device = Y2Partitioner::DeviceGraphs.instance.current.find_by_name("/dev/sda1")
+      filesystem = device.direct_blk_filesystem
+
+      # same sequence as in Controllers::BtrfsDevices.remove_device
+      filesystem.remove_device(device)
+      Y2Partitioner::BlkDeviceRestorer.new(device).restore_from_checkpoint
```
joseivanlopez: In case you consider that rewrite the test is not needed, please, at least use here an expectation to check that no exception is raised:

```
expect { Y2Partitioner::BlkDeviceRestorer.new(device).restore_from_checkpoint }.to_not raise_error
```

PR #951, opened 2019-08-02
fix NilClass issue when calculating proposal on RAID (bsc#1139808)

wfeldt:
## Problem

https://bugzilla.suse.com/show_bug.cgi?id=1139808

Calculating a proposal that involves resizing a partition on a RAID runs into a `NilClass` issue.

The problem basically starts here:

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/partition.rb#L121-L127

where it's explicitly stated that `Partition#disk` is `nil` when the partition is on a RAID. But the semantics used in the code assume that `Partition#disk` gets you to the storage object holding the partition.

Adding to that, 'partition-like' objects as `Planned::Partition` or `FreeDiskSpace` use `#disk` to refer to the object holding the partition.

The fix is to use `Partition#partitionable` **in this case**. But it's not clear where else in the code this misconception is used.

## Solution

Use `Partition#partitionable` instead.

## Testing

Unit testing is a bit tricky as we mock so much. The added test at least fails with the old code and works with the fixed code.

- added unit test
- integration tested manually (Tumbleweed)

## Future

Check whether the distinction  of `Partition#disk` vs.  `Partition#partitionable` is really that useful.


src/lib/y2storage/free_disk_space.rb
```diff
@@ -34,6 +34,10 @@ class FreeDiskSpace
     #   @return [DiskSize]
```
joseivanlopez: NP: I would take the opportunity to fix this documentation: it should be `@return [Partitionable]`.

joseivanlopez: And also in the constructor of this class: `@param disk [Partitionable]`

test/data/devicegraphs/windows-pc-raid1.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0"?>
```
joseivanlopez: NP: For this trivial scenarios, I would use yml files (easier to read, write and maintain), see for example: https://github.com/yast/yast-storage-ng/blob/master/test/data/devicegraphs/md_raid.yml

joseivanlopez: In fact, the test could be confusing because a Software RAID is used, but the error should not happen with Software RAIDs (Mds are not considered disk devices, so the proposal does not use them).

Why not to use the xml file from the logs for this regression test?

wfeldt: This **is** the xml from the logs.

wfeldt: The yml did not work. Now that I looked at it a bit closer I think the yml reader has a bug.

joseivanlopez: Ok, thanks.

joseivanlopez:
I only miss the version bump and the changelog entry.

PR #957, opened 2019-08-28
Partitioner: display the encryption type

ancorgs:
We recently introduced support for LUKS2 encryption in libstorage-ng. Everything keeps working out of the box, as tested manually and documented here.

https://trello.com/c/NkE2aoyy/1246-1-read-only-support-for-luks2-in-the-partitioner

But in addition to that, this pull request adds the encryption type to the Partitioner UI.

## Screenshots

Non encrypted device (nothing changes).

enc_sda1

Encrypted device (now includes the "LUKS1" type).

enc_sda2


test/y2storage/encryption_type_test.rb
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env rspec
+# Copyright (c) [2017-2019] SUSE LLC
```
joseivanlopez: NP: 2019

PR #958, opened 2019-08-29
bind-mount /run from inst-sys to target system during install (bsc#1136463)

wfeldt:
## Problem

https://bugzilla.suse.com/show_bug.cgi?id=1136463

Files in `/run` are needed also in the target system during installation.

That's basically [bnc#717321](https://bugzilla.suse.com/show_bug.cgi?id=1136463) but again for storage-ng.

## Solution

Bind-mount `/run` into target system.

## See also

https://bugzilla.suse.com/show_bug.cgi?id=717321

Same issue in SLE12, fixed in old storage stack.

https://github.com/yast/yast-storage/blob/SLE-12-SP5/src/clients/inst_prepdisk.rb#L143-L166

## Testing

Tested manually.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Aug 30 09:25:52 UTC 2019 - Steffen Winterfeldt <snwint@suse.com>
+
+- bind-mount /run from inst-sys to target system during install (bsc#1136463)
+- 4.2.37
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


kobliha:
Just a note: See https://github.com/yast/yast-packager/blob/d4a21e0ac882b23b68a935367b89e8a1ca6fb273/src/clients/inst_kickoff.rb#L32-L45

We already similarly mount /dev/ at a different place. I even thought we did it for more...

wfeldt:
I've seen it (and was not amused).

Still, the canonical place seems to be `inst_prepdisk` as that's the place used also in the old storage times.

Of course, we do unmount `/run` in `umount_finish`...

PR #959, opened 2019-08-29
Improve behavior when editing encrypted devices

ancorgs:
## Problem

Editing a block device that already contains an encryption layer has some problems, described at
https://trello.com/c/qeSZSCxQ/1244-5-partitioner-fix-behavior-when-selecting-deselecting-encryption-checkbox

This PR specifically targets the following problems:

- If the password was already provided, editing again looses it and asks for a new password. That will become a bigger problem soon, since we plan to add more types of encryption (random, pervasive, maybe LUKS2) each of them with their own arguments. Having to re-enter the type and all the arguments would be even a better PITA than it's today.
- If there is already a existing encryption in the system (e.g. a LUKS2 one created with other tool), it cannot be reused. This will also become a bigger problem soon with pervasive encryption, in which likely the user doesn't want to re-encrypt the device, but simply use the existing one and (re)format the resulting device.

## Solution

When encrypting for the first time a device that was not originally encrypted in the system (or encrypting a new device just created using the Partitioner), nothing changes apart from minor adjustments in the labels. The user simply sees a form with an empty "password" field.

enc_new_3

When editing for a second time a device that was already marked for encryption during the current execution of the Partitioner, the form is already prefilled with the password entered before. In the past, the previous encryption layered was ditched (so it's password and other arguments were forgotten) and the user had to define the encryption again from scratch. That will become even more relevant soon, when the form for encryption becomes more than just a "password" field. See #960.

enc_modify_3

Last but not least, when editing a device that is already encrypted in the system (that is, it contains an active encryption layer that was there already when the Partitioner was executed), an option is offered to just use the existing encryption layer instead of replacing it with a (likely more limited) encryption created by the Partitioner.

enc_keep_3

Note the option is only possible if the preexisting encryption device is active (e.g. the user already entered its passphrase during boot or during the initial hardware probing performed by the Partitioner).

## Testing

- Unit test coverage extended
- Tested manually with the `partitioner_testing` client
- Not done yet: manually test a full installation with reused encryption

### DIFF
--- a/src/lib/y2partitioner/dialogs/encryption.rb
+++ b/src/lib/y2partitioner/dialogs/encryption.rb
@@ -20,10 +20,12 @@
 require "yast"
 require "y2partitioner/dialogs/base"
 require "y2partitioner/widgets/encrypt_password"
+require "y2partitioner/widgets/controller_radio_buttons"
 
 module Y2Partitioner
   module Dialogs
-    # Ask for a password to assign to an encrypted device.
+    # Ask for the concrete action to perform when the goal is to obtain an
+    # encrypted device, including details like the password.
     # Part of {Actions::AddPartition} and {Actions::EditBlkDevice}.
     # Formerly MiniWorkflowStepPassword
     class Encryption < Base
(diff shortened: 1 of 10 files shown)


src/lib/y2partitioner/dialogs/encryption.rb
```diff
@@ -35,13 +37,78 @@ def initialize(controller)
...
+          current_widget.store if current_widget.respond_to?(:store)
+          @controller.action = value
+        end
+
+        # @macro seeAbstractWidget
+        def help
+          # helptext
+          _(
+            "<p>Choose the encryption layer.</p>\n" \
+            "<p>Bla.</p>"
```
dgdavid: Just a note to not forget to update the help.

test/y2partitioner/actions/controllers/encryption_test.rb
```diff
@@ -74,17 +79,34 @@
...
 
-          it "returns false" do
-            expect(subject.to_be_encrypted?).to eq(false)
+            it "returns true" do
+              expect(subject.show_dialog?).to eq(true)
+            end
+          end
+
+          context "with a new encryption" do
+            let(:in_system) { true }
```
dgdavid: It seems to me that you actually wanted to set `in_system` to `false`, isn't it?

test/y2partitioner/dialogs/encryption_test.rb
```diff
@@ -25,9 +25,76 @@
...
-  include_examples "CWM::Dialog"
+  context "when :encrypt is the only possible action" do
+    let(:actions) { [:encrypt] }
+
+    include_examples "CWM::Dialog"
+
+    describe "#contents" do
+      it "delegates everything to an EncryptPassword widget " do
+        content = dialog.contents.params
+        expect(content.size). to eq 1
```
dgdavid: NP: I'd delete the space between `).` and `to`.

test/y2partitioner/dialogs/encryption_test.rb
```diff
@@ -25,9 +25,76 @@
...
+
+  context "when :keep and :encrypt are both possible actions" do
+    let(:actions) { [:keep, :encrypt] }
+
+    include_examples "CWM::Dialog"
+
+    describe "#contents" do
+      it "delegates everything to an Encryption::Action widget " do
+        content = dialog.contents.params
+        expect(content.size). to eq 1
```
dgdavid: NP: I'd delete the space between `).` and `to`.

PR #960, opened 2019-08-30
Partitioner support for encrypting swap with random password 

dgdavid:
## Problem

The *crypttab* file supports a *swap* option to encrypt a swap device with a randomly generated password, for example:

```
cr_swap /dev/sda1 /dev/urandom swap
```

The swap device (*/dev/sda1* in the example above) if re-encrypted with plain encryption by using a new password at every system boot.

Configuring such a random encryption password for swap was possible in the old Partitioner by leaving the password text-fields empty.

This PR brings such feature back, but offering a more intuitive interface to achieve it.

* https://bugzilla.suse.com/show_bug.cgi?id=1088641
* Requires https://github.com/openSUSE/libstorage-ng/pull/666
* Related to https://jira.suse.com/browse/IBM-495
* Includes changes from closed https://github.com/yast/yast-storage-ng/pull/961
* PBI: https://trello.com/c/z9KlLE1m/1243-5-sles15-p3-1088641-partitioner-encrypted-swap-with-random-password

## Solution

* A new widget was added to select the encryption method when we are encrypting a swap device. Now two options are available: "LUKS1" and "Random Swap".

* When "Random Swap" is selected, */dev/urandom* is used as *key_file* to generate the password.

* *Import Mount Points* feature was improved to support importation of swap devices with random password.

* Previous encryption layer can now be preserved when editing an existing encrypted device.

* Bug [bsc#1151079](https://bugzilla.suse.com/show_bug.cgi?id=1151079) was also fixed.

## Testing

* Added a new unit test
* Tested manually

## Screenshots

<details>
<summary>Toggle</summary>

---
<p align="center"><em>Editing a proposed swap partition to add the encryption</em></p>

VirtualBox_SLE-12-SP5 Clone_18_09_2019_14_44_35

---
<p align="center"><em>After selecting the "Random Swap" option</em></p>

VirtualBox_SLE-12-SP5 Clone_18_09_2019_14_44_54

---
<p align="center"><em>Help text</em></p>

VirtualBox_SLE-12-SP5 Clone_18_09_2019_15_04_20

---
<p align="center"><em>The partitioning summary shows that an encrypted partition will be created for swap</em></p>

VirtualBox_SLE-12-SP5 Clone_18_09_2019_14_45_23

---
<p align="center"><em>Encrypting an existing encrypted device</em></p>

VirtualBox_openSUSE Tumbleweed_18_09_2019_14_52_39

---
<p align="center"><em>journalctl -b</em><p>

Screenshot_sles12sp4-lvm10l3_2019-09-12_15:55:15

---
<p align="center"><em>/etc/fstab and /etc/cryptab files</em><p>

Screenshot_sles12sp4-lvm10l3_2019-09-12_15:55:54

</details>


src/lib/y2partitioner/actions/controllers/filesystem.rb
```diff
@@ -184,6 +188,19 @@ def partition_id
           blk_device.is?(:partition) ? blk_device.id : nil
         end
 
+        def encrypt_types
+          types = [Y2Storage::EncryptionType::LUKS1]
+
+          # FIXME: PLAIN instead of TWOFISH (TWOFISH only for testing purposes)
+          types << Y2Storage::EncryptionType::TWOFISH if swap?
```
joseivanlopez: Replace with the type provided by the library for plain encryption: I guess something like `::PLAIN`.

src/lib/y2partitioner/dialogs/encrypt.rb
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) [2017] SUSE LLC
+# Copyright (c) [2017-2019] SUSE LLC
```
joseivanlopez: Be careful with this file rename when rebasing.

src/lib/y2partitioner/widgets/encrypt_type.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+
+      def label
+        _("Encryption type")
+      end
+
+      def items
+        controller.encrypt_types.map do |type|
+          # FIXME
+          if type.is?(:twofish)
```
joseivanlopez: Replace with the new encryption type provided by the library.

src/lib/y2storage/blk_device.rb
```diff
@@ -294,11 +295,15 @@ def encrypted?
     # @param dm_name [String, nil] DeviceMapper table name of the new device
     # @param password [String, nil] password of the new device
     # @return [Encryption]
-    def encrypt(dm_name: nil, password: nil)
+    def encrypt(dm_name: nil, password: nil, type: EncryptionType::LUKS1, key_file: nil)
+      # TODO: create type encryption
```
joseivanlopez: Adapt this code to the new library API for creating a plain encryption.

dgdavid:
@ancorgs 

What do you think about the Code Climate complain? Is it right?

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -402,7 +403,14 @@ def original_filesystem(filesystem)
         #   not found in the devicegraph.
         def entry_device(entry, devicegraph)
           device = entry.device(devicegraph_with_crypttab_names(devicegraph))
-          device ? devicegraph.find_device(device.sid) : nil
+
+          return nil unless device
+
+          if missing_swap_encryption?(device, devicegraph)
+            create_swap_encryption(device, devicegraph)
```
ancorgs: @joseivanlopez, While trying to understand this code, I realize that what it does can be described as copying `device` into `devicegraph`. Since `device` comes from another (so-called fake) graph.

Maybe it would be more obvious if that would be reflected in the name of `create_swap_encryption`. Or maybe even going one step further and using `copy_to_devicegraph` so the method does exactly that (instead of creating an encryption object in `devicegraph` just to "imitate" how another equivalent encryption object was created in the fake devicegraph [here](https://github.com/yast/yast-storage-ng/pull/960/files#diff-0e53ffcdfa995a7567219ea93442be9bR285)).

src/lib/y2storage/encryption_method.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def is?(value)
+      id == value.to_sym
+    end
+
+    def used_for?(encryption)
+      process_class.used_for?(encryption)
+    end
+
+    def available_for?(blk_device)
```
ancorgs: @dgdavid This is a leftover, please remove it while tidying up the PR.

dgdavid: Ok @ancorgs, I'll do it. Thanks!

src/lib/y2storage/encryption_method.rb
```diff
@@ -34,6 +34,10 @@ class EncryptionMethod
     extend Yast::I18n
 
     # Constructor
+    #
+    # @param id [Symbol]
+    # @param label [String]
+    # @param process_class [Y2Storage::EncryptionProcesses]
```
joseivanlopez: `process_class` is a class name. I am not sure how to specify it. Maybe, something like the following?

```
# @param process_class [#new] class name of the encryption process (e.g., EncryptionProcesses::Luks1)
```

src/lib/y2storage/encryption_method.rb
```diff
@@ -53,14 +57,24 @@ def initialize(id, label, process_class)
...
 
+    # Looks for the encryption method used for the given encryption device
+    #
+    # @param encryption [Y2Storage::Encryption]
+    # @return [Y2Storage::EncryptionMethod, nil]
     def self.for_device(encryption)
       all.find { |m| m.used_for?(encryption) }
     end
 
+    # Looks for the encryption method by its value
```
joseivanlopez: NP: I would say "by its symbol representation" instead of "its value".

src/lib/y2storage/encryption_method.rb
```diff
@@ -76,16 +90,28 @@ def to_human_string
       _(@label)
     end
 
+    # Compare two encryption methods
```
joseivanlopez: NP: please, use third person "Compares".

src/lib/y2storage/encryption_method.rb
```diff
@@ -76,16 +90,28 @@ def to_human_string
...
+    #
+    # @param other [Y2Storage::EncryptionMethod]
+    # @return [Boolean] true if compared encryption methods have the same class and id; false if not
     def ==(other)
       other.class == self.class && other.id == id
     end
 
     alias_method :eql?, :==
 
+    # Whether the given value matches with the method id
```
joseivanlopez: NP: I would say "Whether the given value matches the symbol representation (id) of the encryption method"

src/lib/y2partitioner/widgets/encrypt.rb
```diff
@@ -66,8 +66,16 @@ def handle(event)
         nil
       end
 
+      # @macro seeCustomWidget
       def help
-        controller.methods.map { |m| send("help_for_#{m.id}", m) }.join
+        help_header =
```
joseivanlopez: I would extract this to a separate method:

```
def help_header
  if controller.methods.size > 1
    _("The following encryption methods can be chosen:")
  else
    _("The following encryption method will be used:")
  end
end
```

src/lib/y2partitioner/widgets/encrypt.rb
```diff
@@ -66,8 +66,16 @@ def handle(event)
...
+          if display_encrypt_method?
+            _("The following encryption methods can be chosen:")
+          else
+            _("The following encryption method will be used:")
+          end
+
+        "<p>#{help_header}</p><ul>#{help_for_methods}</ul>"
```
joseivanlopez: In general we are also translating the html tags, so I suggest to use something like:

```
format(_("<p>%{header}</p><ul>%{content}</ul>"), header: help_header, content: help_for_methods)
```

src/lib/y2partitioner/widgets/encrypt.rb
```diff
@@ -107,6 +115,22 @@ def display_encrypt_method?
         controller.methods.size > 1
       end
 
+      # Help texts for available encryption methods
+      #
+      # @return [String]
+      def help_for_methods
+        texts = controller.methods.map do |m|
```
joseivanlopez: You can simply:

```
text = send("help_for_#{m.id}", m)
"<li>#{text}</li>"
```

src/lib/y2partitioner/widgets/encrypt.rb
```diff
@@ -122,8 +146,15 @@ def help_for_random_swap(encrypt_method)
         )
       end
 
+      # Help text for the Regular Luks1 encryption method
+      #
+      # @return [String]
       def help_for_luks1(encrypt_method)
-        _("<p>TODO</p>")
+        format(
+          _("<p><b>%{label}</b>: allows to encrypt the device using the LUKS " \
```
joseivanlopez: I would re-word it a little bit:

``` 
"<p><b>%{label}</b>: allows to encrypt the device using LUKS1 (Linux Unified Key Setup). You have to provide the encryption password.</p>"
```

test/y2storage/encryption_method_test.rb
```diff
@@ -0,0 +1,152 @@
...
+    end
+
+    context "when looking for an unknown method" do
+      it "returns nil" do
+        expect(described_class.find("unknown value")).to be_nil
+      end
+    end
+  end
+
+  describe "#create_device" do
```
joseivanlopez: `luks1` and `random_swap` encryption methods have quite different behavior when creating a device. I would add tests for each one, checking for example the type of encryption it creates (plain or luks).

src/lib/y2storage/encryption_method.rb
dgdavid: :sweat_smile:

src/lib/y2partitioner/widgets/encrypt.rb
dgdavid: Right! I don't understand why I added the not necessary line of code there :/

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -83,6 +91,16 @@ def actions
             end
         end
 
+        def methods
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -83,6 +91,16 @@ def actions
...
+          encrypt_methods = [Y2Storage::EncryptionMethod::LUKS1]
+          encrypt_methods << Y2Storage::EncryptionMethod::RANDOM_SWAP if swap?
+          encrypt_methods
+        end
+
+        def method
+          @method || Y2Storage::EncryptionMethod::LUKS1
```
joseivanlopez: Constructor already assigns a default value when there is no encryption. I would simply define `#method` as an `attr_reader`.

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -131,6 +149,10 @@ def wizard_title
         #   device that will be modified by this controller
         attr_reader :fs_controller
 
+        def swap?
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -147,7 +169,11 @@ def blk_device
         # @return [Boolean] false if the device is formatted in the system and
         #   the user wants to preserve that filesystem
         def can_change_encrypt?
-          blk_device.filesystem.nil? || new?(blk_device.filesystem)
+          filesystem.nil? || new?(filesystem)
+        end
+
+        def filesystem
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -296,6 +315,29 @@ def import_mount_point(entry)
           end
         end
 
+        # Device version from the current devicegraph
+        #
+        # @param [Y2Storage::Device] device
```
joseivanlopez: NP: `# @param device [Y2Storage::Device]`

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -296,6 +315,29 @@ def import_mount_point(entry)
...
+            current_graph.find_device(device.sid)
+          end
+        end
+
+        # Copies a plain encryption for swap into the current devicegraph
+        #
+        # @param device [Y2Storage::Encryption]
+        # @return [Y2Storage::Encryption]
+        def copy_swap_encryption(device)
+          blk_device = current_graph.find_device(device.blk_device.sid)
```
joseivanlopez: Use `blk_device = device_from_current_graph(device.blk_device)`

src/lib/y2partitioner/actions/controllers/fstabs.rb
```diff
@@ -385,24 +427,20 @@ def original_filesystem(filesystem)
...
-        # When the device name in the fstab entry corresponds to a encryption device, the device
-        # could not be found by that fstab name. In general, the encryptions might be probed with
-        # a different name, so before searching for the device, the encryption names from the crypttab
-        # file are saved into the proper encryption device. That changes are made in a temporary
-        # devicegraph, so the original one is never altered.
+        # When the device name in a fstab entry corresponds to a encryption device, the device could be
+        # not found by its fstab name. In general, encryptions might be probed with a different name, so
+        # before searching for the device, the encryption names from the crypttab file are saved into the
+        # proper encryption device. Also note that a random encryption layer can be created when a
+        # crypttab entry points to a device encrypted as a random swap. All these changes are made in a
```
joseivanlopez: "points to a swap device encrypted with random password"

src/lib/y2partitioner/widgets/encrypt.rb
```diff
@@ -0,0 +1,168 @@
...
+require "y2partitioner/widgets/encrypt_method_options"
+require "y2partitioner/widgets/helpers"
+
+module Y2Partitioner
+  module Widgets
+    # Widget to set the encryption method and options
+    class Encrypt < CWM::CustomWidget
+      include Helpers
+
+      def initialize(controller)
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2partitioner/widgets/encrypt.rb
```diff
@@ -0,0 +1,168 @@
...
+        format(
+          _("<p>%{header}</p><ul>%{content}</ul>"),
+          header:  help_header,
+          content: help_content
+        )
+      end
+
+      private
+
+      attr_reader :controller
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2partitioner/widgets/encrypt_method.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      # @macro seeAbstractWidget
+      def store
+        controller.method = value
+      end
+
+      private
+
+      attr_reader :controller
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2partitioner/widgets/encrypt_method_options.rb
```diff
@@ -0,0 +1,106 @@
...
+      # Redraws the widget to show the options for given encryption method
+      #
+      # @param encrypt_method [YStorage::EncryptionMethod]
+      def refresh(encrypt_method)
+        replace(options_for(encrypt_method))
+      end
+
+      private
+
+      attr_reader :controller
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2storage/blk_device.rb
```diff
@@ -244,6 +245,7 @@ def udev_full_ids
     #   in the real system. It will fail during commit.
     #
     #   @param dm_name [String] see #dm_table_name
+    #   @param type [EncryptionType] optional encryption type of the new device
```
joseivanlopez: As `type` is optional, please, indicates that the default value is LUKS1.

src/lib/y2storage/encryption.rb
```diff
@@ -57,6 +57,23 @@ class Encryption < BlkDevice
     #   @return [Boolean] whether the device is included in /etc/crypttab
     storage_forward :in_etc_crypttab?
 
+    # @!attribute key_file
+    #   @return [nil] the encryption key file
```
joseivanlopez: `nil`? I guess it returns a String, see https://github.com/openSUSE/libstorage-ng/blob/master/storage/Devices/Encryption.h#L77

src/lib/y2storage/encryption_method.rb
```diff
@@ -0,0 +1,134 @@
...
+      @label = label
+      @process_class = process_class
+    end
+
+    LUKS1 = new(
+      :luks1, N_("Regular LUKS1"), EncryptionProcesses::Luks1
+    )
+    RANDOM_SWAP = new(
+      :random_swap, N_("Random Swap"), EncryptionProcesses::Swap
+    )
```
joseivanlopez: I guess both constants, `LUKS1` and `RANDOM_SWAP` should be private.

src/lib/y2storage/encryption_method.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    ALL = [LUKS1, RANDOM_SWAP].freeze
+    private_constant :ALL
+
+    # Sorted list of all possible settings
```
joseivanlopez: NP: all possible encryption methods.

src/lib/y2storage/encryption_method.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    # Looks for the encryption method by its symbol representation
+    #
+    # @param value [#to_sym]
+    # @return [Y2Storage::EncryptionMethod, nil] the encryption method found if any; nil otherwise
+    def self.find(value)
+      all.find { |i| i.is?(value) }
+    end
+
+    # @return [Symbol]
```
joseivanlopez: Please, document it. For example: "name to represent the encryption method (e.g., :luks1, :random_swap)"

src/lib/y2storage/encryption_method.rb
```diff
@@ -0,0 +1,134 @@
...
+    # @param blk_device [Y2Storage::BlkDevice]
+    # @param dm_name [String]
+    # @return [Y2Storage::Encryption]
+    def create_device(blk_device, dm_name)
+      process_class.new(self).create_device(blk_device, dm_name)
+    end
+
+    private
+
+    attr_accessor :process_class
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2storage/encryption_processes/base.rb
```diff
@@ -0,0 +1,69 @@
...
+# find current contact information at www.suse.com.
+
+require "yast"
+
+module Y2Storage
+  module EncryptionProcesses
+    # Base class to perform the creation of an encrypted device
+    #
+    # @note This is a base class. To really created an encrypted device use any
+    # of its subclasses, which must define an #enccryption_type
```
joseivanlopez: Typo: `#enccryption_type`

joseivanlopez: I would comment that `EncryptionProcesses` are intended to be used through a `EncryptionMethod`.

src/lib/y2storage/encryption_processes/base.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      # Constructor
+      #
+      # @param method [Y2Storage::EncryptionMethod]
+      def initialize(method)
+        @method = method
+      end
+
+      attr_reader :method
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2storage/encryption_processes/base.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      # Creates and returns an encryption device with given params
```
joseivanlopez: NP: I would re-word it as: Creates an encryption layer over the given block device.

src/lib/y2storage/encryption_processes/base.rb
```diff
@@ -0,0 +1,69 @@
...
+      # @param dm_name [String]
+      def create_device(blk_device, dm_name)
+        enc = blk_device.create_encryption(dm_name || "", encryption_type)
+        enc.encryption_process = self
+        enc
+      end
+
+      private
+
+      # Returns the Y2Storage::EncryptionType to be used
```
joseivanlopez: NP: I would not mention the real type in the description text: Returns the encryption type to be used. The real type is indicated with the `@return` tag.

joseivanlopez: BTW, we are using `abstract_method` for these "abstracts methods" (actually, ruby lacks of the abstract method concept): https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2partitioner/actions/base.rb#L58

src/lib/y2storage/encryption_processes/luks1.rb
```diff
@@ -0,0 +1,43 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage/encryption_type"
+require "y2storage/encryption_processes/base"
+
+module Y2Storage
+  module EncryptionProcesses
+    # The enryption process that allows to create and identify an encryped
+    # device using EncryptionType::LUKS1
```
joseivanlopez: NP: please, avoid to mention our ruby type and indicates the real technology name: "device using LUSK1".

src/lib/y2storage/encryption_processes/luks1.rb
```diff
@@ -0,0 +1,43 @@
...
+    class Luks1 < Base
+      # Whether the process was used for the given encryption device
+      #
+      # @param encryption [Y2Storage::Encryption] the encryption device to check
+      # @return [Boolean] true when the Y2Storage::EncryptionType is LUKS1; false otherwise
```
joseivanlopez: NP: when the encryption type is LUKS1; false otherwise.

src/lib/y2storage/encryption_processes/luks1.rb
```diff
@@ -0,0 +1,43 @@
...
+      def self.used_for?(encryption)
+        encryption.type.is?(:luks1)
+      end
+
+      private
+
+      def encryption_type
```
joseivanlopez: NP: please, document it as `# @see EncryptionProcesses::Base#encryption_type`.

src/lib/y2partitioner/actions/controllers/encryption.rb
dgdavid: But, if for any reason the controller ends receiving 

```ruby
controller.method = nil
````

then it will lost the _default_ value. 

What do you think?

File: src/lib/y2storage/encryption_processes/swap.rb [hunk 40]
```diff
@@ -0,0 +1,64 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2storage/encryption_type"
+require "y2storage/encryption_processes/base"
+
+module Y2Storage
+  module EncryptionProcesses
+    # The enryption process that allows to create and identify a randomly
```
joseivanlopez: Typo: enryption.

src/lib/y2partitioner/actions/controllers/encryption.rb
joseivanlopez: With current implementation:

```
controller.method = nil
controller.method.nil? #=> false

controller.method = :foo
controller.method #=> :foo
```

I think this is wrong. If we really want to avoid nil values, I would re-define the setter as:

```
def method=(value)
  raise Exception unless value.is_a?(EncryptionMethod)
end
``` 

For now, I would simply use `attr_accessor :method`. If the user assigns something strange, then the value will not selected, but right now we are the only ones who assign such value.

src/lib/y2storage/encryption_processes/swap.rb
joseivanlopez: I would re-word it a little: Encryption process that allows to encrypt a swap device by using plain encryption and random password.

src/lib/y2storage/encryption_processes/swap.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      # Whether the process was used for the given encryption device
+      #
+      # @param encryption [Y2Storage::Encryption] the encryption device to check
+      # @return [Boolean] true when "swap" is found in the crypt options; false otherwise
+      def self.used_for?(encryption)
+        encryption.crypt_options.any? { |opt| opt.downcase == SWAP }
+      end
+
+      def create_device(blk_device, dm_name)
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2storage/encryption_processes/swap.rb
```diff
@@ -0,0 +1,64 @@
...
+        enc = super
+        enc.key_file = key_file
+        enc.crypt_options = crypt_options + enc.crypt_options
+        enc
+      end
+
+      private
+
+      def encryption_type
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2storage/encryption_processes/swap.rb
```diff
@@ -0,0 +1,64 @@
...
+        EncryptionType::PLAIN
+      end
+
+      def crypt_options
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2storage/encryption_processes/swap.rb
```diff
@@ -0,0 +1,64 @@
...
+        [SWAP]
+      end
+
+      def key_file
```
joseivanlopez: Please, add documentation for the method.

src/lib/y2storage/encryption_method.rb
joseivanlopez: Ok, I see these constants are directly used from outside.

test/y2partitioner/widgets/encrypt_method_test.rb
```diff
@@ -0,0 +1,98 @@
...
+    devicegraph_stub("empty_hard_disk_50GiB.yml")
+
+    enc_controller.method = luks1
+  end
+
+  describe "#init" do
+    it "sets the current encryption method value" do
+      expect(widget).to receive(:value=).with(enc_controller.method)
+
+      widget.init
```
joseivanlopez: Why not this?

```
widget.init
expect(widget.value).to eq(luks1)
```

test/y2partitioner/widgets/encrypt_method_test.rb
```diff
@@ -0,0 +1,98 @@
...
+    before do
+      allow(Yast::UI).to receive(:QueryWidget)
+        .with(Id(widget.widget_id), :Value)
+        .and_return(selected_method)
+    end
+
+    it "sets the selected encryption method" do
+      expect(enc_controller).to receive(:method=).with(widget.value)
+
+      widget.store
```
joseivanlopez:
```
widget.store
expect(enc_controller.method).to eq(luks1)
```

joseivanlopez: BTW, luks1 is de default value, so I would use another method to really check that the controller is updated.

test/y2partitioner/widgets/encrypt_test.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+
+    context "when there are more than one encryption method" do
+      let(:methods) { [luks1, random_password] }
+
+      it "includes the encryption method selector" do
+        expect(encrypt_method_term).to_not be_nil
+      end
+    end
+  end
```
joseivanlopez: What about adding some test for the help? For example, to check that one header is used when there is only one encryption method and another header otherwise.

src/lib/y2storage/encryption.rb
dgdavid: Good catch! Thanks!

src/lib/y2storage/encryption.rb
```diff
@@ -57,6 +57,23 @@ class Encryption < BlkDevice
...
+    storage_forward :key_file
+    storage_forward :key_file=
+
+    # @!attribute key_file
```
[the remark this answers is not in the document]
dgdavid: Copy&paste :(

src/lib/y2storage/encryption_processes/base.rb
dgdavid: Thank you!

test/y2partitioner/widgets/encrypt_test.rb
dgdavid: Good point, thank you

test/y2partitioner/widgets/encrypt_method_test.rb
dgdavid: Because it doesn't work, since the `#value` method querying the UI. That's why I decided to check that, at least, it receives the `.value=` instead of just mocking the Query UI.

joseivanlopez: Ahh, ok.

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -91,12 +91,26 @@ def actions
             end
         end
 
+        # Whether there are more than one encryption methods available
+        #
+        # @return [Boolean] true when there are sevral encrypt methods; false if
```
joseivanlopez: Typo: several encryption methods

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -186,6 +182,9 @@ def can_change_encrypt?
           filesystem.nil? || new?(filesystem)
         end
 
+        # Returns the filesystem
```
joseivanlopez: Please, re-word: Filesystem from the plain block device

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -186,6 +182,9 @@ def can_change_encrypt?
...
+        #
+        # @return [Y2Storage::Filesystem, nil] block device filesystem, if any
```
joseivanlopez: Please, re-word: `# @return [Y2Storage::Filesystem, nil] nil if the plain block device is not formatted`

src/lib/y2storage/encryption.rb
```diff
@@ -172,11 +172,20 @@ def crypttab_name=(value)
       save_userdata(:crypttab_name, value)
     end
 
+    # Returns the encryption method used
+    #
+    # Note that it could be infered is there is not an EncryptionProccess available yet
```
joseivanlopez: Some typos: inferred if there is no encryption process available yet

src/lib/y2storage/encryption_processes/base.rb
```diff
@@ -58,9 +61,12 @@ def create_device(blk_device, dm_name)
...
       private
 
-      # Returns the Y2Storage::EncryptionType to be used
+      # Returns the encryption type to be used
+      #
+      # This method must be defined by derived class.
       #
       # @return [Y2Storage::EncryptionType]
+      abstract_method :encryption_type
       def encryption_type
```
joseivanlopez: Did you not removed the previous definition of `#encryption_type`?

dgdavid: Sorry so much. Clearly a left over because of rush. Nice catch!

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -38,6 +39,12 @@ class Encryption < Base
         # @return [Symbol] :keep, :encrypt, :remove
         attr_accessor :action
 
+        # @return [Boolean] Whether the user wants to encrypt the device
+        attr_accessor :encrypt
```
ancorgs: At first glance, I don't see where this is used. Is the value ever set?

joseivanlopez: You are right. That value is controlled by the controller `Filesystem`.

dgdavid: Seems to be a leftover from the migration from [`Filesystem` controller](https://github.com/yast/yast-storage-ng/blob/7892ca30d46c32206af69f4fe862a4cac39f645f/src/lib/y2partitioner/actions/controllers/filesystem.rb#L64). It's used by the [`FormatAndMount`](https://github.com/yast/yast-storage-ng/blob/80745be7874014ec769465018b212f856280f71e/src/lib/y2partitioner/widgets/format_and_mount.rb#L716) widget there, but it is not necessary here :) I'll remove it.

Thanks!

src/lib/y2partitioner/widgets/encrypt.rb
```diff
@@ -0,0 +1,172 @@
...
+      #
+      # @return [String]
+      def help_for_random_swap(encrypt_method)
+        format(
+          _("<p><b>%{label}</b>: this encryption method uses randomly generated keys at boot and it " \
+            "will not support Hibernation to hard disk. The swap device is re-encrypted during every " \
+            "boot, and its previous content is destroyed. You should disable Hibernation through your " \
+            "respective DE Power Management Utility and set it to Shutdown on Critical to avoid Data " \
+            "Loss!</p>" \
+            "<p>Also note the device UUID changes every time the swap is encrypted, and the file " \
```
ancorgs: I would reorganize the information in this paragraph a bit. Something like:

Please, make sure your swap device is mounted by a stable name that is not subject to change on every reboot (e.g. for a swap partition use the udev device id instead of the partition device name). Otherwise a wrong device could be encrypted instead of your swap! In that regard, note both the file system label and the UUID change every time the swap is re-encrypted, so they are not valid options to mount a randomly encrypted swap device.

src/lib/y2storage/device.rb
```diff
@@ -263,6 +263,29 @@ def potential_orphans
       []
     end
 
+    # Copies the device to a given devicegraph, connecting it to its corresponding parent
+    #
+    # @raise [RuntimeError] if the device has more than one parent
```
ancorgs: This restriction made sense in the original scope of the method (`Y2Partitioner::BlkDeviceRestorer`). But here it looks hard to justify. Why a generic method called `Device#copy_to` has such restriction?

src/lib/y2partitioner/widgets/encrypt.rb
joseivanlopez: The only think I disagree is "the file system label and the UUID change every time ". Label does not change every time, but it is not set. Well, we can see it as it changes the first time to "my label" to "". But the rest of times it is always "". What about saying something like the label is not set?

src/lib/y2storage/device.rb
joseivanlopez: To be honest, I didn't remember the reason of this restriction. I think we can drop the restriction off. I guess `Y2Partitioner::BlkDeviceRestorer` still has to avoid copying an object with several parents, hasn't it?

ancorgs: Yes. We should not change the `BlkDeviceRestorer` behavior. And we should change the implementation of this (and test it) when there are several parents.

We can also rename this method to make very clear that is a limited copier.

src/lib/y2partitioner/widgets/encrypt.rb
ancorgs: I see the difference, but I wanted to keep the explanation simple. Saying "both change so they are not reliable" sounded more straightforward than "one changes and the other is bumped so none of them are reliable for their own reasons".

In other words, I found reasonable to sacrifice precision in order to keep the message straight. But I don't have a strong opinion, TBH.

src/lib/y2partitioner/blk_device_restorer.rb
```diff
@@ -191,9 +191,17 @@ def copy_subvolumes(source)
     #
     # @see Device#copy_to
     #
+    # @raise [RuntimeError] if the device has more than one parent
+    #
     # @param source [Y2Storage::Device]
     # @return [Y2Storage::Device] device copied to the current graph
     def copy_device(source)
+      if source.to_storage_value.in_holders.size != 1
```
ancorgs: Why not simply `source.parents.size != 1`?

joseivanlopez: Copying&pasting blindly :(

PR #963, opened 2019-09-04
added translation for new EncryptionType::PLAIN

aschnell:
The new enum value has to be handled, otherwise the test suite fails.

### DIFF
--- a/test/y2storage/dialogs/proposal_exception_test.rb
+++ b/test/y2storage/dialogs/proposal_exception_test.rb
@@ -54,8 +54,8 @@ describe Y2Storage::Dialogs::Proposal do
       end
 
       it "displays an informative pop-up about the libstorage-ng exception" do
-        expect(Yast2::Popup).to receive(:show).with(/installation may fail/, details: /Encryption \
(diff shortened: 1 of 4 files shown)


test/y2storage/dialogs/proposal_exception_test.rb
```diff
@@ -54,7 +54,8 @@
       end
 
       it "displays an informative pop-up about the libstorage-ng exception" do
-        expect(Yast2::Popup).to receive(:show).with(/installation may fail/, details: /Luks bigger/)
+        expect(Yast2::Popup).to receive(:show).with(/installation may fail/, details: /Encryption \
+bigger/)
```
ancorgs: Maybe that new formatting pleases Rubocop, but wouldn't this be much more readable?

```ruby
it "displays an informative pop-up about the libstorage-ng exception" do
  expect(Yast2::Popup).to receive(:show)
    .with(/installation may fail/, details: /Encryption bigger/)
  dialog.run
end
```

Or any other alternative... but that "bigger" at the very left, out of any indenting...

PR #964, opened 2019-09-12
add warning if /boot is on a luks2 encrypted partition

wfeldt:
## Task

Add warning if `/boot` is luks2 encrypted as grub does only support luks1.

## Solution

Warning added. 

This also implies that the proposal will go for a separate `/boot` partition if `/` is on luks2.

## Testing

- manual integration testing
- unit tests

Note:

> I've added only two tests for LUKS2 in the proposal (to see if it results in a  `/boot` partition). It's my impression the boot requirements document would quite blow up otherwise.

## Screenshots

bar

### DIFF
--- a/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
+++ b/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
@@ -168,17 +168,6 @@ module Y2Storage
         encrypted?(device_for_boot)
       end
 
-      # Whether the filesystem containing /boot is going to be readable by grub
-      #
-      # We might need to check if the filesystem is actually supported by grub
-      # but currently all storage-ng supported filesystems are.
-      #
-      # @return [Boolean] true if the filesystem where /boot resides is going to
-      #   be readable by grub
-      def grub_can_read_boot?
-        encrypted_for_grub?(device_for_boot)
-      end
-
       # Whether the EFI system partition (/boot/efi) is in a LVM logical volume
       #
       # @return [Boolean] true if the filesystem where /boot/efi resides is going to
@@ -260,6 +249,15 @@ module Y2Storage
       #   in the planned devices or in the devicegraph
       def boot_filesystem_type
         filesystem_type(device_for_boot)
+      end
+
+      # Encryption type of boot device
+      #
+      # The device can be a planned one or filesystem from the devicegraph.
+      #
+      # @return [Y2Storage::EncryptionType] Encryption type
+      def boot_encryption_type
+        encryption_type(device_for_boot)
       end
 
       # Whether the partition table of the disk used for booting matches the
@@ -550,17 +548,6 @@ module Y2Storage
         !encryption_type(device).is?(:none)
       end
 
-      # Whether the device is encrypted and grub can decrypt it
-      #
-      # The device can be a planned one or filesystem from the devicegraph.
-      #
-      # @param device [Filesystems::Base, Planned::Device, nil]
-      # @return [Boolean] device encryption state; return false if device is nil
-      def encrypted_for_grub?(device)
-        t = encryption_type(device)
-        t.is?(:none) || t.is?(:luks1)
-      end
-
       # Encryption type of device
       #
       # The device can be a planned one or filesystem from the devicegraph.
@@ -570,11 +557,9 @@ module Y2Storage
       # @param device [Filesystems::Base, Planned::Device, nil]
       # @return [Y2Storage::EncryptionType] Encryption type
       def encryption_type(device)
-        return Y2Storage::EncryptionType::NONE if device.nil?
-
         if device.is_a?(Planned::Device)
           (device.respond_to?(:encrypt?) && device.encrypt?) ? Y2Storage::EncryptionType::LUKS1 : nil
-        else
+        elsif device.respond_to?(:plain_blk_devices)
           device.plain_blk_devices.map { |d| d.encryption&.type }.compact.first
         end || Y2Storage::EncryptionType::NONE
       end
--- a/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
+++ b/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
@@ -20,6 +20,7 @@
 require "yast"
 require "pathname"
 require "y2storage/planned"
+require "y2storage/encryption_type"
 
 module Y2Storage
   module BootRequirementsStrategies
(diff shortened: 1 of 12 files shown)


src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -81,7 +81,18 @@ def needed_partitions(target)
...
       # @return [Array<SetupError>]
       def warnings
-        []
+        res = []
+
+        if !grub_can_read_boot?
+          error_message =
+            _(
+              "The boot loader cannot access the file system mounted at /boot. " \
+              "LUKS2 encryption is currently not supported by grub."
```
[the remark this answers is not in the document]
ancorgs: Well, as originally conceived, Analyzer's responsibilities should not include understanding Grub2 capabilities. Its public methods should simply describe how the final system will look like: `#boot_disk`, `#root_in_lvm?`, `#root_in_software_raid?`, `#encrypted_root?`, `#boot_in_lvm?`, `#boot_in_bcache?`, `#free_mountpoint?`...

The strategies are the right place to translate that into something meaningful about Grub2 and how a Linux system boots.

That isolation between the two layers got broken in this pull request https://github.com/yast/yast-storage-ng/pull/788. But IMO this discussion just proves we should go back to the original design, in which the Analyzer didn't overdo and the strategies were the ones with the knowledge.

test/y2storage/boot_requirements_strategies/analyzer_test.rb
```diff
@@ -767,4 +767,24 @@ def create_filesystem(device_name, mount_point)
       end
     end
   end
+
+  describe "#grub_can_read_boot?" do
+    subject(:analyzer) { described_class.new(devicegraph, planned_devs, boot_name) }
+
+    context "if grub supports the encryption method" do
+      before { allow(analyzer).to receive(:encrypted_for_grub?).and_return true }
```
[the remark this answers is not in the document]
ancorgs: I would expect something kind of similar to this:

```ruby
describe "#grub_can_read_boot?" do
  context "if /boot is in a planned encrypted partition" do
    context "if it's encrypted by the proposal itself (so LUKS1 is used)"
      let(:planned_root) { planned_partition(mount_point: "/", encryption_password: "12345678") }

      it "returns true" do
      end
    end

    context "if AutoYaST plans to reuse a previously encrypted partition" do
      # Note: I'm not sure if these settings would reuse the existing encryption or even whether
      # it's possible at all to reuse an existing encryption with AutoYaST
      let(:planned_root) { planned_partition(mount_point: "/", reuse: "/dev/sda1" }

      context "if the reused partition is LUKS1" do
        let(:scenario) { "any_scenario_in_which_the_reused_is_luks1" }

        it "returns true" do
        end
      end

      context "if the reused partition is LUKS2" do
        let(:scenario) { "any_scenario_in_which_the_reused_is_luks2" }

        it "returns false" do
        end
      end
    end
  end

  context "if /boot is a partition of the devicegraph" do
    # Something similar, with several possible unencrypted/LUKS1/LUKS2 scenarios
  end
end
```

That is, testing how the method result looks from outside when calling on different mocked scenarios. Not mocking the method internals.

src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -201,6 +212,25 @@ def unknown_boot_disk_error
...
       end
+
+      # Whether the filesystem containing /boot is readable by grub
+      #
+      # We might need to check if the filesystem is actually supported by grub
+      # but currently all storage-ng supported filesystems are.
+      #
+      # @return [Boolean] true if the filesystem where /boot resides is going to
+      #   be readable by grub
+      def grub_can_read_boot?
```
ancorgs: No sure about this level of indirection. Why do we need `#grub_can_read_boot?` if it only calls `#encrypted_for_grub?`. Both are private and none of them is redefined in a derivate class. So I don't see the point right now.

Moreover, directly using a method that includes "encrypted" in its name would make the part above more understandable. I mean this:

```ruby
if !encrypted_for_grub?
  error_message = _("....Only LUKS1 encryption is supported...")
  res << SetupError.new(message: error_message)
end
```

test/y2storage/boot_requirements_checker_ppc_test.rb
```diff
@@ -83,6 +83,44 @@
       end
     end
 
+    RSpec.shared_examples "PReP + /boot partition" do
```
ancorgs: Why is this extracted to a `shared_examples` block if it's actually not shared?

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -533,15 +543,25 @@ def in_bcache?(device)
...
+      def encryption_type(device)
         if device.is_a?(Planned::Device)
-          device.respond_to?(:encrypt?) && device.encrypt?
-        else
-          device.plain_blk_devices.any? { |d| d.respond_to?(:encrypted?) && d.encrypted? }
-        end
+          (device.respond_to?(:encrypt?) && device.encrypt?) ? Y2Storage::EncryptionType::LUKS1 : nil
+        elsif device.respond_to?(:plain_blk_devices)
+          device.plain_blk_devices.map { |d| d.encryption&.type }.compact.first
+        end || Y2Storage::EncryptionType::NONE
```
ancorgs: I though you wanted to kill this `if` by unifying the APIs of the descendants of `Y2Storage::Device` and the descendants of `Y2Storage::Planned::Device`.

src/lib/y2storage/boot_requirements_strategies/base.rb
wfeldt: fair enough

test/y2storage/boot_requirements_checker_ppc_test.rb
wfeldt: See my 'Note' in the pr description. I wasn't sure I'm going to need it.

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
wfeldt: I would like to see it changed. But not as part of this rather specific pr.

test/y2storage/boot_requirements_checker_ppc_test.rb
wfeldt: Anyway, does it hurt? The test looks more readable this way, imho.

src/lib/y2storage/boot_requirements_strategies/analyzer.rb
ancorgs: ok

test/y2storage/boot_requirements_checker_ppc_test.rb
ancorgs: Fine for me. I was just curious

PR #968, opened 2019-09-19
Pervasive encryption

ancorgs:
## Goal

Allow to encrypt volumes using pervasive encryption via the Partitioner.

- https://www.ibm.com/support/knowledgecenter/linuxonibm/liaaf/lnz_r_dccnt.html
- https://jira.suse.com/browse/SLE-7376
- https://trello.com/c/HKaDMPL6/1297-5-pervasive-encryption-partitioner

## Solution

Implemented a new PERVASIVE_LUKS2 encryption method.

That method is exposed in the Partitioner when the system has at least one online Crypto Express cryptographic coprocessor configured in CCA mode and the needed tools installed.

Otherwise (most users on earth) nothing changes.

This is how it looks in the partitioner, including the help.

pervasive1

and

pervasive2

It works and it generates secure keys like the ones showed here.

pervasive3

As you can see, the devices are registered by id if possible.

If a key already exists for the volume, then it's reused instead of generating a new one. In that case, the DeviceMapper name specified in the volume entry (if any) is used for the encrypted device.

## Testing

Unit tests included.

Tested manually in an installed system. See screenshots.

Pending: test a full installation process.

### DIFF
--- a/src/lib/y2partitioner/widgets/encrypt_method_options.rb
+++ b/src/lib/y2partitioner/widgets/encrypt_method_options.rb
@@ -54,10 +54,11 @@ module Y2Partitioner
       # @return [CWM::CustomWidget] a widget containing the options for the
       # given encryption method
       def options_for(encrypt_method)
-        if encrypt_method.is?(:random_swap)
+        case encrypt_method.to_sym
+        when :random_swap
           RandomOptions.new(controller)
-        elsif encrypt_method.is?(:luks1)
-          Luks1Options.new(controller)
+        when :luks1, :pervasive_luks2
+          LuksOptions.new(controller)
         end
       end
     end
@@ -88,8 +89,8 @@ module Y2Partitioner
       end
     end
 
-    # Internal widget to display the Luks1 encryption options
-    class Luks1Options < CWM::CustomWidget
+    # Internal widget to display the Luks encryption options
+    class LuksOptions < CWM::CustomWidget
       # Constructor
       #
       # @param controller [Actions::Controllers::Encryption]
--- a/src/lib/y2storage/encryption.rb
+++ b/src/lib/y2storage/encryption.rb
@@ -185,6 +185,10 @@ module Y2Storage
 
     # Saves the given encryption process
     #
+    # @note Keeping the state of this object is important, so see {#encryption_process}
+    #   and {#save_encryption_process} for some considerations about how it is
+    #   persisted into the devicegraph and how the current state is held.
+    #
     # @param value [Encryption::Processes]
     def encryption_process=(value)
       save_userdata(:encryption_process, value)
@@ -219,12 +223,20 @@ module Y2Storage
 
     # Returns the process used to perform the encryption
     #
+    # @note The EncryptionProcess object is persisted using the userdata mechanism,
+    #   but it's also cached in an instance variable because, otherwise, every call
+    #   to #encryption_process would be accessing the object that was stored in the
+    #   devicegraph, with its former state. Thus, all the modifications would be lost.
+    #   See {#encryption_process=} and {#save_encryption_process}.
+    #
     # @return [EncryptionProcess::Base, nil]
     def encryption_process
       @encryption_process ||= userdata_value(:encryption_process)
     end
 
     # Updates the userdata with an up-to-date version of the encryption process
+    #
+    # @see #encryption_process
     def save_encryption_process
       self.encryption_process = encryption_process
     end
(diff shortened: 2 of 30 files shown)


src/lib/y2storage/encryption_method.rb
```diff
@@ -151,6 +151,16 @@ def available?
       process_class.available?
     end
 
+    # Whether the encryption method is useful only for swap
+    #
+    # Some encryption methods are mainly useful for encrypting swap disks since they produdce a new key
```
ancorgs: produdce

dgdavid: :see_no_evil:

src/lib/y2partitioner/widgets/encrypt_method_options.rb
```diff
@@ -54,10 +54,11 @@ def empty_widget
       # @return [CWM::CustomWidget] a widget containing the options for the
       # given encryption method
       def options_for(encrypt_method)
```
dgdavid: I was tempted to use a guard clause here, 


```ruby
def options_for(encrypt_method)
  return RandomOptions.new(controller) if encryp_method.is?(:random_swap)

  LuksOptions.new(controller)
end
```
But then I decided to go ahead with a `case` statement because of the upcoming _random swap_ encryption methods.

dgdavid:
@ancorgs 

During manual testing, I realize that we are _reusing_ the password when editing an encrypted device even when changing the encryption method. I didn't see a clear way to _fix_ this a the UI level but I'd like to know your POV about this behavior.

Thanks!

ancorgs:
> @ancorgs
> 
> During manual testing, I realize that we are _reusing_ the password when editing an encrypted device even when changing the encryption method. I didn't see a clear way to _fix_ this a the UI level but I'd like to know your POV about this behavior.

Well, it's not necessarily bad. I would leave it like that.

src/lib/y2storage/encryption_processes/secure_key_volume.rb
```diff
@@ -0,0 +1,91 @@
...
+      def initialize(plain_name, dm_name)
+        @plain_name = plain_name
+        @dm_name = dm_name
+      end
+
+      # Creates a new object based on the portion of the output of
+      # "zkey list" that represents a concrete volume entry
+      #
+      # @param string [String] name of the plain device and, optionally, also
+      #   DeviceMapper of the encrypted device
```
joseivanlopez: NP: Please, document an example of this param: (e.g., "foo:/dev/mapper/cr_foo")

src/lib/y2storage/encryption.rb
```diff
@@ -188,6 +188,22 @@ def method
     # @param value [Encryption::Processes]
     def encryption_process=(value)
       save_userdata(:encryption_process, value)
+      @encryption_process = value
```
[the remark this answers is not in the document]
ancorgs: Because we modify the state of the `EncryptionProcess` data. If we wouldn't use an instance variable, every time we call `#encryption_process` we would be accessing the object that was stored in the devicegraph, with its former state. Thus, all the modifications would be lost.

That's also why we need `#save_encryption_process`, to persist the state.

Think about the userdata value as it was a database, every time you call `#userdata_value` you are reading the version of the object that was saved in the database, not the current version with the latest changes.

test/y2storage/pervasive_encryption_test.rb
```diff
@@ -0,0 +1,166 @@
...
+
+  let(:manager) { Y2Storage::StorageManager.instance }
+  let(:blk_device) { manager.staging.find_by_name("/dev/dasdc1") }
+  let(:pervasive) { Y2Storage::EncryptionMethod::PERVASIVE_LUKS2 }
+  let(:zkey_list) { zkey_output("list-dasdc1") }
+
+  describe "Y2Storage::BlkDevice#encrypt" do
+    it "creates an encryption device with type LUKS2 and method PERVASIVE_LUKS2" do
+      enc = blk_device.encrypt(method: pervasive)
+      expect(enc.type).to eq Y2Storage::EncryptionType::LUKS2
```
joseivanlopez: NP: `expect(enc.type.is?(:luks2)).to eq(true)`

test/y2storage/pervasive_encryption_test.rb
```diff
@@ -0,0 +1,166 @@
...
+    end
+
+    RSpec.shared_examples "zkey cryptsetup actions" do
+      before do
+        allow(Yast::Execute).to receive(:locally)
+          .with(/zkey/, "cryptsetup", "--volumes", anything, stdout: :capture)
+          .and_return zkey_cryptsetup
+      end
+      # I'm not sure why this default value is needed
+      let(:zkey_cryptsetup) { "" }
```
joseivanlopez: Me either.

joseivanlopez:
Only missing changelog entry and new version (apart of NPs).

PR #969, opened 2019-09-20
Protected and Secure swap encryption

joseivanlopez:
## Problem

IBM z Systems supports *randomly generated protected and secure key* that can be use to encrypt a swap device with a new password at every system boot. That's basically the same as [randomly encrypted swap](https://github.com/yast/yast-storage-ng/pull/960) but using a specific key file/cipher for generating a protected or secure key. 

For a full description see https://www.ibm.com/support/knowledgecenter/en/linuxonibm/com.ibm.linux.z.lxdc/lxdc_swapdisks.html

* Related to https://jira.suse.com/browse/SLE-7376
* Related to https://github.com/yast/yast-storage-ng/pull/960
* https://trello.com/c/KcXHj1RY/1248-5-encrypted-swap-with-random-protected-keys-s390
* Requires https://github.com/openSUSE/libstorage-ng/pull/672
* Also see https://github.com/yast/yast-update/pull/132

## Solution

Added new encryption methods for protected and secure swap encryption.

It also ensures that current plain encryption is not kept when the current device is not a swap anymore.

## Testing

- Added unit tests
- Tested manually

## Screenshots

Screenshot from 2019-09-30 15-57-57

Screenshot from 2019-09-30 15-58-34


src/lib/y2storage/encryption_processes/swap.rb
```diff
@@ -17,60 +17,169 @@
...
+        # @return [String, nil] nil if no specific size
+        def key_size
+          nil
+        end
+
+        # Encryption options to add to the encryption device
+        #
+        # @return [Array<String>]
+        def crypt_options
+          options = [swap_option]
```
dgdavid: NP: since those _*\_option_ methods are actually returning `nil` when no specific value available, I think that simply 

```ruby
[swap_option, cipher_option, key_size_option].compact
```
should be enough.

But probably it's just a matter of taste :)

src/lib/y2partitioner/widgets/encrypt.rb
```diff
@@ -159,30 +192,36 @@ def help_for_random_swap(encrypt_method)
...
-          _("<p><b>%{label}</b>: allows to encrypt the device using LUKS2 with a master secure key " \
-            "processed by a Crypto Express cryptographic coprocessor configured in CCA mode.</p>" \
-           "<p>If the cryptographic system already contains a secure key associated to this " \
-           "volume, that key will be used. Otherwise, a new secure key will be generated and " \
-           "registered in the system. You need to provide an encryption password that will be " \
-           "used to protect the access to that master key.</p>"),
+          # TRANSLATORS: Protected/Secure encryption terminology. For the English version see
+          # https://www.ibm.com/support/knowledgecenter/en/linuxonibm/com.ibm.linux.z.lxdc/
+          # lxdc_swapdisks_scenario.html
+          _("<p><b>%{label}</b>: the key for %{label} is volatile and cannot be recreated if lost, " \
```
ancorgs: I don't think this information is relevant here. The Partitioner cannot be used to encrypt non-transient data with volatile keys. We don't have any text about not using `/dev/urandom` for home either.

In my opinion, the help text for these methods should explain this is basically an improvement over regular random swap encryption so all the considerations for such method still stand.

src/lib/y2storage/encryption.rb
```diff
@@ -322,16 +322,63 @@ def save_crypttab_name(devicegraph, entry)
...
+        encryption_method = encryption_method_for_entry(entry)
+
+        return false unless encryption_method_for_swap?(encryption_method)
+
+        encrypt_with_encryption_method?(device, encryption_method)
+      end
+
+      # Whether the given encryption method corresponds to a swap related encryption method
+      #
+      # @param encryption_method [EncryptionMethod]
```
ancorgs: The type is `EncryptionMethod, nil`. I would say that's why you decided to have this method instead of just checking `#only_for_swap?`

src/lib/y2storage/encryption.rb
```diff
@@ -322,16 +322,63 @@ def save_crypttab_name(devicegraph, entry)
...
+      # entry
+      #
+      # @param device [BlkDevice]
+      # @param entry [SimpleEtcCrypttabEntry]
+      #
+      # @return [Boolean]
+      def create_swap_encryption_for_entry?(device, entry)
+        encryption_method = encryption_method_for_entry(entry)
+
+        return false unless encryption_method_for_swap?(encryption_method)
```
ancorgs: Couldn't this be simply?

```
return false unless encryption_method&.only_for_swap?
```

Sounds simpler than defining an additional method.

src/lib/y2storage/encryption_method.rb
```diff
@@ -60,14 +62,24 @@ def initialize(id, label, process_class)
...
       :pervasive_luks2, N_("Pervasive Volume Encryption"), EncryptionProcesses::Pervasive
     )
+
     RANDOM_SWAP = new(
-      :random_swap, N_("Random Swap"), EncryptionProcesses::Swap
+      :random_swap, N_("Random Swap"), EncryptionProcesses::RandomSwap
+    )
+
+    PROTECTED_SWAP = new(
+      :protected_swap, N_("Protected Swap"), EncryptionProcesses::ProtectedSwap
```
ancorgs: I don't think "Protected Swap" is a good description. What about "Random Swap with Protected Key"?

src/lib/y2storage/encryption_method.rb
```diff
@@ -60,14 +62,24 @@ def initialize(id, label, process_class)
...
     )
 
-    ALL = [LUKS1, PERVASIVE_LUKS2, RANDOM_SWAP].freeze
+    SECURE_SWAP = new(
+      :secure_swap, N_("Secure Swap"), EncryptionProcesses::SecureSwap
```
ancorgs: Same than before. I think "Random Swap with Secure Key" describes this much better. Or maybe something like "Volatile Encryption with Random/Protected/Secure Key"

src/lib/y2storage/encryption.rb
joseivanlopez: Yes, `nil` is a valid param value. And well, I prefer to have it as a separate method because I consider that it is different to `#only_for_swap?` from a semantic point of view. Maybe, in the future, we can "re-create" a swap encryption layer from a method that applies to swap but not only to swap.

src/lib/y2storage/encryption.rb
joseivanlopez: Right now, we could because `#only_for_swap?` means basically the same as `#encryption_method_for_swap?`, but I would prefer to have them as separate methods.

joseivanlopez: I have tried to improve the method description.

src/lib/y2storage/encryption_processes/swap.rb
```diff
@@ -17,60 +17,169 @@
...
+        def used?(key_file, crypt_options)
+          contain_swap_option?(crypt_options) && use_key_file?(key_file)
+        end
+
+        # Whether the given encryption device contains the swap option
+        #
+        # @param crypt_options [Array<String>]
+        # @return [Boolean]
+        def contain_swap_option?(crypt_options)
+          crypt_options.any? { |o| o.downcase == swap_option }
```
ancorgs: Just to be extra safe against future changes, we could use `o.downcase == swap_option.downcase`, or even better `swap_option.casecmp?(o)`

src/lib/y2storage/encryption_processes/swap.rb
```diff
@@ -17,60 +17,169 @@
...
+        end
+
+        # Whether the given encryption device is using the key file for this process
+        #
+        # @param key_file [String]
+        # @return [Boolean]
+        def use_key_file?(key_file)
+          key_file == send(:key_file)
```
ancorgs: Doesn't this work with `key_file == self.key_file`? Looks more natural to me than using `send`

src/lib/y2storage/encryption_processes/swap.rb
```diff
@@ -17,60 +17,169 @@
...
+        end
       end
 
-      # Returns the encryption key file
-      #
-      # @return [String]
+      # @see Swap.key_file
       def key_file
-        URANDOM
+        self.class.send(:key_file)
```
ancorgs: As I see it, it makes more sense to make those methods public than making them private and then using `send` to bypass the protection.

src/lib/y2partitioner/widgets/encrypt.rb
```diff
@@ -159,30 +192,36 @@ def help_for_random_swap(encrypt_method)
...
-      def help_for_pervasive_luks2(encrypt_method)
+      def help_for_secure_swap(encrypt_method)
+        help_for_s390_swap_method(encrypt_method)
+      end
+
+      # Help text for s390 specific swap encryption methods (i.e., protected and secure swap)
+      #
+      # @param encrypt_method [Y2Storage::EncryptionMethod]
+      # @return [String]
+      def help_for_s390_swap_method(encrypt_method)
```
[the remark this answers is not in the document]
joseivanlopez: Yes, but I have fixed this and now the help for both methods is not exactly the same.

src/lib/y2storage/encryption.rb
joseivanlopez: After "face to face" discussion, we have decided to move to `Crypttab` class all the logic related to `#save_crypttab_names`.

src/lib/y2storage/encryption_processes/swap.rb
joseivanlopez: Nop. This works:

```
class Foo
  def self.foo
    "foo" + self.bar
  end

  private

  def self.bar
    "bar"
  end
end
```

But this does not work:

```
class Foo
  class << self
    def foo
      "foo" + self.bar
    end

    private

    def bar
      "bar"
    end
  end
end
```

It returns `private method 'bar' called for Foo:Class (NoMethodError)`. I suspect `self` refers to a new eigenclass class that will become the superclass of the original `Foo`.

joseivanlopez: Anyway, now `#key_file` is a public method and we can do `self.key_file`.

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -69,16 +72,19 @@ def show_dialog?
...
         #
         # @return [Array<Symbol>]
         def actions
           return @actions if @actions
 
           @actions =
             if fs_controller.encrypt
-              if can_keep?
+              if sanitize_encryption?
+                [:sanitize, :encrypt]
```
ancorgs: I find this part quite confusing. In `[:keep, :encrypt]` it's obvious we are offering both options to the user. But here that's not the case... Even if I would manage to understand it now, looks like one of those things that nobody will see clearly in the future.

joseivanlopez: Yes, similar to *:remove* action. There is a clear translation between actions and what the user can select. Do you propose another approach?

joseivanlopez: I see actions as things that can be done over an encryption. Sometimes, the actions are selected by the user, but other times they are automatically selected.

ancorgs: As I can see, this works because the widget contains logic to enforce the `encrypt` option... which again makes this a little bit... unobvious

PR #973, opened 2019-10-01
implemented detection of pervasive encryption

aschnell:
For https://trello.com/c/B72IxUOM/1294-3-probing-of-cipher-and-recognize-preexisting-pervasive-encryption.

## Problem

The detection was so far missing.

## Solution

Detection implemented.

## Testing

- Added a new unit test.
- Tested manually.

### DIFF
--- a/src/lib/y2storage/encryption_processes/pervasive.rb
+++ b/src/lib/y2storage/encryption_processes/pervasive.rb
@@ -36,9 +36,7 @@ module Y2Storage
 
       # @see Base.used_for?
       def self.used_for?(encryption)
-        # TODO: it should be possible to detect a preexisting pervasive encryption
-        # after extending a bit the libstorage-ng probing capabilities.
-        super
+        encryption.type.is?(:luks2) && encryption.cipher == "paes-xts-plain64"
       end
 
       # @see Base.available?
--- a/test/y2storage/encryption_processes/luks2_pervasive_test.rb
+++ b/test/y2storage/encryption_processes/luks2_pervasive_test.rb
@@ -1,0 +1,58 @@
+#!/usr/bin/env rspec
+# Copyright (c) [2019] SUSE LLC
+#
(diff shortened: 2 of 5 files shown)


src/lib/y2storage/encryption_processes/pervasive.rb
```diff
@@ -36,9 +36,7 @@ class Pervasive < Base
 
       # @see Base.used_for?
       def self.used_for?(encryption)
-        # TODO: it should be possible to detect a preexisting pervasive encryption
-        # after extending a bit the libstorage-ng probing capabilities.
-        super
+        encryption.type.is?(:luks2) && encryption.cipher == "paes-xts-plain64"
```
joseivanlopez: NP: I would prefer to have the cipher value as a private constant:

```
CIPHER = "paes-xts-plain64"
private_constant :CIPHER

def cipher
  CIPHER
end

def self.used_for?(encryption)
  encryption.type.is?(:luks2) && encryption.cipher == cipher
end
```

test/y2storage/encryption_processes/luks2_pervasive_test.rb
```diff
@@ -0,0 +1,58 @@
...
+# find current contact information at www.suse.com.
+
+require_relative "../spec_helper"
+require "y2storage"
+
+describe Y2Storage::EncryptionProcesses::Pervasive do
+  subject(:process) { described_class.new(method) }
+  let(:method) { double }
+
+  describe ".used_for?" do
```
ancorgs: Since this is quite internal thing, we are so far testing this from a more external entry point here
https://github.com/yast/yast-storage-ng/blob/master/test/y2storage/encryption_method_test.rb#L193

I find the alternative in this pull request to be basically a fully mocked repetition of the implementation.

I think I would prefer a couple of contexts in the above-mentioned test of `EncryptionMethod.for_device`.

PR #978, opened 2019-10-09
Ensure suitable mount_by when possible

ancorgs:
## Problem

We have several problems regarding mount_by. For details see these two cards or keep reading.

* https://trello.com/c/GrmrcH1n/1337-3-sles15-sp1-p3-1151075-sles-15-sp1-yast-mounts-nvme-partitions-by-kernel-name-per-default-upon-reboot-they-are-randomly-enumera#
* https://trello.com/c/rJAfgxcb/1295-3-improve-relationship-between-encryption-and-mountby

Check the document included in this pull request to see a description of the relatively simplistic system used by libstorage-ng to set the value of the mount_by attributes and to report which values are possible for each device.

There was no additional logic in YaST to try to ensure a reasonable (or even acceptable) mount_by value. The YaST module just relied on the library and the library (in general) just relies on the global default.

Moreover, the Partitioner was relying on `MountPoint#possible_mount_bys` to disable the non-acceptable values in the corresponding widget displayed when the user clicks on "Fstab Options". Again, that was not enough.

The problem is that every time we set a mount_by that makes no sense for the device, we are sentencing it be mounted by device name (which is what the library uses as fallback). That's true for both `MountPoint` and `Encryption` and leaded to different kind of problems.

**Problems with the mount_by of `MountPoint`** 

When the global default value is not adequate for a given device (for example, setting `MountBy::PATH` for a device that has no by-path ids) we fallback to the device name which should be the last resort, not the second option. This was reported in [bug#1151075](https://bugzilla.suse.com/show_bug.cgi?id=1151075), where it was requested to use UUID as a fallback for those devices without an udev path (mounting by path is the default in s390 architecture).

Falling back from a mount_by that cannot be applied is not something that should happen at the very end, during the commit phase. If we already know that the default mount_by will not be used for a device, it should be visible from the start. For example, we should never set `UUID` or `LABEL` as the mount_by value for any randomly encrypted swap. They simply cannot work.

Whatever we do to ensure the consistency of the mount_by when random encryption is used should be reverted if possible when the users change the encryption method to a non-random one or simply change their minds and decide to not encrypt the device.

Moreover, the mount_by types that has no choice to become real should not be offered in the Partitioner. For that, the Partitioner should take the current value of the "Encrypt Device" checkbox into account. No matter whether the encryption device exists or not in the current devicegraph, it should list the options that will be valid after applying the changes specified by that checkbox.

**Problems with the mount_by of `Encryption`** 

If any of the random encryption methods for swap is used, we have to minimize the possibility of using an unstable mount_by. Since the final user cannot see or modify this value, this have no implications in the Partitioner.

## A second problem 

A `BtrfsSubvolume#copy_mount_by_from_filesystem` method exists in order to keep Btrfs consistent. As expected, that method is called internally every time a mount point is created for a subvolume. 

Fun enough, there is a symmetric `Btrfs#copy_mount_by_to_subvolumes` BUT it was not called automatically after every change in the `#mount_by` value of a filesystem. Instead, an explicit extra call to `#copy_mount_by_to_subvolumes` was expected after every change. 

## Solution

This PR adds some logic that makes sure we are proposing the best possible value for both occurrences of mount_by.

The new logic is applied:

- When using `Mountable#create_mount_point`
- When using `BlkDevice#encrypt` (since it can modify the device to which the `Mountable` is attached)
- When the encryption settings are changed and the user has not set the mount_by manually


src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -224,6 +224,8 @@ def perform_finish_action
           else
             finish_remove
           end
+
+          adjust_mount_by
```
imobachgs: `adjust_mount_by` or `adjust_mount_point`?

ancorgs:
Closed in favor of #985

PR #982, opened 2019-10-16
Set encryption sector size from the underlying device

imobachgs:
Set the encryption sector size according to the underlying device.

* If it is 4k or bigger, set the sector-size to 4k for performance reasons.
* If it is smaller than 4k, just do the default (do not set any sector size).

### DIFF
--- a/src/lib/y2storage/encryption_processes/secure_key.rb
+++ b/src/lib/y2storage/encryption_processes/secure_key.rb
@@ -45,15 +45,20 @@ module Y2Storage
       # @return [String] name of the secure key
       attr_reader :name
 
+      # @return [Integer, nil] sector size in bytes
+      attr_reader :sector_size
+
       # Constructor
       #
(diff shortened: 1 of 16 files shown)


src/lib/y2storage/encryption_processes/secure_key.rb
```diff
@@ -209,9 +214,9 @@ def available?
         # @param volumes [Array<Encryption>] encryption devices to register in
         #   the "volumes" section of the new key
         # @return [SecureKey] an object representing the new key
-        def generate(name, volumes: [])
+        def generate(name, sector_size: nil, volumes: [])
```
ancorgs: document the new argument here.

ancorgs:
As mentioned offline, I miss a bit of documentation about the level of support/usage of the sector size in each technology/method. Supported in open vs supported in configuration.

PR #983, opened 2019-10-16
SLE-15-SP1: ensure suitable mount_by when possible

ancorgs:
## Problem

We have several problems regarding mount_by. This pull request is for SLE-15-SP1 and only addresses the most basic one described at the following Trello card:

* https://trello.com/c/GrmrcH1n/1337-3-sles15-sp1-p3-1151075-sles-15-sp1-yast-mounts-nvme-partitions-by-kernel-name-per-default-upon-reboot-they-are-randomly-enumera

Check the document included in this pull request to see a description of the relatively simplistic system used by libstorage-ng to set the value of the mount_by attributes and to report which values are possible for each device.

There was no additional logic in YaST to try to ensure a reasonable (or even acceptable) mount_by value. The YaST module just relied on the library and the library (in general) just relies on the global default.

The problem is that every time we set a mount_by that makes no sense for the device, we are sentencing it be mounted by device name (which is what the library uses as fallback).

When the global default value is not adequate for a given device (for example, setting `MountBy::PATH` for a device that has no by-path ids) we fallback to the device name which should be the last resort, not the second option. This was reported in [bug#1151075](https://bugzilla.suse.com/show_bug.cgi?id=1151075), where it was requested to use UUID as a fallback for those devices without an udev path (mounting by path is the default in s390 architecture).

Moreover, falling back from a mount_by that cannot be applied is not something that should happen at the very end, during the commit phase. If we already know that the default mount_by will not be used for a device, it should be visible from the start.

## Solution

This PR adds some logic that makes sure we are proposing the best possible value for both occurrences of mount_by.

The new logic is applied:

- When using `Mountable#create_mount_point`
- When using `BlkDevice#encrypt` (since it can modify the device to which the `Mountable` is attached)
- When the encryption settings are changed and the user has not set the mount_by manually


jreidinger:
@ancorgs BTW bootloader already have algorithm for it, so maybe it can be moved to storage-ng and bootloader just use it? https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/udev_mapping.rb#L47 and implementation at https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/udev_mapping.rb#L110

test/y2partitioner/actions/controllers/filesystem_test.rb
```diff
@@ -307,10 +307,28 @@
...
 
     RSpec.shared_examples "default_mount_by" do
-      it "sets mount by to the default value" do
-        subject.apply_role
-        expect(subject.filesystem.mount_point.mount_by)
-          .to eq(Y2Storage::StorageManager.instance.configuration.default_mount_by)
+      before do
+        Y2Storage::StorageManager.instance.configuration.default_mount_by = default_mount_by
+      end
+      let(:default_mount_by) { Y2Storage::Filesystems::MountByType::UUID }
```
imobachgs: np: you can remove this line as it is defined in each context.

src/lib/y2storage/encryption.rb
```diff
@@ -159,6 +205,33 @@ def assign_etc_attribute(value)
       self.storage_in_etc_crypttab = value
     end
 
+    # Whether the given type makes sense for the {#mount_by} attribute of the
+    # encryption device
+    #
+    # Using a value that is not suitable would lead to libstorage-ng ignoring
+    # that value during the commit phase. In such case, DEVICE} is used by the
```
joseivanlopez: Typo: Device}

doc/mount_by.md
```diff
@@ -0,0 +1,78 @@
...
+The two mount_by attributes are the following.
+
+* `MountPoint#get_mount_by` defines the way the `mount` command identifies the
+  filesystem (or subvolume) to mount. In other words, it defines the form of the
+  first field in the `fstab` file. `MountPoint` contains a couple of methods to
+  set the mount_by.
+
+  * `#set_default_mount_by` is used internally when creating a new mount point
+    with `Mountable#create_mount_point`. It relies on
+    `Mountable#get_default_mount_point` which almost always just returns the
```
joseivanlopez: I guess you mean `Mountable::get_default_mount_by`.

test/y2partitioner/actions/controllers/filesystem_test.rb
ancorgs: Not really. With shared_examples this is required.  That's why I added it.

src/lib/y2storage/storage_manager.rb
```diff
@@ -89,37 +88,15 @@ class StorageManager
     # @param storage_environment [::Storage::Environment]
     def initialize(storage_environment)
       @storage = Storage::Storage.new(storage_environment)
-      apply_storage_defaults
+      @configuration = nil
```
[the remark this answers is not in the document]
ancorgs: Hmmm, no good reason, indeed

PR #988, opened 2019-10-23
Speed up initial proposal on systems with many disks

ancorgs:
## Problem

As reported in [bug#1154070](https://bugzilla.suse.com/show_bug.cgi?id=1154070), the initial storage proposal, before the user has had any chance to select the disks, is very slow in SLE 15 for SAP in systems with many disks (40 in the case of the report). That's a consequence of SLE for SAP using LVM by default, which is much slower than the partition-based proposal.

## Solution

This limits that initial attempt to the first 10 disks on the system in all cases, not only SLE for SAP. To be honest, in any system with so many disks it's kind of expected the user will run the Guided Setup or the Expert Partitioner anyways (the possibilities for the automatic initial proposal to match user expectations with so many disks is almost zero).

## Testing

Tested manually. With 10 disks the calculation is much, much faster.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Oct 23 13:35:49 UTC 2019 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Proposal: in the initial attempt, consider only up to ten disks
+  (bsc#1154070).
+- 4.0.222
+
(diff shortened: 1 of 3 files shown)


ancorgs:
I haven't merged this yet because I was waiting for feedback from the reporter. But that's not happening and the reporters of bug#1156167 have confirmed this fixes the problem for them. We better don't make the latter wait just because the former are not that responsive. So merging.

PR #991, opened 2019-10-28
fix creation of secure key for new partitions (bsc#1154267)

aschnell:
## Problem

The secure key creation fails if the block device does not exist yet.

## Solution

The secure key is first created without the device information and later the device information is added to the secure key.

## Testing

- Updated and added unit tests.
- Tested manually.

### DIFF
--- a/src/lib/y2storage/encryption_processes/pervasive.rb
+++ b/src/lib/y2storage/encryption_processes/pervasive.rb
@@ -51,10 +51,7 @@ module Y2Storage
       # @see Base#pre_commit
       #
       # If there is no secure key to be used for this device, a new key is
-      # generated. In addition to that, the "zkey cryptsetup" command is
-      # executed to know the sequence of commands that must be used to set the
-      # pervasive encryption and the LUKS format options are adjusted
-      # accordingly.
+      # generated.
       #
       # @param device [Encryption] encryption that will be created in the system
       def pre_commit(device)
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/encryption_processes/pervasive.rb
```diff
@@ -63,19 +60,30 @@ def pre_commit(device)
...
+      # extra commands reported by a call to "zkey cryptsetup".
       #
       # @param device [Encryption] encryption that has just been created in the system
       def post_commit(device)
+        if !@secure_key.for_device?(device)
+          secure_key_volume = @secure_key.add_device(device)
+
+          Yast::Execute.locally(ZKEY, "change", "--name", secure_key.name, "--volumes",
+            "+#{secure_key_volume.plain_name.shellescape}:#{secure_key_volume.dm_name.shellescape}")
+        end
```
ancorgs: This looks like logic that should live in the SecureKey class. That's the class that contains all the logic to register the key in the system (by calling "zkey generate") or to copy the keys to the final system at the end of the installation. Having "zkey" commands here to modify in the system the key generated by the `SecureKey` object without telling that to that object looks quite wrong.

src/lib/y2storage/encryption_processes/pervasive.rb
```diff
@@ -63,19 +60,30 @@ def pre_commit(device)
         # API to allow sharing a new key among several volumes.
         @secure_key ||= generate_secure_key(device)
 
-        @zkey_cryptsetup_output = execute_zkey_cryptsetup(device)
-        return if @zkey_cryptsetup_output.empty?
+        master_key_file = "/etc/zkey/repository/#{@secure_key.name}.skey"
```
ancorgs: The `SecureKey` class already contains the knowledge about the location of the repository (`/etc/zkey/repository` by default but maybe orverwritten by `ENV["ZKEY_REPOSITORY"]`) and its format. It would make sense to not duplicate that knowledge/logic here.

src/lib/y2storage/encryption_processes/pervasive.rb
aschnell: But the SecureKey class has no concept of whether the device is added to the key only in memory or on-disk. Prior to the call to add the volume on-disk is a call to the class to add it in-memory. So that looks fine to me.

ancorgs: So maybe we should make the `SecureKey` class smarter or change its API a bit. But generating the system key inside the object, then modifying it from outside and finally using the object again to copy the system key to the final system cannot be fine, IMO.

Moreoever, in the initial implementation, the `SecureKeyVolume` object that is stored inside the `SecureKey` object encapsulated the logic about how the "zkey" command expresses the mapping of devices ("plain_name:dm_name"). Now we are bypassing all that, directly constructing the "zkey" command out of everything. That's why the coverage decreased, because the class that was created to contain the logic about how "zkey" stores the devices is not being asked for that information any longer.

In general, this change is quite disruptive with the original distribution of responsibilities among `Pervasive`, `SecureKey` and `SecureKeyVolume`.

ancorgs: I have just seen that the second part of my comment has been addressed by a subsequent commit.

PR #993, opened 2019-11-06
AutoYaST: avoid repeating filesystem elements for Btrfs multidevice filesystems

imobachgs:
When exporting a Btrfs multidevice filesystem, the filesystem related information will be repeated for each component. For instance, the list of subvolumes will appear in all `<partition>` elements which are part of the Btrfs filesystem. See [bsc#1148578](https://bugzilla.suse.com/show_bug.cgi?id=1148578).

Additionally, this PR prevents exporting the `enable_snapshots` element when the `drive` section does not contain the root filesystem. The reason for this change is that snapshots will be applied [only to the root filesystem](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/proposal/autoinst_drive_planner.rb#L163), so it does not make sense in any other drive.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Wed Nov  6 11:49:35 UTC 2019 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- AutoYaST: do not repeat filesystem related information when
+  cloning multidevice Btrfs filesystems (bsc#1148578).
+- AutoYaST: do not export the enable_snapshots element for drive
+  which do not contain the root filesystem.
+- 4.2.54
+
+-------------------------------------------------------------------
 Wed Oct 30 09:55:48 UTC 2019 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - AutoYaST: add support to set the encryption method (related to
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -3,7 +3,7 @@ Wed Nov  6 11:49:35 UTC 2019 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - AutoYaST: do not repeat filesystem related information when
   cloning multidevice Btrfs filesystems (bsc#1148578).
-- AutoYaST: do not export the enable_snapshots element for drive
+- AutoYaST: do not export the enable_snapshots element for drives
   which do not contain the root filesystem.
 - 4.2.54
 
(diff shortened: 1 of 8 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Wed Nov  6 11:49:35 UTC 2019 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- AutoYaST: do not repeat filesystem related information when
+  cloning multidevice Btrfs filesystems (bsc#1148578).
+- AutoYaST: do not export the enable_snapshots element for drive
+  which do not contain the root filesystem.
```
joseivanlopez: NP: Typo: for drives which do not contain ...

src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -572,13 +572,18 @@ def use_value_from_string(use)
...
+      # for the root filesystem and this setting is specified at
+      # drive section level.
       #
       # @param filesystems [Array<Y2Storage::Filesystem>] Filesystems to evaluate
-      # @return [Boolean] true if snapshots are enabled
+      # @return [Boolean,nil] true if snapshots are enabled; false if they are not enabled;
+      #   nil if the root filesystem is not applicable.
       def enabled_snapshots?(filesystems)
-        filesystems.any? { |f| f.respond_to?(:snapshots?) && f.snapshots? }
+        root_fs = filesystems.find { |f| f.respond_to?(:root?) && f.root? }
```
joseivanlopez: As discussed offline, checking if the filesystem responds to `:root?` can be omitted.

src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -572,13 +572,18 @@ def use_value_from_string(use)
...
+        return nil if root_fs.nil? || (@type != :CT_BTRFS && root_fs.multidevice?)
```
joseivanlopez: NP: please, add a comment to explain the rationale of this condition.

imobachgs: I've have decided to improve code readability. Let me know if it fixes the issue or if a comment is still required.

src/lib/y2storage/autoinst_profile/drive_section.rb
imobachgs: Actually, I found a potential problem when calling `enabled_snapshots?` from `#init_from_disk` after removing the condition.

src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -622,6 +622,13 @@ def use_value_from_storage(disk, partitions)
           "all"
         end
       end
+
+      # Determined whether the section is describine a multi-device Btrfs filesystem
```
joseivanlopez: Typo: describine

joseivanlopez: I would also use: "Determines ..."

PR #996, opened 2019-11-12
AutoYaST: inject proposal settings

imobachgs:
`Y2Storage::AutoinstProposal` receives now the proposal settings. The idea is that, instead of using `Y2Storage::ProposalSettings.new_for_current_product`, it can receive a customized one.

Needed for https://github.com/yast/yast-autoinstallation/pull/544.


imobachgs:
Thanks!

PR #999, opened 2019-11-22
Error message for inactive root

joseivanlopez:
## Problem

Under certain scenarios, the root mount point could be probed as inactive (which does not make sense). One situation where this happens is when the user executes a snapshot rollback and the system has not been rebooted yet. In that case, */etc/fstab* and */proc/mounts* files would contain something like the following:

~~~
$ cat /etc/fstab

/dev/sda1  /  btrfs  defaults  0  0

$ cat /proc/mounts

dev/sda1 / btrfs rw,relatime,space_cache,subvolid=268,subvol=/@/.snapshots/1/snapshot 0 0
~~~

Note that subvolid 268 in */proc/mounts* entry is the Id of the default subvolume before the rollback process. The new default subvolume would have a different Id, and for that reason, the root mount point is probed as inactive (there is no entry in */proc/mounts* for root with the new default subvolume Id).

- Related to [bsc#1124581](https://bugzilla.suse.com/show_bug.cgi?id=1124581)
- See also https://github.com/openSUSE/libstorage-ng/pull/678
- https://trello.com/c/5lTMnxHm/688-3-tw-p2-1124581-problem-with-mount-bind-and-or-docker-and-or-rollback

## Solution

A new check has been added to the Devicegraph sanitizer process. Now, when an inactive root mount point is detected, the user is alerted about that. In case Btrfs is used, a tip about rollback without rebooting is reported as possible reason. 

## Testing

- Added new unit tests
- Tested manually after a rollback

## Screenshots

<details>

<summary>Show/hide</summary>

<h3>Error about inactive root</h3>

VirtualBox_openSUSE Tumbleweed 2_25_11_2019_12_09_35

<h3>Mount point probed as inactive (asterisk in path)</h3> 

VirtualBox_openSUSE Tumbleweed 2_25_11_2019_12_10_35

<h3>Help already explains what asterisk means</h3>

VirtualBox_openSUSE Tumbleweed 2_25_11_2019_12_11_03

</details>

### DIFF
--- a/src/lib/y2storage/devicegraph_sanitizer.rb
+++ b/src/lib/y2storage/devicegraph_sanitizer.rb
@@ -23,6 +23,8 @@ require "y2storage/devicegraph"
 require "y2storage/bcache"
 require "y2storage/bcache_cset"
 
+require "abstract_method"
+
 Yast.import "Mode"
 
 module Y2Storage
@@ -35,8 +37,6 @@ module Y2Storage
   #   sanitizer = DevicegraphSanitizer.new(devicegraph)
   #   new_devicegraph = sanitizer.sanitized_devicegraph
   class DevicegraphSanitizer
-    include Yast::I18n
-
     # @return [Devicegraph]
     attr_reader :devicegraph
 
@@ -44,15 +44,12 @@ module Y2Storage
     #
     # @param devicegraph [Devicegraph] devicegraph to sanitize
     def initialize(devicegraph)
-      textdomain "storage"
-
       @devicegraph = devicegraph
     end
 
     # Errors that need to be fixed in order to obtain a sanitized devicegraph
     #
-    # @return [Array<DevicegraphSanitizer::Error>] empty if the devicegraph is
-    #   already sanitized.
+    # @return [Array<DevicegraphSanitizer::Error>] empty if the devicegraph is already sanitized.
     def errors
       @errors ||= errors_for(devicegraph)
     end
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/devicegraph_sanitizer.rb
```diff
@@ -35,24 +37,19 @@ module Y2Storage
...
-    include Yast::I18n
-
     # @return [Devicegraph]
     attr_reader :devicegraph
 
     # Constructor
     #
     # @param devicegraph [Devicegraph] devicegraph to sanitize
     def initialize(devicegraph)
-      textdomain "storage"
```
shundhammer: Are you sure this is no longer needed?
Aren't there still translated user messages in this file?
Won't those translations break when this `textdomain` is removed?

Not 100% sure, but IIRC we need that `textdomain` once for giving the translation extraction tools (`xgettext`) a hint what `.po` file to put the message to and also for the `_(...)` call to know what file to extract the message from.

Does this still work here without the `textdomain`?

joseivanlopez: Yes, we still need it. But it was simply moved to a more internal class, which has now the translations, see `Error` class in that same file.

shundhammer: At first glance it had appeared to me that this class still has user messages. But you are right: The messages are now all in different classes; in the same file, but in different classes.

PR #1004, opened 2019-12-05
Extend support for Raspberry Pi in the storage proposal

ancorgs:
## Previous status

Since SLE-15-SP1 and openSUSE Leap 15.1, the storage proposal (i.e. the partitioning guided setup) can handle the installation on Raspberry Pi devices, as long as there is an existing firmware partition in the target system. YaST proposes to reuse that partition and to create all the other required partitions. See the "expected workflow" section in [fate#323484](https://fate.suse.com/323484) .

But the technologies included in the upcoming SLE-15-SP2 (and, thus, Leap 15.2) will allow to support more scenarios. (open)SUSE will not longer need a separate firmware partition in all cases. Now it should be possible to make a reasonable installation proposal in all situations, even if the system doesn't contain a firmware partition in advance.

That implies, among other things, enforcing the usage of a MS-DOS partition table in the device used for booting (GPT is useless in that regard).

## Goal

This pull request implements changes in the strategy used to calculate the required boot-related partitions in Raspberry Pi. See, for example, what this new strategy would suggest by default for SLE-15-SP2 Beta1 when installing in a 32 GiB sdcard that contained initially a GPT partition table.

proposal_raspi_32

Being more precise and comprehensive, the strategy in this pull request implements the following behavior, that should allow to propose a proper partitioning layout in all situations:

- if the target boot disk initially contains a MBR partition table
  - and the first partition in the boot disk has id DOS32 and a FAT filesystem
    - and it contains an EFI directory
      - **only requires to use the existing EFI partition (bootcode will be installed there)**
    - and it contains a Raspberry Pi boot code but no EFI directory
      - and there is also a suitable standard EFI partition in the target disk
        - **requires to mount at /boot/vc the firmware partition from the target disk**
        - **requires to use the existing EFI partition**
      - and there is no suitable EFI partition in the target disk
        - **requires to mount at /boot/vc the firmware partition from the target disk**
        - **requires a new /boot/efi partition**
        - **does not enforce the /boot/efi partition to be the first**
        - **requires the /boot/efi partition to have id ESP (according to the EFI standard)**
    - and the first partition contains no boot code or EFI files
      - **requires only a new /boot/efi partition**
      - **requires /boot/efi to be at the start of an MBR partition table, bumping any existing table if needed**
      - **requires the /boot/efi partition to have id DOS32 (bootcode will be installed there)**
    - and there are no partitions in the boot disk
      - **requires only a new /boot/efi partition**
      - **requires /boot/efi to be at the start of an MBR partition table, bumping any existing table if needed**
      - **requires the /boot/efi partition to have id DOS32 (bootcode will be installed there)**
  - and the first partition in the boot disk is a standard EFI with id ESP
    - **requires only a new /boot/efi partition**
    - **requires /boot/efi to be at the start of an MBR partition table, bumping any existing table if needed**
    - **requires the /boot/efi partition to have id DOS32 (bootcode will be installed there)**
- if the target boot disk initially contains a GPT partition table
  - even if the first partition has id DOS32 and a FAT filesystem with an EFI system
    - **requires only a new /boot/efi partition**
    - **requires /boot/efi to be at the start of an MBR partition table, bumping any existing table if needed**
    - **requires the /boot/efi partition to have id DOS32 (bootcode will be installed there)**
  - if there are no partitions in the boot disk
    - **requires only a new /boot/efi partition**
    - **requires /boot/efi to be at the start of an MBR partition table, bumping any existing table if needed**
    - **requires the /boot/efi partition to have id DOS32 (bootcode will be installed there)**
- if the target boot disk contains no partition table initially
  - **requires only a new /boot/efi partition**
  - **requires /boot/efi to be at the start of an MBR partition table, bumping any existing table if needed**
  - **requires the /boot/efi partition to have id DOS32 (bootcode will be installed there)**

- when proposing a new EFI partition
	- **requires /boot/efi to be on the boot disk**
	- **requires /boot/efi to be a non-encrypted vfat partition**
	- **requires /boot/efi to be close enough to the beginning of disk**
	- **requires /boot/efi to be exactly 256 MiB large (min size for FAT32)**

The description above is an excerpt of the specification auto-generated by the test suite included in this pull request.

Relevant links describing the goal:

- Current Jira entry: [SLE-7321](https://jira.suse.com/browse/SLE-7321)
- Original Fate entry: [Fate#327155](https://fate.suse.com/327155)
- Also reported as [bug#1129939](https://bugzilla.suse.com/show_bug.cgi?id=1129939)

## Testing

This pull request includes several unit tests.

Manual tests done with the dedicated YaST client we usually use to test the proposal logic (see screenshot above).

Tested on real hardware with a patched (dud) SLE-15-SP2 Beta1. It works as expected. The resulting system boots... as long as some packages (raspberrypi-* and u-boot-rpi3) are manually selected for installation.


src/lib/y2storage/partition_tables/msdos.rb
```diff
@@ -49,6 +49,13 @@ class Msdos < Base
       # chosen big enough to be on the safe side.
       MBR_GAP_GRUB_LIMIT = DiskSize.KiB(256)
 
+      # Default value used by libstorage-ng to set the minimal MBR grap
+      #
+      # This is equivalent to the static const default_minimal_mbr_gap that is
+      # used internally in Storage::Msdos but not exposed in the API.
+      DEFAULT_MBR_GAP = DiskSize.MiB(1)
```
joseivanlopez: NP: freeze ?

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Dec 13 16:37:32 UTC 2019 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Improved support for Raspberry Pi in the Guided Proposal: now it
+  also works without preexisting firmware partition (jsc#SLE-7298)
+
```
joseivanlopez: Please, indicate the new version here.

ancorgs: Uups

PR #1007, opened 2019-12-11
Improve callback for LUKS activation

joseivanlopez:
NOTE FOR REVIEWERS: unit tests require https://github.com/openSUSE/libstorage-ng/pull/688 to pass.

## Problem

When a LUKS needs to be activated, the user is prompted to give the encryption password. But the popup only shows the encryption UUID, making very difficult to identify the device when the systems contains several encrypted devices. 

* Part of https://trello.com/c/MR24HhqL/1486-2-research-p4-1125774-sles-15-sp1-beta3-improve-naming-of-crypt-devices-in-the-passphrase-prompt

* Related to https://bugzilla.suse.com/show_bug.cgi?id=1125774

* Requires https://github.com/openSUSE/libstorage-ng/pull/688

## Solution

Now on, the activation popup also shows the name of the underlying device.

## Testing

- Unit tests adapted.
- Tested manually

## Screenshots

<details>

<summary>Show/Hide</summary>

VirtualBox_openSUSE Tumbleweed_11_12_2019_15_43_29

VirtualBox_openSUSE Tumbleweed 2_12_12_2019_17_09_29

</details>

### DIFF
--- a/src/lib/y2storage/callbacks/activate.rb
+++ b/src/lib/y2storage/callbacks/activate.rb
@@ -18,10 +18,10 @@
 # find current contact information at www.suse.com.
 
 require "yast"
+require "yast2/popup"
+
 require "y2storage/dialogs/callbacks/activate_luks"
 require "y2storage/callbacks/libstorage_callback"
-
-Yast.import "Popup"
 
 module Y2Storage
   module Callbacks
@@ -52,12 +52,12 @@ module Y2Storage
         return true if forced_multipath?
         return false unless looks_like_real_multipath
 
-        Yast::Popup.YesNo(
-          _(
-            "The system seems to have multipath hardware.\n"\
-            "Do you want to activate multipath?"
-          )
+        message = _(
+          "The system seems to have multipath hardware.\n" \
+          "Do you want to activate multipath?"
         )
+
+        Yast2::Popup.show(message, buttons: :yes_no) == :yes
       end
 
       # Decides whether a LUKS device should be activated
@@ -68,6 +68,9 @@ module Y2Storage
       # @return [Storage::PairBoolString]
       def luks(info, attempt)
         log.info("Trying to open luks UUID: #{info.uuid} (#{attempt} attempts)")
+
+        luks_error(attempt) if attempt > 1
+
         dialog = Dialogs::Callbacks::ActivateLuks.new(info, attempt)
         result = dialog.run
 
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/callbacks/activate.rb
```diff
@@ -56,9 +60,15 @@ def multipath(looks_like_real_multipath)
...
-        log.info("Trying to open luks UUID: #{uuid} (#{attempt} attempts)")
-        dialog = Dialogs::Callbacks::ActivateLuks.new(uuid, attempt)
+      # Decides whether a LUKS device should be activated
+      #
+      # @param info [Storage::LuksInfo]
+      # @param attempt [Numeric]
+      #
+      # @return [Storage::PairBoolString]
+      def luks(info, attempt)
+        log.info("Trying to open luks UUID: #{info.uuid} (#{attempt} attempts)")
```
dgdavid: NP: shall we uses `attempt(s)` instead?

```
Trying to open luks UUID: 2c8bc367-1058-46c5-8d3b-f8b75242dc97 (1 attempt(s))
```

Just asking, I have no strong opinion. Although I think we already discussed something similar in the past...

dgdavid: Forget it, I just realize it is a `log` line.

joseivanlopez: Yes, right now only the log informs about the number of attempts. In fact, nothing indicates to the user whether the password was wrong, but simply the popup is shown again. I think we can do it better when the user gives a wrong password:

a) explicitly telling to the user that the password was wrong and
b) informing about how many attempts have been performed.

@ancorgs what do you think?

ancorgs: I agree

dgdavid: > I agree

Me too

src/lib/y2storage/callbacks/activate.rb
```diff
@@ -48,24 +52,50 @@ def multipath(looks_like_real_multipath)
...
+
+      private
+
+      # Error popup when the LUKS could not be activated
+      #
+      # @param attempt [Numeric] current attempt
+      def luks_error(attempt)
+        message = format(
+          _("The encrypted volume could not be activated (attempt number %{attempt}).\n\n" \
+            "Please, make sure you are giving the correct password."),
```
ancorgs: "you are giving" is a little bit weird. I would say "you entered"

PR #1012, opened 2019-12-17
Wrapping error

jreidinger:
## Problem

stderr of some programs can be so long that it does not fit on the screen and so it is useless for users.

- https://bugzilla.suse.com/show_bug.cgi?id=1085468

## Solution

Use shared functionality to [wrap text](https://github.com/yast/yast-yast2/blob/6bc7bd9f1ef03f783ed269cd59174fe9e8a25c78/library/general/src/lib/ui/text_helpers.rb#L34).

## Testing

- Unit tests included.

## Screenshots

before:
Snímek obrazovky_2019-12-17_14-07-15

after:
Snímek obrazovky_2019-12-17_14-09-06

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Tue Dec 17 12:48:40 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
 
-- Wrap long details message when error happen (bsc#1085468)
+- Wrap long details message when error happens (bsc#1085468)
 - 4.2.64
 
 -------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


test/y2storage/callbacks/callbacks_examples.rb
```diff
@@ -52,6 +52,29 @@
     end
   end
 
+  # see https://bugzilla.suse.com/show_bug.cgi?id=1085468
+  context "with an long error" do
```
joseivanlopez: NP: typo "with a~n~ long error"

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Dec 17 12:48:40 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
+
+- Wrap long details message when error happen (bsc#1085468)
```
joseivanlopez: NP: "error happens"

PR #1013, opened 2019-12-19
improved sorting of devices by device size

aschnell:
So far devices in tables were sorted alphabetical when the user selected sorting by size. That does not give good results. Now the sorting is done numerical.

Note: There is already some code included to also better sort by name. The missing code is easy but adapting the testsuite needs a lot more work.

size

Also see https://github.com/libyui/libyui/pull/156.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Dec 19 17:51:09 CET 2019 - aschnell@suse.com
+
+- Improved sorting of devices by device size (bonus to bsc#1140018)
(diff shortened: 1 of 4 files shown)


aschnell:
Locally the unit tests work fine. Analyzing the failure is a task for next year or never.

jreidinger:
@aschnell few notes to failure on travis:

1. segfault is caused that unit tests hold failing object and try to write object.inspect for details of failure, but underlaying libstorage object is already released. So it access 0 pointer.
2. What usually helps if it build locally and not on travis, is to run rspec tests in same order as on travis as some tests somehow incorrectly depends on certain state.

PR #1014, opened 2019-12-20
Unit test to demostrate bug#1115749

ancorgs:
## Problem

We were not sure whether [bug#1115749](https://bugzilla.suse.com/show_bug.cgi?id=1115749) was been solved or was still current.

## Solution

Added a pull request to prove the bug is still present in recent versions of yast2-storage-ng


ancorgs:
Closed in favor of #1056 which includes this together with the fix.

PR #1017, opened 2019-12-23
Improved sorting by device name and size in tables

aschnell:
So far devices in tables were sorted alphabetical when the user selected sorting by name or size. That does not give good results. Now the sorting is done better, e.g. numerical for size.

name

size

Also see libyui/libyui#156.

### DIFF
--- a/test/y2partitioner/test_helper.rb
+++ b/test/y2partitioner/test_helper.rb
@@ -1,5 +1,5 @@
 #!/usr/bin/env rspec
-# Copyright (c) [2017] SUSE LLC
+# Copyright (c) [2017-2019] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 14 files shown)


aschnell:
Travis is using libstoage-ng 4.2.40 which is more than a week old. Locally the tests work fine.

test/y2partitioner/test_helper.rb
```diff
@@ -19,3 +19,27 @@
...
+
+  term.params.delete_if { |param| param.is_a?(Yast::Term) && param.value == :sortKey }
+  return term if term.params.size > 1
+
+  term.params[0]
+end
+
+# Removes the :sortKey term from :cell terms somewhere inside (to the
+# extent needed so far) of value. Also see remove_sort_key.
+def remove_sort_keys(value)
```
mchf: is it really needed only in tests or I've missed something?

mchf: ah forgot it ... I see now.

aschnell: Yes, it is really only needed in tests.

When not running tests the sort-key should be passed to libyui.

PR #1018, opened 2020-01-09
NFS Client: update the UI after discarding a created but not reachable entry

dgdavid:
####  :warning: Closed after [move the "RefreshUI" action to the NfsClient4partClient](https://github.com/yast/yast-nfs-client/pull/87/commits/ec9bb889bf73c0c91024f6b696ab3de208f40d62) itself. Motivation in [comments](https://github.com/yast/yast-storage-ng/pull/1018#discussion_r364660436)

---

## Problem

Partitioner ends with a crash when the user edits a _phantom_ NFS entry.

For more information, see the description of https://github.com/yast/yast-nfs-client/pull/87, which is ~~a required part of~~ the solution.

  * Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1156446
  * Trello: https://trello.com/c/WGmxRkxu

## Solution

~~To update the UI after discarding a not reachable NFS entry, calling to the _RefreshUI_ action added in [yast2-nfs-client 4.1.6](https://github.com/yast/yast-nfs-client/pull/87). As nfs entries are previously sync via _FromStorage_ action, the useless entry will be removed from the UI.~~

## Testing

- Tested manually

### DIFF
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -32,6 +32,8 @@ Requires:	libstorage-ng-ruby >= 4.1.89
 Requires:	rubygem(ruby-dbus)
 # Y2Packager::Repository
 Requires:	yast2-packager >= 3.3.7
+# NfsClient4partClient RefreshUI action
+Requires:	yast2-nfs-client >= 4.1.6
 # findutils for xargs
 Requires:	findutils
 
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:        4.1.90
+Version:        4.1.91
 Release:	0
 
 BuildRoot:	%{_tmppath}/%{name}-%{version}-build
--- a/src/lib/y2partitioner/yast_nfs_client.rb
+++ b/src/lib/y2partitioner/yast_nfs_client.rb
@@ -213,6 +213,9 @@ module Y2Partitioner
 
     # Handler for the 'add' button
     #
+    # Note that UI will reflect the added entry immediately, even before test if it is reachable.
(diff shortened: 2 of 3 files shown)


package/yast2-storage-ng.spec
```diff
@@ -32,6 +32,8 @@ Requires:	libstorage-ng-ruby >= 4.1.89
 Requires:	rubygem(ruby-dbus)
 # Y2Packager::Repository
 Requires:	yast2-packager >= 3.3.7
+# NfsClient4partClient RefreshUI action
+Requires:	yast2-nfs-client >= 4.1.6
```
dgdavid: As discussed in IRC, this require will produce a cyclic dependency since `yast-nfs-client` [already requires `yast-storage-ng`](https://github.com/yast/yast-nfs-client/blob/23eb048f5f52629d0e5fba2e7c9e773c0fa1d469/package/yast2-nfs-client.spec#L45).

src/lib/y2partitioner/yast_nfs_client.rb
```diff
@@ -230,6 +233,7 @@ def newbut_handler(legacy)
       log.info "Test mount failed, so removing NFS from current graph: #{nfs.inspect}"
       current_graph.remove_nfs(nfs)
       set_client_list
+      call_client("RefreshUI")
```
dgdavid: Due to the cyclic dependency commented above, we are going to move this call to the NfsClient4partClient itself, just at the end of [FromStorage](https://github.com/yast/yast-nfs-client/blob/1e2fadaaf16f2d2eddd87065fdfed3f05785fc7a/src/clients/nfs-client4part.rb#L38) action. If it works, this PR can be discarded and the fix will live only in yast-nfs-client.

dgdavid:
No longer needed. See description and comments to understand why.

PR #1019, opened 2020-01-09
disallow to mount BitLocker

aschnell:
## Problem

Now that libstorage-ng can represent BitLocker in the devicegraph it is possible to try to mount it in the expert partitioner.

## Solution

Disallow mounting of BitLocker in the expert partitioner using the new ```supports_mount?``` function from libstorage-ng.

## Testing

- Tested manually

### DIFF
--- a/src/lib/y2storage/filesystems/base.rb
+++ b/src/lib/y2storage/filesystems/base.rb
@@ -45,6 +45,10 @@ module Y2Storage
       #   @return [Filesystems::Type]
       storage_forward :type, as: "Filesystems::Type"
 
+      # @!method supports_mount?
+      #   @return [Boolean]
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/filesystems/base.rb
```diff
@@ -45,6 +45,10 @@ class Base < Mountable
       #   @return [Filesystems::Type]
       storage_forward :type, as: "Filesystems::Type"
 
+      # @!method supports_mount?
+      #   @return [Boolean]
+      storage_forward :supports_mount?
```
joseivanlopez: I expected this forward method directly in `Mountable` class.

aschnell: Indeed.

PR #1021, opened 2020-01-14
Improve integration of NFS client

joseivanlopez:
## Problem

The NFS client (yast2-nfs-client) can be used standalone or through the Partitioner. But the behaviour is not exactly the same in both cases.

With the standalone mode:

* All nfs shares are deleted and created again.
* Current unmounted shares are mounted after saving the changes, even though the nfs entry is not edited.
* Current shares that are not included in the fstab are written to fstab after saving the changes.

With the Expert Partitioner:

* Only edited entries are modified after saving.
* Current unmounted shares are mounted only if their server or path change.
* Current shares that are not included in the fstab are written to fstab only if their server or path change.

Related bugs:

* https://bugzilla.suse.com/show_bug.cgi?id=1006815
* https://bugzilla.suse.com/show_bug.cgi?id=1151426

Part of PBI: https://trello.com/c/Z8gqtcC1/1484-3-nfs-client-and-unmounting-bug1006815-bug1151426

## Solution

This PR unifies the behavior. Now, in both cases (standalone and inside the Partitioner):

* Current unmounted shares keep unmounted after editing them, even when the server or path changes.
* Current shares that are not included in the fstab are never written to the fstab, even when the server or path changes.

These changes imply that an unmounted share cannot be mounted. For that, the user should delete the entry and create it again (from the UI, of course).

Note that libstorage-ng 4.1.107 is required to avoid some mount/unmount issues, see https://github.com/openSUSE/libstorage-ng/pull/699.

## Testing

- Added new unit tests
- Tested manually

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jan 16 11:48:39 UTC 2020 - José Iván López González <jlopez@suse.com>
+
+- Improve integration with yast2-nfs-client.
+- Related bugs: bsc#1006815, bsc#1151426.
+- 4.1.91
+
+-------------------------------------------------------------------
 Wed Oct 23 11:51:58 UTC 2019 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - AutoYaST: consider CT_DMMULTIPATH an alias of CT_DISK (related
--- a/test/y2partitioner/widgets/pages/nfs_mounts_test.rb
+++ b/test/y2partitioner/widgets/pages/nfs_mounts_test.rb
@@ -434,11 +434,8 @@ describe Y2Partitioner::Widgets::Pages::NfsMounts do
         end
 
         it "edits the NFS mount in the current devicegraph" do
-          expect(nfs_object.mount_point.path).to_not eq("/foo")
-
-          nfs_page.handle(event)
-
-          expect(nfs_object.mount_point.path).to eq("/foo")
+          expect { nfs_page.handle(event) }.to change(nfs_object.mount_point, :path)
+            .from("/test1").to("/foo")
         end
 
         it "keeps the mount status" do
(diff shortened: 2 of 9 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Jan 16 11:48:39 UTC 2020 - José Iván López González <jlopez@suse.com>
+
+- Improve integration with yast-nfs-client.
```
jreidinger: I would use rpm name here. so `yast2-nfs-client`

test/y2partitioner/widgets/pages/nfs_mounts_test.rb
```diff
@@ -409,6 +415,81 @@ def includes_error_msg?(term)
...
+
+          expect(previous_sid).to eq(sid)
+        end
+
+        it "edits the NFS mount in the current devicegraph" do
+          expect(nfs_object.mount_point.path).to_not eq("/foo")
+
+          nfs_page.handle(event)
+
+          expect(nfs_object.mount_point.path).to eq("/foo")
```
jreidinger: you can use here nicer "expect().to change" - https://relishapp.com/rspec/rspec-expectations/v/2-0/docs/matchers/expect-change

test/y2partitioner/widgets/pages/nfs_mounts_test.rb
```diff
@@ -409,6 +415,81 @@ def includes_error_msg?(term)
...
+        end
+
+        it "creates a new NFS mount in the current devicegraph" do
+          nfs = Y2Storage::Filesystems::Nfs.find_by_server_and_path(device_graph, "srv", "/home/b")
+          expect(nfs).to be_nil
+
+          nfs_page.handle(event)
+
+          nfs = Y2Storage::Filesystems::Nfs.find_by_server_and_path(device_graph, "srv", "/home/b")
+          expect(nfs).to_not be_nil
```
jreidinger: also here change expectation can be used.

test/y2partitioner/widgets/pages/nfs_mounts_test.rb
```diff
@@ -434,11 +434,8 @@ def includes_error_msg?(term)
         end
 
         it "edits the NFS mount in the current devicegraph" do
-          expect(nfs_object.mount_point.path).to_not eq("/foo")
-
-          nfs_page.handle(event)
-
-          expect(nfs_object.mount_point.path).to eq("/foo")
+          expect { nfs_page.handle(event) }.to change(nfs_object.mount_point, :path)
```
jreidinger: change have to use `{}`

jreidinger: ah, maybe not in this case.

PR #1023, opened 2020-01-16
treat partitions with BitLocker as Windows

aschnell:
## Problem

Partitions with BitLocker were not treated as Windows. One notable consequence of that is that BitLocker partitions could be removed by the storage proposal even when the user instructed YaST to not modify Windows.

## Solution

Treat partitions with BitLocker as Windows.

## Testing

- Added a new unit test
- Tested manually

## See also

https://trello.com/c/urYYuPsW/1562-3-detecting-bitlocker-part-of-bug1159318-expert-partitioner-let-me-resize-bitlocker-partition-cause-data-lost

### DIFF
--- a/test/y2storage/filesystems/type_test.rb
+++ b/test/y2storage/filesystems/type_test.rb
@@ -260,7 +260,7 @@ describe Y2Storage::Filesystems::Type do
       expect(described_class.windows_filesystems).to be_a(Array)
     end
 
-    it "only includes ntfs and vfat" do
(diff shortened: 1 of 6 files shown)


test/y2storage/filesystems/type_test.rb
```diff
@@ -261,7 +261,8 @@
     end
 
     it "only includes ntfs and vfat" do
```
jreidinger: this needs to be also updated

PR #1026, opened 2020-01-20
Split partitioner into another package

hellcp:
## Problem

boo#1160101

## Solution

Added new package and transferred desktop file there

## Testing

- *Tested manually*

### DIFF
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -68,13 +68,13 @@ This package contains the files for YaST2 that handle access to disk
 devices during installation and on an installed system.
 This YaST2 module uses libstorage-ng.
 
-%package -n yast2-partitioner-ng
-Summary:        YaST2 - Storage Configuration
+%package -n yast2-partitioner
+Summary:        YaST2 - Partitioning Client
 Group:          System/YaST
 
 Obsoletes:      yast2-partitioner
 
-%description -n yast2-partitioner-ng
+%description -n yast2-partitioner
 This module provides partitioner for yast2-storage-ng
 
 %prep
@@ -109,7 +109,7 @@ rake test:unit
 %doc README.md
 %doc CONTRIBUTING.md
 
-%files -n yast2-partitioner-ng
+%files -n yast2-partitioner
 %{yast_desktopdir}
 
 %changelog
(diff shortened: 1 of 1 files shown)


package/yast2-storage-ng.spec
```diff
@@ -101,4 +109,7 @@ rake test:unit
 %doc README.md
 %doc CONTRIBUTING.md
 
+%files -n yast2-partitioner-ng
```
ancorgs: I would remove the "-ng" part here. There was never a yast2-partitioner package in the past, so no need to distinguish it.

package/yast2-storage-ng.spec
```diff
@@ -68,6 +68,15 @@ This package contains the files for YaST2 that handle access to disk
 devices during installation and on an installed system.
 This YaST2 module uses libstorage-ng.
 
+%package -n yast2-partitioner-ng
```
ancorgs: I would remove the "-ng" part here. There was never a yast2-partitioner package in the past, so no need to distinguish it. Not everything relying on storage-ng has to receive the "-ng" suffix. :-)

ancorgs:
So the only goal of this is to "hide" the Partitioner from the end users? The first thing the user sees when executing the Partitioner is a quite intimidating and verbose text explaining how wrong can everything go. And the user has to click on "yes" to the question: "Continue despite this warning?"

Do we need to, in addition, "hide" the icon in a separate package?

hellcp:
Yes, it is mainly the issue of duplication, partitioning tools that don't require root or don't have to have those warning already exist in the desktop managers and aren't needed for the desktop use cases.

ancorgs:
I understand you point, but I find the above reasoning a bit weak to justify a change in how we package the YaST tools (is one of those changes that may look trivial but ends up causing problems to more than one user and/or tool).

We are already considering to change that initial warning to make it nicer and more useful. When doing so, we can mention that there are desktop tools more adequate for smaller tasks.

But the package split wouldn't be our preferred solution. So closing the PR for the time being.

hellcp:
Eh, I disagree, there is no valid reasons in showing the user the option to launch the program if the first thing they see is a warning they shouldn't use it.

PR #1027, opened 2020-01-20
Test to demostrate bsc#1161331

ancorgs:
See https://bugzilla.suse.com/show_bug.cgi?id=1161331


ancorgs:
Closed in favor of #1043 that shows the expected behavior in a more detailed way... and that proves that was indeed the behavior in SLE-15-SP1

PR #1028, opened 2020-01-21
added warning before resizing block devices which were probed empty

aschnell:
## Problem

We had a bug report where a user shrunk a partition with BitLocker on it (which caused data loss since BitLocker was not detected).

## Solution

The data loss is now prevented by two changes:

- BitLocker is detected (other PRs)
- A warning is presented to the user when trying to resize a device which seems to be empty (this PR).

## Testing

- Added a new unit test.
- Tested manually.

## See also

- https://bugzilla.suse.com/show_bug.cgi?id=1159318
- https://trello.com/c/CEJP4jFv/1575-1-warn-when-resizing-empty-partitions-2nd-part-of-bug1159318-expert-partitioner-let-me-resize-bitlocker-partition-cause-data-loss
- https://trello.com/c/urYYuPsW/1562-3-detecting-bitlocker-part-of-bug1159318-expert-partitioner-let-me-resize-bitlocker-partition-cause-data-lost

### DIFF
--- a/src/lib/y2partitioner/actions/resize_blk_device.rb
+++ b/src/lib/y2partitioner/actions/resize_blk_device.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2017-2019] SUSE LLC
+# Copyright (c) [2017-2020] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -48,12 +48,15 @@ module Y2Partitioner
 
       # Checks whether it is possible to resize the device, and if so, the action is performed.
       # It also checks whether the device needs to be unmounted before resizing (e.g., NTFS).
-      # When unmounted device is required, it offers the option for unmounting it.
+      # When unmounted device is required, it offers the option for unmounting it. If the
+      # device seems to be empty (see confirm_empty) the user may cancel the action after
+      # a warning.
       #
-      # @note An error popup is shown when the device cannot be resized.
+      # @note An error popup is shown when the device cannot be resized or seems empty.
       #
       # @return [Symbol, nil]
       def run
+        return :back unless confirm_empty
         return :back unless try_unmount && validate
 
         resize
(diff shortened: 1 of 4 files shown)


src/lib/y2partitioner/actions/resize_blk_device.rb
```diff
@@ -76,6 +79,37 @@ def resize
...
+      # @return [Boolean] true iff there is no high risk or user accepted high risk
+      def confirm_empty
+        # If the device has descendants now resizing is ok even if
+        # nothing was probed on the device since the user already made
+        # an operation that overrides the device.
+        return true if @device.descendants.any?
+
+        # If the device does not exist in probed there cannot be an
+        # undetected file system or storage system on it.
+        return true if !@device.exists_in_probed?
```
joseivanlopez: I would avoid to directly use the probed devicegraph when working with the Expert Partitioner. The Partitioner is designed to work with two given devicegraphs, one representing the system status (normally the probed one) and the second one is the working graph (equivalent to staging). So, I would modify the code a little bit, for example:

```
system_device = DeviceGraphs.instance.system.find_device(@device.sid)

return true if !system_device || system_device.descendants.any?

```

test/y2partitioner/actions/resize_blk_device_test.rb
```diff
@@ -275,6 +275,17 @@ def create_partition(disk_name)
       end
     end
 
+    context "when executed on an 'empty' partition" do
+      let(:scenario) { "mixed_disks.yml" }
+
+      let(:device_name) { "/dev/sdb7" }
+
+      it "shows a warning popup" do
+        expect(Yast2::Popup).to receive(:show)
```
joseivanlopez: I would make sure that the correct popup is used:

```
        expect(Yast2::Popup).to receive(:show).with(/No file system nor any storage system was detected/, anything)
```

aschnell: Sure, I will also make that 20 lines further down.

PR #1031, opened 2020-01-22
Partitioner: remember active tab and selected row per node

dgdavid:
## Problem

The UI only can remember the selected tab for the current node. Thus, if the user changes from, let's say, _Hard Disks -> sda -> `Partitions`_ to _Hard Disks -> sdb_ and then comes back to _Hard Disks -> sda_, the _`Overview`_ tab is shown instead of _`Partitions`_. Exactly the same happens for the selected row: only is remembered when the current node is _redrawn_, being forgotten when changing to another node.

- https://bugzilla.suse.com/show_bug.cgi?id=1159883

## Solution

Keep the _status_ per node to be able to restore, when possible, the UI as it was before leaving the node. The solution includes

* To "register" the system page too, needed to track its status
* To restore `.label` class method for some pages (kind of revert of https://github.com/yast/yast-storage-ng/pull/752/commits/b107fc5b13662b10fe8904f7d6331a2b057eaa61)
* To fix the _dependency cycles in Ruby requires_ between UIState and some actions 
* To "discard" the `UIState` before re-scanning devices
* An UI blink mitigation

## Testing

- Added unit tests
- Tested manually (both, during installation and in a running system)

## Screencast

<details>

---

ezgif com-video-to-gif
</details>

---

<details>
<summary>Show/hide some notes about UI blinks</summary>

---

This set of changes also includes a commit mitigating the UI blinks originated by the fact that tabs are being rendered in two steps and the first one (before rendering the content from the _initial tab_):

* was being shown in a vertically centered position.
  
  <details>

    ---
    
    vertically_centered_first_render
  </details>
  
* does not occupy the full available space.
  
  <details>

    ---
    
    dumbtab_first_render

  </details>
  

Starting always in the top left corner of its container avoids the _first blink_. Actually, we only can see one of them because the _second render step_ fits all the space with the tab content. Sadly, fixing the first makes the second "visible", which probably will require changes in YaST2 CWM library.

---

Note: you cannot see the mentioned blink in the attached screencast because of the frame rate. But you can see it in the original video  (if available yet) [here](https://www.veed.io/download/d4b7ad53-596b-4e74-a10f-6e60fa4791eb).

</details>

### DIFF
--- a/src/lib/y2partitioner/actions/go_to_device_tab.rb
+++ b/src/lib/y2partitioner/actions/go_to_device_tab.rb
@@ -46,7 +46,7 @@ module Y2Partitioner
 
         # First, pretend the user visited the device and then the tab...
         state = UIState.instance
-        state.go_to_tree_node(target_page)
+        state.select_page([device.sid])
         state.switch_to_tab(tab_label)
 
         # ...then trigger a redraw
--- a/src/lib/y2partitioner/actions/go_to_device_tab.rb
+++ b/src/lib/y2partitioner/actions/go_to_device_tab.rb
@@ -46,7 +46,7 @@ module Y2Partitioner
 
         # First, pretend the user visited the device and then the tab...
         state = UIState.instance
-        state.select_page([target_page.id])
+        state.select_page(target_page.tree_path)
         state.switch_to_tab(tab_label)
 
         # ...then trigger a redraw
--- a/src/lib/y2partitioner/actions/go_to_device_tab.rb
+++ b/src/lib/y2partitioner/actions/go_to_device_tab.rb
@@ -46,7 +46,7 @@ module Y2Partitioner
 
         # First, pretend the user visited the device and then the tab...
         state = UIState.instance
-        state.select_page([device.sid])
+        state.select_page([target_page.id])
         state.switch_to_tab(tab_label)
 
         # ...then trigger a redraw
--- a/src/lib/y2partitioner/ui_state.rb
+++ b/src/lib/y2partitioner/ui_state.rb
@@ -24,9 +24,6 @@ module Y2Partitioner
   class UIState
     # A collection holding a PageStatus for each Page visited by the user
     #
-    # The Widgets::Pages::Base#id is used as index
-    # @see #status_for
-    #
     # @return [Array<PageStatus>]
     attr_reader :statuses
 
@@ -59,7 +56,7 @@ module Y2Partitioner
     # Method to be called when the user operates in a row of a table of devices
     # or creates a new device.
     #
-    # @param device [Y2Storage::Device, Integer] sid or device object
+    # @param device [Y2Storage::Device, Integer] a device or its sid
     def select_row(device)
       sid = device.respond_to?(:sid) ? device.sid : device
 
@@ -72,7 +69,7 @@ module Y2Partitioner
     # It remembers the decision so the user is taken back to a sensible point of
     # the tree (very often the last he decided to visit) after redrawing.
     #
-    # @param pages_ids [Array<String, Integer>] a list of Page#id identifying the path to the page
+    # @param pages_ids [Array<String, Integer>] the path to the selected page
     def select_page(pages_ids)
       self.current_status = status_for(pages_ids)
     end
@@ -96,10 +93,12 @@ module Y2Partitioner
     # Select the page to open in the general tree after a redraw
     #
     # @param pages_ids [Array<String, Integer>] all pages ids in the tree
-    # @return [page_id, nil] the page to be opened; the initial one when nil
+    # @return [String, Integer, nil] the page id to be opened or nil
     def find_page(pages_ids)
       return nil unless current_status
 
+      # Let's pick the more accurate, which means that it is present in both the tree and the
+      # candidates. See PageStatus#candidate_pages.
       (current_status.candidate_pages & pages_ids).last
     end
 
@@ -142,18 +141,16 @@ module Y2Partitioner
 
     # Returns the status representation for a page
     #
-    # @param pages_ids [Array<String, Integer>] a list of Page#id identifying the path to the page
+    # @param pages_ids [Array<String, Integer>] the path to the page. See PageStatus#candidate_pages
     # @return [PageStatus] the current status if it already exists; a new one when not.
     def status_for(pages_ids)
       id = pages_ids.last
       status = statuses.find { |s| s.page_id == id }
 
-      if status.nil?
-        status = PageStatus.new(id, pages_ids)
-        statuses << status
-      end
+      return status unless status.nil?
 
-      status
+      statuses << PageStatus.new(id, pages_ids)
+      statuses.last
     end
 
     class << self
@@ -173,7 +170,7 @@ module Y2Partitioner
       private :new, :allocate
     end
 
-    # Represent the UI status for a CWM::Page
+    # Represents the UI status for a CWM::Page
     #
     # For the time being, it is able to keep
     #
@@ -192,26 +189,29 @@ module Y2Partitioner
       # @return [String]
       attr_accessor :active_tab
 
-      # The Page#id
+      # The Widgets::Pages::Base#id
       #
-      # @return [String, Integet>]
+      # @return [String, Integer]
       attr_reader :page_id
 
-      # The path to a page, useful to correctly place the user within the tree after redrawing the
-      # UI and also to remove useless statuses after deleting a device.
+      # A partial path to a page, useful to correctly place the user within the tree after
+      # redrawing the UI and also to remove useless statuses after deleting a device.
+      #
+      # This path always contains the page parent id and the page id itself. So, taking a partition
+      # as an example, it will be [disk_page_id, partition_page_id]
       #
       # @see UIState#find_page
       # @see UIState#clear_statuses_for
       #
       # It stores page ids, see Pages::Base#id
       #
-      # @return [Array<Integer, String>]
+      # @return [Array<String, Integer>]
       attr_reader :candidate_pages
 
       # Constructor
       #
-      # @param page_id [String, Integer] the Page#id identifying the page
-      # @param candidate_pages_ids [Array<String, Integer>] a list of Page#id for candidate pages
+      # @param page_id [String, Integer] the Widgets::Pages::Base#id identifying the page
+      # @param candidate_pages_ids [Array<String, Integer>] a list of Widgets::Pages::Base#id
       def initialize(page_id, candidate_pages_ids)
         @page_id = page_id
         @candidate_pages = candidate_pages_ids
--- a/src/lib/y2partitioner/ui_state.rb
+++ b/src/lib/y2partitioner/ui_state.rb
@@ -22,9 +22,9 @@ module Y2Partitioner
   # information that needs to be remembered across UI redraws to give the user a
   # sense of continuity.
   class UIState
-    # A collection holding a PageStatus for each CWM::Page visited by the user
+    # A collection holding a PageStatus for each Page visited by the user
     #
-    # The CWM::Page#widget_id is used as index
+    # The Widgets::Pages::Base#id is used as index
     # @see #status_for
     #
     # @return [Hash{String => PageStatus}]
@@ -77,9 +77,9 @@ module Y2Partitioner
       self.current_status = status_for(pages_ids)
     end
 
-    # Method to be called when the user switches to a tab within a tree node
+    # Method to be called when the user switches to a tab within a page
     #
-    # It remembers the decision so the same tab is showed when the same node will be redraw.
+    # It records the decision, so the last active tab is displayed when the page will be redraw.
     #
     # @param label [String]
     def switch_to_tab(label)
@@ -96,7 +96,7 @@ module Y2Partitioner
     # Select the page to open in the general tree after a redraw
     #
     # @param pages_ids [Array<String, Integer>] all pages ids in the tree
-    # @return [CWM::Page, nil] the page to be opened; the initial one when nil
+    # @return [page_id, nil] the page to be opened; the initial one when nil
     def find_page(pages_ids)
       return nil unless current_status
 
@@ -185,13 +185,13 @@ module Y2Partitioner
       # @return [String]
       attr_accessor :active_tab
 
-      # The path to a node, useful to correctly place the user within the tree after redrawing the
+      # The path to a page, useful to correctly place the user within the tree after redrawing the
       # UI and also to remove useless statuses after deleting a device.
       #
       # @see UIState#find_page
       # @see UIState#clear_statuses_for
       #
-      # It could hold both, a device id (sid, Integer) or a page label (String).
+      # It stores page ids, see Pages::Base#id
       #
       # @return [Array<Integer, String>]
       attr_reader :candidate_pages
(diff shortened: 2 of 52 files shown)


src/lib/y2partitioner/widgets/tabs.rb
```diff
@@ -11,7 +11,7 @@ module TabsWithState
       # Overrides default behavior of tabs to register the new state before
       # doing the real switch
       def switch_page(page)
-        UIState.instance.switch_to_tab(page)
+        UIState.instance.switch_to_tab(page.label)
```
joseivanlopez: Why not page.id?

src/lib/y2partitioner/widgets/pages/page.rb
```diff
@@ -0,0 +1,48 @@
...
+
+require "cwm/widget"
+require "cwm/pager"
+require "cwm/tree_pager"
+
+module Y2Partitioner
+  module Widgets
+    module Pages
+      # A base class for partitioner pages
+      class Page < CWM::Page
```
joseivanlopez: NP: I would call this class Pages::Base, similar to we have done for other base clases.

src/lib/y2partitioner/widgets/pages/page.rb
```diff
@@ -0,0 +1,48 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "cwm/widget"
```
joseivanlopez: Are all these requires needed?

src/lib/y2partitioner/ui_state.rb
```diff
@@ -19,22 +19,16 @@
...
-    # Called through {.create_instance}, starts with a blank situation (which
-    # means default for each widget will be honored).
-    def initialize
-      textdomain "storage"
-
-      @candidate_nodes = []
-      @open_items = {}
-      @overview_tree_pager = nil
-    end
+    # The CWM::Page#widget_id is used as index
```
joseivanlopez: Page#id, isn’t it?

src/lib/y2partitioner/ui_state.rb
```diff
@@ -51,124 +45,102 @@ def initialize
...
+    # @param page_id [String, Integer] id to identify the page status to track
+    def select_page(page_id)
+      self.current_status = status_for(page_id)
     end
 
     # Method to be called when the user switches to a tab within a tree node.
     #
-    # It remembers the decision so the same tab is showed in case the user stays
-    # in the same node after redrawing.
+    # It remembers the decision so the same tab is showed when the same node will be redraw.
```
joseivanlopez: NP: ... when the current node will be redraw...

src/lib/y2partitioner/ui_state.rb
```diff
@@ -51,124 +45,102 @@ def initialize
...
     #
     # @param [CWM::Page, String] page associated to the tab (or just its label)
     def switch_to_tab(page)
-      self.tab = page.respond_to?(:label) ? page.label : page
+      current_status&.active_tab = page.respond_to?(:label) ? page.label : page
```
joseivanlopez: I guess the parameter is the page is, so we can assign it directly.

src/lib/y2partitioner/ui_state.rb
```diff
@@ -51,124 +45,102 @@ def initialize
...
+    def row_sid
+      current_status&.selected_device
     end
 
     # Select the page to open in the general tree after a redraw
     #
-    # If the first candidate page/node (the "current" one) is not available anymore, the tab name
-    # and selected row are cleared. See #tab=
-    #
     # @param pages [Array<CWM::Page>] all the pages in the tree
```
joseivanlopez: The parameter should be a list of page ids.

src/lib/y2partitioner/ui_state.rb
```diff
@@ -51,124 +45,102 @@ def initialize
...
     # @return [CWM::Page, nil] the page to be opened; the initial one when nil
```
joseivanlopez: And the result a page id, or nil.

src/lib/y2partitioner/ui_state.rb
```diff
@@ -51,124 +45,102 @@ def initialize
...
       nil
     end
 
-    # Select the tab to open within the node after a redraw
+    # Select the tab to open within the page after a redraw
     #
     # @param pages [Array<CWM::Page>] pages for all the possible tabs
     # @return [CWM::Page, nil]
     def find_tab(pages)
+      tab = current_status&.active_tab
```
joseivanlopez: The same here.

src/lib/y2partitioner/ui_state.rb
```diff
@@ -19,22 +19,16 @@
...
-      textdomain "storage"
-
-      @candidate_nodes = []
-      @open_items = {}
-      @overview_tree_pager = nil
-    end
+    # The CWM::Page#widget_id is used as index
+    # @see #status_for
+    #
+    # @return [Hash{String => PageStatus}]
```
joseivanlopez: Why a hash? I would say we can use an array of PageStatus. We can retrieve a PageStatus by the page id.

src/lib/y2partitioner/ui_state.rb
```diff
@@ -266,5 +191,136 @@ def create_instance
...
+      # @return [String]
+      attr_accessor :active_tab
+
+      # The path to a node, useful to correctly place the user within the tree after redrawing the
+      # UI and also to remove useless statuses after deleting a device.
+      #
+      # @see UIState#find_page
+      # @see UIState#clear_statuses_for
+      #
+      # It could hold both, a device id (sid, Integer) or a page label (String).
```
joseivanlopez: I simply would say that it stores page ids, see Pages::Base#id.

src/lib/y2partitioner/ui_state.rb
```diff
@@ -266,5 +191,136 @@ def create_instance
...
+        else
+          device_page_section(device)
+        end
+      end
+
+      # @see #device_page_candidates
+      # @see #device_page_parent
+      #
+      # @return [Integer, String, nil] nil if there is no parent tree entry
+      def device_page_section(device)
```
joseivanlopez: Oh no! We still have a pages dependency!

src/lib/y2partitioner/ui_state.rb
dgdavid: Right. This was a leftover. Fixed, thanks.

src/lib/y2partitioner/ui_state.rb
dgdavid: Yes, you are right.

But, after having a look, I'd keep the current approach because it looks more performant to me. Changing to a not indexed structure force us to iterates over it until find the desired PageStatus.

Do you agree?

joseivanlopez: Umm, I would prefer a plain array. It is easier to understand, and there will not be performance issues in that find actions. Only in case of having hundreds of pages we would notice some difference, otherwise such performance penalty would not be noticeable.

src/lib/y2partitioner/ui_state.rb
dgdavid: ### :skull:

src/lib/y2partitioner/ui_state.rb
dgdavid: I'm partially agree with you. Let's go for a plain array.

Thanks!

src/lib/y2partitioner/widgets/pages/page.rb
dgdavid: I think so. They were there in almost all pages inheriting from CWM::Page.

Anyway, let's re-check it.

src/lib/y2partitioner/widgets/tabs.rb
dgdavid: Because `page` here is not a `Widgets::Pages::Base` but `Widgets::PartitionsTab` (a `CWM::Tab`). And tabs has not `.id` method.

dgdavid: So... 

The method should be `switch_tab` and the param `tab`

Isn't it?

src/lib/y2partitioner/widgets/pages/page.rb
dgdavid: We need at least to require [`cwm/tree_pager`](https://github.com/yast/yast-yast2/blob/master/library/cwm/src/lib/cwm/tree_pager.rb), which requires `cwm/page`, `cwm/pager`, and `cwm/tree`. Or require them explicitly.

dgdavid: I mean, to require both, `cwm/page` and `cwm/tree_pager`

src/lib/y2partitioner/widgets/pages/lvm_vg.rb
```diff
@@ -60,20 +60,31 @@ def label
...
-            Left(
-              HBox(
-                Image(Icons::LVM, ""),
-                Heading(format(_("Volume Group: %s"), "/dev/" + @lvm_vg.vg_name))
+          Top(
+            VBox(
+              Left(
+                HBox(
+                  Image(Icons::LVM, ""),
+                  Heading(format(_("Volume Group: %s"), "/dev/" + @lvm_vg.vg_name))
```
joseivanlopez: Please, use `@lvm_vg.name`.

src/lib/y2partitioner/widgets/pages/base.rb
```diff
@@ -0,0 +1,82 @@
...
+          end
+        end
+
+        # The full path to reach the page within the tree
+        #
+        # Useful to know where to place the user after redrawing the UI. See
+        # {Y2Partitioner::UIState::PageStatus#candidate_pages}
+        #
+        # @return [Array<String, Integer>]
+        def parents
```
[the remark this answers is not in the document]
dgdavid: As we agreed in a chat after this review, we are going to rename it to `tree_path` (among other changes related to this methods)

src/lib/y2partitioner/widgets/pages/base.rb
```diff
@@ -0,0 +1,82 @@
...
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "cwm/widget"
+require "cwm/pager"
+require "cwm/tree_pager"
```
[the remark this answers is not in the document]
dgdavid: I tried, but seems not to be enough

```ruby
NameError:
  uninitialized constant CWM::TreePager
  Did you mean?  CWM::TreeItem
```

See https://github.com/yast/yast-yast2/blob/master/library/cwm/src/lib/cwm.rb and https://github.com/yast/yast-yast2/blob/master/library/cwm/src/lib/cwm/widget.rb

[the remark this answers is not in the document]
joseivanlopez: Oh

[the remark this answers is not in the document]
dgdavid: So, we need to require "cwm/page" and "cwm/tree_pager"

src/lib/y2partitioner/widgets/overview.rb
```diff
@@ -67,15 +77,25 @@ def initialize(hostname)
         super(OverviewTree.new(items))
       end
 
-      # Pages whose cached content should be considered outdated
+      # Ensures that UIState clears obsolte statuses and is aware of current page
```
joseivanlopez: Typo: ... obsolete statuses and it is aware ...

src/lib/y2partitioner/widgets/overview.rb
```diff
@@ -67,15 +77,25 @@ def initialize(hostname)
...
       #
-      # This is a hack introduced because the NFS page works in a completely
-      # different way in which triggering a full redraw every time something
-      # changes is not an option. This way, the NFS page can invalidate the
-      # cached contents of other pages supporting this mechanism.
+      # That's especially needed when initial page is a candidate of a no longer exist one
```
joseivanlopez: Typo: of a no longer existing one.

src/lib/y2partitioner/widgets/overview.rb
```diff
@@ -67,15 +77,25 @@ def initialize(hostname)
...
       #
-      # @return [Array<Symbol>] only :system supported so far
-      attr_accessor :invalidated_pages
+      # @see #initial_page
+      def init
+        super
+
+        UIState.instance.clear_dead_statuses(@pages.map(&:id))
+        UIState.instance.select_page(@current_page.tree_path)
```
joseivanlopez: NP: thinking about the method name, what about a `UIState#prune` method with parameters?

~~~
def prune(keep: [])
  statuses.select! { |s| keep.include?(s.page_id) }
end
~~~

By default, `#prune` would delete everything, but we can use `:keep` parameter to filter out the pages to keep.

Anyway, this is a NP, I am also fine with current version.

src/lib/y2partitioner/widgets/pages/base.rb
```diff
@@ -0,0 +1,84 @@
...
+        # {Y2Partitioner::UIState::PageStatus#candidate_pages}
+        #
+        # @return [Array<String, Integer>]
+        def tree_path
+          [parent, id].compact
+        end
+
+        private
+
+        # The parent page i
```
joseivanlopez: Leftover i.

src/lib/y2partitioner/actions/go_to_device_tab.rb
```diff
@@ -46,7 +46,7 @@ def run
 
         # First, pretend the user visited the device and then the tab...
         state = UIState.instance
-        state.go_to_tree_node(target_page)
+        state.select_page([target_page.id])
```
joseivanlopez: Isn't it possible to use here `#tree_path`? `        state.select_page(target_page.tree_path)`

dgdavid: Sure!

src/lib/y2partitioner/actions/go_to_device_tab.rb
```diff
@@ -46,7 +46,7 @@ def run
 
         # First, pretend the user visited the device and then the tab...
         state = UIState.instance
-        state.select_page([target_page.id])
+        state.select_page([target_page.tree_path])
```
joseivanlopez: I think you don't need to create a hash here because `#tree_path` already returns a hash.

dgdavid: Sorry, I overlooked that. You are right, #tree_path already provides the needed structure. 

Thanks for catching it!

rwx788:
In case anyone is interested, this PR broke Factory First strategy and we got this change in SLES already, but still in review for TW. So can be used as a real example for the further process improvements.

joseivanlopez:
> In case anyone is interested, this PR broke Factory First strategy and we got this change in SLES already, but still in review for TW. So can be used as a real example for the further process improvements.

To be honest, all that OBS/IBS stuff is a mystery for me. We did two request (as everyday):

* From YaST:Head to openSUSE:Factory
* From Devel:YaST:Head to SUSE:SLE-15-SP2:GA

But IMHO, we should have a direct request from openSUSE:Factory to SUSE:SLE-15-SP2:GA. Maybe something to discuss in the future.

This problem would be more evident in case of SUSE would release its own rolling release.

PR #1033, opened 2020-01-24
Select APQNs for a new secure key

joseivanlopez:
## Problem

When a device is going to be encrypted with pervasive encryption (s390), a new secure key is generated for the device (if there is none yet). By default, this new secure key will be generated for all adapters/domains available in the system. But this could fail when the adapters/domains are set with different master keys. In that case, only adapters using the same master key should be taken into account for generating the new secure key.  

* https://jira.suse.com/browse/IBM-495?focusedCommentId=974832&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-974832

* https://trello.com/c/KwTmeWhi/1587-5-sle-7376-pervasive-encryption-second-round-specifying-the-apqns

## Solution

Now, the dialog for pervasive encryption allows to select APQNs when a new secure key is going to be generated. An error is shown when the secure key cannot be generated with the selected APQNs.

Theoretically, *zkey generate* command should fail when the APQNs used to generate the secure key are configured with different master keys. But, according to my tests, the command does not fail in that case. It only fails when none of the given APQNs are configured with a master key.   

## Testing

* Added unit tests.
* Manually tested.

## Screenshots

<details>
<summary>Show/Hide</summary>

Screenshot from 2020-01-27 15-15-45

Screenshot from 2020-02-28 17-34-28

</details>

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/encryption.rb
+++ b/src/lib/y2partitioner/actions/controllers/encryption.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2019] SUSE LLC
+# Copyright (c) [2019-2020] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -20,6 +20,8 @@
 require "y2storage"
 require "y2partitioner/actions/controllers/base"
 require "y2storage/encryption_type"
+require "y2storage/encryption_processes/secure_key"
+require "y2storage/encryption_processes/apqn"
 
 module Y2Partitioner
   module Actions
@@ -48,6 +50,11 @@ module Y2Partitioner
         # @return [String] Password for the encryption device
         attr_accessor :password
 
+        # Selected APQNs to generate a new secure key for pervasive encryption
+        #
+        # @return [Array<Y2Storage:.EncryptionProcesses::Apqn>]
+        attr_accessor :apqns
+
         # Contructor
         #
         # @param fs_controller [Filesystem] see {#fs_controller}
@@ -59,6 +66,7 @@ module Y2Partitioner
           @action = actions.first
           @password = encryption&.password || ""
           @method = initial_method
+          @apqns = initial_apqns
         end
 
         # Whether the dialog to select and configure the action makes sense
@@ -123,6 +131,52 @@ module Y2Partitioner
           blk_device.encryption
         end
 
+        # Existing secure key for the encrypted device
+        #
+        # @return [Y2Storage::EncryptionProcesses::SecureKey, nil] nil if no secure key is associated to
+        #   the device.
+        def secure_key
+          Y2Storage::EncryptionProcesses::SecureKey.for_device(blk_device)
+        end
+
+        # Currently available APQNs for generating a new secure key
+        #
+        # @return [Array<Y2Storage::EncryptionProcesses::Apqn>]
+        def online_apqns
+          @online_apqns ||= Y2Storage::EncryptionProcesses::Apqn.online
+        end
+
+        # Finds an online APQN by its name
+        #
+        # @param name [String] APQN name (e.g., "01.0001")
+        # @return [Y2Storage::EncryptionProcesses::Apqn, nil]
+        def find_apqn(name)
+          online_apqns.find { |a| a.name == name }
+        end
+
+        # Tests the generation of a secure key
+        #
+        # Note that the command for generating a secure key might fail when the APQNs are not correctly
+        # configured. An APQN without a master key makes the command to tail. Theoretically, all the
+        # APQNs used for generating the secure key must have the same master key, but for some reason
+        # the command is not failing when different master keys are used.
+        #
+        # @param apqns [Array<Y2Storage::EncryptionProcesses::Apqn>]
+        # @return [String, nil] error message or nil if the secure key can be generated
+        def test_secure_key_generation(apqns: [])
+          key_name = "yast2_tmp_secure_key_test"
+
+          begin
+            key = Y2Storage::EncryptionProcesses::SecureKey.generate!(key_name, apqns: apqns)
+          rescue Cheetah::ExecutionFailed => e
+            return e.message
+          end
+
+          key.remove
+
+          nil
+        end
+
         # Title to display in the dialog during the process
         #
         # @return [String]
@@ -156,6 +210,17 @@ module Y2Partitioner
           else
             Y2Storage::EncryptionMethod::LUKS1
           end
+        end
+
+        # Currently used APQNs when the device is encrypted with pervasive encryption
+        #
+        # @return [Array<Y2Storage::EncryptionProcesses::Apqn>]
+        def initial_apqns
+          process = encryption&.encryption_process
+
+          return [] unless process&.respond_to?(:apqns)
+
+          process.apqns
         end
 
         # Calculate actions that make sense for the block device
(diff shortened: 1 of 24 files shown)


src/lib/y2partitioner/widgets/apqn_selector.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      # @macro seeAbstractWidget
+      def init
+        enable_on_init ? enable : disable
+        self.value = controller.apqns
+      end
+
+      # @return [Array<String, String>]
+      def items
+        controller.online_apqns.map { |d| [d.name, d.name] }
```
imobachgs: I see this is a common pattern, so I guess it is OK as it is. However, I do not see the point in having to rely on a controller to get the items. Why do not pass the list of apqns in the constructor? IMHO, it would make the widget more reusable: it simply gets a list of apqns and the `#value` method returns the selected ones. No controller needed at all except for `#store`, and that could takes us to another discussion :)

src/lib/y2partitioner/widgets/apqn_selector.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+
+      # Saves the selected APQNs into the controller
+      def store
+        controller.apqns = value
+      end
+
+      private
+
+      # @return [Actions::Controllers::ClonePartitionTable]
```
imobachgs: I guess the type should be `Actions::Controllers::Encryption`, right?

src/lib/y2storage/blk_device.rb
```diff
@@ -305,10 +305,19 @@ def encrypted?
     # @param method [EncryptionMethod, Symbol] encryption method to create the new device
     # @param dm_name [String, nil] DeviceMapper table name of the new device
     # @param password [String, nil] password of the new device
+    # @param apqns [Array<EncryptionProcesses::Apqn>] APQNs to use when generating a secure key for
+    #   pervasive encryption.
+    #
     # @return [Encryption]
-    def encrypt(method: EncryptionMethod::LUKS1, dm_name: nil, password: nil)
+    def encrypt(method: EncryptionMethod::LUKS1, dm_name: nil, password: nil, apqns: [])
```
imobachgs: np: So the `apqns` argument is only used depending on the `method` value. For the time being I guess it is OK, but in the future (if we get more specific options for each method) we could think about doing some refactoring (like having specific classes or methods for encryption method options. But, for the time being, let's keep it simple.

src/lib/y2storage/encryption_processes/apqn.rb
```diff
@@ -0,0 +1,147 @@
...
+      # Location of the lszcrypt command
+      LSZCRYPT = "/sbin/lszcrypt".freeze
+      private_constant :LSZCRYPT
+
+      class << self
+        # All APQNs found in the system
+        #
+        # @return [Array<Apqn>]
+        def all
+          create_apqns
```
imobachgs: Using `create_` as the method name is misleading, as it makes me think that we are actually creating the APQNs in the system. But what we are doing is creating the objects, right?

src/lib/y2storage/encryption_processes/apqn.rb
```diff
@@ -0,0 +1,147 @@
...
+        # @return [Array<Array<String>>]
+        def apqns_data
+          data = execute_lszcrypt.split("\n").map(&:split)
+          return [] if data.size == 2
+
+          # remove header
+          data.shift(2)
+
+          # select card.domain entries
+          data.select { |d| d.first.match?(/.*\..*/) }
```
imobachgs: Are `.01` or `01.` valid names? If you always need something before and after the point, you could consider using `.+` instead of `.*`. If it does not matter, I would simply use `d.first.include?(".")`.

src/lib/y2storage/encryption_processes/apqn.rb
```diff
@@ -0,0 +1,147 @@
...
+      attr_reader :status
+
+      # Constructor
+      #
+      # @param name [String]
+      # @param type [String]
+      # @param mode [String]
+      # @param status [String]
+      # @param _others [Array<String>]
+      def initialize(name, type, mode, status, *_others)
```
imobachgs: You are introducing a `*_others` argument just for convenience, to be able to do `new(*d)`. To be honest, I would rather prefer a more explicit approach and not polluting the arguments list. For instance (there are more ways to do so):

```ruby
name, type, mode, status, _ = d
new(name, type, mode, status)
```

If you still want to have such a convenience constructor, I would write it as a separate one that does the lifting and finally relies on the regular `new`.

test/y2partitioner/widgets/encrypt_method_options_test.rb
```diff
@@ -46,10 +48,12 @@
...
     let(:fake_luks1_options) { CWM::Empty.new("__fake_luks1_options__") }
+    let(:fake_pervasive_options) { CWM::Empty.new("__fake_pervasive_options__") }
 
     before do
       allow(Y2Partitioner::Widgets::SwapOptions).to receive(:new).and_return(fake_random_swap_options)
       allow(Y2Partitioner::Widgets::LuksOptions).to receive(:new).and_return(fake_luks1_options)
+      allow(Y2Partitioner::Widgets::PervasiveOptions).to receive(:new).and_return(fake_pervasive_options)
     end
 
     it "generates the new content based on the selected method" do
```
imobachgs: np: I would probably write separate tests for each case. But they are simple enough, so let's keep them as they are.

src/lib/y2partitioner/actions/controllers/encryption.rb
```diff
@@ -123,6 +131,52 @@ def encryption
...
+        def find_apqn(name)
+          online_apqns.find { |a| a.name == name }
+        end
+
+        # Tests the generation of a secure key
+        #
+        # Note that the command for generating a secure key might fail when the APQNs are not correctly
+        # configured. An APQN without a master key makes the command to tail. Theoretically, all the
+        # APQNs used for generating the secure key must have the same master key, but for some reason
+        # the command is not failing when different master keys are used.
```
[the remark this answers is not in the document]
joseivanlopez: You are right. Actually, this problem is already commented in the Jira entry, see https://jira.suse.com/browse/SLE-7376?focusedCommentId=985691&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-985691. So, maybe we can simply drop it out from here. In the end, it should be a "temporary" misbehavior.

src/lib/y2partitioner/widgets/apqn_selector.rb
joseivanlopez: I don't like either to have a widget depending on a controller. But as you say, it is the common pattern we have decided to follow to avoid having logic scattered in all widgets around. Anyway, this widget will not be re-usable without a controller, so I would prefer to simply pass the controller in the constructor (as for the rest of widgets) and use that controller to get/set all the information.

src/lib/y2storage/encryption_processes/apqn.rb
joseivanlopez: Yes. In fact, I was trying to find a good name during a while, but I see that I was not able. What about `read_apqns` ?

src/lib/y2storage/encryption_processes/apqn.rb
joseivanlopez: Yes, `.+` is the correct regexp here.

PR #1034, opened 2020-01-24
reduce size of dialogs

aschnell:
## Problem

Some dialogs do not fit on small screens, neither ncurses nor qt UI.

- https://trello.com/c/YWiwQnng/1580-3-leap-151-p3-1135723-installation-fstab-options-window-extended-below-screen-bottom-on-1280x800-screen
- https://bugzilla.suse.com/show_bug.cgi?id=1135723
- https://bugzilla.suse.com/show_bug.cgi?id=1161651

## Solution

Drop some obsolete options:

- noatime (likely not use often since relatime is the default for years)
- acl and xattrs (default for years; not supported for UDF)

Replace the five radio buttons for the mount by method by a combobox.

User can still enable options using the text field for arbitrary options.

## Testing

- Adapted unit test.
- Tested manually.

## Screenshots

ncurses

### DIFF
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2017] SUSE LLC
+# Copyright (c) [2017-2020] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -212,7 +212,6 @@ module Y2Partitioner
             VSpacing(1),
             Left(GeneralOptions.new(@controller)),
             Left(FilesystemsOptions.new(@controller)),
-            Left(AclOptions.new(@controller)),
             * ui_term_with_vspace(JournalOptions.new(@controller)),
             Left(ArbitraryOptions.new(@controller, self))
           )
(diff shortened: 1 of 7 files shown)


src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -381,59 +380,41 @@ def label
...
-                  Left(RadioButton(Id(:path), _("Device &Path")))
-                )
-              )
-            )
-          )
-        )
-      end
+      def items
+        # CWM does not support symbols for entries in ComboBoxes
+        # contrary to libyui. Otherwise a few conversations between
```
dgdavid: NP: do you mean "conversions"?

src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -381,59 +380,41 @@ def label
...
+        # string and symbol below could be avoided.
 
-      def selected_mount_by
-        Y2Storage::Filesystems::MountByType.find(value)
-      end
+        suitable = mount_point.suitable_mount_bys(label: true, encryption:
+          @controller.encrypt).map { |mount_by| mount_by.to_sym.to_s }
```
dgdavid: If I'm not wrong, you can avoid `#to_sym` here.

```ruby 
mount_point.suitable_mount_bys(...).map(&:to_s)
```

should work.

PR #1035, opened 2020-01-28
Added package requirement for UF_PLAIN_ENCRYPTION (bsc#1160773)

cwh42:
## Problem

When user tries to encrypt a swap partition using YaST the error message "Command not found: "/sbin/cryptsetup…" appears instead of a request to install that package.

libstorage suggests UF_PLAIN_ENCRYPTION for encrypting swap, but the package requirement for that feature was missing in yast-storage-ng. So the package `cryptsetup` was not installed before commiting the changes.

- [bsc#1160773](https://bugzilla.suse.com/show_bug.cgi?id=1160773)

## Solution

Added package requirement for `cryptsetup` to UF_PLAIN_ENCRYPTION in yast-storage-ng.

By this opportunity removed the obsolete feature UF_CRYPT_TWOFISH which libstorage does not use any more.

## Testing

- Tested manually

## Screenshots

image

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jan 29 10:00:18 UTC 2020 - Christopher Hofmann <cwh@suse.com>
+
+- Added package requirement for UF_PLAIN_ENCRYPTION (bsc#1160773)
+- 4.2.80
+
+-------------------------------------------------------------------
 Tue Jan 28 20:20:50 CET 2020 - aschnell@suse.com
 
 - also mask swap units in mask-systemd-units (bsc#1152545)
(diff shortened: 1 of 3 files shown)


ancorgs:
The fix works... but the explanation is a little bit misleading. The Partitioner does not try to use any UsedFeature. That is not how it works.

It's always libstorage-ng who reports to the Partitioner which UsedFeatures will be used during the commit phase. And then the Partitioner uses that information to tell the user that some packages need to be installed.

The problem here is that libstorage-ng was reporting that the feature UF_PLAIN_ENCRYPTION was going to be used and the Partitioner didn't know that the "cryptsetup" package was a requisite to use that feature, so it didn't tell the user to install the package.

UF_CRYPT_TWOFISH may be a leftover and it's good that we used this as a chance to remove it. But it has nothing to do with the problem. The Partitioner cannot have decided to use UF_CRYPT_TWOFISH... or any other feature. It only works the other way around (is the library who decides which features are used).

cwh42:
Now I thing I got it and fixed the commit desctription accordingly.

PR #1037, opened 2020-01-29
Add commit progress dialog

joseivanlopez:
## Problem

When using the Expert Partitioner in an already installed system, the last screen is a summary about all the actions to apply. But once this summary is accepted, the Expert Partitioner is automatically closed, and there is no hint about what is happening under the hood. Were all the changes correctly applied? Has it already finished? Or is it still working? No way to know.

* https://bugzilla.suse.com/show_bug.cgi?id=1135366 

* https://trello.com/c/Xpjm7Kdr/1579-5-1135366-ux-issue-when-applying-changes-with-expert-partitioner

## Solution

The old Partitioner had an extra dialog after the summary to show the progress of the changes being applied to the system. A similar dialog has been brought back.

These changes require the new progress bar widget defined here https://github.com/yast/yast-yast2/pull/1022.

## Testing

- Added unit tests
- Tested manually

## Screenshots

<details>

<summary>Show/Hide</summary>

VirtualBox_openSUSE Tumbleweed_04_02_2020_13_52_33

</details>

### DIFF
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -28,8 +28,8 @@ Source:         %{name}-%{version}.tar.bz2
 # Fix mount/unmount
 BuildRequires:	libstorage-ng-ruby >= 4.2.54
 BuildRequires:  update-desktop-files
-# for CWM sort_key helper
-BuildRequires:  yast2 >= 4.2.48
+# CWM::DynamicProgressBar
+BuildRequires:  yast2 >= 4.2.63
 BuildRequires:  yast2-devtools >= 4.2.2
 # for AbortException and handle direct abort
 BuildRequires:  yast2-ruby-bindings >= 4.0.6
@@ -49,8 +49,8 @@ BuildRequires:  rubygem(%{rb_default_ruby_abi}:parallel_tests)
 Requires:       findutils
 # Fix mount/unmount
 Requires:       libstorage-ng-ruby >= 4.2.54
-# for CWM sort_key helper
-Requires:       yast2 >= 4.2.48
+# CWM::DynamicProgressBar
+Requires:       yast2 >= 4.2.63
 # Y2Packager::Repository
 Requires:       yast2-packager >= 3.3.7
 # for AbortException and handle direct abort
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-storage-ng
-Version:        4.2.81
+Version:        4.2.82
 Release:        0
 Summary:        YaST2 - Storage Configuration
 License:        GPL-2.0-only OR GPL-3.0-only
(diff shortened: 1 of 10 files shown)


src/lib/y2partitioner/dialogs/commit.rb
```diff
@@ -0,0 +1,70 @@
...
+    # Dialog to show the progress of the storage changes while they are being applied
+    class Commit < CWM::Dialog
+      include Yast::I18n
+
+      def initialize
+        textdomain "storage"
+      end
+
+      def title
+        _("Performing Installation")
```
dgdavid: As far as I understood, this new dialog is only shown in a running system. Isn't it? If so, should we choose a different title?

package/yast2-storage-ng.spec
```diff
@@ -28,8 +28,8 @@ Source:         %{name}-%{version}.tar.bz2
 # Fix mount/unmount
 BuildRequires:	libstorage-ng-ruby >= 4.2.54
 BuildRequires:  update-desktop-files
-# for CWM sort_key helper
-BuildRequires:  yast2 >= 4.2.48
+# CWM::DynamicProgessBar
```
jreidinger: typo Progess

src/lib/y2partitioner/dialogs/commit.rb
joseivanlopez: Yes, it is only used in a running system. I have simply used the same title as the old partitioner, but I agree, it seems this is not the most accurate name in the world. What about "Applying Changes To The System".

I am not sure about capitalizing "To The", what do you think?

dgdavid: "Applying Changes to the System" looks good to me. 

https://english.stackexchange.com/questions/14/which-words-in-a-title-should-be-capitalized

src/lib/y2storage/storage_manager.rb
```diff
@@ -278,20 +278,21 @@ def committed?
...
     #
-    # @return [Boolean] whether commit was successfull, false if libstorage-ng
-    #   found a problem and it was decided to abort due to that
-    def commit(force_rw: false)
+    # @param force_rw [Boolean] if mount points should be forced to have read/write permissions.
+    # @param callbacks [Y2Storage::Callbacks::Commit]
+    #
+    # @return [Boolean] whether commit was successful, false if libstorage-ng found a problem and it was
+    #   decided to abort.
+    def commit(force_rw: false, callbacks: nil)
```
jreidinger: question is if you want here to allow passing explicit nil or not. If not, then you can add this call to method signature like `def commit(force_rw: false, callbacks: Callbacks::Commit.new)` and it is still backward compatible.

joseivanlopez: Well, libstorage-ng does not force to pass a callbacks, so maybe in the future we decide to allow `nil` (although is not very likely).

PR #1038, opened 2020-01-31
Partitioner: fix button to delete all partitions (bsc#1162290) 

ancorgs:
## Problem

The "Delete all" button in the "Partitions" tab of any partitionable device failed when there were logical partitions involved.

- https://bugzilla.suse.com/show_bug.cgi?id=1162290

The Partitioner got the list of all partitions in the device and then iterated over that list deleting each. But after having deleted the extended partitions, the logical partitions were not longer there. Still, the Partitioner tried to delete the not-longer-existent partitions because they were in its list.

## Solution

Simply use the method `PartitionTables:Base#delete_all_partitions` that already have all that into account.

## Testing

- Added a new unit test
- Tested manually also

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan 31 16:49:28 UTC 2020 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Partitioner: the button to delete all partitions does no longer
+  fail for devices containing logical partitions (bsc#1162290).
+- 4.2.81
+
+-------------------------------------------------------------------
 Tue Jan 29 10:00:18 UTC 2020 - Christopher Hofmann <cwh@suse.com>
 
 - Added package requirement for UF_PLAIN_ENCRYPTION (bsc#1160773)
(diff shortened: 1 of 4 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Jan 31 16:49:28 UTC 2020 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Partitioner: the button to delete all partitions do not longer
```
jreidinger: typo "no longer"

ancorgs: "does no longer", in fact :-)

src/lib/y2partitioner/actions/delete_partitions.rb
```diff
@@ -39,7 +39,7 @@ def initialize(*args)
       def delete
         log.info "deleting partitions from #{device}"
 
-        device.partitions.each { |p| device.partition_table.delete_partition(p) }
```
jreidinger: good catch. So it delete extended and then it tries to delete logical which is no longer there.

PR #1039, opened 2020-02-03
improved message

aschnell:
Improved recently added message.

### DIFF
--- a/src/lib/y2partitioner/actions/resize_blk_device.rb
+++ b/src/lib/y2partitioner/actions/resize_blk_device.rb
@@ -101,7 +101,7 @@ module Y2Partitioner
         return true if probed.find_device(@device.sid).descendants.any?
 
         message = _(
-          "No file system nor any storage system was detected on the\n" \
+          "Neither a file system nor a storage system was detected on the\n" \
           "device. In case the device does contain a file system or a\n" \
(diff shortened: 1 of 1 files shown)


src/lib/y2partitioner/actions/resize_blk_device.rb
```diff
@@ -101,7 +101,7 @@ def confirm_empty
         return true if probed.find_device(@device.sid).descendants.any?
 
         message = _(
-          "No file system nor any storage system was detected on the\n" \
+          "Neither a file system nor a storage system was detected on the\n" \
```
ancorgs: This one is more correct. From that point of view, it LGTM.

But I think "No storage system or file system was detected" is more straightforward to read. I personally find the proposed solution to be a little bit too formal compared to the language we usually use at pop-ups. I think an english-oriented Rubocop would say "_Perceived complexity for the sentence is too high_"... although it's maybe my very own impression.

shundhammer: Arvin's version is more precise.

And yes, quite a number of our pop-up messages leave a lot to be desired; we are sorely missing regular proofreading from a native English speaker. Back many years ago Rebecca Walther did that on a regular basis, but AFAIK she no longer works for us.

shundhammer:
LGTM

PR #1041, opened 2020-02-06
allow to install packages needed for probing

aschnell:
For https://trello.com/c/ae4cAtFe/1573-5-improve-probing-when-there-are-missing-tools.

## Problem

Probing can fail if programs are missing.

## Solution

Allow user to install missing packages.

## Testing

- Added a new unit test.
- Tested manually.

## Picture

popup

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Feb 06 15:09:57 CET 2020 - aschnell@suse.com
+
+- allow to install packages needed for probing
+- 4.2.84
+
+-------------------------------------------------------------------
 Wed Feb  5 12:38:51 UTC 2020 - David Diaz <dgonzalez@suse.com>
 
 - Partitioner: usability improved by remembering the last active
--- a/src/lib/y2storage/callbacks/probe.rb
+++ b/src/lib/y2storage/callbacks/probe.rb
@@ -35,7 +35,7 @@ module Y2Storage
       #   (in the ASCII-8BIT encoding! see https://sourceforge.net/p/swig/feature-requests/89/)
       # @param what [String] details coming from libstorage-ng (in the ASCII-8BIT encoding!)
       # @param command [String] missing command coming from libstorage-ng (in the ASCII-8BIT encoding!)
-      # @param used_features used features bit field coming from libstorage-ng
+      # @param used_features [Integer] used features bit field as integer coming from libstorage-ng
       #
       # @return [Boolean] true will make libstorage-ng ignore the error, false
       #   will result in a libstorage-ng exception
--- a/src/lib/y2storage/callbacks/probe.rb
+++ b/src/lib/y2storage/callbacks/probe.rb
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 SUSE LLC
+# Copyright (c) [2018,2020] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -19,11 +19,116 @@
 
 require "y2storage/callbacks/libstorage_callback"
 
+Yast.import "Mode"
+Yast.import "Label"
+Yast.import "Popup"
+
 module Y2Storage
   module Callbacks
     # Class to implement callbacks used during libstorage-ng probe
-    class Probe < Storage::ProbeCallbacks
+    class Probe < Storage::ProbeCallbacksV3
       include LibstorageCallback
+
+      # Callback for missing commands during probing.
+      #
+      # @param message [String] error title coaming from libstorage-ng
+      #   (in the ASCII-8BIT encoding! see https://sourceforge.net/p/swig/feature-requests/89/)
+      # @param what [String] details coming from libstorage-ng (in the ASCII-8BIT encoding!)
+      # @param command [String] missing command coming from libstorage-ng (in the ASCII-8BIT encoding!)
+      # @param used_features used features bit field coming from libstorage-ng
+      #
+      # @return [Boolean] true will make libstorage-ng ignore the error, false
+      #   will result in a libstorage-ng exception
+      #
+      def missing_command(message, what, command, used_features)
+        textdomain "storage"
+
+        # force the UTF-8 encoding to avoid Encoding::CompatibilityError exception
+        message.force_encoding("UTF-8")
+        what.force_encoding("UTF-8")
+        command.force_encoding("UTF-8")
+
+        log.info "libstorage-ng reported a missing command, asking the user whether to continue"
+        log.info "Error details. message: #{message}. what: #{what}. command: #{command}. "\
+                 "used_features: #{used_features}."
+
+        # Redirect to error callback if no packages can be installed.
+        return error(message, what) if used_features == 0 || Yast::Mode.installation
+
+        package_handler = Y2Storage::PackageHandler.new
+        package_handler.add_feature_packages(used_features)
+
+        # Redirect to error callback if no packages can be installed.
+        return error(message, what) if package_handler.compact.empty?
+
+        description = _("An external command required for probing is missing. When\n"\
+                        "continuing despite the error, the presented system information\n"\
+                        "will be incomplete. You may also install the required packages\n"\
+                        "and restart probing.")
+
+        what += "\n\n" + missing_command_handle_packages_text(package_handler)
+
+        question = _("Continue despite the error, install required packages or abort?")
+        buttons = { continue: Yast::Label.ContinueButton, install: _("Install Packages"),
+                    abort: Yast::Label.AbortButton }
+
+        result = Yast2::Popup.show(full_message(message, description, question, what),
+          details: wrap_text(what), buttons: buttons, focus: :continue)
+        log.info "User answer: #{result}"
+
+        missing_command_handle_user_decision(result, package_handler)
+      end
+
+      # Initialization.
+      #
+      def begin
+        @again = false
+      end
+
+      # Should probing be run again?
+      #
+      # @return [Boolean] Whether probing should be run again.
+      #
+      def again?
+        @again
+      end
+
+      private
+
+      # Generates the text for the packages that must be installed.
+      #
+      # @return [String] The text.
+      #
+      def missing_command_handle_packages_text(package_handler)
+        packages = package_handler.compact
+
+        n_("The following package needs to be installed:",
+          "The following packages need to be installed:", packages.size) + "\n" +
+          packages.sort.join(", ")
+      end
+
+      # Handles the result from the popup.
+      #
+      # @return [Boolean] Whether probing should continue.
+      #
+      def missing_command_handle_user_decision(result, package_handler)
+        case result
+
+        when :install
+          package_handler.commit
+          @again = true
+          false
+
+        when :continue
+          @again = false
+          true
+
+        when :abort
+          @again = false
+          false
+
+        end
+      end
     end
   end
 end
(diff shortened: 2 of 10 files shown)


src/lib/y2storage/callbacks/probe.rb
```diff
@@ -19,11 +19,116 @@
...
+    class Probe < Storage::ProbeCallbacksV3
       include LibstorageCallback
+
+      # Callback for missing commands during probing.
+      #
+      # @param message [String] error title coaming from libstorage-ng
+      #   (in the ASCII-8BIT encoding! see https://sourceforge.net/p/swig/feature-requests/89/)
+      # @param what [String] details coming from libstorage-ng (in the ASCII-8BIT encoding!)
+      # @param command [String] missing command coming from libstorage-ng (in the ASCII-8BIT encoding!)
+      # @param used_features used features bit field coming from libstorage-ng
```
jreidinger: used features type missing. Is it integer or bitmap?

jreidinger: I mean ideally we should specify its ruby type. I expect it is `Integer` right?

src/lib/y2storage/callbacks/probe.rb
```diff
@@ -19,11 +19,116 @@
...
+                        "and restart probing.")
+
+        what += "\n\n" + missing_command_handle_packages_text(package_handler)
+
+        question = _("Continue despite the error, install required packages or abort?")
+        buttons = { continue: Yast::Label.ContinueButton, install: _("Install Packages"),
+                    abort: Yast::Label.AbortButton }
+
+        result = Yast2::Popup.show(full_message(message, description, question, what),
+          details: wrap_text(what), buttons: buttons, focus: :continue)
```
jreidinger: I am not sure about focus here. Why not having install as default option? It should be the ideal path for user, not?

src/lib/y2storage/used_storage_features.rb
```diff
@@ -112,8 +112,29 @@ class UsedStorageFeatures
...
-    def initialize(devicegraph)
-      @devicegraph = devicegraph
+    # Initialize object.
+    #
+    # @param arg [Integer or ::Storage::Devicegraph] Either a integer with the feature
+    #   bits or a devicegraph from which to get the feature bits.
+    #
+    def initialize(arg)
+      @devicegraph = nil
+      @used_features = nil
```
jreidinger: this initialization is not needed as it is default value in newly created object.

test/y2storage/callbacks/probe_test.rb
```diff
@@ -29,4 +29,83 @@
...
+
+  describe "#begin" do
+    it "clears again flag" do
+      subject.begin
+      expect(subject.again?).to be false
+    end
+  end
+
+  describe "#again?" do
+    it "is unset after calling begin" do
```
jreidinger: I found a bit confusing to use "unset" here as I would expect `nil` there. If I get logic here.  It is `nil` for newly created object, `false` after begin and `true`/`false` after user decision. So I would use explicit `returns false after calling begin`

test/y2storage/callbacks/probe_test.rb
```diff
@@ -29,4 +29,83 @@
...
+    before do
+      allow(Yast::Mode).to receive(:installation).and_return false
+      allow(Y2Storage::PackageHandler).to receive(:new).and_return package_handler
+    end
+
+    let(:package_handler) do
+      double("PackageHandler", add_feature_packages: packages, compact: packages)
+    end
+
+    let(:packages) { ["btrfsprogs", "e2fsprogs"] }
```
jreidinger: order here is a bit confusing, as it is used even for first `it` statement. So usual order is `let`, then `before`, `after`, `around` and then list of `it` statements.

Try yourself in simple rspec:
```
describe "A" do
  it "says B" do
    puts "B"
  end

  before do
    puts "A"
  end
end
```

test/y2storage/callbacks/probe_test.rb
```diff
@@ -29,4 +29,83 @@
...
+    before do
+      allow(Y2Storage::PackageHandler).to receive(:new).and_return package_handler
+    end
+
+    let(:package_handler) do
+      double("PackageHandler")
+    end
+
+    context "when user selected install" do
+      it "returns false, sets again flag and installs packages" do
```
jreidinger: Usual convention is to have each statement in separate it, so it document behavior nicer in output. I am not sure how strict is this conventions in yast2-storage-ng.

test/y2storage/used_storage_features_test.rb
```diff
@@ -104,4 +104,16 @@
         .to contain_exactly("device-mapper", "multipath-tools", "fcoe-utils")
     end
   end
+
+  context "Created from integer representing UF_BTRFS" do
```
jreidinger: I would not mention here specific flag, but to be real documention it is better to use something like `Created from integer representing used features`

src/lib/y2storage/callbacks/probe.rb
aschnell: I presented a screenshot yesterday and did not get any feedback.

src/lib/y2storage/used_storage_features.rb
aschnell: I think it helps to understand the code.

src/lib/y2storage/callbacks/probe.rb
aschnell: Indeed, the type is missing. Why does yardoc not compain?

test/y2storage/callbacks/probe_test.rb
aschnell: I consider ```again``` to be a flag and there "unset" and "set" is common. But I can change it.

test/y2storage/callbacks/probe_test.rb
aschnell: At least is seems common to first define ```before``` and then ```let``` inside ```it```.

test/y2storage/callbacks/probe_test.rb
aschnell: That would make the code longer (more ```it```s) and kind of rip things apart AFAIS.

aschnell:
The last Travis failure is not related to the PR.

src/lib/y2storage/callbacks/probe.rb
```diff
@@ -19,11 +19,116 @@
...
 module Y2Storage
   module Callbacks
     # Class to implement callbacks used during libstorage-ng probe
-    class Probe < Storage::ProbeCallbacks
+    class Probe < Storage::ProbeCallbacksV3
       include LibstorageCallback
+
+      # Callback for missing commands during probing.
+      #
+      # @param message [String] error title coaming from libstorage-ng
```
joseivanlopez: Typo: coming from

test/y2storage/callbacks/probe_test.rb
dgdavid: I think is a matter of taste. I also like to put `let`s before `before`, but it's true that there are so many tests in storage-ng where `let` s are placed after `before`.

test/y2storage/callbacks/probe_test.rb
dgdavid: > of rip things apart AFAIS.

I disagree.

I'm with @jreidinger about convention. It's _expected_ to have an expectation per example.  True, it makes the code longer, but I don't se the problem in that case.

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Feb 06 15:09:57 CET 2020 - aschnell@suse.com
+
+- allow to install packages needed for probing
```
dgdavid: We need a reference to a bug or Jira entry in the changelog, isn't it?

src/lib/y2storage/used_storage_features.rb
```diff
@@ -112,8 +112,29 @@ class UsedStorageFeatures
     #----------------------------------------------------------------------
     #
 
-    def initialize(devicegraph)
-      @devicegraph = devicegraph
+    # Initialize object.
+    #
+    # @param arg [Integer or ::Storage::Devicegraph] Either a integer with the feature
```
joseivanlopez: NP: I would say yardoc expects comma instead or *or* to separate different types: `@param arg [Integer, ::Storage::Devicegraph]`

src/lib/y2storage/exceptions.rb
```diff
@@ -39,4 +39,7 @@ class DeviceNotFoundError < Error
   # Requested access mode is incompatible with current mode
   class AccessModeError < Error
   end
+  # Operation was aborted
+  class Aborted < Error
```
joseivanlopez: Where is this used?

test/y2storage/callbacks/probe_test.rb
```diff
@@ -29,4 +29,83 @@
...
+
+  describe "#begin" do
+    it "clears again flag" do
+      subject.begin
+      expect(subject.again?).to be false
+    end
+  end
+
+  describe "#again?" do
+    it "returns false after calling begin" do
```
dgdavid: What about before calling `#begin`?

dgdavid: NP: BTW, I'd use `context`s

```ruby
context "when #begin has been called" do
  before { subject.begin }

  it "returns false" do
    expect(subject.again?).to be false
  end
end
```

src/lib/y2storage/callbacks/probe.rb
joseivanlopez: I agree with @jreidinger, IMHO *install* should be the default option. Sorry @aschnell if you didn't get feedback earlier. Sometimes we don't realize about some details until we have a deeper revision. Changing the default option should be easy to do, right?

dgdavid:
> But I'm going to block the PR for now because I miss the Bugzilla or Jira reference.

:thinking: :thinking:  Although I see that there is none in the linked Trello card.

test/y2storage/callbacks/probe_test.rb
jreidinger: I do not have strong opinion about `let` and `before` order, but having `it` before them is confusing for readers as it affect also code in `it`.

dgdavid: Ah! I overlooked that `it` placed before `before`.

I agree with @jreidinger again

test/y2storage/callbacks/probe_test.rb
joseivanlopez: The idea of rpec is to test code (of course) but also to document the behavior. In general, I try to achieve both things:

* by adding atomic tests (a test only check one thing)
* by adding good behavior descriptions

In this test, the first point is not fulfill, because it is testing three things at once. But, to be honest, I am not sure if we have been very strict with this along the storage-ng testsuite.

test/y2storage/callbacks/probe_test.rb
joseivanlopez: Yes, a matter of taste. What I usually do:

~~~
context "when sky is grey" do
  let(:sky) { "grey" } # this let is setting the condition of this context

  before do
    allow(foo).to receive(:bar).and_return(foobar)
  end

  let(:foobar) { whatever } # this let is defining something used previously in the *before* block
~~~

But this is only my own convention.

test/y2storage/callbacks/probe_test.rb
joseivanlopez: Yes, this looks like more rspec style.

aschnell:
Will have to wait until after Hackweek.

PR #1045, opened 2020-02-18
Add not usable as block device error

joseivanlopez:
## Problem

Some block devices cannot used as block devices (e.g., DASD devices). Therefore, these devices cannot be directly formatted, but the Expert Partitioner does not prevent it.

* https://bugzilla.suse.com/show_bug.cgi?id=1163597
* https://trello.com/c/NsJWnEWd/1651-sles15-sp2-p2-1163597-partitioner-allows-to-install-on-an-unpartitioned-dasd

## Solution

A new error was added to the action for editing block devices. Now, an error popup is shown when the user tries to edit a device that cannot be used as block device.

Note that `BootRequirementsChecker` was working correctly and some warnings were generated when a DASD device was directly formatted for root. See screenshots.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

<details>
<summary>Show/Hide</summary>

* New error

Screenshot from 2020-02-18 09-46-26

* `BootRequirementsChecker` was showing warnings correctly

Screenshot from 2020-02-18 09-06-50

</details>

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Feb 18 10:26:02 UTC 2020 - José Iván López González <jlopez@suse.com>
+
+- Partitioner: do not allow to edit a device when the device cannot
+  be used as block device (bsc#1163597).
+- 4.2.85
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


ancorgs:
I don't get it. Why do we offer the option if it cannot be used? Shouldn't we simply not display the option to format such devices?

PR #1047, opened 2020-02-18
Hide edit button for DASD devices

joseivanlopez:
## Problem

Some block devices cannot be used as block devices (e.g., DASD devices). Therefore, these devices cannot be directly formatted, but the Expert Partitioner does not prevent it.

* https://bugzilla.suse.com/show_bug.cgi?id=1163597
* https://trello.com/c/NsJWnEWd/1651-sles15-sp2-p2-1163597-partitioner-allows-to-install-on-an-unpartitioned-dasd

## Solution

A solution was already proposed in the PR https://github.com/yast/yast-storage-ng/pull/1045, but now a more convenient approach is implemented. The buttons for editing a device are hidden (i.e., when the device cannot be used as block device) instead of showing an error popup.

## Testing

- Added a new unit tests
- Tested manually

## Screenshots

<details>

<summary>Show/Hide</summary>

* Before:

Screenshot from 2020-02-18 15-10-09

Screenshot from 2020-02-18 15-09-48

* After:

Screenshot from 2020-02-18 15-08-39

Screenshot from 2020-02-18 15-08-01

</details>

### DIFF
--- a/src/lib/y2partitioner/actions/edit_blk_device.rb
+++ b/src/lib/y2partitioner/actions/edit_blk_device.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2017-2020] SUSE LLC
+# Copyright (c) [2017-2019] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -116,8 +116,7 @@ module Y2Partitioner
           used_device_error,
           partitions_error,
           extended_partition_error,
-          lvm_thin_pool_error,
-          no_usable_as_blk_device_error
+          lvm_thin_pool_error
         ].compact
       end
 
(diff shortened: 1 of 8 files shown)


src/lib/y2partitioner/actions/edit_blk_device.rb
```diff
@@ -206,7 +205,7 @@ def extended_partition_error
         _("An extended partition cannot be edited")
       end
 
-      # Error message if trying to edit an LVM thin pool
+      # Error message is trying to edit an LVM thin pool
```
dgdavid: Maybe we can keep that small fix :)

dgdavid:
```suggestion
      # Error message if trying to edit an LVM thin pool
```

joseivanlopez: Ups, right!

PR #1049, opened 2020-02-20
allow to install packages needed for probing

aschnell:
For https://trello.com/c/ae4cAtFe/1573-5-improve-probing-when-there-are-missing-tools.

## Problem

Probing can fail if programs are missing.

## Solution

Allow user to install missing packages.

## Testing

- Added a new unit test.
- Tested manually.

## Screenshot

See trello card.

### DIFF
--- a/src/lib/y2storage/used_storage_features.rb
+++ b/src/lib/y2storage/used_storage_features.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2016-2017,2019] SUSE LLC
+# Copyright (c) [2016-2017,2019-2020] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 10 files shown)


src/lib/y2storage/used_storage_features.rb
```diff
@@ -112,8 +112,29 @@ class UsedStorageFeatures
...
     #
 
-    def initialize(devicegraph)
-      @devicegraph = devicegraph
+    # Initialize object.
+    #
+    # @param arg [Integer, ::Storage::Devicegraph] Either an integer with the feature
+    #   bits or a devicegraph from which to get the feature bits.
+    #
+    def initialize(arg)
```
ancorgs: I know it's a little bit late since this is already merged and in general it works. So take this just as a note for the future. :wink: And, of course, it's only a personal opinion.

This looks a bit like misusing a class for something different it was designed to.

I mean, originally we have a class with the description: "Class that collects information about which storage features are used in the current target machine's storage setup". And, as it makes a lot of sense based on that description, its constructor receives a devicegraph which describes such storage setup.

Now the class can also receive in its constructor something as different as an Integer. Which, of course, makes the description not apply any longer and changes the intention of the class.

In general, a method that receives an argument simply called "arg" and than then, based on the content/type of "arg" does something different is a design smell, in my opinion[1]. If it's a constructor, I would say that's even worse.

Of course, we can argue the design of the class was bad to begin with, mixing the responsibilities of processing the target devicegraph and of installing the packages. So I would have refactored things a big to have single responsibility classes, instead of adding just another responsibility to a class that already had too many.

[1] Yes, Ruby is a dynamically typed language. But that's done to make duck typing possible. Using that dynamic nature to do things like an argument called "arg" with different meanings based on its type is, IMO, wrong.

PR #1051, opened 2020-03-03
bsc#1159201: export unpartitioned RAID when used for another device (LVM, RAID, bcache)

imobachgs:
## Problem

As seen in [bsc#1159201](https://bugzilla.suse.com/show_bug.cgi?id=1159324), AutoYaST omits the `<partitions>` list when cloning an unpartitioned RAID that does not have partitions nor a file system. It may happen that the RAID is used as part of another device (e.g., as part of an LVM volume group).

Additionally, while debugging the issue, I have found that the `<disklabel>` element was not set.

## Solution

This PR fixes both problems, setting the `<disklabe>` to the appropriate value and exporting the `<partitions>` section when the RAID device is used.

## Links

- https://bugzilla.suse.com/show_bug.cgi?id=1159324
- Trelo card: https://trello.com/c/lLRXuABJ/1673-3-sles15-sp1-p5-1159201-autoyast-does-not-get-encrypted-setup-with-a-unusual-partition-scheme


imobachgs:
Thanks for the review. However, I fixed the bug for SLE 15 SP1 too and I am [merging the fix](https://github.com/yast/yast-storage-ng/pull/1053) into the `master` branch.

PR #1052, opened 2020-03-03
 bsc#1159201 (SP1): export unpartitioned RAID when used for another device (LVM, RAID, bcache)

imobachgs:
## Problem

As seen in [bsc#1159201](https://bugzilla.suse.com/show_bug.cgi?id=1159324), AutoYaST omits the `<partitions>` list when cloning an unpartitioned RAID that does not have partitions nor a file system. It may happen that the RAID is used as part of another device (e.g., as part of an LVM volume group).

Additionally, while debugging the issue, I have found that the `<disklabel>` element was not set.

## Solution (SP1)

This PR fixes both problems, setting the `<disklabe>` to the appropriate value and exporting the `<partitions>` section when the RAID device is used.

## Links

- https://bugzilla.suse.com/show_bug.cgi?id=1159324
- Trello card: https://trello.com/c/lLRXuABJ/1673-3-sles15-sp1-p5-1159201-autoyast-does-not-get-encrypted-setup-with-a-unusual-partition-scheme


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Tue Mar  3 12:56:34 UTC 2020 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- AutoYaST: export the 'disklabel' attribute for software RAID
+  devices.
+- AutoYaST: includes a 'partitions' section when exporting an
```
ancorgs: include (or remove the preceding colon)

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,13 @@
...
+  unpartitioned software RAID which is not used as as filesystem
```
ancorgs: as as

test/data/devicegraphs/md_raid_lvm.xml
```diff
@@ -0,0 +1,277 @@
+<?xml version="1.0"?>
```
ancorgs: At least in master I would expect to be able to define an LVM on top of a RAID using the yml format. And to be honest, it's much more human-compatible (and you could omit many unnecessary details.

Can you check if that's possible in this branch as well?

We even have a conversion tool

PR #1057, opened 2020-03-09
AutoYaST: Display an error when no suitable PVs are found

imobachgs:
## Problem

AutoYaST crashes (not exactly, as it simply catches the exception and displays
the problem in a controlled way) when no PVs are available for a volume group.

- [bsc#1162043](https://bugzilla.suse.com/show_bug.cgi?id=1162043)

## Solution

This PR adds a *fatal AutoYaST issue* when no suitable PVs are found for a
volume group. It fixes the problem and it is the way to go in SP2 too.

However, we should consider an additional top level validation layer for the
future. The problem reported in the bug could have been easily detected at the
very beginning of the *import* process (no need to go down to the
`AutoinstDevicesCreator` class). However, as we check each `<drive>` in
isolation (and the correspondence of VGs and PVs may be scattered through
different `<drive>` sections), we miss some information during validation.

## Testing

I have manually tested the fix (see the screenshot below) and, moreover, I added
some unit tests.

## Screenshots

<details>
<summary>"No physical volumes" error using Qt</summary>

no-pvs-issue
</details>

<details>
<summary>"No physical volumes" error in text mode</summary>

no-pvs-issue-textmode
</details>


src/lib/y2storage/autoinst_issues/no_pvs.rb
```diff
@@ -0,0 +1,54 @@
...
+      # @param planned_vg [Planned::LvmVg] Planned volume group
+      def initialize(planned_vg)
+        textdomain "storage"
+
+        @planned_vg = planned_vg
+      end
+
+      # Fatal problem
+      #
+      # @return [Symbol] :warn
```
jreidinger: copy and paste error, it return `:fatal`

src/lib/y2storage/autoinst_issues/no_pvs.rb
```diff
@@ -0,0 +1,54 @@
...
+        :fatal
+      end
+
+      # Return the error message to be displayed
+      #
+      # @return [String] Error message
+      # @see Issue#message
+      def message
+        format(
+          _("Could not find a suitable physical volume for\nvolume group '%{vg_name}'."),
```
jreidinger: Maybe add comment for translators to add \n after X characters? if translation is longer?

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -62,11 +62,16 @@ module Proposal
...
       include Yast::Logger
 
+      # @return [AutoinstIssues::List] List of found AutoYaST issues
+      attr_reader :issues_list
+
       # Constructor
       #
       # @param original_graph [Devicegraph] Devicegraph to be used as starting point
-      def initialize(original_graph)
+      # @param issues_list [AutoinstIssues::List] List of AutoYaST issues to register them
```
jreidinger: documentation of parameter is not much clear. Maybe mention that it is write only parameter that is modified and issues found during creating device is added there.

src/lib/y2storage/proposal/autoinst_devices_creator.rb
```diff
@@ -358,6 +363,11 @@ def set_up_lvm(vgs, devs_to_reuse)
         vgs.reduce(creator_result) do |result, vg|
           pvs = creator_result.created_names { |d| d.pv_for?(vg.volume_group_name) }
           pvs += devs_to_reuse.select { |d| d.pv_for?(vg.volume_group_name) }.map(&:reuse_name)
+          if pvs.empty?
+            issues_list.add(:no_pvs, vg)
+            next result
```
jreidinger: NP: always when I see this reduce and next handling I am a bit scared that it is forgotten. What about using instead each_with_object that is safer for next and break, as it does not need to have parameter?

imobachgs: For the time being, I trust in the test suite to catch that problem. But I will have it into account in the future.

src/lib/y2storage/autoinst_issues/no_pvs.rb
imobachgs: The initial message was longer but now it is only 60 characters (without the VG name), so I think we can drop the `\n`.

PR #1060, opened 2020-03-10
Simpler probing callback and refactoring of storage features

ancorgs:
## Problem

The handler implemented in #1049 presented serious usability problems, summarized in [this gist](https://gist.github.com/ancorgs/d2565bf505834329946a998500a326c8). So the YaST Team decided it needed a second iteration to improve the UX.

In addition, there was a lot of room for improvement in the way yast-storage-ng managed the so-called "used storage features" coming from libstorage-ng:

- It was very poorly object oriented
- The API contained some "design smells", like the one [commented here](https://github.com/yast/yast-storage-ng/pull/1049#discussion_r382509498) with a class receiving a generic argument called `arg` in its constructor (that could be a devicegraph or an integer) and having a different behavior depending on the type of that argument.
- yast-storage-ng contained several made-up features that were actually not defined in libstorage-ng (and, thus, made no sense).
- Some of the features reported by libstorage-ng were unknown to yast-storage-ng, which, as a consequence, ignored the recommended packages for those features.
- There was no test to detect when a new storage feature is added or removed from libstorage-ng (so both yast-storage-ng and libstorage-ng were out of sync, as explained in the two points above).
- Some clients installed the packages on their own, instead of relying in the class `PackageHandler` (which, btw, had several other responsibilities apart from installing the packages).

Trello card: https://trello.com/c/CA7oLu3O/1677-3-usability-follow-up-of-improve-probing-when-there-are-missing-tools

### The refactoring

The first three commits include a refactoring of the former `StorageFeatures` and `PackageHandler` classes to fix all the problems mentioned above. They introduce no change in behavior, but all the functionality in those classes is moved to a real object-oriented model with the following classes and responsibilities:

- `StorageFeature`: each instance represents one of the features known to libstorage-ng and is responsible for knowing which packages are associated to that feature.
- `StorageFeaturesList`: makes it possible to know the list of packages needed to support for each set of features, it's also responsible for transforming the bit-fields coming from libstorage-ng into proper collection of objects.
- `PackageHandler`: installs or marks for installation a set of packages. It is not longer responsible of making calculations on top of a devicegraph to get the list of needed packages or to get such list from a bit-field obtained from libstorage-ng.
- `Devicegraph`: it now can report which features it uses (as a list of objects), instead of reporting a plain bit-field.

Since the last step of the Partitioner now relies on `PackageHandler` to install the packages instead of doing it by itself, the behavior in case of error is now a bit better. In the past, if the installation failed it just returned to the same screen without displaying any information. It was kind of confusing. Now it shows the `PackageHandler` error pop-up.

error-install

Last but not least (about the refactoring), the following fake storage features were removed: `UF_EFIBOOT`, `UF_DM`.  They were in fact never reported by libstorage-ng.

And the following features were added, together with their corresponding lists of packages: `UF_DASD`, `UF_BITLOCKER`, `UF_JFS`, `UF_F2FS`, `UF_EXFAT`. Those features were reported by libstorage-ng but completely ignored by yast-storage-ng.

### The simplified UX for the callback 

The fourth commit simplifies the probing callback to look like the screenshot below with only two buttons:

- Install: installs the mentioned packages and restarts the probing process.
- Ignore: continues the probing process.

probe-callback

## Testing

- Reorganized the unit tests, including one to verify when new storage features are added or removed in libstorage-ng
- Tested manually


src/lib/y2storage/package_handler.rb
```diff
@@ -63,6 +63,7 @@ def reset
     # @return [Array<String>] new package list (may contain duplicates)
     def add_packages(pkg_list)
       @pkg_list.concat(pkg_list)
+      compact
```
imobachgs: np: I would call this method `compact!` instead.

src/lib/y2partitioner/widgets/summary_text.rb
```diff
@@ -100,9 +100,7 @@ def calculate_actions
 
       # Updates the value of {#packages}
       def calculate_packages
-        handler = Y2Storage::PackageHandler.new
-        handler.add_feature_packages(current_graph)
-        @packages = handler.pkg_list.uniq
+        @packages = current_graph.used_features.pkg_list
```
imobachgs: AFAIS, `#calculate_packages` is only called in the `#init` method, right? Why not simply replace this method with a `#packages` one:

```ruby
def packages
  # feel free to remove memoization if it is not needed
  @packages ||= current_graph.used_features.pkg_list
end
```

You could drop the `attr_reader :packages` and simplify the `#init` method.

src/lib/y2storage/callbacks/probe.rb
```diff
@@ -55,18 +57,17 @@ def missing_command(message, what, command, used_features)
         # Redirect to error callback if no packages can be installed.
         return error(message, what) if used_features == 0 || Yast::Mode.installation
 
-        package_handler = Y2Storage::PackageHandler.new
-        package_handler.add_feature_packages(used_features)
+        packages = StorageFeaturesList.new(used_features).pkg_list
```
imobachgs: This API looks confusing to me. Let's put aside that we are getting an integer from an OOP library to get the features that are used. Having a class called *BlaBlaList*, I would expect the constructor to receive an array (or a Set or "splat" args, etc.) of `StorageFeature` objects.

I had a look at the `StorageFeatureList` implementation and I would rather prefer an special constructor `from_bits` (or probably a better name).

However, it is just a matter of preference, and I can live with the current implementation.

src/lib/y2storage/storage_feature.rb
```diff
@@ -0,0 +1,226 @@
...
+  # Class representing the storage features defined by libstorage-ng
+  #
+  # In libstorage-ng each feature is represented by a bit-mask identified
+  # by a constant with name UF_XXX (with XXX being the name of the feature).
+  # A set of features is represented by a bit-field of those masks.
+  #
+  # Thus, one of those bit-fields is used to specify which features are
+  # used in the current target machine's storage setup. Additionally, a similar
+  # bit-field is used to report features that have been detected in the system
+  # but that cannot be probed because some command in missing.
```
imobachgs: "is missing"

src/lib/y2storage/storage_feature.rb
```diff
@@ -0,0 +1,226 @@
+# Copyright (c) [2016-2017,2019-2020] SUSE LLC
```
imobachgs: Are these years OK?

src/lib/y2storage/storage_features_list.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    def_delegators :@features, :each, :empty?
+
+    # Constructor
+    #
+    # @param bits [Integer, nil] bit-field representing all the features that
+    #   must be part of the list. If nil, the list will contain all the known
+    #   storage features.
+    def initialize(bits = nil)
+      @features = StorageFeature.all.dup
```
imobachgs: np: I usually prefer passing the list of features as an argument as it makes it easier to test even without mocking. However, in this case, the constructor would look even more weird :).

test/y2storage/storage_features_list_test.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      it "returns an empty list" do
+        expect(described_class.new(bits)).to be_empty
+      end
+
+      # The combination of this test and the corresponding unit test for
+      # StorageFeature#in_bitfield? ensures yast-storage-ng does not contain any
+      # feature that is not defined in libstorage-ng (a NameError exception
+      # would be raised).
+      it "calls #in_bitfield? for all registered features" do
```
imobachgs: np: This description basically reveals an implementation detail.  Let's keep it as it is because, at least, we have a nice text documenting the reason.

imobachgs:
Oh, additinally, is the https://github.com/yast/yast-storage-ng/blob/184365e389167b8b5549def52c10c3d1c2cde7ac/src/lib/y2partitioner/widgets/summary_text.rb#L112 method still needed?

src/lib/y2partitioner/widgets/summary_text.rb
ancorgs: All that is improved in a subsequent commit. This whole API made very little sense and this pull request was just intended as an initial minimal polishing (which is exactly the name of the commit).

src/lib/y2storage/callbacks/probe.rb
ancorgs: You are right, I will change it.

test/y2storage/storage_features_list_test.rb
ancorgs: Thinking it better, I could move the code raising the `NameError` exception directly to the constructor of `StorageFeature`. There is no need to defer the check until `#in_bitfield?` is called. Failing right away when the object is created would be ok and everything (including the tests) would look way more natural.

ancorgs:
BTW. I have just realized the old `PackageHandler` API is still used in AutoYaST. I will open a pull request about it:

https://github.com/yast/yast-autoinstallation/blob/648ba399f6825628eff1b5946910afff310a60f8/src/modules/AutoinstSoftware.rb#L864

src/lib/y2storage/storage_feature.rb
ancorgs: Well, the original `StorageFeatures` class was basically copied from the old storage. And some people tries to always add the exact year when they touch a file... but TBH I also find it weird. I will change it.

test/y2storage/storage_features_list_test.rb
ancorgs: Done in a later commit already added to this PR.

src/lib/y2storage/storage_features_list.rb
ancorgs: Fixed in a later commit already added to this PR.

src/lib/y2storage/storage_feature.rb
ancorgs: Done in a later commit already added to this PR.

src/lib/y2storage/storage_feature.rb
ancorgs: Fixed in a later commit already added to this PR.

src/lib/y2storage/callbacks/probe.rb
ancorgs: Done in a later commit already added to this PR.

src/lib/y2storage/package_handler.rb
ancorgs: I'm not so sure. In theory methods with an exclamation mark are only intended as a "more aggressive" counterpart of another method without the mark. There is no such counterpart here.

src/lib/y2storage/storage_features_list.rb
```diff
@@ -38,13 +34,22 @@ class StorageFeaturesList
     def_delegators :@features, :each, :empty?, :size, :length
 
     # Constructor
+    def initialize(*features)
+      @features = [*features].flatten
```
imobachgs: np: Not sure, but maybe you can achieve the same with `Array(features)`.

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,15 @@
+-------------------------------------------------------------------
+Wed Mar 11 12:53:13 UTC 2020 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Usability improvements when asking the user to install packages
+  during the probing phase (related to bsc#1140040).
+- More flexible API for PackageHandler and related classes
+- More libstorage-ng "storage features" now recognized and handled
+  if needed: UF_DASD, UF_BITLOCKER, UF_JFS, UF_F2FS, UF_EXFAT.
+- Partitioner: better error reporting when the installation of
+  a needed packages fails.
```
imobachgs: "a needed package~s~ fails"

src/lib/y2storage/storage_features_list.rb
ancorgs: No (that was also my first attempt). ;-P

ancorgs:
Jenkins problems fixed by https://github.com/yast/yast-storage-ng/pull/1061

PR #1062, opened 2020-03-12
AutoYaST: warn about conflicting elements

imobachgs:
## Problem

[bsc#1165907](https://bugzilla.suse.com/show_bug.cgi?id=1165907) revealed a problem when setting two mutually exclusive options in an AutoYaST partition section. In that case, setting the `lvm_group` and `raid_name` caused `libstorage-ng` to crash.

## Solution

There are a few elements that determine how is the device going to be used: `mount`, `raid_name`, `lvm_name`, `btrfs_name`, `bcache_backing_for` and `bcache_caching_for`. The truth is that all those elements should be exclusive.

Now, if more than one of those attributes are present, AutoYaST will select just one and ignore the rest. For the time being, the logic is pretty simple (just a matter of precedence), but in the future, we might consider a smarter mechanism.

## Testing

The PR contains some unit tests and I have tried the changes manually.

## Screenshots

<details>
<summary>AutoYaST warning</summary>

conflicting-attrs
</details>

### DIFF
--- a/src/lib/y2storage/autoinst_issues/conflicting_attrs.rb
+++ b/src/lib/y2storage/autoinst_issues/conflicting_attrs.rb
@@ -1,0 +1,68 @@
+# Copyright (c) [2020] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+require "y2storage/autoinst_issues/issue"
+
+module Y2Storage
+  module AutoinstIssues
+    # Conflicting attributes where specified for the given section.
+    #
+    # The conflict is resolved and the 'selected_attr' is honored while the rest
+    # is ignored.
+    class ConflictingAttrs < Issue
+      # @return [Symbol] Selected attribute
+      attr_reader :selected_attr
+      # @return [Array<Symbol>] List of ignored attributes
+      attr_reader :ignored_attrs
+
+      # @param section [#parent,#section_name] Section where it was detected (see {AutoinstProfile})
+      # @param selected_attr [Symbol] Name of the attribute that will be used
+      # @param ignored_attrs [Array<Symbol>] List of attributes to be ignored
+      def initialize(section, selected_attr, ignored_attrs = [])
+        textdomain "storage"
+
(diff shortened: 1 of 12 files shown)


src/lib/y2storage/autoinst_issues/conflicting_attrs.rb
```diff
@@ -0,0 +1,68 @@
...
+    class ConflictingAttrs < Issue
+      # @return [Symbol] Selected attribute
+      attr_reader :selected_attr
+      # @return [Array<Symbol>] List of ignored attributes
+      attr_reader :ignored_attrs
+
+      # @param section [#parent,#section_name] Section where it was detected (see {AutoinstProfile})
+      # @param selected_attr [Symbol] Name of the attribute that will be used
+      # @param ignored_attrs [Array<Symbol>] List of attributes to be ignored
+      def initialize(section, selected_attr, ignored_attrs = [])
```
jreidinger: what is purpose of default attribute?

src/lib/y2storage/autoinst_issues/conflicting_attrs.rb
```diff
@@ -0,0 +1,68 @@
...
+        textdomain "storage"
+
+        @section = section
+        @selected_attr = selected_attr
+        @ignored_attrs = ignored_attrs
+      end
+
+      # Fatal problem
```
jreidinger: fatal problem and just warning?

src/lib/y2storage/proposal/autoinst_drive_planner.rb
```diff
@@ -64,11 +64,60 @@ def planned_devices(_drive)
...
+      private_constant :USAGE_ATTRS_MAP
+
+      # Returns the attributes which determines how the device will be used
+      #
+      # @param partition_section [AutoinstProfile::PartitionSection] Partition specification
+      # @return [Array<Symbol>] List of usage related attributes which are defined
+      def usage_attrs_in(partition_section)
+        USAGE_ATTRS.reject do |e|
+          value = partition_section.public_send(e)
+          value.nil? || (value.is_a?(Array) && value.empty?)
```
jreidinger: so empty hash does not make sense? If makes, then maybe respond_to?(:empty) and call it can make more sense.

test/support/autoinst_devices_planner_conflicts.rb
```diff
@@ -0,0 +1,128 @@
...
+      let(:missing_attrs) { ["mount", "raid_name", "lvm_group", "btrfs_name"] }
+
+      it "prefers the bcache_backing_for" do
+        disk = planner.planned_devices(drive).first
+        expect(planned_device(disk).bcache_backing_for).to eq("/dev/bcache0")
+      end
+
+      it "registers an issue" do
+        expect(issues_list).to receive(:add)
+          .with(:conflicting_attrs, drive.partitions.first, :bcache_backing_for, Array)
```
jreidinger: NP: this is kind of strange testing. I expect more like checking after call that issues list contain specific Issue, so it does not need mocking of add method.

src/lib/y2storage/proposal/autoinst_drive_planner.rb
imobachgs: No, it does not make sense. They can only be arrays and scalar values.

PR #1063, opened 2020-03-12
integrate new #adjust_block_size method (bsc#1164295)

wfeldt:
## Task 1

- https://github.com/openSUSE/libstorage-ng/pull/709

`libstorage-ng` got a new method in its `Region` class.

## Task 2

Show RAID size and block size also in YAML file.

Note
> I've seen no good way to set the values in FakeDeviceFactory (because devices are added to the RAID at the end and overwrite the values), so they are just ignored when reading YAML.

### DIFF
--- a/src/lib/y2storage/region.rb
+++ b/src/lib/y2storage/region.rb
@@ -88,6 +88,15 @@ module Y2Storage
     #   @param delta [Integer]
     storage_forward :adjust_length
 
+    # @!method adjust_block_size(block_size)
+    #   Sets new block size without changing the region size
+    #
+    #   @raise [storage::InvalidBlockSize] if region start or region end are not
+    #   aligned to block_size
+    #
+    #   @param block_size [DiskSize]
+    storage_forward :adjust_block_size
+
     # @!method <(other)
     #   Checks whether the region starts before the other
     #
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/region.rb
```diff
@@ -88,6 +88,15 @@ def cover?(sector)
     #   @param delta [Integer]
     storage_forward :adjust_length
 
+    # @!method adjust_block_size(block_size)
```
ancorgs: I would not expose this method in `Y2Storage`. We don't need to call it and I expect we will never do.

ancorgs:
I'm not sure if it's a good idea to print values in the YAML that can be inferred from other values and that will be ignored when importing. In other words, values that are not exactly something that can be defined, but rather a consequence of other values.

The (original) goal of the YAML format is not to include all the information (we have the XML for that), but to specify scenarios for the tests that a human can read and modify. In that regard, the point was always to write the minimal possible information and let the library calculate all the gaps (contrary to the XML, that forces you to specify every single value, even the calculated ones).

I know that some people tend to use the YAML files also for debugging and I see that having the values written directly there can potentially save some brain processing time in those cases. But to be honest, if you are looking for that level of precision you should not use the YAML files. They are good to have a quick overview, but they are imprecise already at many levels.

wfeldt:
That was also my concern. And as the device factory does not allow to implement this easily, I left it at that.

I'm fine with dumping all this. But I'd keep fix for the `nocow` issue. Strange that this stayed unnoticed so far.

I now know that the RAID block size is stored in the XML. And you can also set the MD size there.

wfeldt:
I've created https://github.com/yast/yast-storage-ng/pull/1064 for the nocow issue.

PR #1064, opened 2020-03-13
write nocow attribute correctly into YAML

wfeldt:
## Problem

Working on https://github.com/openSUSE/libstorage-ng/pull/709 I noticed that the `nocow` btrfs attribute is not written correctly into YAML, resulting in invalid YAML files.

It stored the value as string (`'true'`) instead of as `true`.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Mar 13 14:02:15 UTC 2020 - Steffen Winterfeldt <snwint@suse.com>
+
+- write nocow attribute correctly into YAML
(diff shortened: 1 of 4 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Mar 13 14:02:15 UTC 2020 - Steffen Winterfeldt <snwint@suse.com>
+
+- write nocow attribute correctly into YAML
```
teclator: Miss the bug number

wfeldt: There's no bug number. I wouldnät mind staying this in some queue som time...

PR #1067, opened 2020-03-20
[bsc#1167053] Show an error when no components are found for multi-device stuff

imobachgs:
## Problem

Similar to #1059, AutoYaST will crash when no suitable devices are found for a Btrfs multi-device filesystem.

- [bsc#1167053](https://bugzilla.suse.com/show_bug.cgi?id=1167053).

## Solution

Apply the solution in #1059 to these cases: Btrfs multi-device filesystems, RAIDs and Bcache (missing backing device).

## Testing

- Added a new unit test
- Tested manually

## Screenshots

<details>
<summary>Errors detected in the profile</summary>

multi-btrfs-errors
</details>


test/y2storage/callbacks/callbacks_examples.rb
```diff
@@ -19,6 +19,8 @@
 # find current contact information at www.suse.com.
 
 RSpec.shared_examples "general #error examples" do
+  ERROR_EXAMPLES_MAX_LENGTH = 80
```
ancorgs: Now that we are touching this... there is any good reason for it to be defined as a constant instead of just using a variable or a `let`?

test/y2storage/proposal/autoinst_devices_creator_test.rb
```diff
@@ -417,6 +432,8 @@
           expect(result.shrinked_partitions.map(&:planned)).to eq([root])
         end
       end
+
+      context "when "
```
ancorgs: when?

imobachgs: Eeerrr never. I was about to add the test there, but I found out that we already were testing that scenarion. Fixed.

test/y2storage/callbacks/callbacks_examples.rb
imobachgs: I do not think so, just wanted to keep the original "intention" (if any). A `let` is OK for me (and a constant too).

PR #1068, opened 2020-03-23
Fix the /boot in bcache detection

dgdavid:
## Problem

The installer allows to proceed even when `/boot` is explicitly or implicitly mounted in a [bcache](https://bcache.evilpiepirate.org/) formatted partition.

However, the installed system is not able to boot.

- https://trello.com/c/KK7xbNps
- https://bugzilla.suse.com/show_bug.cgi?id=1165903
- Related to https://github.com/yast/yast-storage-ng/pull/944

## Solution

To show an alert when `/boot` will be mounted in a bcache.

## Testing

- Added a new unit test
- Tested manually

## Screenshots

<details>

---

Screenshot_sl5sp2_2020-03-23_20:44:18

Screenshot_sl5sp2_2020-03-23_20:44:25
</details>

### DIFF
--- a/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
+++ b/src/lib/y2storage/boot_requirements_strategies/analyzer.rb
@@ -570,9 +570,9 @@ module Y2Storage
           # If this is not a BlkFilesystem (e.g. NFS), it can't be in a BCache
           return false unless device.respond_to?(:plain_blk_devices)
 
-          # Let's trust in the hierarchy to determine if the device belongs to a Bcache, although
-          # there may be some advanced configuration for which this is not totally true because the
-          # device is reachable from a not BCache ancestor.
+          # Strictly speaking, with very advanced storage configurations it may be possible to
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/boot_requirements_strategies/analyzer.rb
```diff
@@ -570,7 +570,10 @@ def in_bcache?(device)
           # If this is not a BlkFilesystem (e.g. NFS), it can't be in a BCache
           return false unless device.respond_to?(:plain_blk_devices)
 
-          device.plain_blk_devices.any? { |dev| dev.is?(:bcache) }
+          # Let's trust in the hierarchy to determine if the device belongs to a Bcache, although
+          # there may be some advanced configuration for which this is not totally true because the
+          # device is reachable from a not BCache ancestor.
```
ancorgs: Not sure if the comment is clear enough. What about this?

Strictly speaking, with very advanced storage configurations it may be possible to access a filesystem
with bcache ancestors in the devicegraph without actually accessing the bcache. But that would be an
extreme case and is not supported by YaST.

PR #1069, opened 2020-03-24
Add preferred_name to BlkDevice and BlkFilesystem (needed for bsc#1166096)

ancorgs:
## Problem

Recently yast2-booloader was modified to not contain the logic about what is the best file path to reference a file, but to delegate that calculation to yast2-storage-ng. See https://github.com/yast/yast-bootloader/pull/591

Unfortunately, the logic implemented in that pull request is not really a direct translation of the previous one. If the device contains no filesystem (for example, PReP partitions and bios_boot ones), the logic implemented in https://github.com/yast/yast-bootloader/pull/591 fallbacks immediately to use the kernel device name.

In fact, the API offered by yast2-storage-ng only contained method to get the preferred name for `BlkFilesystem` objects, but not for `BlkDevice` objects themselves. Moreover, the distinction was not so clear and the method `BlkDevice#path_for_mount_by` was in fact calculating a path to point to the filesystem (if any), not to the block device.

**References:**

- https://bugzilla.suse.com/show_bug.cgi?id=1166096
- https://trello.com/c/SFYLkAIX/1703-3-tw-1166096-openqa-yast2bootloader-invalid-path-by-id-string 

## Solution

This pull request extends the yast2-storage-ng API to get udev names for a block device:

- `BlkFilesystem#mount_by_name` allows to get an udev name that matches the `mount_by` attribute of the filesystem (if any).
- `BlkFilesystem#preferred_name` allows to get the best (most stable) udev name to reference a filesystem. Useful for the bootloader when the previous one is not available.
- `BlkDevice#preferred_name` allows to get the best (most stable) udev name to reference a block device. Useful for the bootloader when the device does not contain any filesystem (e.g. PReP).

The new API is then used from https://github.com/yast/yast-bootloader/pull/593 to effectively fix the bug.

As part of the fix, `BlkDevice#path_for_mount_by` now always returns a file path that points to the device itself and `BlkFilesystem#path_for_mount_by` was added, for cases in which a path to the filesystem is needed. Note that, with the new API, the bootloader module doesn't need to call these methods anymore (there are more high-level methods to directly get the names).

## Testing

Added a new unit tests
Verified in the yast2-booloader side with a regression test.

### DIFF
--- a/src/lib/y2storage/blk_device.rb
+++ b/src/lib/y2storage/blk_device.rb
@@ -288,6 +288,15 @@ module Y2Storage
     #   @return [Encryption] nil if the device is not encrypted
     storage_forward :encryption, as: "Encryption", check_with: :has_encryption
 
+    # @!method possible_mount_bys
+    #   Possible mount-by methods to reference the block device itself, regardless of its content
+    #
+    #   @see #preferred_name
+    #
+    #   @return [Array<Filesystems::MountByType>]
+    storage_forward :possible_mount_bys, as: "Filesystems::MountByType"
+    private :possible_mount_bys
+
     # Checks whether the device is encrypted
     #
     # @return [boolean]
@@ -658,6 +667,30 @@ module Y2Storage
     # @return [Boolean]
     def windows_suitable?
       false
+    end
+
+    # Most suitable mount by option to reference the block device itself,
+    # regardless of its content
+    #
+    # This would return the same result if the device is formatted or if it's
+    # empty. To determine the MountByType to reference a filesytem (e.g. in
+    # fstab), call {Mountable#preferred_mount_by} on the filesystem object.
+    #
+    # Only the options that are indeed available are considered, which means
+    # this method always returns an option that can be used safely.
+    #
+    # @return [Filesystems::MountByType]
+    def preferred_mount_by
+      Filesystems::MountByType.best_for(self, possible_mount_bys)
+    end
+
+    # Most suitable name to reference the block device
+    #
+    # @see #preferred_mount_by
+    #
+    # @return [String]
+    def preferred_name
+      path_for_mount_by(preferred_mount_by)
     end
 
     protected
--- a/src/lib/y2storage/blk_device.rb
+++ b/src/lib/y2storage/blk_device.rb
@@ -183,7 +183,7 @@ module Y2Storage
     # @see #udev_paths
     # @return [Array<String>]
     def udev_full_paths
-      udev_paths.map { |path| File.join("/dev", "disk", "by-path", path) }
+      udev_paths.map { |path| Filesystems::MountByType::PATH.udev_name(path) }
     end
 
     # @!method udev_ids
@@ -208,7 +208,7 @@ module Y2Storage
     # @see #udev_ids
     # @return [Array<String>]
     def udev_full_ids
-      udev_ids.map { |id| File.join("/dev", "disk", "by-id", id) }
+      udev_ids.map { |id| Filesystems::MountByType::ID.udev_name(id) }
     end
 
     # @!attribute dm_table_name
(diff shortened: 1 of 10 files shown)


src/lib/y2storage/blk_device.rb
```diff
@@ -682,6 +698,15 @@ def volume_match_values
       }
     end
 
+    def available_mount_bys
```
ancorgs: This method is kind of temporary. I will now open a pull request against libstorage-ng to make `BlkDeviceImpl#possible_mount_bys` public. That would make more sense than duplicating logic here.

test/y2storage/blk_device_test.rb
```diff
@@ -445,72 +445,81 @@
...
-        it "returns the by-uuid udev path" do
-          expect(subject.path_for_mount_by(mount_by)).to eq(path_by_uuid)
+        it "returns nil" do
+          expect(subject.path_for_mount_by(mount_by)).to be_nil
         end
       end
 
-      context "and the device has no by-uuid udev path" do
-        let(:path_by_uuid) { nil }
+      context "and the device constains no filesystem" do
```
imobachgs: typo: contains

test/y2storage/blk_device_test.rb
```diff
@@ -445,72 +445,81 @@
...
     end
 
-    context "when mounting by id" do
-      let(:mount_by) { Y2Storage::Filesystems::MountByType::ID }
+    context "when mounting by label" do
+      let(:mount_by) { Y2Storage::Filesystems::MountByType::LABEL }
 
-      before do
-        allow(subject).to receive(:udev_full_ids).and_return(paths_by_id)
+      context "and the device contains a filesystem a label" do
```
imobachgs: "a filesystem a label" ?

test/y2storage/blk_device_test.rb
```diff
@@ -445,72 +445,81 @@
...
-        it "returns the first by-id udev path" do
-          expect(subject.path_for_mount_by(mount_by)).to eq(path_by_id1)
+        it "returns nil" do
+          expect(subject.path_for_mount_by(mount_by)).to be_nil
         end
       end
 
-      context "and the device has no by-id udev paths" do
-        let(:paths_by_id) { [] }
+      context "and the device constains no filesystem" do
```
imobachgs: "contains"

test/y2storage/blk_device_test.rb
```diff
@@ -1426,4 +1451,42 @@ def create_encrypted_partition(disk, slot_index)
...
+
+    before do
+      allow(device).to receive(:udev_paths).and_return ["udev_path"]
+    end
+
+    let(:by_device) { Y2Storage::Filesystems::MountByType::DEVICE }
+    let(:by_path) { Y2Storage::Filesystems::MountByType::PATH }
+    let(:all_suitable) { [by_device, by_path] }
+
+    it "returns the best mount by from all the suitable ones" do
```
imobachgs: To be honest, I would rather prefer "mount-by" (or even *"mount by"*) instead of "mount by" ("by from" looks wrong). However, I know that we use that expression in many places and perhaps it is not worth changing it now.

ancorgs: I fully agree, I just tried to respect the original ones written by @joseivanlopez

src/lib/y2storage/blk_device.rb
ancorgs: This is the due pull request https://github.com/openSUSE/libstorage-ng/pull/713

src/lib/y2storage/blk_device.rb
```diff
@@ -508,11 +508,7 @@ def filesystem_label
     # @see #udev_paths
     # @return [String, nil]
     def udev_full_label
-      label = filesystem_label
-
-      return nil if label.nil? || label.empty?
-
-      File.join("/dev", "disk", "by-label", label)
+      Filesystems::MountByType::LABEL.udev_name(filesystem_label)
```
joseivanlopez: Now, methods like `#preferred_mount_by`, `#preferred_name` or `path_for_mount_by` do not take into account the device content. But methods `#udev_full_label` and `#udev_full_uuid` does. Maybe, those methods should not consider its content either. I.e., `#udev_full_label` should return nil by default (only luks could return something) and `#udev_full_uuid` should be nil too (only MD could return something). What do you think? Maybe this change does not pay off because it requires to adapt logic in some parts of the code.

ancorgs: Yes, those old methods look inconsistent now. But changing them is simply too risky. 

To be honest, even the fact that `MountByType` contains the word "mount" in the name and that it lives under the `Filesystem` namespace looks wrong now.

Things to consider for the future.

src/lib/y2storage/blk_device.rb
ancorgs: libstorage-ng PR already merged.

PR #1080, opened 2020-04-08
WIP: Speed up proposal

ancorgs:
## Problem

To be written.

## Solution

To be written.

## Testing

Adapted unit tests.


ancorgs:
Close in favor of #1083 (I messed up with git history).

PR #1081, opened 2020-04-09
rework VFAT mount option handling (bsc#1161771)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1161771
- and also revisiting https://bugzilla.suse.com/show_bug.cgi?id=1080731
- https://trello.com/c/vBCU6H1R

TL;DR: VFAT mounts with `iocharset=utf8` are implicitly case-sensitive. There's a separate `utf8` option (without `iocharset=`) that is to be used instead. `utf8` is case-insensitve in some limited fashion which is good enough to avoid the `EFI` spelling ambiguity that caused [bsc#108073]( https://bugzilla.suse.com/show_bug.cgi?id=108073).

## Solution

The goal is to avoid `iocharset=utf8` in favor of `utf8`.

This also changes the defaults of `iocharset` and `codepage` for newly created vfat filesystems to "" (empty) if the default encoding and code page are used. This has the advantage that entering the fstab option dialog no longer adds `iocharset` and `codepage` options with default values.

The net effect is that for a utf8 locale the default vfat mount option is just `utf8` and for a latin1 locale even `defaults`.

Users can still set any combination they like in the fstab option dialog. Existing values from `/etc/fstab` are kept unchanged. Even combining `iocharset=utf8` with `utf8` is possible.

## Todo

- MAYBE? (rather not): add warning popup if the user goes for `iocharset=utf8` for /boot - this looks like something for the boot requirements checker

### DIFF
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -869,7 +869,7 @@ module Y2Partitioner
           "<p><b>Charset for File Names:</b>\n" \
           "Set the charset used for display of file names in Windows partitions.<br>\n" \
           "Note that choosing 'utf8' here will implicitly make " \
-          "filename comparision case-sensitve.<br>\n" \
+          "filename comparison case-sensitve.<br>\n" \
           "To avoid this, leave this field empty and instead add 'utf8' " \
           "to the Arbitrary Option Value field." \
           "</p>\n"
--- a/src/lib/y2partitioner/widgets/fstab_options.rb
+++ b/src/lib/y2partitioner/widgets/fstab_options.rb
@@ -722,10 +722,14 @@ module Y2Partitioner
 
       # @macro seeAbstractWidget
       def help
-        _("<p><b>Arbitrary Option Value:</b> " \
+        _(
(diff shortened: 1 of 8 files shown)


shundhammer:
LGTM so far.

src/lib/y2partitioner/widgets/fstab_options.rb
```diff
@@ -859,8 +865,15 @@ def label
 
       # @macro seeAbstractWidget
       def help
-        _("<p><b>Charset for File Names:</b>\nSet the charset used for display " \
-        "of file names in Windows partitions.</p>\n")
+        _(
+          "<p><b>Charset for File Names:</b>\n" \
+          "Set the charset used for display of file names in Windows partitions.<br>\n" \
+          "Note that choosing 'utf8' here will implicitly make " \
+          "filename comparision case-sensitve.<br>\n" \
```
shundhammer, line 872: Typo: comparison

ancorgs:
This change should not be targeting SLE-15-SP2, but only TW and future SLE releases.

Thus, this pull request should not have been merged in the master branch before the SLE-15-SP2 branching and it should not have been versioned as 4.2.x.

I will try to fix all that.

PR #953, opened 2019-08-08
Initial guided proposal: multidisks first

dgdavid:
## Problem

Currently, the initial guided proposal tries first to install the product into individual disks, adjusting the default settings as much as possible (disabling snapshots, separated home, etc) before attempt to use several disks to fulfill the requirements. This is not useful in certain scenarios in which the product has been designed to be expanded over as many disks as possible, like SUSE Manager.

- https://trello.com/c/tSrlEnnb/
- Related to https://github.com/yast/yast-storage-ng/pull/936

## Solution

Add a new configuration value to the control file and use it to decide how many groups of candidate devices will be generated.

It has been called `multidisk_first` and it's `false` by default. When set to `true`, the initial proposal will be tried using only one group containing all available candidate devices, resulting in a proposal which will use several disks before adjusting the settings.

## Testing

- Added unit tests
- Also tested manually via _dud_.

## Screenshots

<details>
<summary>Click to show/hide screenshots</summary>

---

Screenshot_test_2019-08-11_01:09:00

<p align="center"><sup>Installing SUSE Manager Server</sup></p>

---

Screenshot_test_2019-08-11_01:30:22

<p align="center"><sup>Initial proposal uses as much devices as possible (1/2).</sup></p>

Screenshot_test_2019-08-11_01:30:30

<p align="center"><sup>Initial proposal uses as much devices as possible (2/2).</sup></p>

---

Screenshot_test_2019-08-11_01:31:07

<p align="center"><sup>Using the expert partitioner to see the initial proposal.</sup></p>

---

Screenshot_test_2019-08-11_01:32:22

<p align="center"><sup>Device graph of the initial proposal.</sup></p>

</details>

PR #1083, opened 2020-04-16
Speed up the initial proposal, specially for suma_multidisc_role

ancorgs:
## Notes for reviewers

This replaces #1080 (I messed up with git)

Review commit by commit, all of them make sense on their own and contain descriptive messages.

This pull request would kickstart the SLE-15-SP2 branch, so it also includes a commit to adapt the Rakefile.

## Problem

Quite some time ago, we implemented several configuration options for the storage proposal to enable the creation of the `suma_multidisc_role` for SUSE Manager Server (see https://github.com/yast/skelcd-control-suse-manager-server/pull/15).

Apart from the three new options (`allocate_volume_mode`, `separate_vgs` and `delete_resize_configurable`) introduced to change how the Guided Setup works, we also added the option `multidisk_first` to influence the initial proposal (the one that is calculated before the user has had any chance to influence the result or to select the disks to use).

But all those changes leaded to two performance problems in some situations.

- In a system with many disks and `multidisk_first` set to true, the initial proposal tried an insane amount of possible combination of the disks. That was very slow. It was also pretty useless, since it's really unlikely that a user installing in a system with more than five disks could be satisfied with the initial proposal.

- In addition, when `allocate_volume_mode` was set to "device" and the disk(s) in the system were small, the initial proposal was also slow because it insisted in trying (by brute-force) combinations that had no chance to work (locating certain volumes in disks that were too small to accommodate them).

Of course, the combination of all these circumstances (`multidisk_first == true` + `allocate_volume_mode == :device` + a system with many small disks) leaded to a dreadfully slow initial proposal that looked like a hung system.

## Solution

Three aspects were improved:

- (A) When `multidisk_first` is set, bring back the limit that was introduced as an installer self-update for SLE-15-GA via #1005. That limit doesn't exist in SLE-15-SP1 because multidisk attempts are almost never tried (since the option `multidisk_first` does not even exist). The limit speeds up  dramatically the calculation of the initial proposal in systems with many disks, both for the `suma_multidisc_role` case and for an hypothetical case in which a more traditional proposal would be configured with `multidisk_first` set.

- (B) When `allocate_volume_mode` is set to "device", many useless combinations of disks and volumes are now discarded much earlier in the initial proposal process. That also leads to a huge speedup.

- (C) Other small fixes and speed improvements implemented along the way (see separate commits).

## Testing

Added a new unit test. It would have been green without the code changes as well, but it would take an insane amount of time to complete. Now is on pair with any other test.

## Benchmark 1

Just executing the unit test mentioned above in a pretty limited machine and verifying the impact of each of the improvements described in "solution" (only one execution of each scenario, so don't take the numbers too literally).

- Without any improvement, the test took >5:00:00 (more than five hours, not sure how much).
- Applying only (A), it took 3:07
- Applying only (B), it took 2:58
- Applying (A) + (C), it took 2:58.
- Applying (A) + (B), it took  0:00.52 (yes, less than one second)
- Applying (A) + (B) + (C), it took also 0:00.52

## Benchmark 2

Yet another setup designed to show the slow points. See all the details of the bechmark [in this gist](https://gist.github.com/dgdavid/ae58a9b0c468fefb974ff3ef3ba19973). In summary:

- Before the changes, it took >11:00:00 (more than eleven hours)
- Applying only (A), it took 45:37
- Applying (A) + (B) + (C), it took 0:14

### DIFF
--- a/src/lib/y2storage/initial_guided_proposal.rb
+++ b/src/lib/y2storage/initial_guided_proposal.rb
@@ -255,9 +255,10 @@ module Y2Storage
 
     # Sorted list of disks to be tried as root device.
     #
-    # If the current settings already specify a root_device, the list will contain only that one.
+    # In the theoretical case in which the current settings already specify a root_device (in practice,
+    # the initial guided proposal never receives such settings), the list will contain only that device.
     #
-    # Otherwise, it will contain all the candidate devices, sorted from bigger to smaller disk size.
+    # Otherwise, it will contain all the candidate devices, sorted in the order they should be tried.
     #
     # @return [Array<String>]
     def candidate_roots
@@ -325,7 +326,7 @@ module Y2Storage
     #
     # @return [Array<VolumeSpecificationsSet>]
     def non_root_volumes_sets
-      settings.volumes_sets.select(&:proposed?).reject(&:root?)
+      proposed_volumes_sets.reject(&:root?)
     end
   end
 end
--- a/src/lib/y2storage/initial_guided_proposal.rb
+++ b/src/lib/y2storage/initial_guided_proposal.rb
@@ -165,6 +165,8 @@ module Y2Storage
     def try_with_each_permutation
       return try_with_each_target_size if settings.allocate_mode?(:auto)
 
+      raise Y2Storage::Error if useless_permutations?
+
       try_with_each(devices_permutations) do |permutation|
         non_root_volumes_sets.each_with_index do |set, idx|
           set.device = permutation[idx]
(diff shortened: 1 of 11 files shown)


src/lib/y2storage/initial_guided_proposal.rb
```diff
@@ -255,32 +282,21 @@ def groups_of_candidate_devices
...
+    # Otherwise, it will contain all the candidate devices, sorted in the order they should be tried.
     #
     # @return [Array<String>]
     def candidate_roots
       return [settings.root_device] if settings.explicit_root_device
 
+      # We use #explicit_candidate_devices because it contains all the devices of the candidate_group.
+      # That's not always true for #candidate_devices. If there are fewer proposed_volumes_sets than
+      # devices in the group, using #candidate_devices could result in giving up before trying all the
+      # possibilities.
```
dgdavid, line 297: Just for the record, this is the missing explanation in the commit https://github.com/yast/yast-storage-ng/commit/61521afeaf85da19af618855c0c7dd0ceec16e77, introduced as part of #953.

Thanks @ancorgs!

ancorgs:
Submitted to SLE-15-SP2 (and thus Leap 15.2) as https://build.suse.de/request/show/216163

PR #1086, opened 2020-04-22
disallow to edit LVM cache pools

aschnell:
## Problem

The expert partitioner allows to edit a LVM cache pool because it does not use the ```is_usable_blk_device()``` function in this code path.

## Solution

For once add a special check for LVM cache pools to get a good error message but also use ```is_usable_blk_device()``` as a safety net.

## Testing

- Added a new unit test.
- Tested manually.

## Also See

- https://trello.com/c/D8BBQ5S3/280-8-ostumbleweed-p3-1099744-follow-up-probe-lvmcache


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Apr 22 13:29:05 CEST 2020 - aschnell@suse.com
+
+- disallow to edit LVM cache pools
```
shundhammer, line 4: The OBS check scripts might want a bug or feature number.

shundhammer: Trello says this is for
https://bugzilla.suse.com/show_bug.cgi?id=1099744

aschnell: It is remotely related to that bug. I have added the number.

PR #1087, opened 2020-04-22
update /etc/crypttab status properly (bsc#1118977)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1118977
- https://trello.com/c/jtojel4R

New encrypted volumes are added to `/etc/crypttab` (even when not mounted).

## Analysis

There are two issues:

1. The default value for `in_etc` is true for newly created encrytpted volumes.  Since the `#update_etc_status` method ensures the status can't be set to false when it was originally true, this unavoidably creates an initial crypttab entry.
To avoid this, set the `in_etc` flag to false explicitly for freshly created encrypted volumes.

2. The `#update_etc_status` method is called when you switch between mounting/not mounting in the fstab dialog.
But when encryption had just been enabled, the encrypted device does not yet exist in the device tree. It's created only after going through the password dialog.
This means `#update_etc_status` has to be called also after creating an encryption device to get a consistent state.

## Comments

- This applies equally to `/etc/mdadm.conf` - currently it's not possible to avoid an entry there - even if the RAID is not mounted. The use-cases are a bit different compared to encryption, so this might be the desired state.

- I've found the logic aroud `etc_status_autoset` very hard to follow and have rewritten that part to something I might still understand in a few years from now...

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Apr 23 14:32:38 UTC 2020 - Steffen Winterfeldt <snwint@suse.com>
+
+- update /etc/crypttab status properly (bsc#1118977)
+- 4.3.2
+
+-------------------------------------------------------------------
 Wed Apr 22 13:29:05 CEST 2020 - aschnell@suse.com
 
 - disallow to edit LVM cache pools (related to bsc#1099744)
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-storage-ng
-Version:        4.3.1
+Version:        4.3.2
 Release:        0
 Summary:        YaST2 - Storage Configuration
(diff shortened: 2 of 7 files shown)


ancorgs:
`etc_status_autoset` is there to ensure we don't remove devices from `/etc/crypttab`, `/etc/mdadm` and friends if those devices were already there at the time of probing.

We only want to revert the decision of listing the devices in those files if adding them was our decision, not if they were there for other circumstances.

ancorgs:
That's why we added the test that is currently broken: to ensure we don't remove stuff that was originally in `/etc/crypttab` and `/etc/mdadm` (ie. found to be there already during probing)

wfeldt:
> That's why we added the test that is currently broken: to ensure we don't remove stuff

The test is wrong: it only ensures `in_etc` is always true. And that's wrong.

And that's also the problem with the existing code: since a new encryption has `in_etc` set to true the code ensures it's never false - so a new encryption is always in crypttab.

ancorgs:
LGTM

PR #1088, opened 2020-04-24
add x-initrd.attach option to crypttab if needed (bsc#1168465)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1168465
- https://trello.com/c/01yaT7SA

The new crypttab option `x-initrd.attach` should be added for filesystems that are attached in the initrd.

See [crypttab(5)](https://www.freedesktop.org/software/systemd/man/crypttab.html).

## Solution

Add this option for newly created filesystems mounted at `/` and all filesytems with `x-initrd.mount` fstab option.


aschnell:
Changelog entry is missing.

wfeldt:
I knoiw, I will add it before moerging.

PR #1090, opened 2020-05-07
remove obsolete proposal_settings_editable

wfeldt:
## Task

- https://bugzilla.suse.com/show_bug.cgi?id=1171423
- https://trello.com/c/BqWDHmp1

Remove obsolete `proposal_settings_editable` setting.

### DIFF
--- a/test/y2storage/clients/inst_disk_proposal_test.rb
+++ b/test/y2storage/clients/inst_disk_proposal_test.rb
@@ -46,17 +46,13 @@ describe Y2Storage::Clients::InstDiskProposal do
 
     let(:partitioning_section) { {} }
 
-    context "the proposal dialog" do
-      before do
-        allow(proposal_dialog).to receive(:run).and_return(:abort)
-      end
+    it "it does not exclude the Guided Setup button" do
+      allow(proposal_dialog).to receive(:run).and_return(:abort)
 
-      it "does not exclude the Guided Setup button" do
-        expect(Y2Storage::Dialogs::Proposal).to receive(:new)
-          .with(anything, anything, excluded_buttons: []).and_return(proposal_dialog)
+      expect(Y2Storage::Dialogs::Proposal).to receive(:new)
+        .with(anything, anything, excluded_buttons: []).and_return(proposal_dialog)
 
-        client.run
-      end
+      client.run
     end
 
     context "when running the client for the first time" do
--- a/test/y2storage/clients/inst_disk_proposal_test.rb
+++ b/test/y2storage/clients/inst_disk_proposal_test.rb
@@ -46,31 +46,14 @@ describe Y2Storage::Clients::InstDiskProposal do
 
     let(:partitioning_section) { {} }
 
-    context "when the Guided Setup can be shown" do
+    context "the proposal dialog" do
       before do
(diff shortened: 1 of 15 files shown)


test/y2storage/clients/inst_disk_proposal_test.rb
```diff
@@ -46,36 +46,19 @@
 
     let(:partitioning_section) { {} }
 
-    context "when the Guided Setup can be shown" do
+    context "the proposal dialog" do
```
ancorgs, line 49: This usage of `context` is a little bit weird. It results in something like "InstDiskProposal#run the proposal dialog does not exclude the buttons".

I would say the whole context could be removed. Something like this would make more sense to me: "InstDiskProposal#run opens the proposal dialog with all the buttons" (or maybe "...opens the proposal dialog without excluding any button").

In any case, I don't understand the need of the context (and no, changing the subject of the sentence is not what a context should be used for).

test/y2storage/dialogs/guided_setup/select_filesystem/ng_test.rb
```diff
@@ -68,55 +68,37 @@
...
+        expect(dialog.skip?).to eq true
       end
     end
 
-    context "when the Guided Setup is allowed" do
-      let(:allowed) { true }
-
-      context "and the proposal settings contain no volumes" do
-        let(:volumes) { [] }
+    context "and none of the volumes are configurable" do
```
ancorgs, line 79: This should start with something line "when" or "if" instead of "and", isn't it? Right now is read as "SelectFilesystem::Ng#skip? and none of the volumes are configurable returns true" instead of "SelectFilesystem::Ng#skip? when none of the volumes are configurable returns true".

test/y2storage/dialogs/guided_setup/select_filesystem/ng_test.rb
```diff
@@ -68,55 +68,37 @@
...
-          [
-            double("VolumeSpecification", configurable?: false),
-            double("VolumeSpecification", configurable?: false)
-          ]
-        end
-
-        it "returns true" do
-          expect(dialog.skip?).to eq true
-        end
+    context "and any volume is configurable" do
```
ancorgs, line 92: Same here with the "and".

test/y2storage/dialogs/guided_setup/select_scheme_test.rb
```diff
@@ -39,21 +39,7 @@
...
-
-      it "returns true" do
-        expect(subject.skip?).to eq(true)
-      end
-    end
-
-    context "when the Guided Setup is allowed" do
-      let(:allowed) { true }
-
+    context "since the Guided Setup is always allowed" do
```
ancorgs, line 42: So the context makes no sense here. Just remove it. Contexts are not there to attach explanatory texts to the output. :wink:

test/y2storage/clients/inst_disk_proposal_test.rb
```diff
@@ -46,17 +46,13 @@
 
     let(:partitioning_section) { {} }
 
-    context "the proposal dialog" do
-      before do
-        allow(proposal_dialog).to receive(:run).and_return(:abort)
-      end
+    it "it does not exclude the Guided Setup button" do
```
ancorgs, line 49: NP: `it "it` looks weird. Not a big deal anyways

PR #1091, opened 2020-05-08
Cleanup AutoYaST issue and profile handling

schubi2:
## Problem

Each module (storage, firewall,network) has his own AutoYaST Issue  and profile handling which is incompatible and doubles the code.
bsc#1171335

## Solution

Bring base classes to yast2 package and cleanup the code of the other packages which will use
these common base classes:
https://github.com/yast/yast-yast2/pull/1050
https://github.com/yast/yast-yast2/pull/1055

## Testing

- Adding test cases for showing the issues
- Tested manually.

Bildschirmfoto_2020-05-12_14-11-12


src/lib/y2storage/autoinst_issues/could_not_calculate_boot.rb
```diff
@@ -17,12 +17,12 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "y2storage/autoinst_issues/issue"
+require "installation/autoinst_issues/issue"
 
-module Y2Storage
+module Installation
```
imobachgs, line 22: I think we should not move this code to a different module/namespace. IMHO, it is OK that specific storage-ng issues live under the `Y2Storage::AutoinstIssues` namespace.

PR #1092, opened 2020-05-12
Avoid crash if no blk_device is reported for an LVM PV (bsc#1170216)

ancorgs:
## Problem

When a system contains multipath devices but multipath is not activated, the devicegraph can end up with a LVM PV associated to several parent block devices, instead of the expected layout that is every PV associated always to a single parent.

four-parents-means-no-parent

That situation is wrongly managed by libstorage-ng (`LvmPv#has_blk_device` returns false) and by the Partitioner, which assumes that every PV is always located into a single block device and crashes if that's not the case.

In the past, when libstorage-ng found a PV associated to several devices it used to raise an exception and then YaST showed [an explanatory warning](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/callbacks/libstorage_callback.rb#L135) to the user, so the crash was kind of expected if the user decided to continue. But since the output of `lvmtools` changed, that's not happening anymore and the crash hits the user by surprise.

- https://bugzilla.suse.com/show_bug.cgi?id=1170216
- https://trello.com/c/5ZxGomMg/1868-2-sles15-sp2-p2-1170216-build-1814-expert-partitioner-fails-to-create-lvm-on-zfcp-with-disabled-multipath

## Solution

This avoids the Partitioner crash by ignoring PVs that are on top of several block devices (to be precise, it actually ignores PVs for which libstorage-ng reports no block device, since that's the current status of the library). That should be enough for the upcoming SLE-15-SP2, which is already in advanced Beta version and, thus, not open for drastic changes.

A more general solution, including fixing the detection of inactive multipath setups, will be developed for the master branch in the short future.

## Testing

- Added a new unit test
- Tested manually (using the devicegraph from the logs attached to the bug to mock the system)


ancorgs:
The rubocop error is a temporary problem with the docker image used by Cirrus CI. As you can see, everything is green in Travis.

src/lib/y2storage/blk_device.rb
```diff
@@ -377,7 +377,9 @@ def mount_point
     # @return [LvmPv] nil if neither the raw device or its encrypted version
     #   are used as physical volume
     def lvm_pv
-      descendants.detect { |dev| dev.is?(:lvm_pv) && dev.plain_blk_device == plain_device }
+      descendants.detect do |dev|
```
imobachgs, line 380: np: I prefer `#find` to `#detect` (just like mentioned in [this guide](https://github.com/rubocop-hq/ruby-style-guide#mapfindselectreduceincludesize)).

src/lib/y2storage/blk_device.rb
```diff
@@ -377,7 +377,9 @@ def mount_point
...
+        dev.is?(:lvm_pv) && dev.blk_device && dev.plain_blk_device == plain_device
```
jreidinger, line 381: I think it deserve explanation why lvm_pv can be non blk device.

jreidinger: I mean not here, but in code.

PR #1095, opened 2020-05-12
ensure crypttab entries use correct mount_by variant (bsc#1165702)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1165702
- https://trello.com/c/RJ2YOnR3

## Solution

There's no way to directly influence the crypttab entry in the UI. So, some heuristics is needed.

The code tries to use the same mount_by for crypttab that was used for fstab - when the mount point is created. This way it's a plain device if the user decides to set the default in sysconfig to `device` and `uuid` otherwise.

## Remark

I would not try to put this into SLE15-SP2.


ancorgs:
I don't feel super-comfortable with this solution which, as I see it, is introduced as a workaround to mask an inconsistency in the behavior of libstorage-ng.

In theory libstorage-ng honors the default mount_by where it makes sense. The value of that default mount_by is set by YaST, as documented [here](https://github.com/yast/yast-storage-ng/blob/master/doc/mount_by.md).

But in fact:

- When a new encryption device is created using `BlkDevice#create_encryption` the library indeed honors the default mount_by. *Good*.
- When an encryption device is probed and it's present in the existing crypttab, the library sets `Encryption#mount_by` based in the value of the crypttab entry, no matter what the default mount_by is. *Good*.
- When an encryption device is probed but it is not present in the existing crypttab, the library creates the `Encryption` object without honoring the default mount_by. *Wrong* (IMHO).

So I would rather make the behavior of libstorage-ng consistent so all the logic we already have in place in YaST would simply work for probed devices in the same way it currently does for created ones.

ancorgs:
> I don't feel super-comfortable with this solution [...]

Among other things, because `MountPoint#mount_by` and `Encryption#mount_by` are quite different things with a different meaning and different implications for each possible value, I wouldn't sync the value of both attributes as a general approach.

So far both attributes share the name, the type and they also share the setting that defines its default value... but the more I work with this part, the more I think that's not 100% correct. In my opinion, whatever we change in this area should go in the direction of separating and differentiating both attributes more, instead of coupling them closer.

aschnell:
We can argue forever about what is good and wrong. I consider consistency important so mount-by during probing should behave the same for fstab and crypttab.

Concerning differentiating that between fstab and crypttab I do not see the point. People who prefer by-id or by-path links likely want the same in both.

ancorgs:
> We can argue forever about what is good and wrong. I consider consistency important so mount-by during probing should behave the same for fstab and crypttab.

But the third case (of the three in me previous comment) does not exist in fstab, isn't it? I mean, when probing a MountPoint you always know the mount_by that you must use. If you are probing it, is because the mount point is active in the system or it's in fstab, so you are:

 - creating the mount point => use the default
 - probing an active mount point from the system => use what the system uses
 - probing a mount point from the system that is only in fstab => use the mount_by from fstab

So there is only one situation in which the value is not defined by the system and, thus, you apply the default.

Only with encryption devices you can face two different situations in which the mount_by is not defined, so you must apply the default in both:

 - creating the encryption => use the default
 - probing an encryption listed in crypttab from the system => use the mount_by from crypttab
 - probing an encryption not listed in crypttab => use the default

Not using the default in that latest case is inconsistent. That's what a default value is for.

wfeldt:
I think that's beside the point.

It's in no way libstorage's buisiness to decide how to access a device. During probing, the entry does not exist yet. It may or may not make something up at this point. But it's entirely up to yast to look at the situation *again* when the mount point is actually created. Whatever yast puts into crypttab must not depend on the initial value libstorage made up. There is simply nothing to 'probe' here.

Particularly as yast itself uses quite an elaborate decision process as this:

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/filesystems/mount_by_type.rb#L39-L66

As long as you don't replicate this in libstorage (for no good reason, imho) you will continue to see differences.

wfeldt:
> Among other things, because MountPoint#mount_by and Encryption#mount_by are quite
> different things with a different meaning and different implications for each possible value,

I feel that's a bit out of scope here, but anyway:

Swap, for example is neither a filesystem nor does it get mounted. Yet I haven't seen you making a point to treat it separately.

What matters is that `mount_by` has the easily to grasp semantics to mean 'how to pick the device name to use'. And that usage is the same both in fstab and crypttab. That a separate crypttab exists at all can be seen as a mistake from the beginning, imho.

Sure, a Java developer (for example) would feel compelled to use a more explicit method name like `getUdevClassByWhichToAccessTheDevice` but I feel that `mount_by` is doing just fine.

wfeldt:
> I wouldn't sync the value of both attributes as a general approach.

We have already things like `adjust_crypt_options` and `update_etc_status` that do exactly that. One could indeed argue that the sync belongs into  `adjust_crypt_options` but that's a bit far reaching, imho, especially as we hide crypttab entirely from users so far and they have no good way of controlling what's going on.

ancorgs:
> Yet I haven't seen you making a point to treat it separately.

Let's not open that can.

ancorgs:
My point is:

- This PR is a potential fix for the reported issue, but it adds complexity and affects all cases in which we create a mount point on top of an encrypted device. We don't even dare to merge the fix into the SLE-15-SP2 branch.

- Another potential fix would be ensuring libstorage-ng uses the default value of mount_by in all the situations in which that value is unknown. That's safe enough. I would feel fully confident merging that in SLE-15-SP2. And, to be honest, I think that it adds consistency to the library (when the value us uncertain, use always the configurable default value, instead of using the default in some situations and `DEVICE` in others).

wfeldt:
- I do not consider this to be suitable for SLE15-SP2 regardless the implementation. As mentioned in the Trello board, there's not even a bug against SLE15 at all and SP2 is in it's final stage. A change in the library would be a much more risky change, btw.
- As I pointed out above it's not libstorage 'simply' taking the default. yast-storage doesn't spent a whole class to do this task for no reason. You'd have to replicate this in libstorage exactly. So this is would be a pointless excercise however you look at this.
- And finally, it is simply *yast*'s responsibility to ensure things are *created* correcty.

ancorgs:
If I got your point, YaST should generally not trust the values libstorage-ng uses to initialize the `mount_by` fields. So, instead of just setting `Storage.default_mount_by` it should include it's own mechanism (likely in a separate class) to handle all that.

Fair enough, but then let's merge this PR and create a separate card for designing and implementing all that properly. Because at the current stage it's quite a mess, with the library offering a mechanism for setting the default value that YaST sometimes trusts and sometimes overrides.

wfeldt:
My point is that yast already **does** do this with `best_for`

https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/filesystems/mount_by_type.rb#L99-L107

maybe not consistently.

And you cannot pick 'just' the default as it may not be applicable. Like `by-label`. So there will be a bit more logic involved.

And, yes, I think yast should not (necessarily) rely on library defaults.

PR #1096, opened 2020-05-18
added translation for new resize blocking reason

aschnell:
## Problem

The message for a new resize blocking reason is missing.

- https://github.com/openSUSE/libstorage-ng/pull/744

## Solution

Add the message.


aschnell:
The build failure is due to Travis CI using an outdated libstorage-ng. But without this PR yast2-storage-ng already fails in OBS.

PR #1098, opened 2020-05-26
New idea for partitioner_ui: overview tab

ancorgs:
We are collecting ideas for the UI of the partitioner at `doc/partitioner_ui.md`

This pull request adds another one.

For a more pleasant review, use the "view file" option on the markdown file.

### DIFF
--- a/doc/partitioner_ui/ideas/overview.md
+++ b/doc/partitioner_ui/ideas/overview.md
@@ -3,7 +3,7 @@
 This is part of the [bigger document](../../partitioner_ui.md) about rethinking the YaST Partitioner
 user interface.
 
-When clicking in a device, the Partitioner has traditionally presented a tabbed interface in which
+When clicking on a device, the Partitioner has traditionally presented a tabbed interface in which
 the first tab is always the so-called "Overview" that presents all kind of technical details about
 the device. The relationship with other devices is then presented in separate tabs like
 "Partitions", "Logical Volumes", "Physical Volumes", "Used Devices", etc. With all the possibilities
@@ -30,7 +30,7 @@ to perform actions.
    │Size: 500.00 GiB - Model: HGST HTS75000          │
    │The device is empty                              │
    │                                                 │
-   │ - Create a file system directly in the disk     │
+   │ - Create a file system directly on the disk     │
    │   [Format/mount]                                │
    │                                                 │
    │ - Partition the disk                            │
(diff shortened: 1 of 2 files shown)


doc/partitioner_ui/ideas/overview.md
```diff
@@ -0,0 +1,193 @@
+## Idea: Adaptative Overview Tab
+
+This is part of the [bigger document](../../partitioner_ui.md) about rethinking the YaST Partitioner
+user interface.
+
+When clicking in a device, the Partitioner has traditionally presented a tabbed interface in which
```
shundhammer, line 6: When clicking **on** a device

doc/partitioner_ui/ideas/overview.md
```diff
@@ -0,0 +1,193 @@
...
+### Overview Tab for a Disk Being Used as Physical Volume
+
+```
+   Hard Disk: /dev/sda
+   ┌Overview──Details ───────────────────────────────┐
+   │                                                 │
+   │Size: 500.00 GiB - Model: HGST HTS75000          │
+   │The device is part of the LVM /dev/system        │
+   │                                                 │
+   │No actions can be performed directly in          │
```
shundhammer, line 96: directly **on** the disk

doc/partitioner_ui/ideas/overview.md
```diff
@@ -0,0 +1,193 @@
...
+"Partitions", "Logical Volumes", "Physical Volumes", "Used Devices", etc. With all the possibilities
+currently offered by the Partitioner, it's usually not so obvious which of those tabs is the most
+useful one for each device.
+
+This idea proposes to present always the same two tabs for all devices.
+
+ - A first one called "Overview" offering the possible actions. Its content would change based on
+   the content of the device.
+ - A second "Details" tab that would contain all the technical details that are currently presented
+   in the "Overview" one, but without any action button.
```
shundhammer, line 18: "Details" tab without any action button? Really? Will that work?
I fear we'll need to have buttons there as well. We may offer different actions in the "Overview" tab; possibly simpler ones or ones that the user can make more sense of. But I think we will sill need lower-level actions in the "Details" tab as well. If we don't, we will always need to present every least little detail in the "Overview" tab just so the user can make any possible changes, and having everything there would pretty much defeat the purpose of the overview.

PR #1107, opened 2020-06-30
install nvme-cli if NVMe devices are present

aschnell:
## Problem

With NVMeOF devices the NVMe tools in the package nvme-cli are needed.

## Solution

Install the tools if libstorage-ng reports the new feature UF_NVME.

## Testing

- Tested manually.

## See

- https://trello.com/c/QIyuWNgI/1911-5-sles15-sp2-p2-1172866-nvme-cli-not-installed-when-root-is-an-nvme-device
- https://bugzilla.suse.com/show_bug.cgi?id=1172866

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 container:
-  image: registry.opensuse.org/yast/head/containers/yast-ruby:latest
+  image: registry.opensuse.org/yast/sle-15/sp2/containers/yast-ruby
 
 task:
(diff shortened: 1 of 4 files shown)


shundhammer:
Please add the two new storage features to that failing unit test.

aschnell:
Cirrus CI is using the libstorage-ng from factory, not SLE 15 SP2. UF_PMEM is not present in SLE 15 SP2.

PR #1109, opened 2020-07-03
AutoYaST delegates to initial proposal

joseivanlopez:
## Problem

AutoYaST proposal falls back to the guided proposal when no partitions are indicated in the AutoYaST profile. This fallback only performs two tries at most: with max and desired device sizes. For example, it does not try with every candidate disk as possible boot disk.

Because of that, the boot disk is always fixed to the default disk reported by the "Boot Requirements Checker" (first disk as general default), making the installation impossible when the boot disk does not belong to the list of candidates devices.

This bug becomes very evident when the AutoYaST profile does not include the first disk as possible candidate disk:

~~~
<partitioning config:type="list">
    <drive>
      <device>/dev/sdc</device>
      <disklabel>gpt</disklabel>
      <initialize config:type="boolean">true</initialize>
      <use>all</use>
    </drive>
</partitioning>
~~~  

In this case, */dev/sdc* is the only one candidate disk to perform the installation, but the BIOS BOOT partition would always be proposed in the first disk (i.e., */dev/sda*).

* https://bugzilla.suse.com/show_bug.cgi?id=1173610

## Solution

Now on AutoYaST falls back to the "Initial Guided Proposal", so more proposal attempts are performed (e.g., by rotating the boot disk among all the candidate disks).

BONUS: AutoYaST profile allows to specify the *enable_snapshots* attribute for a *drive*, but this attribute is not properly transferred to the proposal settings. This was also fixed as part of this PR.

## Testing

* Added unit tests.

### DIFF
--- a/src/lib/y2storage/autoinst_proposal.rb
+++ b/src/lib/y2storage/autoinst_proposal.rb
@@ -283,7 +283,7 @@ module Y2Storage
     # @see Y2Storage::BlkDevice#name
     def proposal_settings_for_disks(drives)
       settings = @proposal_settings || ProposalSettings.new_for_current_product
-      settings.use_snapshots = drives.use_snapshots?
+      drives.use_snapshots? ? settings.force_enable_snapshots : settings.force_disable_snapshots
       settings.candidate_devices = drives.disk_names
       settings
     end
--- a/src/lib/y2storage/autoinst_proposal.rb
+++ b/src/lib/y2storage/autoinst_proposal.rb
@@ -22,6 +22,7 @@ require "y2storage/proposal_settings"
 require "y2storage/exceptions"
 require "y2storage/planned"
 require "y2storage/proposal"
+require "y2storage/guided_proposal"
 
 module Y2Storage
   # Class to calculate a storage proposal for autoinstallation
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -190,23 +191,24 @@ def root?(devices)
...
+    #
+    # @return [Devicegraph] Proposed devicegraph using the guided proposal approach
     def propose_guided_devicegraph(devicegraph, drives)
       devicegraph = clean_graph(devicegraph, drives, [])
-      begin
-        guided_devicegraph_for_target(devicegraph, drives, :desired)
-      rescue Error
-        guided_devicegraph_for_target(devicegraph, drives, :min)
-      end
+
```
imobachgs, line 206: np: Why those blank lines?

src/lib/y2storage/autoinst_proposal.rb
```diff
@@ -296,6 +283,8 @@ def add_reduced_devices_issues(result)
     # @see Y2Storage::BlkDevice#name
     def proposal_settings_for_disks(drives)
       settings = @proposal_settings || ProposalSettings.new_for_current_product
+      # FIXME: #use_snapshots is a property of legacy settings format. To configure snapshots with the
```
imobachgs, line 286: In this case, we need to 1) open a new bug report or 2) fix it now. It is a real problem, although I would say that this is not the most common use case for AutoYaST. We already have 35 TODOs and 83 FIXMEs :smile:. Not to mention the 4 DEPRECATED notes.

src/lib/y2storage/autoinst_proposal.rb
joseivanlopez: Only a matter of styling. I tend to separate lines that are no a direct consequence of the previous one (more or less).

src/lib/y2storage/autoinst_proposal.rb
joseivanlopez: After discussing about it, finally I have also fixed this issue.

src/lib/y2storage/autoinst_proposal.rb
imobachgs: Well, in that case, 209 and 211 looks quite related to me :)

PR #1113, opened 2020-08-05
Detect subvolumes shadowed by a separate_vg (bug#1174475)

ancorgs:
## Problem

The proposals contain logic to avoid creating Btrfs subvolumes if their mount points are shadowed by any other filesystem. But that was not working properly for roles using `separate_vgs` in the configuration of the proposal.

So far, that setting is used only by the following SUSE Manager roles: `suma_multidisc_role` and `suma_retail_multidisc_role`. When one of such roles was selected, the proposal created a separate `/srv` LVM logical volume in its own dedicated volume group... but also a `/srv` subvolume for the root filesystem. That, of course, caused problems.

- https://bugzilla.suse.com/show_bug.cgi?id=1174475
- https://trello.com/c/b2NAhJzv/1990-susemanager-41-p5-1174475-srv-mountpoint-is-declared-twice-in-etc-fstab

## Solution

Make sure the detection of shadowed subvolumes also has into account the mount points of separate LVs located in their own VGs (represented by the class `Y2Storage::Planned::LvmVg`).

## Testing

Added a new unit test

### DIFF
--- a/src/lib/y2storage/planned/can_be_formatted.rb
+++ b/src/lib/y2storage/planned/can_be_formatted.rb
@@ -137,7 +137,7 @@ module Y2Storage
         return [] if subvolumes.nil?
 
         other_devices = all_devices - [self]
-        other_mount_points = other_devices.map { |dev| mount_point_for(dev) }.compact
+        other_mount_points = other_devices.flat_map { |dev| mount_points_for(dev) }
         subvolumes.select { |s| s.shadowed?(mount_point, other_mount_points) }
       end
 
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/planned/can_be_formatted.rb
```diff
@@ -254,6 +254,18 @@ def assign_mount_point(filesystem)
         filesystem.mount_point.mount_by = mount_by if mount_by
       end
 
+      # @param device [Planned::Device]
+      def mount_points_for(device)
+        points =
```
joseivanlopez, line 259: NP: I would prefer to name it `mount_points` :)

PR #1117, opened 2020-08-17
[bsc#1115749]  AutoYaST: modify the LVM VG name only when really needed

imobachgs:
Backport of https://github.com/yast/yast-storage-ng/pull/1056 as [suggested in the bug report](https://bugzilla.suse.com/show_bug.cgi?id=1115749#c14).

## Problem

Given a system where there is already an LVM VG named (e.g., `system`), if you try to reinstall using a new VG with the same name, it might happen that AutoYaST modifies the wanted name (like `system0`) to avoid *colliding* with the old name. It does not make sense, as you do not want to keep the old one.

- [bsc#1115749](https://bugzilla.suse.com/show_bug.cgi?id=1115749)

## Solution

The problem happens when you are using a full disk as a PV (it will not happen when using partitions). The reason is that `AutoinstSpaceMaker` will not have full disks into account, even when you have set the `use` to `all`.

Additionally, this PR also fixes the reusing of Bcache devices when, for some reason, the user does not mark the caching/backing devices to be reused.

## Screenshots

Before the fix:

reuse-lvm-ko

After the fix:

reuse-lvm-ok

### DIFF
--- a/src/lib/y2storage/proposal/autoinst_space_maker.rb
+++ b/src/lib/y2storage/proposal/autoinst_space_maker.rb
@@ -53,13 +53,14 @@ module Y2Storage
       def cleaned_devicegraph(original_devicegraph, drives_map, planned_devices)
         devicegraph = original_devicegraph.dup
 
-        reused_partitions = reused_partitions_by_disk(devicegraph, planned_devices)
+        reused_devices = reused_devices_by_disk(devicegraph, planned_devices)
         sid_map = partitions_sid_map(devicegraph)
 
         drives_map.each_pair do |disk_name, drive_spec|
           disk = BlkDevice.find_by_name(devicegraph, disk_name)
           next unless disk
-          delete_stuff(devicegraph, disk, drive_spec, reused_partitions[disk.name])
+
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/proposal/autoinst_space_maker.rb
```diff
@@ -72,42 +73,55 @@ def cleaned_devicegraph(original_devicegraph, drives_map, planned_devices)
...
-      # @param disk         [Disk]
-      # @param drive_spec   [AutoinstProfile::DriveSection]
-      # @param reused_parts [Array<String>] Reused partitions names
-      def delete_stuff(devicegraph, disk, drive_spec, reused_parts)
-        reused_parts ||= []
-        if drive_spec.initialize_attr && reused_parts.empty?
+      # @param devicegraph    [Devicegraph]
+      # @param disk           [Disk]
+      # @param drive_spec     [AutoinstProfile::DriveSection]
+      # @param reused_devices [Array<String>] Reused disks and partitions names
```
jreidinger, line 79: I would document that this method allows nil ( as later ones that use identical parameter reused_devices crash if nil is passed ).

PR #1118, opened 2020-08-18
Merge SLE-15-SP1 into SLE-15-SP2

imobachgs:
Merge #1117 into SP2. It adds a line to the changes log that was omitted in the original #1056 and it should have been added in the #1057.


imobachgs:
> Just a question. You said, "merge into SP2" but commits will be merged to "master". Is that ok?

No, it is not. Thanks.

PR #1121, opened 2020-08-26
[AutoYaST] Fix profile element paths

imobachgs:
## Problem

The new error reporting mechanism and the <ask-list> features use different syntax to point to a specific element in a profile (`partitioning,0,partitions,1` vs `drives [1] > partitions [2]`).

## Solution

Use the same syntax that the ask-list. This PR relies on https://github.com/yast/yast-yast2/pull/1090.

## Screenshots

ask-list-autoinst-issues


dgdavid (approved):
LGTM

PR #1123, opened 2020-09-15
Reduce tree nesting

joseivanlopez:
## Problem

The tree options *Device Graphs*, *Installation Summary* and *Settings* were moved from the left tree to the new menu bar. Now, having a first level in the tree with everything hanging on it simply does not make sense. We can remove the outer nesting of the left tree, so the first level would be "Hard disks", "Volume Management", etc.

See https://github.com/yast/yast-storage-ng/blob/master/doc/partitioner_ui.md#agreed-plan-so-far for a sketch of the expected result.

* Related to: https://bugzilla.suse.com/show_bug.cgi?id=1175489
* PBI: https://trello.com/c/9fXaJujC/2044-2-partitioner-left-tree-less-nesting

## Solution

The first level was removed.

Note: ruby 2.7 deprecates the usage of last argument as keyword parameters. Because that, we start to see a lot warnings like this:

~~~
warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
~~~

This was fixed to avoid warnings. See https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/ for more details.

## Testing

* Adapted unit tests
* Tested manually

## Screenshots

Screenshot from 2020-09-15 11-49-28

### DIFF
--- a/src/lib/y2partitioner/widgets/bcache_edit_button.rb
+++ b/src/lib/y2partitioner/widgets/bcache_edit_button.rb
@@ -27,8 +27,8 @@ module Y2Partitioner
   module Widgets
     # Button for editing a bcache
     class BcacheEditButton < DeviceButton
-      def initialize(*args)
-        super
+      def initialize(args = {})
+        super(**args)
         textdomain "storage"
       end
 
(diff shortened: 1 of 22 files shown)


joseivanlopez:
I am not sure whether we are using the best labeling. We have *View* in the menu bar, *View* as label of the tree, and then *System Overview*. It sounds very redundant to me. Apart of that, maybe we should take the opportunity to rename *Volume Management* as simply *LVM*.

dgdavid (approved):
LGTM

PR #1124, opened 2020-09-15
Bcache Caching Sets popup

joseivanlopez:
## Problem

A menu bar is being added to the Expert Partitioner. We are taking the opportunity for improving some other parts, for example the Bcache presentation. Right now, the Bcache page contains two tabs, one to list the Bcache devices and a second one to list the Bcache Caching Sets. This last tab is only informative (no actions can be done over the Bcache Caching Sets). Note that this is the only general page with two tabs. For example, Hard Disk, LVM, etc have no tabs. For a matter of homogeneity, it would be better to remove the tabs in this Bcache page and show that information as a popup by using an option of the new menu bar. 

* Part of PBI: https://trello.com/c/NITXJRBE/2055-menu-driven-interface-finish-it

## Solution

Removed tabs in the Bcache page and added a new option in the menu bar that opens a popup with the Bcache Caching Sets.

## Testing

- Added unit test
- Tested manually

## Screenshots

Screenshot from 2020-09-17 09-45-25

Screenshot from 2020-09-17 09-46-01

### DIFF
--- a/src/lib/y2partitioner/dialogs/bcache_csets.rb
+++ b/src/lib/y2partitioner/dialogs/bcache_csets.rb
@@ -1,0 +1,86 @@
+# Copyright (c) [2020] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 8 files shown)


joseivanlopez:
Tests are failing after rebasing these changes: https://github.com/yast/yast-storage-ng/pull/1125. This is expected and it will fixed in following-up PRs for https://trello.com/c/NITXJRBE/2055-menu-driven-interface-finish-it.

dgdavid (approved):
:+1:

PR #1125, opened 2020-09-16
First version of a menu bar with all the entries

ancorgs:
## Problem

The previous version of the Partitioner menu bar was intentionally incomplete since it was just a proof of concept.

## Solution

After some meetings and discussions (see, for example, [this mail thread](https://lists.opensuse.org/yast-devel/2020-09/msg00004.html)) we have agreed on the general approach the menu should follow.

This pull request implements a first version of that menu, in which all possible Partitioner actions are represented in the most consistent way we have been able to come up with.

#### System

With all the options that were below the system table, plus two extra entries to quit (with and without saving).
system

#### Modify

With a few generic actions for the device currently selected in the table. We are considering to call it "Edit", since its a more common term for a menu. But "edit" has another historical meaning in the Partitioner and we would need to change that first.
modify

#### Add

For creating new devices. The first four actions are always possible. Creating partitions and LVs is only enabled when it makes sense for the selected device.
add

#### View

With all the options that were (misplaced) in the left tree although they didn't correspond to devices in the system. In addition, the menu includes an entry to display the read-only list of csets that was available as a tab in the Bcache section.
view

#### Go

To navigate through the interface. If the page currently open refers to a device, this switches among its tabs. If not (e.g. in the "System Overview" page), the menu entries can be used to go directly to the corresponding tab of the device that is selected in the table.
go

#### Actions

Options to modify the current device. In the "System Overview" page or any other of the first level pages, the action is performed on the device selected in the table. In a page corresponding to a device, the action is performed on that device.
actions

## Still pending

The goal of this pull request is not to provide a totally polished solution, but a checkpoint from which we can continue distributing the remaining work. For a full list of everything that is still pending, see the "Tasks" section of https://trello.com/c/NITXJRBE/2055-8-menu-driven-interface-finish-it

In summary:

  - Not all menu entries actually trigger the corresponding action. There are still some wires to connect.
  - Unit tests need to be adapted (and new tests created)
  - We currently have a menu called "Modify" that contains our historical "Edit" action. We should maybe find a better name for the action. Then we could rename the menu entry to the more common "Edit".

### DIFF
--- a/src/lib/y2partitioner/widgets/main_menu_bar.rb
+++ b/src/lib/y2partitioner/widgets/main_menu_bar.rb
@@ -19,101 +19,26 @@
 
 require "yast"
 require "cwm"
-require "y2partitioner/icons"
-require "y2partitioner/execute_and_redraw"
-require "y2partitioner/actions/rescan_devices"
-require "y2partitioner/actions/configure_actions"
-require "y2partitioner/actions/import_mount_points"
-require "y2partitioner/dialogs/summary_popup"
-require "y2partitioner/dialogs/settings"
-require "y2partitioner/dialogs/device_graph"
+require "y2partitioner/widgets/menus/system"
+require "y2partitioner/widgets/menus/view"
+require "y2partitioner/widgets/menus/configure"
 
 module Y2Partitioner
   module Widgets
     # Main menu bar of the partitioner
     class MainMenuBar < CWM::CustomWidget
       Yast.import "UI"
-      Yast.import "Stage"
       include Yast::Logger
-      include ExecuteAndRedraw
-
-      # Class to represent each one of the entries in the 'Configure' menu
-      class ConfigureEntry
-        include Yast::I18n
-        extend Yast::I18n
-
-        # Constructor
-        def initialize(action_class_name, label, icon)
-          @action_class_name = action_class_name
-          @label = label
-          @icon = icon
-        end
-
-        # All possible entries
-        ALL = [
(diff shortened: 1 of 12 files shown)


src/lib/y2partitioner/widgets/main_menu_bar.rb
```diff
@@ -19,246 +19,117 @@
...
-      # Called by CWM after the widgets are created
-      def init
-        enable_items
+      # @see UIState#select_row
+      def select_row(id)
+        @device = find_device(id)
+        refresh
       end
 
+      # @see UIState#select_row
```
dgdavid, line 54: s/select_row/select_page

src/lib/y2partitioner/widgets/main_menu_bar.rb
```diff
@@ -19,246 +19,117 @@
...
-          Item(Id(:settings), _("Se&ttings...")),
-          Item("---"),
-          Item(Id(:abort), _("Abo&rt (Abandon Changes)")),
-          Item("---"),
-          Item(Id(:next), _("&Finish (Save and Exit)"))
-        ].freeze
-      end
-
-      def edit_menu
+      # List of buttons that make sense for the current target device
```
dgdavid, line 121: s/buttons/menus ?

dgdavid (approved):
LGTM.

Just two minor comments and also a suggestion about menu order:

`System` - `View` -  `Go` - `Add` - `Modify` - `Actions` 

I have no strong opinion about it, but right now it looks as more _natural_ order to me. But, as you said in the PR description, this is just a checkpoint. So, we can discuss the menu order later.

PR #1135, opened 2020-09-18
Remove titles and add nesting in used devices

joseivanlopez:
Part of the dirty prototype at `partitioner-ui-03`: remove titles from list pages and add nesting in used devices/physical volumes.

The "Used Devices" tab for Bcache is not using nesting because it is not clear how to do it. Right now, this tab is showing both: caching and backing device. 

Screenshot from 2020-09-18 23-16-16

Screenshot from 2020-09-19 00-03-41


ancorgs (approved):
LGTM

PR #1136, opened 2020-09-21
Fix test suite

dgdavid:
Let's test suite _to be green_ as intermediate step of the [menu-driven interface](https://trello.com/c/NITXJRBE/2055-menu-driven-interface-finish-it).

For sure, better unit tests will be added in follow up PRs.


src/lib/y2partitioner/widgets/menus/configure.rb
```diff
@@ -30,8 +30,12 @@ class ConfigureEntry
         include Yast::I18n
         extend Yast::I18n
 
+        textdomain "storage"
```
ancorgs, line 33: I don't think this call to textdomain is needed

src/lib/y2partitioner/widgets/menus/modify.rb
```diff
@@ -27,6 +27,13 @@ module Widgets
     module Menus
       # Class to represent the Modify/Edit menu
       class Modify < Device
+        # Constructor
+        def initialize(*args)
```
joseivanlopez, line 31: Having constructor only because `textdomain` is annoying. I would extend the class with `Yast::I18n` and then we can call `textdomain` inside the class:

~~~
class Modify < Device
  extend Yast::I18n
  
  textdomain "storage"
~~~

dgdavid: If it works, fine with me.

joseivanlopez (approved):
LGTM

src/lib/y2partitioner/widgets/menus/add.rb
```diff
@@ -27,6 +27,10 @@ module Widgets
     module Menus
       # Class to represent the Add menu
       class Add < Device
+        extend Yast::I18n
```
ancorgs, line 30: Why extend?

dgdavid: As already discussed offline, because the @joseivanlopez's suggestion.

PR #1140, opened 2020-09-22
Partitioner: remove tab state

joseivanlopez:

This PR modifies `UIState` avoiding it to memorize the latest tab when we go back to a previous visited page. With new UI, the first tab will be always the default one. For example, a MD RAID will have two tabs: *Overview Device* and *Used Devices*. *Overview Device* will always be the landing tab when go to the page of a specific MD. The *Used Devices* tab will only be automatically open when we are managing the used devices.


ancorgs (approved):
LGTM

PR #1142, opened 2020-09-23
Partitioner: add menu actions

joseivanlopez:

This PR makes the *Device* and *Add* menus completely functional by calling the proper action for every item of the menus.

### DIFF
--- a/src/lib/y2partitioner/widgets/menus/add.rb
+++ b/src/lib/y2partitioner/widgets/menus/add.rb
@@ -19,8 +19,12 @@
 
 require "yast"
 require "y2partitioner/widgets/menus/device"
-require "y2partitioner/actions/add_partition"
 require "y2partitioner/actions/add_md"
+require "y2partitioner/actions/add_lvm_vg"
+require "y2partitioner/actions/add_btrfs"
+require "y2partitioner/actions/add_bcache"
+require "y2partitioner/actions/add_partition"
+require "y2partitioner/actions/add_lvm_lv"
 
 module Y2Partitioner
   module Widgets
@@ -42,7 +46,7 @@ module Y2Partitioner
         # @see Base
         def items
           @items ||= [
-            Item(Id(:menu_add_raid), _("&RAID...")),
+            Item(Id(:menu_add_md), _("&RAID...")),
             Item(Id(:menu_add_vg), _("LVM &Volume Group...")),
             Item(Id(:menu_add_btrfs), _("&Btrfs...")),
             Item(Id(:menu_add_bcache), _("B&cache...")),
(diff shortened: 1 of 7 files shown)


src/lib/y2partitioner/widgets/menus/add.rb
```diff
@@ -54,26 +58,62 @@ def items
...
+          items << :menu_add_partition unless support_add_partition?
+          items << :menu_add_lv unless support_add_lv?
+          items
         end
 
         private
 
         # @see Base
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # Is this a complex method actually?
```
ancorgs, line 71: Well, it likely could be substituted by a hash and/or a little bit of metaprogramming (generating the name of the method and then using `send`).

src/lib/y2partitioner/widgets/menus/modify.rb
```diff
@@ -75,12 +81,34 @@ def disabled_for_device
         end
 
         # @see Base
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # Is this a complex method actually?
```
ancorgs, line 85: Like before, this could be probably improved with a bit of `send` or similar.

src/lib/y2partitioner/widgets/menus/add.rb
joseivanlopez: Yes, sure. In fact, I did it with metaprogramming, but the result is more difficult to read/understand. In this kind of things, I am not able to understand why a simple `case` has high complexity.

ancorgs: Well, if you tried and it looked worse, then I guess this is fine. I didn't actually tried, I was just assuming it could look good. But experience beats assumptions. :-)

test/y2partitioner/widgets/menus/add_test.rb
```diff
@@ -0,0 +1,267 @@
...
+  end
+
+  describe "#handle" do
+    shared_examples "no action" do |scenario, device_name|
+      context "and another type of device is selected" do
+        let(:scenario) { scenario }
+
+        let(:device_name) { device_name }
+
+        it "calls none action" do
```
ancorgs, line 159: NP: "none" looks weird here. I would simply say "calls no action". But it's not a big deal.

ancorgs (approved):
LGTM

src/lib/y2partitioner/widgets/menus/add.rb
jreidinger: well, in general case is usually hidden class separation. Here it is done for actions. So if all actions has simple format of class.new or class.new(device). you can use something like `MENU_ADD_MAPPING[event].new(device)` So hash based factory.

jreidinger: and btw if item is proper object, code can look like `items.find { |i| i.id == event }.action` and also above it can create list of items from object like `@items ||= [add_md.item, add_vg.item, ..]`

PR #1143, opened 2020-09-24
Partitioner: improvements

joseivanlopez:

This PR:

* Avoids some rubocop complaints by using metaprogramming.
* Adds some missing actions to the *Device->Delete* menu. 
* Uses a better wording in the Type column for the Bcache caching and backing devices.

Screenshot from 2020-09-24 13-39-30

### DIFF
--- a/src/lib/y2partitioner/widgets/menus/add.rb
+++ b/src/lib/y2partitioner/widgets/menus/add.rb
@@ -66,37 +66,36 @@ module Y2Partitioner
 
         private
 
-        # @see Base
-        # rubocop:disable Metrics/CyclomaticComplexity
-        # Is this a complex method actually?
-        def action_for(event)
-          case event
-          when :menu_add_md
-            Actions::AddMd.new
-          when :menu_add_vg
-            Actions::AddLvmVg.new
-          when :menu_add_btrfs
-            Actions::AddBtrfs.new
-          when :menu_add_bcache
-            Actions::AddBcache.new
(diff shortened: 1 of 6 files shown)


src/lib/y2partitioner/widgets/menus/add.rb
```diff
@@ -66,37 +66,36 @@ def disabled_for_device
...
-            Actions::AddBtrfs.new
-          when :menu_add_bcache
-            Actions::AddBcache.new
-          when :menu_add_partition
-            add_partition_action
-          when :menu_add_lv
-            add_lv_action
-          end
+        # @see Device#action_for
+        def menu_add_md_action
```
ancorgs, line 70: NP: maybe we could omit the "menu_" part of all the method names. It's useful in the ID of the UI to avoid collision with other UI elements, but once we are in the context of Menu class it looks unnecessary.

ancorgs: But that's a minor detail. This LGTM as it is.

ancorgs (approved):
LGTM

src/lib/y2partitioner/widgets/menus/add.rb
joseivanlopez: Yes, I did that initially, but then I added the "menu" prefix to have a direct relation between the event name and its action. It only makes the "handler" a bit more readable.

PR #1145, opened 2020-09-24
Pending UI stuff

ancorgs:

Turns out we had forgotten to delete several components that are not longer necessary and/or references to them.

On the other hand, the Device Overview was still not being used in the case of XEN partitions (ie. StrayBlkDevices).

This pull request fixes both problems.

Note the test coverage decrease is just a consequence of the reduction of the code base (all the code we are dropping was fully tested).


jreidinger (approved):
looks good

PR #1147, opened 2020-09-25
Partitioner: double click navigation

joseivanlopez:

This PR adapts the behavior of the double-click over a device in a table row.

Before this change, the double-click always navigated to the page of the selected device. For example, double-click over a disk went to the page of the disk. But now, with the new UI, this is not possible for some devices (e.g., partitions and logical volumes) because the new UI does not show a specific page for such devices.

The new behavior is:

* When the current page is a "list" page (e.g., Hard Disks, RAID, etc):
  * and the selected device has a page
    * go to the page of the selected device
  * and the selected device has no page
    * go to  page of the parent of the selected device
* When the current page is a device page:
  * show a popup with the description of the selected device


src/lib/y2partitioner/widgets/overview.rb
```diff
@@ -152,6 +152,22 @@ def validate
         valid_setup? && packages_installed?
       end
 
+      # Checks whether it is possible to jump to another page from the given page
+      #
+      # For example, when the current page is Hard Disks, we can go to the page of the selected disk.
+      #
+      # @param page [CWM::Page]
+      # @return [Boolean]
+      def allow_jump_from?(page)
```
ancorgs, line 161: I find this a bit confusing. First of all the argument `page` is a bit unnecessary, the method is always called on `current_page` which is already known by `Overview`.

Even the name/semantic of the method is a bit strange. It basically returns false unless the current page corresponds to a concrete device. Isn't it?

Why not having a method called `device_page?` or something similar, without argument.

joseivanlopez: You are absolutely right! It is much easier (and correct) to simply check whether the page has an associated device. Thanks!

ancorgs (approved):
LGTM

PR #1148, opened 2020-09-28
More robust menus

ancorgs:

## Problem

The "Add" and "Device" menus act on the device currently selected in the table. But it was failing in several situations:

- With a completely empty devicegraph, no device is ever selected and the menu bar was always empty.
- When clicking in a section with no elements (eg. clicking "RAID") there is no selected device (empty table), so the menu still referenced to the selected device in the previous table.
- When the process of adding a new device was canceled, the temporary new device (that didn't exist anymore after cancelling) was considered as the current one. Producing a crash when any action was selected in the menu.
- When managing NFS, all the menu entries were always enabled. Selecting any action that refers to a concrete device produced a crash.

## Solution

This fixes the first three problems.

To mitigate the fourth problem, now the menus simply ignore actions executed on unknown devices instead of crashing. Due to the nature of NFS management (with another client being embedded inside the Partitioner), the main problem is still not solved and all actions are still enabled.

## Testing

- Tested manually
- Updated and extended unit tests

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/btrfs_devices.rb
+++ b/src/lib/y2partitioner/actions/controllers/btrfs_devices.rb
@@ -19,7 +19,6 @@
 
 require "yast"
 require "y2storage"
-require "y2partitioner/ui_state"
 require "y2partitioner/blk_device_restorer"
 require "y2partitioner/actions/controllers/base"
 require "y2partitioner/actions/controllers/available_devices"
@@ -55,8 +54,6 @@ module Y2Partitioner
 
           @metadata_raid_level = Y2Storage::BtrfsRaidLevel::DEFAULT
           @data_raid_level = Y2Storage::BtrfsRaidLevel::DEFAULT
-
-          UIState.instance.select_row(filesystem.sid) if filesystem
         end
 
         # Metadata RAID level for the filesystem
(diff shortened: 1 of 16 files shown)


src/lib/y2partitioner/actions/controllers/btrfs_devices.rb
```diff
@@ -225,8 +222,6 @@ def create_filesystem(device)
           filesystem.metadata_raid_level = @metadata_raid_level
           filesystem.data_raid_level = @data_raid_level
 
-          UIState.instance.select_row(filesystem.sid)
```
joseivanlopez, line 228: How is the row of the new filesystem selected now? For example, when we go to the list of Btrfs and we add a new one.

joseivanlopez (commented):
In general LGTM. Only a couple of questions.

src/lib/y2partitioner/actions/controllers/btrfs_devices.rb
ancorgs: Because the latest steps of the process are not performed on this controller, but by the `Filesystem` one. See
https://github.com/yast/yast-storage-ng/blob/c28f15ad9cec7a4e011ad9de6c39e8e08efb1ac3/src/lib/y2partitioner/actions/add_btrfs.rb#L51

joseivanlopez (approved):
LGTM

PR #1150, opened 2020-10-07
Btrfs quotas

imobachgs:
## Problem

Implement support for Btrfs subvolumes quotas.

See: https://trello.com/c/d2i9pCgP/2079-8-subvolume-quotas

Finally, we decided to change the approach and implement support for reading and setting quota groups information in `libstorage-ng` itself. However, this PR is still remain open as it already defines the API.


ancorgs:
Closing this in favor of #1163, which offers a very similar API but implemented on top of libstorage-ng's support for quotas.

PR #1132, opened 2020-09-18
Nested tables

joseivanlopez:
Part of the dirty prototype at `partitioner-ui-03`: add nested tables.

This requires these other changes to work:

* https://github.com/libyui/libyui/pull/171
* https://github.com/libyui/libyui-qt/pull/132
* https://github.com/yast/yast-ycp-ui-bindings/pull/55
* https://github.com/yast/yast-yast2/pull/1099 

Screenshot from 2020-09-18 13-28-21

PR #1151, opened 2020-10-15
Nested devices

ancorgs:
This replaces #1132 

## Goal

In order to implement the interface described in [doc/partitioner_ui.md](https://github.com/yast/yast-storage-ng/blob/master/doc/partitioner_ui.md), we want to represent nesting of devices in several tables.

But we were using plain arrays to represent the rows of a table, instead of the `CWM::TableItem` class introduced at https://github.com/yast/yast-yast2/pull/1111

## Solution

Introduce a new `Y2Partitioner::Widgets::DeviceTableEntry` class to represent each entry of a table of devices, including the device itself and all the corresponding nested entries (like the partitions of a given disk). For each entry it's possible to specify if full names of devices must be enforced (useful for the tables at the "Used Devices" tab).

## Screenshots

nested1b

nested2b

nested3

## Testing

- Unit tests adapted.
- Manually tested in the installed system
- Missing: manual test during installation (can likely be postponed until we merge everything to master)

### DIFF
--- a/src/lib/y2partitioner/widgets/blk_devices_table.rb
+++ b/src/lib/y2partitioner/widgets/blk_devices_table.rb
@@ -42,12 +42,10 @@ module Y2Partitioner
       class DeviceTree
         attr_accessor :device
         attr_accessor :children
-        attr_accessor :status
 
-        def initialize(device, children: [], status: :open)
+        def initialize(device, children: [])
           @device = device
           @children = children
-          @status = status
         end
 
         def sid
@@ -98,16 +96,17 @@ module Y2Partitioner
       end
 
       def row_for(device)
-        d = device.is_a?(DeviceTree) ? device.device : device
-
-        row = [row_id(d)] + cols.map { |c| c.value_for(d) }
-
         if device.is_a?(DeviceTree)
-          row << device.children.map { |c| row_for(c) }
-          row << device.status
+          tree = device
+          device = device.device
+        else
+          tree = DeviceTree.new(device)
         end
 
-        row
+        values = cols.map { |c| c.value_for(device) }
+        children = tree.children.map { |c| row_for(c) }
+
+        item(row_id(device), values, children: children)
       end
 
       # LibYUI id to use for the row used to represent a device
--- a/src/lib/y2partitioner/widgets/blk_devices_table.rb
+++ b/src/lib/y2partitioner/widgets/blk_devices_table.rb
@@ -30,34 +30,13 @@ module Y2Partitioner
     # The subclasses must define the following methods:
     #
     #   * #columns returning an array of {Y2Partitioner::Widgets::Columns::Base}
-    #   * #devices returning a collection of {Y2Storage::BlkDevice}
+    #   * #entries returning a collection of {DeviceTableEntry}
     #
     class BlkDevicesTable < CWM::Table
       include Help
       extend Yast::I18n
 
       textdomain "storage"
-
-      # Helper class to create a row with children
-      class DeviceTree
-        attr_accessor :device
-        attr_accessor :children
-
-        def initialize(device, children: [])
-          @device = device
-          @children = children
-        end
-
-        def sid
-          device.sid
-        end
-
-        def to_list
-          list = [device] + children.map { |c| c.is_a?(DeviceTree) ? c.to_list : c }
-
-          list.flatten
-        end
-      end
 
       # @see CWM::Table#header
       def header
@@ -66,7 +45,7 @@ module Y2Partitioner
 
       # @see CWM::Table#items
       def items
-        devices.map { |d| row_for(d) }
+        entries.map { |e| e.table_item(cols) }
       end
 
       # Updates table content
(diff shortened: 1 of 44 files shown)


shundhammer:
LGTM in general so far.

Just a few minor points:

- It would be useful to point out the difference between a _device_ as used now (which is a flat list of _all_ devices belonging to a disk) and an _entry_ (which are only the ones on the current level or top level). Right now that is not obvious when reading the class documentation (or I overlooked it).

- I recall we also have that pathological case of Xen virtualization where we only have partition devices without a corresponding disk device. Does that fit into this model?

ancorgs:
> LGTM in general so far.
> 
> Just a few minor points:
> 
>     * It would be useful to point out the difference between a _device_ as used now (which is a flat list of _all_ devices belonging to a disk) and an _entry_ (which are only the ones on the current level or top level). Right now that is not obvious when reading the class documentation (or I overlooked it).

Ok. I will try to make it more clear in future versions of the doc with an example.

>     * I recall we also have that pathological case of Xen virtualization where we only have partition devices without a corresponding disk device. Does that fit into this model?

Yes, because in our model those are not really partitions. They are devices of type `StrayBlkDevice`... which is essentially a disk that cannot be partitioned.

ancorgs:
I have updated the pull request to represent logical partitions as nested within their corresponding extended one, as suggested by @shundhammer in the IRC channel. Screenshots updated according.

src/lib/y2partitioner/widgets/configurable_blk_devices_table.rb
```diff
@@ -169,8 +170,8 @@ def help
       #   partitioner
       attr_reader :pager
 
-      # @return [Array<Y2Storage::Device>] list of devices to display
-      attr_reader :devices
+      # @return [Array<DeviceTableEntry>] list of devices to display
```
joseivanlopez, line 173: NP: list of entries to display

src/lib/y2partitioner/widgets/device_table_entry.rb
```diff
@@ -0,0 +1,166 @@
...
+      # @return [Array<DeviceTableEntry>]
+      attr_reader :children
+
+      # Whether the device should be represented in the table using the full name
+      #
+      # When this is false, the device may be represented with the full name or
+      # with a short version, depending on the circumstances.
+      #
+      # @return [Boolean] true to enforce the usage of full name
+      def full_name?
```
joseivanlopez, line 47: NP: this method is intended to "force" a full name, but not to decide whether to use a short or full name. I would use a more specific name to avoid misleading, e.g., `#force_full_name?`.

src/lib/y2partitioner/widgets/pages/bcaches.rb
```diff
@@ -105,9 +105,8 @@ def table_buttons
         #
         # @return [Array<Y2Storage::BlkDevice>]
```
joseivanlopez, line 106: Please, fix return type.

src/lib/y2partitioner/widgets/overview_tab.rb
```diff
@@ -68,7 +69,7 @@ def contents
       #
       # @return [Array<Y2Storage::Device>]
```
joseivanlopez, line 70: Please, fix return type.

src/lib/y2partitioner/widgets/pages/disks.rb
```diff
@@ -51,9 +51,8 @@ def label
         #
         # @return [Array<Y2Storage::BlkDevice>]
```
joseivanlopez, line 52: Please, fix return type.

src/lib/y2partitioner/widgets/overview_tab.rb
joseivanlopez: And maybe rename it as `#entries` ?

src/lib/y2partitioner/widgets/pages/lvm.rb
```diff
@@ -68,9 +68,8 @@ def table
         #
         # @return [Array<Y2Storage::LvmVg, Y2Storage::LvmLv>]
```
joseivanlopez, line 69: Please, fix return type.

src/lib/y2partitioner/widgets/pages/md_raids.rb
```diff
@@ -66,9 +66,8 @@ def table
         # @return [Array<Y2Storage::Md>]
```
joseivanlopez, line 66: Please, fix return type.

src/lib/y2partitioner/widgets/pages/system.rb
```diff
@@ -110,9 +110,8 @@ def disk_devices
           # Since XEN virtual partitions are listed at the end of the "Hard
```
joseivanlopez, line 110: Please, fix return type.

src/lib/y2partitioner/widgets/pages/system.rb
```diff
@@ -123,37 +122,37 @@ def disk_devices
         #
         # @return [Array<Y2Storage::LvmVg, Y2Storage::LvmLv>]
```
joseivanlopez, line 123: Please, fix return type.

src/lib/y2partitioner/widgets/pages/system.rb
```diff
@@ -123,37 +122,37 @@ def disk_devices
...
         def lvm_vgs
-          device_graph.lvm_vgs.reduce([]) do |devices, vg|
-            devices << vg
-            devices.concat(vg.all_lvm_lvs)
+          device_graph.lvm_vgs.map do |vg|
+            DeviceTableEntry.new_with_children(vg)
           end
         end
 
         # @return [Array<Y2Storage::Device>]
```
joseivanlopez, line 130: Please, fix return type.

src/lib/y2partitioner/widgets/pages/system.rb
```diff
@@ -123,37 +122,37 @@ def disk_devices
...
         def software_raids
-          device_graph.software_raids.reduce([]) do |devices, raid|
-            devices << raid
-            devices.concat(raid.partitions)
+          device_graph.software_raids.map do |raid|
+            DeviceTableEntry.new_with_children(raid)
           end
         end
 
         # @return [Array<Y2Storage::Device>]
```
joseivanlopez, line 137: Please, fix return type.

src/lib/y2partitioner/widgets/pages/system.rb
```diff
@@ -123,37 +122,37 @@ def disk_devices
...
         def nfs_devices
-          device_graph.nfs_mounts
+          device_graph.nfs_mounts.map do |nfs|
+            DeviceTableEntry.new(nfs)
+          end
         end
 
         # @return [Array<Y2Storage::Device>]
```
joseivanlopez, line 144: Please, fix return type.

src/lib/y2partitioner/widgets/pages/system.rb
```diff
@@ -123,37 +122,37 @@ def disk_devices
...
-          all = device_graph.bcaches
-          all.each_with_object([]) do |bcache, devices|
-            devices << bcache
-            devices.concat(bcache.partitions)
+          device_graph.bcaches.map do |bcache|
+            DeviceTableEntry.new_with_children(bcache)
           end
         end
 
         # @return [Array<Y2Storage::Filesystems::Base>]
```
joseivanlopez, line 151: Please, fix return type.

src/lib/y2partitioner/widgets/device_table_entry.rb
ancorgs: That was my first intention, but I found it to be too long

src/lib/y2partitioner/widgets/blk_devices_table.rb
```diff
@@ -45,24 +45,32 @@ def header
...
-      def valid_sid?(device)
-        return false if device.nil?
+      # @return [DeviceTableEntry, nil]
+      def entry(device)
+        return nil if device.nil?
 
         sid = device.respond_to?(:sid) ? device.sid : device.to_i
-
-        devices.any? { |d| d.sid == sid }
+        entries.find { |entry| entry.all_entries.find { |e| e.sid == sid } }
```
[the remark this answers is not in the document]
ancorgs: You are right, that line is wrong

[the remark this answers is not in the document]
ancorgs: I fixed it (together with its unit test) via `git commit --amend`

joseivanlopez (approved):
LGTM. I guess the idea is to bump the version and add the changelog entry at the moment of merging the *partitioner-ui-02* branch, right?

PR #1153, opened 2020-10-20
L3 Veritas Volume Manager: Allow to Ignore Probe Errors

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1177332

**This is the PR against SLE-15-SP1**

## Trello

https://trello.com/c/rjFULQ97/

## Problem

Some third-party storage software like the Veritas Volume Manager use a disk label (partitioning scheme) that we don't support in libstorage-ng and yast-storage-ng. In this particular case, they use a Sun disk label for disks below 2 TB, and GPT above that size.

When starting the YaST partitioner, libstorage-ng detects the unsupported disk label and reports it as an error. That error is displayed in an error pop-up dialog, and the user can choose to continue anyway (in which case that disk is treated as if it were empty, i.e. no partitions are shown).

In this particular case, the customer had 29 disks and found the sheer number of reported errors overwhelming. They wanted to have a mechanism to speed this up; even more so since this happens every time the YaST partitioner is started.

## Solution

This PR introduces a new environment variable `$LIBSTORAGE_IGNORE_PROBE_ERRORS`. If set, all errors found during probing are suppressed and treated as if the user had chosen "Continue" in the error pop-up. No error pop-up is shown.

## Usage

```shell
xhost +
sudo LIBSTORAGE_IGNORE_PROBE_ERRORS=1 yast2 disk
```

or

```
export LIBSTORAGE_IGNORE_PROBE_ERRORS=1
sudo yast2 disk
```

# Test

## Unit Test

```
cd yast-storage-ng/src/test/lib/callbacks
rspec probe_test.rb
cd ../../..
rake test:unit
```

## Manual Test

- In the VM manager, add a new virtual disk
- Start the VM
- Log in
- Create a Sun disk label on that disk:

      sudo parted /dev/sdb mklabel sun

- Start the YaST partitioner normally; it will complain about that disk.

      xhost +
      sudo yast2 disk

- Start it with the new environment variable; it will not complain:

      sudo LIBSTORAGE_IGNORE_PROBE_ERRORS=1 yast2 disk

# Related PRs

- Original: SLE-15 SP1: This PR
- Port to SLE-15 SP2:
- Port to _master_:  https://github.com/yast/yast-storage-ng/pull/1152

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct 20 15:36:03 UTC 2020 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Added $LIBSTORAGE_IGNORE_PROBE_ERRORS environment variable
+  to ignore storage probing errors (bsc#1177332)
+- 4.1.97
+
+-------------------------------------------------------------------
 Mon Aug 17 12:37:24 UTC 2020 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Do not append a suffix to LVM Volume Group names unless it
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:		yast2-storage-ng
-Version:        4.1.96
+Version:        4.1.97
 Release:	0
 
 BuildRoot:	%{_tmppath}/%{name}-%{version}-build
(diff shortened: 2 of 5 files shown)


mvidner:
Thanks!

`LIBSTORAGE_IGNORE_PROBE_ERRORS=1` removes the popups from the UI. What about the logs? Should YaST log them, or does libstorage log them anyway?

shundhammer:
It's logged on the libstorage-ng already. You can see an exception that is thrown and then caught inside libstorage-ng.

mvidner:
BTW the correct commands for the new tests are:

```
cd test/y2storage/callbacks/
rspec probe_test.rb 
```

mvidner (approved):
As discussed on yesterday's call, this is a good approach for a quick fix / maintenance update.

(Alternatives to be considered for the master branch: aggregate the errors to a single popup, add a "Skip all messages like this one" button)

PR #1155, opened 2020-10-22
Adapt UIState to changes in the partitioner-ui-02 branch

ancorgs:
## Problem

On one hand, after all the recent changes in the the `partitioner-ui-02` branch, the correct row of each table was not always properly selected in the UI:

- After deleting a LV in the "All Devices" table or in the "LVM Volume Groups" one, the corresponding VG was not being selected.
- After creating a new multi-device Btrfs filesystem, that new device was not selected.
- After creating a single (not multi-device) Btrfs filesystem through the "Add -> Btrfs" menu option, in a table containing the block device being formatted, such device was not selected.
- When moving to the "Used Devices" tab of a device and then back to the "Overview" one without performing any chance, the selection was always changed to the first row.

On the other hand, now that all tables in the Partitioner contain nested elements, it's necessary to remember which parts of each table has being collapsed by the user and ensure they remain collapsed in subsequent redraws of the same table.

## Solution

First of all, this pull request contains fixes for all the mentioned problems.

It also includes a mechanism to store into `UIState` the state information of the current page (and tab) and uses that mechanism to ensure collapsed/expanded table sections are properly kept (the table state is stored before switching to another page or tab and before executing any action) and displayed (the initial value of `TableItem#open` for each item is set based on the stored state).

Bonus: this also adds a non-related missing "require" found during some manual testing. Apparently it was not needed in the `master` branch, but its absence was producing a crash in the `partitioner-ui-02` branch.

## Testing

- Tested manually, so far.
- Added unit tests for the new parts.

## Review

It's probably easier to review commit by commit, since each of them addresses a different issue.

### DIFF
--- a/src/lib/y2partitioner/actions/base.rb
+++ b/src/lib/y2partitioner/actions/base.rb
@@ -20,7 +20,7 @@
 require "yast"
 require "yast/i18n"
 require "yast2/popup"
-
+require "y2partitioner/ui_state"
 require "abstract_method"
 
 module Y2Partitioner
@@ -42,6 +42,7 @@ module Y2Partitioner
       def run
         return :back unless run?
 
+        UIState.instance.save_extra_info
         action_result = perform_action
 
         result(action_result)
--- a/src/lib/y2partitioner/actions/add_btrfs.rb
+++ b/src/lib/y2partitioner/actions/add_btrfs.rb
@@ -19,6 +19,7 @@
 
 require "yast"
 require "yast2/popup"
+require "y2partitioner/ui_state"
 require "y2partitioner/actions/transaction_wizard"
 require "y2partitioner/actions/controllers/filesystem"
 require "y2partitioner/actions/controllers/btrfs_devices"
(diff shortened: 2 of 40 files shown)


src/lib/y2partitioner/actions/base.rb
```diff
@@ -42,6 +42,7 @@ def initialize
       def run
         return :back unless run?
 
+        UIState.instance.save_extra_info
```
joseivanlopez, line 45: With the changes in my PR, all the buttons and menu options are using `#execute_and_redraw` to run the action:

~~~
execute_and_redraw { action.run }
~~~

Maybe we can move the `#save_extra_info` call to `#execute_and_redraw` method.

joseivanlopez: As discussed at jangouts, I will adapt it in https://github.com/yast/yast-storage-ng/pull/1156 after rebasing these changes.

joseivanlopez (approved):
LGTM

PR #1159, opened 2020-11-03
Revisit the size and format of the proposed /boot/efi

ancorgs:
## Problem

The installer was proposing a `/boot/efi` partition in which both the size and FAT size were subject to improvement.

First of all, we were proposing a partition with a size between 256MiB and 500MiB. The reason for such minimum was that in Advanced Format 4K Native drives (4-KB-per-sector), a FAT32 file-system cannot work properly on smaller partitions.

But we were not enforcing the FAT size to 32. We were delegating the decision of the FAT size to `mkfs.vfat`. Turns out that, for whatever reason, that tool only uses FAT32 for partitions equal or bigger than 512MiB... which we were never creating by default.

As a result, our `/boot/efi` partition was always FAT16. The EFI standard recommends FAT32 when possible.

On the other hand, some people claim that 256 MiB is too big as a minimum size, specially in architectures like aarch64 in which sharing the computer with MS Windows or other operating systems is less common (although not impossible).

Related bug reports:

https://bugzilla.suse.com/show_bug.cgi?id=1177358
https://bugzilla.suse.com/show_bug.cgi?id=1170625
https://bugzilla.suse.com/show_bug.cgi?id=1119318

## Solution

Adjust the limits and enforce the FAT size when possible, as explained by the auto-generated documentation:

### When proposing a new /boot/efi in aarch64 (including Raspberry Pi)

  - when aiming for the recommended  or the minimal size (it doesn't make a difference)
    - **does not enforce FAT32 or 16 for /boot/efi (FAT size will be decided by mkfs.vfat)**
    - **requires /boot/efi to be exactly 128 MiB large**

### When proposing a new /boot/efi in x86

  - when aiming for the recommended size
    - **requires /boot/efi to use FAT32**
    - **requires it to be at least 256 MiB (min size for FAT32 in drives with 4-KiB-per-sector)**
    - **requires it to be at most 512 MiB (enough space for several operating systems)**
  - when aiming for the minimal size
    - **does not enforce FAT32 or 16 for /boot/efi (FAT size will be decided by mkfs.vfat)**
    - **requires it to be at least 128 MiB (MS Windows requires 100 MiB for itself)**
    - **requires it to be at most 512 MiB (enough space for several operating systems)**

## Testing

- Updated unit tests and documentation
- Tested manually, see screenshot below.

small-efi

### DIFF
--- a/doc/boot-requirements.md
+++ b/doc/boot-requirements.md
@@ -118,37 +118,45 @@
 				- **requires the /boot/efi partition to have id ESP (according to the EFI standard)**
 		- and the first partition contains no boot code or EFI files
 			- **requires only a new /boot/efi partition**
-			- **requires /boot/efi to be at the start of an MBR partition table, bumping any existing table if needed**
+			- **requires /boot/efi to be the first partition of the device**
+			- **requires /boot/efi to be in a MBR partition table, bumping any existing table if needed**
 			- **requires the /boot/efi partition to have id DOS32 (bootcode will be installed there)**
 		- and there are no partitions in the boot disk
 			- **requires only a new /boot/efi partition**
-			- **requires /boot/efi to be at the start of an MBR partition table, bumping any existing table if needed**
+			- **requires /boot/efi to be the first partition of the device**
+			- **requires /boot/efi to be in a MBR partition table, bumping any existing table if needed**
 			- **requires the /boot/efi partition to have id DOS32 (bootcode will be installed there)**
 	- and the first partition in the boot disk is a standard EFI with id ESP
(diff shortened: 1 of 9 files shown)


shundhammer:
Is it really a good idea to enforce or not enforce FAT32 depending if we are going for recommended or minimal on x86_64? The result may seem pretty random to a user.

ancorgs:
If we go for 128MiB, we should not enforce FAT32. It's not 100% to force FAT32 in partitions < 256MiB (as already documented, even in the Windows documentation).

shundhammer (approved):
LGTM

PR #1161, opened 2020-11-04
WIP: Disable 'Add' menu actions on NFS page

shundhammer:
## Trello

https://trello.com/c/MRJhR5x6/2080-1-partitioner-disable-menu-entries-from-device-and-add-when-an-nfs-device-is-selected

## Problem

The NFS page in the partitioner is really only an embedded version of the NFS client module. It does not fit into the scheme of thing in the partitioner, in particular not how menus and menu actions work. That NFS client module/page uses its own buttons; it really only pretends to be a part of the partitioner.

## Fix

Disable menu actions that won't work on that page.

### "Device" Menu

All actions are disabled anyway since no appropriate device can be selected on this page; no matter if an NFS mount is the current device or nothing.

### "Add" Menu

All items are disabled.

## Implementation

**This is an ugly hack** since there is (so far) no way for the menus to find out what page is currently selected.

What this does is to query the tree pager widget what its current value is and check that value against the well-known item ID of the NFS page in that tree. This uses UI calls directly, completely bypassing the entire object structure in the partitioner.

I tried to find an official way, but the partitioner practices information hiding way too much: There is no connection between the menu classes and the pager or the view. We need to fix that.

### DIFF
--- a/src/lib/y2partitioner/widgets/menus/add.rb
+++ b/src/lib/y2partitioner/widgets/menus/add.rb
@@ -60,6 +60,8 @@ module Y2Partitioner
 
         # @see Device
         def disabled_for_device
+          return all_item_ids if disable_all?
+
           items = []
           items << :menu_add_partition unless support_add_partition?
           items << :menu_add_lv unless support_add_lv?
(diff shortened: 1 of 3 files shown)


shundhammer:
We decided that we can safely leave the "Add" menu alone on the NFS page in the partitioner.

Closing.

PR #1162, opened 2020-11-05
Disable "Device" Menu Items for NFS Shares

shundhammer:
## Trello

https://trello.com/c/MRJhR5x6/2080-1-partitioner-disable-menu-entries-from-device-and-add-when-an-nfs-device-is-selected

## Problem

The NFS page in the partitioner is really only an embedded version of the NFS client module. It does not fit into the scheme of thinge in the partitioner, in particular not how menus and menu actions work. That NFS client module/page uses its own buttons; it really only pretends to be a part of the partitioner.

## Fix

Disable menu actions that won't with NFS.

On the NFS page, all menu items that would cause trouble are already disabled.
That leaves the "All Devices" page where some actions in the "Device" menu were still enabled:

- Edit
- Show Details
- Delete

This PR disables them.

The user has to go to the NFS page to perform those actions via buttons in the embedded NFS client module.

## Show Details for NFS Shares?

While it might be possible to support at least "Show Details" even for NFS shares, AFAICS everything that we know about an NFS share is already displayed in the table; we don't seem to have any additional information that would warrant a separate pop-up for an NFS share.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Nov  5 10:26:55 UTC 2020 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Disable "Device" menu items for NFS shares
+  (related to jsc#PM-73, jsc#SLE-7742, jsc#SLE-15283)
+- 4.3.19
+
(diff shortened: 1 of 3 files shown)


ancorgs (approved):
LGTM

ancorgs (approved):

PR #1163, opened 2020-11-06
Initial wrapper for BtrfsQgroup and associated methods

ancorgs:
## Problem

Support for Btrfs qgroups was recently added to libstorage-ng, but it's not yet exposed in yast2-storage-ng.

## Solution

Create the wrappers and all the methods needed in order to offer an API similar to the one at #1150.

When finished, this pull request will obsolete #1150 

## Testing

- Tested manually.
- Unit tests included


src/lib/y2storage/btrfs_qgroup.rb
```diff
@@ -0,0 +1,115 @@
...
+    #
+    # @return [Array<Integer>] includes two terms, level and id
+    def id
+      [storage_id.first, storage_id.second]
+    end
+
+    # @!method referenced
+    #   Size of the referenced space of the qgroup
+    #
+    #   @return [DiskSize] zero if the size is not known (the qgroup was created after probing)
```
shundhammer, line 52: This means that you can't tell if the qgroup ran out of disk space or if it's just unknown. In the general context of quotas this sounds like an accident waiting to happen. Wouldn't it be better to return `nil` if it's unknown? or -1?

src/lib/y2storage/btrfs_qgroup.rb
```diff
@@ -0,0 +1,115 @@
...
+module Y2Storage
+  # A qgroup of a Btrfs filesystem
+  #
+  # This is a wrapper for Storage::BtrfsQgroup.
+  #
+  # In Btrfs, the qgroups hold the quota information for a given set of subvolumes.
+  # Although they can be organized in a tree hierarchy, we are only considering the
+  # first level (level 0, actually), so they are associated to a single subvolume.
+  #
+  # @see man btrfs-qgroup
```
shundhammer, line 32: Please add `@see man btrfs-quota` which contains some more high-level documentation to make sense of what this is all about.

I looked at `man btrfs-qgroup`, and it's a particular useless piece of documentation, written for those who already know everything (like so much of the Btrfs documentation).

src/lib/y2storage/btrfs_qgroup.rb
ancorgs: In fact, it means you can't distinguish between this two cases:

- the subvolume exists for real (ie. is probed) but is empty
- the subvolume is still not created in the system (so the space it will use is unknown yet)

First of all, that's consistent with what libstorage-ng returns. On the other hand, I don't think it's specially dangerous.

shundhammer (approved):
Just the two small things, otherwise LGTM.

src/lib/y2storage/btrfs_qgroup.rb
shundhammer: OK, if libstorage-ng doesn't make the difference, we can't do much here anyway.

src/lib/y2storage/btrfs_qgroup.rb
shundhammer: :+1:

shundhammer (approved):
:+1:

src/lib/y2storage/btrfs_subvolume.rb
```diff
@@ -109,6 +126,67 @@ def can_be_auto_deleted=(value)
...
+    # The information is obtained from the qgroup of level 0 associated to the subvolume,
+    # which implies it can only be known if quotas are enabled for the filesystem.
+    # If quota support is disabled, this method returns nil.
+    #
+    # If the filesystem already existed during probing but got no quota support enabled at
+    # that moment, this method returns zero if quota support was enabled after probing.
+    #
+    # @return [DiskSize, nil] nil if quota support is disabled, zero if it was enabled only
+    #   after probing, the known size in any other case
+    def referenced
```
joseivanlopez, line 140: AFAIS, when there is no quota or no qgroup:

~~~
#referenced #=> nil
#exclusive #=> nil
#referenced_limit #=> unlimited
#exclusive_limit #=> unlimited
~~~

I would expect to either return a size or nil in all these methods. Maybe it does not make sense to use those values when there is no quota, so maybe these methods should return nil in that case. What do you think?

src/lib/y2storage/filesystems/btrfs.rb
```diff
@@ -127,6 +127,35 @@ class Btrfs < BlkFilesystem
...
+      # @!method btrfs_qgroups
+      #   Collection of Btrfs qgroups of the filesystem
+      #   @return [Array<BtrfsQgroup>]
+      storage_forward :btrfs_qgroups, as: "BtrfsQgroup"
+
+      # @!method has_quota
+      #   Whether quota support is enabled for this btrfs filesystem
+      #
+      #   @return [Boolean]
+      storage_forward :has_quota
```
joseivanlopez, line 139: What about making `#has_quota` private?

joseivanlopez (approved):
LGTM. Only a NP about returning nil.

src/lib/y2storage/btrfs_subvolume.rb
ancorgs: If I'm querying the `#referenced_limit`, I want to know what the max size is. And I want a consistent response to that in all situations (ie. `unlimited`if there is no limit). I didn't ask about the reason. **So far**, there are two situations in which the size is unlimited: quotas are disabled or quotas are enabled but a limit was not set. When asking about the limit, I don't care in which situation I am. I only want to know what the limit is (`unlimited` if none).

If I want to know the status of quotas there are other methods for that.

The case of `referenced` is different. It returns `nil` simply because there is no information. Not to indicate that quotas are disabled.

src/lib/y2storage/filesystems/btrfs.rb
ancorgs: `#has_quota` is a pretty stable method, since it comes from libstorage-ng. I don't see any problem in offering it (it's maybe convenient for people used to the libstorage-ng API). Its only "problem" is that the name is not as rubist as the alias, but I'm not sure whether it deserves to be hidden just for that. :wink:

PR #1160, opened 2020-11-04
Improve Btrfs subvols management

joseivanlopez:
## Problem

Btrfs subvolumes are not well integrated in the Expert Partitioner UI. To see the subvolumes of a Btrfs you have to edit the filesystem and use the *Handle Subvolumes* button. It is not easy to simple check the list of subvolumes without editing the filesystem. Moreover, the dialog to create/edit a subvolume is open as a popup, instead of a wizard like in the rest of devices.

current-subvols

This PR is about improving the integration of the Btrfs subvolumes in the Expert Partitioner. The idea is to convert Btrfs subvolumes to first class citizen, similar to partitions or LVM LVs.

* PBI: https://trello.com/c/OQGxR8Eo/2077-5-partitioner-more-integrated-subvolumes-management.

Related bugs:

* https://bugzilla.suse.com/show_bug.cgi?id=996598
* https://bugzilla.suse.com/show_bug.cgi?id=1175139
* https://bugzilla.suse.com/show_bug.cgi?id=1161408

## Solution

Btrfs subvolumes are now presented in a similar way to other devices like partitions or LVM logical volumes:

* In the tables, their are nested to its filesystem (or to the block device when the filesystem is single-device).
* There is a new single-step wizard to create/edit a subvolume.
* Buttons and menus are adapted to also work with a Btrfs subvolume.
* A description for the subvolume is shown when using "Device->Show Details" menu.

Because sometimes there could be too many subvolumes (e.g., when snapper is used), a table entry is automatically closed by default when it has more than 10 children. This affects to all entries, for example a disk with many partitions, no only to subvolumes. 

Bonus: some refactoring was done to buttons and actions.

## Testing

- Added new unit tests.
- Tested manually with the *partitioner_testing* client.

## Screenshots

<details><summary>Show/Hide</summary>

Screenshot from 2020-11-05 12-37-34

Screenshot from 2020-11-05 12-38-34

Screenshot from 2020-11-05 12-39-08

Screenshot from 2020-11-05 12-39-44

Screenshot from 2020-11-05 12-40-43

Screenshot from 2020-11-05 12-44-27
</details>

PR #1164, opened 2020-11-10
Improve automatically added subvolumes

joseivanlopez:
## Problem

This is a follow-up of https://github.com/yast/yast-storage-ng/pull/1160.

The control file can define some default subvolumes for a specific file system. This is typically used to define the subvolumes of the root file system, but it is not limited to root. When you use the Expert Partitioner to create or edit a Btrfs, such list of default subvolumes is automatically added to the file system. But this is done without noticing to the user. This could lead to very unexpected results. For example, let's say you have created a Btrfs, then you add the subvolumes you want, and then you edit the Btrfs again to set a mount point. This last action could ruin all the work you did with with your subvolumes, because the current subvolumes will be automatically replaced by the default list.  

* PBI: https://trello.com/c/Atgk0DCx/2155-5-automatically-added-btrfs-subvolumes
* Also fixes this: https://bugzilla.suse.com/show_bug.cgi?id=1004783

## Solution

Now, the user is asked before replacing the current subvolumes with the default list. Moreover, the default list of subvolumes is only applied when the Btrfs does not exist on disk yet.

This is the new behavior:

~~~
* When adding or editing a new Btrfs
  * When the Btrfs is set to be mounted
    * And the Btrfs has no subvolumes yet
      * Adds the default subvolumes
    * And the Btrfs has subvolumes
      * And there is default subvolumes for the selected mount location
        * Ask to replace with the default subvolumes.
      * And there is no default subvolumes for the new mount location
        * Ask to remove the current subvolumes
  * When the Btrfs is set to not be mounted
    * And the Btrfs has subvolumes
      * Ask to remove the current subvolumes

* When editing an existing Btrfs
  * Do not ask to the user
  * Do not add subvolumes
~~~

## Testing

* Added a new unit tests
* Tested manually by using the *partitioner_testing* client

## Screenshots

<details><summary>Show/Hide</summary>

Screenshot from 2020-11-18 15-34-07

Screenshot from 2020-11-18 15-34-53

Screenshot from 2020-11-18 15-36-06

</details>


src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -249,12 +248,62 @@ def handle(event)
...
+        return unless controller.new_btrfs? && controller.mount_path_modified?
+
+        controller.restore_btrfs_subvolumes = true
+
+        # It does not have subvolumes, so let's simply restore the default list.
+        return unless controller.btrfs_subvolumes?
+
+        if controller.default_btrfs_subvolumes?
+          # TRANSLATORS: message to ask to the user about restoring the default list of subvolumes.
+          message = _("Do you want to restore the default list of subvolumes?")
```
ancorgs, line 285: I find this message to be too succinct. Shouldn't we explain the situation a bit better?

Something in the line of:

```
You have chosen to mount the device at #{mount_point}.

Do you want to remove the current subvolumes and
create the suggested subvolumes for #{mount_point}?
```

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -249,12 +248,62 @@ def handle(event)
...
+      def btrfs_subvolumes_question
+        return unless controller.new_btrfs? && controller.mount_path_modified?
+
+        controller.restore_btrfs_subvolumes = true
+
+        # It does not have subvolumes, so let's simply restore the default list.
+        return unless controller.btrfs_subvolumes?
+
+        if controller.default_btrfs_subvolumes?
+          # TRANSLATORS: message to ask to the user about restoring the default list of subvolumes.
```
ancorgs, line 284: I find this kind of "clarifications" for the translators to be almost annoying. :wink: It adds no information when compared to the string they already have to translate.

joseivanlopez: Totally agree.

src/lib/y2partitioner/widgets/format_and_mount.rb
```diff
@@ -249,12 +248,62 @@ def handle(event)
...
+          message = _("Do you want to restore the default list of subvolumes?")
+        else
+          # TRANSLATORS: message to ask to the user about deleting the current subvolumes.
+          message = _("Do you want to delete the current subvolumes?")
```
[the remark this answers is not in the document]
joseivanlopez: Ok, but take into account that this message is also shown when the mount path is modified, not only when it was unmounted. So, maybe something like this?

~~~
You have changed the mount point of the device.

Do you want to delete the current subvolumes?
~~~

doc/btrfs-subvolumes.md
```diff
@@ -0,0 +1,183 @@
...
+    </volume>
+  </volumes>
+</partitioning>
+~~~
+
+This configuration of subvolumes is taken into accout by the Storage Proposal when calculating the partitioning proposal. But it also affects to the Expert Partitioner, see *Expert Partitioner* section.
+
+Note that there is a `btrfs_default_subvolume` option. This option allows to configure a kind of "prefix" subvolume, and it has some implications over the rest of subvolumes:
+
+* All the subvolumes are created as children of the default one.
```
ancorgs, line 42: We should clarify that, in this context, when we say "default subvolume"  we are referring to that special prefix subvolume... which may end up being the real default subvolume or not (is very unfortunate that somebody decided to misuse a term that already existed in Btrfs).

src/lib/y2partitioner/widgets/columns.rb
```diff
@@ -29,7 +29,6 @@ module Columns
 require "y2partitioner/widgets/columns/chunk_size"
 require "y2partitioner/widgets/columns/device"
 require "y2partitioner/widgets/columns/encrypted"
-require "y2partitioner/widgets/columns/filesystem_id"
```
ancorgs, line 32: I think `filesystem_devices` must also be removed. It's pretty much the same case than `filesystem_id`.

src/lib/y2storage/btrfs_subvolume.rb
```diff
@@ -187,69 +207,57 @@ def exclusive_limit=(limit)
...
     #
     # @return [Boolean]
-    def self.shadowed?(devicegraph, mount_path)
-      !shadowers(devicegraph, mount_path).empty?
+    def snapshot?
+      path.match?(/^.snapshots/)
     end
 
-    # Returns the current shadowers for a specific mount point
+    # Whether the subvolume is the used as prefix (typically @)
```
ancorgs, line 243: Maybe we should mention that this is only trustworthy if we have volume specifications (ie. during installation).

src/lib/y2storage/shadower.rb
```diff
@@ -0,0 +1,261 @@
...
+
+require "y2storage/subvol_specification"
+require "y2storage/mountable"
+
+module Y2Storage
+  # Class to manage the shadowing of Btrfs subvolumes
+  #
+  # This class is used to automatically shadow Btrfs Subvolumes. In some cases, Btrfs subvolumes are
+  # automatically added when creating a Btrfs. Those subvolumes should also be automatically hidden when
+  # other device shadows them. Moreover, subvolumes added by the user should be automatically mounted or
```
ancorgs, line 28: subvolumes added by the user and also subvolumes that come from probing. Isn't it?

ancorgs (changes requested):
In general, looks pretty ok-ish. I just added some comments.

src/lib/y2storage/btrfs_subvolume.rb
joseivanlopez: Actually not. The control file is also accessible in the running system.

src/lib/y2storage/shadower.rb
joseivanlopez: Yes, right.

src/lib/y2storage/btrfs_subvolume.rb
ancorgs: But is trustworthy in that case? Does it really represent the product/role that was installed and all that?

ancorgs (approved):
LGTM

src/lib/y2storage/btrfs_subvolume.rb
joseivanlopez: I am not sure, but I think that the control file stored in the installed system represents the final version. Anyway, as we have commented on IRC, in case of AutoYaST we should take into account the AutoYaST profile to decide whether a subvolume is the prefix. I will improve that as part of https://github.com/yast/yast-storage-ng/pull/1168.

PR #1166, opened 2020-11-17
Partial fix for bug#176140

ancorgs:
## Problem

The logic for adding _netdev needs to be refined... once again!

Now we have been told that `_netdev` should be skipped not only for "/" and all its subvolumes, but also for all the mount points that are based on the same disk than "/".

See https://bugzilla.suse.com/show_bug.cgi?id=1176140

## Partial Solution

This expands the logic of `#set_default_mount_options` to honor the new proposed (and approved) logic described here:
https://gist.github.com/ancorgs/435328f6f16920a17b0b8cca7500526e#proposed-rspec-output-for-y2storagemountpoint

Unfortunately, this is only a solution if the root mount point is the first to be created. For example, imagine we create a mount point for "/var" in sda2. In that moment, `#set_default_mount_options` will add _netdev to it since there is no evidence of that partition being in the root disk. If at a later point, sda3 is created and mounted as "/", the mount options of sda2 will not be re-adjusted (so far, we have never needed to adjust the options of one mount point based on the presence of another one, so we have no such mechanism).

## Testing

- Added a new unit test

## Do not merge

We need a definitive solution.

### DIFF
--- a/src/lib/y2storage/blk_device.rb
+++ b/src/lib/y2storage/blk_device.rb
@@ -644,6 +644,13 @@ module Y2Storage
       false
     end
 
+    # Whether this device contains, directly or indirectly, the root mount point
+    #
+    # @return [Boolean]
+    def contain_root?
+      ([self] + descendants).any? { |desc| desc.is_a?(MountPoint) && desc.root? }
+    end
+
     # Whether the block device fulfills conditions to be used for a Windows system
     #
     # Note that only partitions can be used for installing Windows.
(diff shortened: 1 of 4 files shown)


dgdavid:
> We need a definitive solution.

@ancorgs which is the status of this PR? Should we keep it open?

ancorgs:
I guess this can be closed until we decide what the best mid-term solution is. I'm trying to re-activate the discussion here https://jira.suse.com/browse/PM-2830

PR #1167, opened 2020-11-18
Improved Error Handling for Mounting efivars

shundhammer:
## Trello

https://trello.com/c/fhCSVHzi/2149-2-tw-1174029-yast2-depends-on-efi-vars

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1174029

## Problem

When we are done creating parititons and filesystems during installation, we mount system mounts such as `/dev`, `/proc`, `/sys`, `/run` to the newly created target system so it can later be used correctly in a `chroot` environment.

On EFI systems, that also includes the EFI variables in `/sys/firmware/efi/efivars` (a separate filesystem / kernel table).

On some platforms (arm64) that is only half-implemented, however: They do use EFI, but there is no support (yet) for the special _efivarfs_ filesystem that is used to read and write EFI variables directly via that system mount.

So attempting to mount that  `/sys/firmware/efi/efivars` filesystem fails, an exception is (intentionally!) thrown, and with that the whole installation fails.

## Fix

- Check not only if `/sys/firmware/efi/efivars` exists (i.e. if the current system uses EFI), but also if the _efivarfs_ filesystem type is supported by the current kernel, i.e. if that filesystem type is in the list of supported filesystems in `/proc/filesystems`.

- More graceful error handling: If any of those mounts fails, don't just throw a hard exception (so the installation fails), but report the problem in a warning pop-up.

### DIFF
--- a/src/lib/y2storage/clients/inst_prepdisk.rb
+++ b/src/lib/y2storage/clients/inst_prepdisk.rb
@@ -90,7 +90,7 @@ module Y2Storage
 
         if !SCR.Execute(path(".target.mount"), [device, target_path], options)
           # TRANSLATORS: %s is the path of a system mount like "/dev", "/proc", "/sys"
-          Yast::Report.Warning(_("Could not mount %s") % path )
+          Yast::Report.Warning(_("Could not mount %s") % path)
         end
 
         nil
@@ -113,12 +113,10 @@ module Y2Storage
       #
       # @return [Boolean] true if efivarfs is supported
       def efivarfs_support?
-        begin
-          File.readlines("/proc/filesystems").any? { |line| line =~ /efivarfs/ }
-        rescue Errno::ENOENT => err
-          log.error("Can't check efivarfss support: #{err}")
-          false
-        end
+        File.readlines("/proc/filesystems").any? { |line| line =~ /efivarfs/ }
+      rescue Errno::ENOENT => e
+        log.error("Can't check efivarfss support: #{e}")
+        false
       end
 
       def manager
--- a/src/lib/y2storage/clients/inst_prepdisk.rb
+++ b/src/lib/y2storage/clients/inst_prepdisk.rb
@@ -106,7 +106,12 @@ module Y2Storage
       #
       # @return [Boolean] true if efivarfs is supported
       def efivarfs_support?
-        File.readlines("/proc/filesystems").any? { |line| line =~ /efivarfs/ }
+        begin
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/clients/inst_prepdisk.rb
```diff
@@ -83,12 +89,38 @@ def mount_in_target(path, device, options)
...
+      # Check if the efivarfs filesystem type is supported on this system,
+      # i.e. if /proc/filesystems contains a line with "efivarfs".
+      #
+      # Notice that a system might have the /sys/firmware/efi/efivars file,
+      # but no support for the efivarfs filesystem to actually mount it.
+      # See https://bugzilla.suse.com/show_bug.cgi?id=1174029
+      #
+      # @return [Boolean] true if efivarfs is supported
+      def efivarfs_support?
+        begin
```
ancorgs, line 116: Rubocop is going to complain because this `begin` is not needed

src/lib/y2storage/clients/inst_prepdisk.rb
```diff
@@ -83,12 +89,38 @@ def mount_in_target(path, device, options)
...
+          File.readlines("/proc/filesystems").any? { |line| line =~ /efivarfs/ }
+        rescue Errno::ENOENT => err
```
ancorgs, line 118: Rubocop is (stupidly, IMO) going to complain because it wants this to be called `e`.

src/lib/y2storage/clients/inst_prepdisk.rb
```diff
@@ -83,12 +89,38 @@ def mount_in_target(path, device, options)
         log.info "Cmd: mount #{options} #{device} #{target_path}"
 
         if !SCR.Execute(path(".target.mount"), [device, target_path], options)
-          raise ".target.mount failed"
+          # TRANSLATORS: %s is the path of a system mount like "/dev", "/proc", "/sys"
+          Yast::Report.Warning(_("Could not mount %s") % path )
```
ancorgs, line 93: Rubocop is going to complain about the extra space after `path`.

ancorgs (changes requested):
Looks good in general, but please fix the rubocop issues.

src/lib/y2storage/clients/inst_prepdisk.rb
shundhammer: OK; fixed.

ancorgs (approved):
LGTM

PR #1168, opened 2020-11-18
Fix subvolumes prefix

joseivanlopez:
## Problem

This is a follow-up of https://github.com/yast/yast-storage-ng/pull/1160.

In SLE and openSUSE, a "@" Btrfs subvolume is used as kind of prefix for the rest of subvolumes in the root file system. All subvolumes proposed in the control file should be created as children of "@". Moreover, the path of all the subvolumes should start with "@/", for example:

~~~
top level
|-- @
  |-- @/home
  |-- @/var
  |-- @/.snapshots
      |-- @/.snapshots/1/snapshot
      |-- @/.snapshots/2/snapshot
~~~

This was designed so to make the snapshots rollback easier, see https://www.spinics.net/lists/linux-btrfs/msg44611.html (although having a prefix is not necessary at all. What really eases to rollback snapshots is the fact of having different subvolumes for root, home, etc).

The Expert Partitioner should be able to recognize a subvolume prefix to create the new Btrfs subvolumes accordingly. But there are some scenarios where the Partitioner is not working as expected:

* When the Btrfs exists on disk and has no subvolumes: it considers the first added subvolume as the prefix.
* When the Btrfs exists on disk and it is using snapshots: it does not recognize "@" as prefix.

Related links: 

* PBI: https://trello.com/c/Ei0AyUtc/2152-5-the-subvolume-prefix
* Bug: https://bugzilla.suse.com/show_bug.cgi?id=1067505
* See also: https://github.com/yast/yast-storage-ng/pull/1164

## Solution

The detection of the subvolumes prefix was improved. Now the prefix is inferred from the subvolumes hierarchy of the file system and the *btrfs_default_subvolume* indicated in the control file. So, given a subvolumes tree, the file system is supposed to be using a prefix when:

* The top level subvolume only has a child.
* The path of such subvolume is "@" or whatever the control file indicates in the *btrfs_default_subvolume* property for the current file system.
* All the children of such subvolume starts with its path.

This logic to infer the prefix is only applied when the subvolumes prefix has not been set yet for the file system. The prefix is directly set in some cases:

* When AutoYaST is creating the subvolumes for a file system (it sets the *subvolumes_prefix* option from the profile).
* When the Storage Proposal is creating the subvolumes for a file system (it sets the *btrfs_default_subvolume* option from the control file).
* When the Expert Partitioner is creating the default list of subvolumes for a file system (it also sets the *btrfs_default_subvolume* option from the control file).

When the prefix is neither directly set nor inferred, the new subvolumes will not use a prefix. 

Moreover, the shadowing process was improved. Now, the mount point is removed from user created subvolumes when the subvolume is shadowed.

And last but not least, now the tables do not show the prefix subvolume. Previously, the tables were hiding the default subvolume instead of the prefix. Note that in an running system with snapshots, @ subvolume is not the default subvolume. In that case, a snapshot is the default subvolume. 

## Testing

* Added unit tests
* Tested manually with the *partitioner_testing* client
* Tested during installation and in a running system

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/btrfs_subvolume.rb
+++ b/src/lib/y2partitioner/actions/controllers/btrfs_subvolume.rb
@@ -126,15 +126,11 @@ module Y2Partitioner
           !new?(subvolume)
         end
 
-        # Whether the filesystem already has a subvolume on disk with the given path
+        # Whether the filesystem already has a subvolume with the given path
         #
         # @return [Boolean]
         def exist_path?(path)
-          subvolume = filesystem.btrfs_subvolumes.find { |s| s.path == path }
-
-          return false unless subvolume
-
-          !new?(subvolume)
+          filesystem.btrfs_subvolumes.any? { |s| s.path == path }
         end
 
         # Whether quota support is enabled for the Btrfs filesystem
--- a/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
+++ b/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
@@ -181,8 +181,7 @@ module Y2Partitioner
         def uniqueness_error
           return nil unless controller.exist_path?(value)
 
-          # TRANSLATORS: error message, where %s is replaced by a path given by the user.
-          format(_("Subvolume name %s already exists."), value)
+          format(_("A subvolume already exists with this path."), value)
         end
 
         # Error when the given path is part of an already existing path
--- a/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
+++ b/src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
@@ -181,7 +181,7 @@ module Y2Partitioner
         def uniqueness_error
           return nil unless controller.exist_path?(value)
 
-          format(_("A subvolume already exists with this path."), value)
+          format(_("There is already a subvolume with that path."), value)
         end
 
         # Error when the given path is part of an already existing path
--- a/src/lib/y2storage/btrfs_subvolume.rb
+++ b/src/lib/y2storage/btrfs_subvolume.rb
@@ -219,6 +219,13 @@ module Y2Storage
       userdata_value(:former_referenced_limit)
     end
 
+    # Whether the subvolume is used as prefix (typically @)
+    #
+    # @return [Boolean]
+    def prefix?
+      path == filesystem.subvolumes_prefix
+    end
+
     protected
 
     # Whether the subvolume requires a default mount point
@@ -250,13 +257,6 @@ module Y2Storage
     # @return [Boolean]
     def for_snapshots?
       path.match?(/.snapshots/)
-    end
-
-    # Whether the subvolume is the used as prefix (typically @)
-    #
-    # @return [Boolean]
-    def prefix?
-      path == filesystem.subvolumes_prefix
     end
 
     # Parent subvolume
--- a/src/lib/y2storage/btrfs_subvolume.rb
+++ b/src/lib/y2storage/btrfs_subvolume.rb
@@ -20,7 +20,6 @@
 require "y2storage/storage_class_wrapper"
 require "y2storage/btrfs_qgroup"
 require "y2storage/mountable"
-require "y2storage/volume_specification"
 
 module Y2Storage
   # A subvolume in a Btrfs filesystem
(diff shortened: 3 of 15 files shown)


src/lib/y2partitioner/dialogs/btrfs_subvolume.rb
```diff
@@ -181,8 +181,7 @@ def content_error
         def uniqueness_error
           return nil unless controller.exist_path?(value)
 
-          # TRANSLATORS: error message, where %s is replaced by a path given by the user.
-          format(_("Subvolume name %s already exists."), value)
+          format(_("A subvolume already exists with this path."), value)
```
ancorgs, line 184: The sentence is correct but sounds a bit weird. It sounds better ordered like this. "A subvolume with that path already exists". But I would go for something more natural like "There is already a subvolume with that path".

src/lib/y2partitioner/actions/controllers/btrfs_subvolume.rb
```diff
@@ -126,15 +126,11 @@ def exist_subvolume?
...
+        # Whether the filesystem already has a subvolume with the given path
         #
         # @return [Boolean]
         def exist_path?(path)
-          subvolume = filesystem.btrfs_subvolumes.find { |s| s.path == path }
-
-          return false unless subvolume
-
-          !new?(subvolume)
+          filesystem.btrfs_subvolumes.any? { |s| s.path == path }
```
ancorgs, line 133: Do you know why in the past we only checked for subvolumes that already existed in disk?

src/lib/y2storage/btrfs_subvolume.rb
```diff
@@ -241,29 +240,23 @@ def former_referenced_limit
...
     end
 
-    # Whether the subvolume is for snapshots
+    # Whether the subvolume is used for snapshots
     #
     # @return [Boolean]
-    def snapshot?
-      path.match?(/^.snapshots/)
+    def for_snapshots?
+      path.match?(/.snapshots/)
```
ancorgs, line 259: Should this rely on `Filesystems::Btrfs#snapshots_root` instead of a hadcoded ".snapshots"?

ancorgs (commented):
Marking it as "partially reviewed" to release the comments. :wink:

src/lib/y2partitioner/actions/controllers/btrfs_subvolume.rb
joseivanlopez: This is used for the validation of the widget that allows to enter the subvolume path. For some reason, we decided that creating a subvolume with an existing path should be possible (if the subvolume does not exist on disk). Maybe the old partitioner allowed that? Anyway, creating a subvolume was only a matter of selecting its path and nocow property. But now, the Btrfs subvolume form has more options. So, silently replacing the subvolume with the same path is not a good idea.

src/lib/y2storage/btrfs_subvolume.rb
joseivanlopez: Yes, in fact, there was a corner case with this hypothetical scenario:

~~~
# top-level                       
# |-- @                           
#    |-- @/.snapshots               
#        |-- @/.snapshots/1/snapshot 
# |-- .snapshots

subvol1.path            #=> @/.snapshots/1/snapshot
subvol1.for_snapshots?  #=> true

subvol2.path            # => .snapshots
subvol2.for_snapshots?  # => true (but it should be false)
~~~

src/lib/y2storage/btrfs_subvolume.rb
```diff
@@ -241,29 +247,16 @@ def former_referenced_limit
...
-    def prefix?
-      return false unless filesystem.root?
-
-      spec = Y2Storage::VolumeSpecification.for("/")
-
-      return false unless spec&.btrfs_default_subvolume
-
-      spec.btrfs_default_subvolume == path
+    def for_snapshots?
+      path.start_with?(filesystem.snapshots_root)
```
ancorgs, line 259: This would return true for `@/.snapshotsisms`, which is wrong.

ancorgs (changes requested):
I think I found an error in the logic of `#for_snapshots?`

src/lib/y2storage/btrfs_subvolume.rb
joseivanlopez: Oh, right!

ancorgs (approved):
LGTM now

PR #1169, opened 2020-11-19
Fix help

jreidinger:
## Problem

double tab on CLI for partitioner opens a module.

- https://bugzilla.suse.com/show_bug.cgi?id=1172340
- https://trello.com/c/8G4FikEf/1988-ostumbleweed-p5-1172340-several-yast-modules-can-be-started-by-typing-yast2-module-tabtab

## Solution

Print properly that CLI is not supported. Also add built in help for partitioner_testing client.

## Testing

- *Tested manually*

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Nov 19 09:10:13 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- print properly help to avoid accidental opening of module by bash
+  completion (bsc#1172340) 
+- 4.3.21
+
+-------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


ancorgs:
I find the proposed solution to be a bit messy:

1) If `yast partitioner` is called, then arguments are processed in `lib/clients/partitioner`. Not in `clients/partitioner`.
2) But if `yast partitioner_testing` is called, then arguments are processed in `clients/partitioner_testing` (different approach)... except one of the possible arguments that is intentionally bypassed (and Rubocop rightfully complains) to the arguments processing from `lib/clients/partitioner`.

BTW, as a consequence of (2), `lib/clients/partitioner` becomes aware of the existence of  `partitioner_testing` and becomes aware of its syntax (to print the help)... although that syntax is actually processed elsewhere.

We can say the dual client was already a bit messy to start with... but this pull request makes that much worse.

jreidinger:
replaced by https://github.com/yast/yast-storage-ng/pull/1170

PR #1170, opened 2020-11-20
Help

jreidinger:
## Problem

double tab on CLI for partitioner opens a module.

- https://bugzilla.suse.com/show_bug.cgi?id=1172340
- https://trello.com/c/8G4FikEf/1988-ostumbleweed-p5-1172340-several-yast-modules-can-be-started-by-typing-yast2-module-tabtab

## Solution

Print properly that CLI is not supported

## Testing

- *Tested manually*

## Screenshots

*If the fix affects the UI attach some screenshots here.*


src/lib/y2partitioner/cli.rb
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) 2014 SUSE LLC.
```
ancorgs, line 1: 2014?

src/lib/y2partitioner/cli.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module Y2Partitioner
+  # CLI support for storage-ng. It just say that there is no one.
+  class CLI
+    extend Yast::I18n
+
+    def self.run
+      cmdline_description = {
+        "id"   => "partitioner"
+      }
```
ancorgs, line 29: Does this actually print something?

jreidinger: yes, it prints all that is needed:
```
localhost:/home/jreidinger # yast2 partitioner help
YaST2 partitioner
-----------------

This YaST2 module does not support the command line interface.
```

for others I use help to suggest CLI tool, but for all partitioner functionality it is hard.

ancorgs (approved):
LGTM

ancorgs (approved):
Still LGTM after rubocop changes

PR #1171, opened 2020-11-24
Partitioner: fix a crash with multi-device filesystems (bsc#1179100)

ancorgs:
## Problem

The Partitioner used to crash when "Do not format device" was selected for a block device that used to be part of a multi-device filesystem. See details at [bsc#1179100](https://bugzilla.suse.com/show_bug.cgi?id=1179100).

## Solution

Since restoring the previous file system never makes sense in case of multi-device file systems (see details in comment#1 of the bug report), we do not try.

## Testing

- Tested manually
- Added new regression unit test and also fixed the unit test for bsc#1142669 (which was very artificial and didn't actually explained the situation).


joseivanlopez (approved):
LGTM

PR #1172, opened 2020-11-25
Merge btrfs_subvols

joseivanlopez:
NOTE: *btrfs_subvols* branch should be removed after merging this.

This PR merges *btrfs_subvols* branch into master. All the code included here was already reviewed. Only the version bump and the chagelog entry is new (last commit).

Here it is included:

* Improved support for Btrfs subvolumes
  * https://github.com/yast/yast-storage-ng/pull/1160
  * https://github.com/yast/yast-storage-ng/pull/1164
  * https://github.com/yast/yast-storage-ng/pull/1168

* Added support for Btrfs quotas in the Partitioner
  * https://github.com/yast/yast-storage-ng/pull/1163
  * https://github.com/yast/yast-storage-ng/pull/1165

PR #1173, opened 2020-11-25
Remove leftover require

ancorgs:
## Problem

Jenkins at #1172 failed because we are, indeed, requiring a file that is not longer there.

## Solution

Remove the leftover call to `require`.

## Bonus

I took the opportunity to improve some unit tests.


joseivanlopez (approved):
LGTM

mvidner (approved):
LGTM, and a local `rake osc:build` passes (in 77s, octacore FTW!)

PR #1174, opened 2020-11-26
Initial wrapper for Storage::Tmpfs

ancorgs:
## Problem

In the past (before storage-ng) it was possible to use the YaST Partitioner to define tmpfs mount points in `/etc/fstab` (although it was quite buggy). It was also possible to create such mount points in AutoYaST using drives of type `CT_TMPFS`.

That was never implemented in storage-ng because the plan was to rely on systemd for the management of tmpfs mount points.

But systemd-based tmp filesystems are not as straightforward to create and maintain as the good old entries in `/etc/fstab`, specially with AutoYaST. Hence, there is a [feature request](https://jira.suse.com/browse/SLE-16313) to resurrect the AutoYaST capability to manage drives of type `CT_TMPFS`.

## Solution

In order to bring back the support for tmpfs in AutoYaST (and the Partitioner), support for fstab-based tmp filesystems was introduced in libstorage-ng 4.3.46.

This pull request makes such support available in YaST by implementing the corresponding `Y2Storage` wrapper.

The wrapper includes some extra functionality to check the size of the resulting filesystem (something that is a bit tricky using libstorage-ng directly).

## Testing

- Added a new unit test
- Tested manually


src/lib/y2storage/filesystems/tmpfs.rb
```diff
@@ -0,0 +1,89 @@
...
+require "y2storage/filesystems/base"
+
+module Y2Storage
+  module Filesystems
+    # Class to represent a tmpfs.
+    #
+    # The Tmpfs object should always have MountPoint as child. So the Tmpfs
+    # must be deleted whenever the MountPoint is removed.
+    #
+    # This a wrapper for Storage::Tmpfs
```
joseivanlopez, line 30: This *is*.

joseivanlopez (approved):
LGTM

src/lib/y2storage/filesystems/tmpfs.rb
ancorgs: This was copy&pasted from `Nfs`. So you found two typos at once. :wink: Commit with the fix added.

PR #1179, opened 2020-12-01
Partitioner: add tmpfs section 

joseivanlopez:
## Problem

In SLE 12 (with old *yast2-storage*) AutoYaST was able to import/export tmpfs file systems. Moreover, the Expert Partitioner included a section for the tmpfs file systems. Both features are missing in SLE 15.

* https://jira.suse.com/browse/SLE-16313
* PBI: https://trello.com/c/JRECcir9/2134-5-tmpfs-in-the-partitioner-and-autoyast

## Solution

This PR brings the tmpfs section back to the Expert Partitioner. Some actions (i.e., create and edit) are not implemented here. They will come with another PR. Including support for tmpfs file systems in AutoYasT will be also done in a separate PR.

* See also: https://github.com/libyui/libyui-qt/pull/142

## Testing

* Added unit tests.
* Tested manually with *partitioner_testing* client.

## Screenshots

<details><summary>Show/Hide</summary>

Screenshot from 2020-12-01 13-16-32

Screenshot from 2020-12-01 13-17-02

Screenshot from 2020-12-01 13-22-34

</details>


src/lib/y2partitioner/icons.rb
```diff
@@ -56,6 +56,8 @@ module Icons
     # icon
     BTRFS = LVM
 
+    TMPFS = BTRFS
+
```
ancorgs, line 60: Reusing the LVM icon for tmpfs doesn't look like a great idea. The current icon consist on several stacked hard disks, which makes a lot of sense for LVM but definitely not for tmpfs.

joseivanlopez: You are right. Maybe this? https://github.com/libyui/libyui-qt/blob/master/src/icons/drive-partition.svg

src/lib/y2storage/filesystems/tmpfs.rb
```diff
@@ -65,6 +68,21 @@ def size
...
       end
 
+      # Name used to idenfity the device
+      #
+      # @return [String]
+      def name
+        # FIXME: wrapper classes should not provide strings to be presented in the UI. Use decorators.
+        textdomain "storage"
+
+        # TRANSLATORS: name used to identify a tmpfs filesystem, where %{fs_type} is replaced by the
```
ancorgs, line 78: it's not `fs_type`, but `type`

ancorgs (approved):
LGTM in general. Just the concerns about the icons that we are already trying to address (in a separate pull request to libyui-qt).

ancorgs:
About the message when deleting... For a probed tmpfs, should we try to umount the filesystem right away, just like we do with other filesystems that are mounted?

If that's the case, then fine. It simply didn't look obvious to me from this PR.

joseivanlopez:
> About the message when deleting... For a probed tmpfs, should we try to umount the filesystem right away, just like we do with other filesystems that are mounted?
> 
> If that's the case, then fine. It simply didn't look obvious to me from this PR.

Yes, it is automatically done by the parent class `Actions::DeleteDevice`.

ancorgs (approved):
LGTM

PR #1181, opened 2020-12-02
Use shared hibernation check

joseivanlopez:
## Problem

`Yast::Kernel` API was extended with a `#propose_hibernation?` method. Such method indicates whether the "resume from RAM" feature should be used for the product being installed product, taking into account different aspects like: the current architecture, whether virtualization is involved, or whether the product should offer hibernation or not. But *yast-storage-ng* has its own simplistic logic to decide whether the resume from RAM feature is supported. *yast2-storage-ng* should be adapted to use the new `Yast::Kernel` API instead of its own logic.

* PBI: https://trello.com/c/Cys7qIrF/2194-1-hibernation-storage-part-of-sle-15921
* https://jira.suse.com/browse/SLE-15921
* See also, https://github.com/yast/yast-bootloader/pull/625

## Solution

The code was adapted to use the new `Yast::Kernel.propose_hibernation?` method.

## Testing

* Adapted unit tests


dgdavid (approved):
LGTM

PR #1178, opened 2020-11-30
Correct calculation of tmpfs size

ancorgs:
## Problem

Calculation of the size limit of a tmpfs is more complex than what our original implementation included. See this examples:

```no options
tmpfs (rw,relatime,inode64)

size=500M
tmpfs (rw,relatime,size=512000k,inode64)

size=30%
tmpfs (rw,relatime,size=2349284k,inode64)

nr_blocks=512
tmpfs (rw,relatime,size=2048k,inode64)

nr_blocks=1M
tmpfs (rw,relatime,size=4194304k,inode64)

nr_blocks=512,size=500M
tmpfs (rw,relatime,size=512000k,inode64)

size=500M,nr_blocks=512
tmpfs (rw,relatime,size=2048k,inode64)

nr_inodes=512
tmpfs (rw,relatime,nr_inodes=512,inode64)

size=500M,nr_inodes=512
tmpfs (rw,relatime,size=512000k,nr_inodes=512,inode64)
```

## Solution

Improve the size calculation to take all possible situations into account.

As a side effect, reorganize a bit the code, so `#ram_size` lives now in `Y2Storage::Arch`.

## Extra

This pull request also includes a rebase on top of master in order to use the new CI configuration.

## Testing

Added a new unit test

PR #1182, opened 2020-12-03
Merge the tmpfs branch into master

ancorgs:
## Problem

In the past (before storage-ng) it was possible to use the YaST Partitioner to define tmpfs mount points in `/etc/fstab` (although it was quite buggy). 

That was never implemented in storage-ng because the plan was to rely on systemd for the management of tmpfs mount points. But in the context of a recent [feature request](https://jira.suse.com/browse/SLE-16313) for AutoYaST, we have decided to resurrect support for tmpfs in the Partitioner.

## Solution

So far, support for tmpfs in the Partitioner has been developed in a separate `tmpfs` branch. This pull request merges all that work into master, including the following pull requests:

 - https://github.com/yast/yast-storage-ng/pull/1178
 - https://github.com/yast/yast-storage-ng/pull/1179
 - https://github.com/yast/yast-storage-ng/pull/1180

For further explanations and screenshots, check the descriptions of those pull requests.


joseivanlopez (approved):
LGTM

PR #1184, opened 2020-12-04
Fix unit tests

joseivanlopez:
## Problem

OBS build is failing after including these changes https://github.com/yast/yast-storage-ng/pull/1182, see https://build.opensuse.org/request/show/852869.

## Solution

Some unit tests were properly adapted.


ancorgs (approved):
LGTM

PR #1185, opened 2020-12-10
Partitioner: only show summary when needed

joseivanlopez:
## Problem

In some cases, the Expert Partitioner is showing the "Summary" step even though nothing was modified.

* https://bugzilla.suse.com/show_bug.cgi?id=1179829

## Solution

There was a problem with the condition to decide whether to show the "Summary". The idea was to show the Summary only if any device was edited. To decide that, a devicegraph comparison was used (`initial != current`). But such comparison takes into account the userdata. So, two devicegraphs with the same devices but different user data would be different.

Because that, when a devicegrah has Btrfs subvolumes, the userdata of Btrfs is filled in the *current* devicegraph to save its `subvolumes_prefix`. This was making *current* to be different to *initial*.

Now on, the Summary is shown only if there is any action to perform during the commit phase.

## Testing

- Added unit test
- Tested manually with the *partitioner_testing* client

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Dec 10 08:45:09 UTC 2020 - José Iván López González <jlopez@suse.com>
+
+- Partitioner: do not show summary when there are no actions to
+  perform (bsc#1179829).
+- 4.3.28
+
+-------------------------------------------------------------------
 Fri Dec  4 08:57:17 UTC 2020 - José Iván López González <jlopez@suse.com>
 
 - Fix unit tests (related to jsc#SLE-11308).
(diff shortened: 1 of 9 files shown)


package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Dec 10 08:45:09 UTC 2020 - José Iván López González <jlopez@suse.com>
+
+- Partitioner: do not show summary when there are no actions to
```
dgdavid, line 4: NP: Actually the fix not only stop displaying the summary when it is not useful. It _stop considering the system edited_ if there are no actions to perform. Although I understand that the bug report is only about the summary.

dgdavid (approved):
It looks good, thank you so much for taking care as well as for the PR explanation.

I tested it manually and it also prevents to display the warning message when aborting if it is not needed.

package/yast2-storage-ng.changes
joseivanlopez: Yes, it has more implications (next/finish button, abort message, etc), but I did not consider relevant to list all them.

PR #1188, opened 2020-12-15
Improve installation of packages

ancorgs:
## Problem

Sometimes yast2-storage-ng decides that some packages need to be installed. But there was room for improvement at least in two situations:

- [bsc#1065588](https://bugzilla.suse.com/show_bug.cgi?id=1065588): During installation, all potentially needed packages were added to the software proposal as mandatory ones, which implies the users cannot install the system if they decide to deselect the packages for installation at a later point (ie. in the "Software" section of the installation summary).

- [bsc#1168077](https://bugzilla.suse.com/show_bug.cgi?id=1168077): In an already installed system, the Partitioner calculated the packages to install based on the content of the system after performing all the changes, instead of calculating that based on the actions the Partitioner was going to actually perform. For example, if the system contained an NTFS filesystem that was not going to be modified in any way, the Partitioner always insisted in installing `ntfs-3g` and `ntfsprogs`.

The described behavior was the best yast2-storage-ng could do based on the old functionality provided by libstorage-ng, that offered no way to distinguish optional and required packages and no way to check which features were needed for a concrete set of actions.

Internal Trello card: https://trello.com/c/SresnvEB/1826-3-osdistribution-p5-1065588-these-packages-need-to-be-selected-to-install-ntfs-3g-ntfsprogs

## Solution

This [pull request](https://github.com/openSUSE/libstorage-ng/pull/774) extended libstorage-ng API, so now it can provide all the information needed by yast2-storage-ng. So the current pull request improves package handling in yast2-storage-ng in several ways.

During installation, some packages are added to the software proposal as required but others are added as optional, which means the user can de-select them without blocking the whole system installation. For example, if the system contains a filesystem that is NOT going to be mounted by default (let's say, a NTFS or JFS partition from another operating system), the user can deselect the corresponding packages and finish the installation without them.

In a running system, the Partitioner will only try to install the packages it needs to perform the requested operations, ignoring devices that are not involved in the changes.

As a nice side effect of the latter change, the Partitioner will cause way fewer repository refreshes. Turns out the Partitioner sometimes triggered unwanted `zypper refresh` operations when it considered that some of these packages could need to be installed (no matter whether they were actually installed or not): "ntfs-3g", "ntfsprogs", "exfat-utils", "f2fs-tools", "jfsutils". Now that will only happen if a NTFS/ExFAT/JFS/f2fs filesystem is indeed going to be created or modified, which is a very minimal subset of all the situations that used to trigger the repository refresh in the past.

## Testing

- Both scenarios tested manually
- Unit tests extended

### DIFF
--- a/src/lib/y2partitioner/widgets/summary_text.rb
+++ b/src/lib/y2partitioner/widgets/summary_text.rb
@@ -100,6 +100,10 @@ module Y2Partitioner
 
       # Updates the value of {#packages}
       def calculate_packages
+        # This is only used during installation (the summary in a installed system does not
+        # include the packages to be installed because they are interactively installed right
+        # away), so we use Devicegraph#used_features (without arguments) to display all packages
+        # that are added to the software proposal, both optional and required.
         @packages = current_graph.used_features.pkg_list
       end
 
--- a/src/lib/y2partitioner/widgets/summary_text.rb
+++ b/src/lib/y2partitioner/widgets/summary_text.rb
@@ -100,7 +100,7 @@ module Y2Partitioner
 
       # Updates the value of {#packages}
       def calculate_packages
-        # This is only used during installation (the summary in a installed system does not
+        # This is only used during installation (the summary in an installed system does not
         # include the packages to be installed because they are interactively installed right
         # away), so we use Devicegraph#used_features (without arguments) to display all packages
         # that are added to the software proposal, both optional and required.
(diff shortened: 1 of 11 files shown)


dgdavid (approved):

src/lib/y2partitioner/widgets/summary_text.rb
```diff
@@ -100,6 +100,10 @@ def calculate_actions
 
       # Updates the value of {#packages}
       def calculate_packages
+        # This is only used during installation (the summary in a installed system does not
```
joseivanlopez, line 103: Typo: *an* installed ...

joseivanlopez (approved):
LGTM

PR #1189, opened 2020-12-17
Use the REMOVE view in remove_with_dependants

ancorgs:
## Problem

The problem described on [bsc#1179590](https://bugzilla.suse.com/show_bug.cgi?id=1179590) was caused because deleting a Btrfs subvolume in the devicegraph was not triggering the corresponding deletion of its qgroups.

That was fixed in libstorage-ng by adding the qgroups to the REMOVE view. See https://github.com/openSUSE/libstorage-ng/pull/787

But that didn't fix the problem in all possible situations because `Devicegraph#remove_with_dependants` was not honoring the REMOVE view.

## Solution

This PR fixes `remove_with_dependants` to use the REMOVE view when descending the hierarchy.

## Future improvements

Now that libstorage-ng has the concept of the REMOVE view, maybe we can get rid of `#remove_with_dependants` if we move some of its logic (eg. the fact that deleting a VG should invalidate all its PVs) to that view, instead of having extra code in the YaST side.

## Testing

- Added a new unit test
- I didn't check whether it actually solves 100% the situation described in [bsc#1179590](https://bugzilla.suse.com/show_bug.cgi?id=1179590), but it should.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Dec 17 11:31:34 UTC 2020 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Fixed deletion of dependant devicegraph nodes (bsc#1179590)
+- 4.3.32
+
+-------------------------------------------------------------------
 Tue Dec 15 15:56:44 UTC 2020 - José Iván López González <jlopez@suse.com>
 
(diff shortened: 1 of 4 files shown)


dgdavid (approved):
LGTM

dgdavid:
> I didn't check whether it actually solves 100% the situation described in bsc#1179590, but it should.  — @ancorgs in the PR's description

I have tested it manually using a [dud](https://en.opensuse.org/SDB:Linuxrc#p_dud) and the installation works as expected, without warnings.

PR #1191, opened 2020-12-18
AutoYaST: disable quotas when <quotas> is set to false

imobachgs:
If `<quotas>` element is set to `false`, AutoYaST disables the quotas no matter whether the `referenced_limit` is specified for some subvolume.

Related to #1186.


ancorgs (approved):
LGTM

PR #1193, opened 2020-12-22
Partitioner: do not allow to modify the path of Btrfs subvolumes (bsc#1180182)

ancorgs:
## Problem

The YaST Partitioner didn't allow the users to modify the path of those subvolumes that already existed on the disk.

But it allowed to modify the path of those subvolumes that existed only in the devicegraph so far. In order to make that possible, the corresponding subvolume was deleted from the devicegraph and re-created.

That had undesired effects, like causing the removal of all the children subvolumes every time a subvolume was modified (even if the path was not altered). That's described in [bsc#1180182](https://bugzilla.suse.com/show_bug.cgi?id=1180182) (which is a spin-off of [bsc#1179590](https://bugzilla.suse.com/show_bug.cgi?id=1179590)).

## Solution

Simply, do not allow to modify the path of a subvolume that already exists in the devicegraph. libstorage-ng doesn't offer a safe way of doing such change and recreating the whole thing doesn't look like a good idea either (more similar problems would likely arise, since we are risking to confuse libstorage-ng).

## Testing

- Improved/added unit tests
- Tested manually with `partitioner_testing`

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Dec 22 13:09:47 UTC 2020 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Partitioner: do not allow to modify the path of Btrfs subvolumes
+  (bsc#1180182)
+- 4.3.35
+
+-------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


imobachgs (commented):
The code looks good. Just missing the entry in the changes file and the version bump.

dgdavid (approved):

PR #1195, opened 2020-12-29
Attempt to fix #1194

### LINKED ISSUE #1194: Sync EFI partition conditions with the "other" world
Hi,

while installing TW 20201223 on a brand new system (ASUS ROG STRIX B550-F, AMD Ryzen 9 3900X, WD Black 1TB SN850 NVMe), that should perform in `both` worlds, I noticed the partitioner complaining about a too small efi partition (<128G) with the wrong type, resulting in a failed `shim-install` call later on. Luckily, I was able to fix this manually, but that shouldn't be necessary for the common case.

Still interested? Here are the details: The usual procedure is: install Windows 10 (Pro), get it to behave, and install TW into a resized primary Win partition. The system parts are pretty current (besides the AMD cpu from the last generation). Consequently, Windows didn't install, unless an 20H2 image was used (created with the latest MediaCreationTool).

I mention this in order to show, that the Microsoft part is current as well. Unfortunately, it seems to be using some default sizes, that YaST doesn't deal with properly, and screwing the OOB experience greatly.

Here's the final layout:
```
Disk /dev/nvme0n1: 931.53 GiB, 1000204886016 bytes, 1953525168 sectors
Disk model: WDS100T1X0E-00AFY0                      
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: FCAE70F3-XXXX-YYYY-ZZZZ-serial obscured

Device             Start        End    Sectors   Size Type
/dev/nvme0n1p1      2048     206847     204800   100M EFI System
/dev/nvme0n1p2    206848     239615      32768    16M Microsoft reserved
/dev/nvme0n1p3    239616  773332991  773093376 368.7G Microsoft basic data
/dev/nvme0n1p4 773332992 1953525134 1180192143 562.8G Linux filesystem
```

If Microsoft thinks, 100M is enough for the EFI partition, we should cope with that fact.
It looks, like things are well after manual intervention:
```
$ df  -h
/dev/nvme0n1p4         563G   55G  509G  10% /
/dev/nvme0n1p1          96M   31M   66M  32% /boot/efi
```

frispete:
By default, Windows 10 (20H2) demands and creates a 100 MiB EFI partition of its own. While it would be preferable to add some margin, reduce minimum size from 128 to 100 MiB in order to allow reusing an existing EFI partition, which should provide a better OOB installation experience in the common case. Windows 10 (20H2) occupies less than 30 MiB, and we're additionally using about 4 MiB for the time being.

I'm not sure to turn the right screw here, but I haven't found a better matching one.

Attempts to fix #1194

### DIFF
--- a/doc/boot-partition.md
+++ b/doc/boot-partition.md
@@ -28,7 +28,7 @@ Boot Partition Layout / Restrictions For Storage Proposal
 
 ### Grub2 and disk abstractions
 
-- Grub natively suppors lvm raid 0/1/4/5/6, encryption with or without lvm
+- Grub natively supports lvm raid 0/1/4/5/6, encryption with or without lvm
 - The problem with disk abstractions like lvm or raid is not exactly about
   booting. The system will boot fine but some features have an additional
   requirement - having a pre-boot environment block writable by grub-once.
@@ -65,7 +65,7 @@ Boot Partition Layout / Restrictions For Storage Proposal
  * generic boot loader installed into mbr
  * we have generic boot code for both dos/gpt partition table
  * stage1 installed into /boot or / partition (if possible, note: not on xfs)
- * **OR** separate grub boot partition for embdding stage1 (like prep on ppc)
+ * **OR** separate grub boot partition for embedding stage1 (like prep on ppc)
 
      > *[mchang]* gpt has bios_grub partition but only if you instruct grub2-install to
 install stage1 on mbr then it will search bios_grub to embed stage2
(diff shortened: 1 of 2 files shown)


ancorgs:
See my comment on #1194, I don't think the correct solution is to lower this number to match whatever Windows is proposing nowadays. That would be hard to keep in sync.

I would prefer to stop using the minimal of 128MiB as an absolute criteria in situations in which that number is indeed not so relevant. In other words, the installer should use existing EFI partitions even if they are smaller than the one the installer would have proposed automatically for a brand empty disk.

frispete:
Probably, you want to cherry pick the typo fixes?

PR #1197, opened 2021-01-14
Do not use the ESP minimal size to discard partitions

ancorgs:
## Problem

The automatic partitioning proposal never creates an EFI system partition (ESP) that is smaller than a certain size (128 MiB) at the time of writing. That's fine so far.

But there are two other situations in which the same minimal size is applied and that's causing problems to some users:

- When the automatic partitioning proposal is deciding whether a pre-existing ESP should be reused instead of proposing a brand new one.
- If the Expert Partitioner was used to create a custom ESP.

So the min size is ok for brand new EFI partitions suggested by the installer, but it should not be used to discard partitions in other scenarios.

See the detailed discussion here: #1194. 

See also (restricted access): https://trello.com/c/eAdUReCf/2236-2-do-not-use-the-efi-minimal-size-to-discard-partitions

## Solution

Stop using the proposal min ESP size to discard partitions.

This pull request replaces #1195. As discussed there, this solution is preferred to lowering the min size used by the proposal.

## Testing

- Adapted the unit tests.
- Manually tested:
  - Verified the proposal indeed reuses a small ESP (before this, it used to propose the creation of a second EFI partition).
  - Verified the guided setup does not longer show the warning pop-up if a small (but correct otherwise) ESP is manually created.

### DIFF
--- a/doc/boot-partition.md
+++ b/doc/boot-partition.md
@@ -28,7 +28,7 @@ Boot Partition Layout / Restrictions For Storage Proposal
 
 ### Grub2 and disk abstractions
 
-- Grub natively suppors lvm raid 0/1/4/5/6, encryption with or without lvm
+- Grub natively supports lvm raid 0/1/4/5/6, encryption with or without lvm
 - The problem with disk abstractions like lvm or raid is not exactly about
   booting. The system will boot fine but some features have an additional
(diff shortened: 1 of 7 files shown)


joseivanlopez (approved):
LGTM

PR #1198, opened 2021-01-18
Use the driver to decide whether _netdev is needed

ancorgs:
## Problem

See [bsc#1176140](https://bugzilla.suse.com/show_bug.cgi?id=1176140).

YaST adds `_netdev` to the default mount options of all file systems that are placed on network-based disks.

Turns out that adding `_netdev` when not really needed is counterproductive and can cause ordering cycles in the systemd units. That's something that would need a more general solution, likely implemented in the systemd side.

On the other hand, YaST considers that all iSCSI and FCoE disks are network-based and, thus, their mount points could need the `_netdev` option. But turns out that is not true, many iSCSI and FCoE drivers can actually initialize the disks and make the corresponding block device available to the operating system before the network is fully configured.

## Solution

Do not blindly trust the data transport to decide if a device deserves the `_netdev` option in their mount points. Check also whether the driver is able to handle the situation on its own. So far, this implements a hard-coded list of drivers that would need `_netdev` and the option is only added if the current driver is in that list.

## Testing

- Added new unit tests
- Tested manually with a DUD on top of SLE-15-SP2
- Grab testing package [yast2-storage-ng-4.2.115](https://build.suse.de/package/show/home:ancorgs:branches:SUSE:SLE-15-SP2:Update/yast2-storage-ng)

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jan 20 07:26:55 UTC 2021 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Improved mechanism to detect whether _netdev is needed for a
+  given disk: use its driver as extra criterion (bsc#1176140).
+- 4.2.115
+
+-------------------------------------------------------------------
 Tue Oct 20 15:36:03 UTC 2020 - Stefan Hundhammer <shundhammer@suse.com>
 
 - Added $LIBSTORAGE_IGNORE_PROBE_ERRORS environment variable
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-storage-ng
-Version:        4.2.114
+Version:        4.2.115
 Release:        0
 Summary:        YaST2 - Storage Configuration
 License:        GPL-2.0-only OR GPL-3.0-only
--- a/src/lib/y2storage/blk_device.rb
+++ b/src/lib/y2storage/blk_device.rb
@@ -611,6 +611,17 @@ module Y2Storage
       hwinfo.bus
     end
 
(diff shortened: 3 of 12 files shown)


joseivanlopez:
Maybe we need a new method to get the "underlying" plain block devices from a specific device, for example:

~~~
LvmVg#underlying_blk_devices #=> [pv1, pv2, pv3]
LvmPv#underlying_blk_devices #=> [partition1]
Partition#underlying_blk_devices #=> [disk1]
~~~

And then, we can have a generic `in_network?` and `systemd_remote?` in `BlkDevice`:

~~~
def systemd_remote?
  underlying_blk_devices.any?(:systemd_remote?)
end

def in_network?
 underlying_blk_devices.any?(:in_network?)
end
~~~

And some specific devices implements the check, for example for `Disk`:

~~~
def systemd_remote?
  # check
end
~~~

I am not sure if this is a good idea or not. Only thinking about avoiding some repetitive code we have now.

ancorgs:
> Maybe we need a new method to get the "underlying" plain block devices from a specific device, for example:
> 
> [...]
> 
> I am not sure if this is a good idea or not. Only thinking about avoiding some repetitive code we have now.

I tried to implement something very similar to what you just suggested. Check #1199. Indeed it results in less code. Although it's only 20 fewer lines of code, the strong point is that the implementation remains in `BlkDevice` and `Disk`. That makes it easier to maintain (actually, the current implementation overlooks `Bcache`, which is covered by #1199 since everything is managed in the base `BlkDevice` class).

ancorgs:
@joseivanlopez What do you think about going forward with this current PR for SLE-15-SP2 and implementing the code unification from #1199 only when merging into master?

joseivanlopez (approved):
LGTM

ancorgs:
The reporter of bsc#1176140 confirmed this fix worked for them. So I will merge now and start working on the (improved) merge to master.

PR #1199, opened 2021-01-19
Alternative implementation of: use the driver to decide whether _netdev is needed

ancorgs:
This is the same than #1198 but with an extra commit to reduce the code duplication based on @joseivanlopez's comment at the original pull request.

Done in a separate pull request just to showcase the idea because I'm not sure it's a good idea to include it as part of the maintenance update.


ancorgs:
Closed in favor of #1209

PR #1200, opened 2021-01-21
Partitioner: Fix subvolume dialog

joseivanlopez:
## Problem

When creating a new subvolume and the "Size Limit" checkbox is marked, the subvolume is created without the indicated limit.

* https://bugzilla.suse.com/show_bug.cgi?id=1181205

## Solution

Fix controller to take into account the limit option.

## Testing

* Added new unit tests
* Tested manually with *partitioner_testing* client


ancorgs (approved):
LGTM

PR #1203, opened 2021-01-28
AutoYaST UI: improved visualization of some partition sections

ancorgs:
## Problem

There are several situations in the AutoYaST profile in which the `<partition>` subsection within a given `<drive>` section does not really represent a partition, but it contains general information about the drive itself. For example, when the drive contains a disklabel set to the value "none" or when the drive represents an NFS or btrfs filesystem.

In those cases, the corresponding entry in the left tree of the AutoYaST UI should read "Partition (Drive)" instead of just "Partition". That was happening for drives with the disklabel "none", but not for drives of type `CT_NFS` or `CT_BTRFS`.

autoyast_tree1

## Solution

Fixed the method used to get the so-called "master" partition section (that is, the partition section that actually is meant to describe the whole drive). Now the tree looks as expected.

autoyast_tree2

## Testing

Tested manually

## See also

https://github.com/yast/yast-autoinstallation/pull/727

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jan 28 17:07:25 UTC 2021 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- AutoYaST UI: improved visualization of some partition sections
+  in the left tree (related to jsc#SLE-11308).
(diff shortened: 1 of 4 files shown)


dgdavid (approved):
LGTM

dgdavid (approved):
:+1:

PR #1205, opened 2021-02-03
Update proposal documentation (bsc#1175098)

ancorgs:
## Problem

The documents about the storage proposal were messy because they were left in a state where they documented the transformation process from yast-storage into yast-storage-ng instead of focusing on the status quo.

1) The reference documentation of the control file (at the yast-installation repo) was still describing the old yast-storage format
2) The more detailed documentation in this repo was written as some kind of list of differences with the obsolete old format, instead of simply describing the present.

## Solution

The solution comes in two pull requests: this one and https://github.com/yast/yast-installation/pull/909

This one updates the information regarding how the proposal works, removing a lot of information that has been moved to the general control file document (see the mentioned https://github.com/yast/yast-installation/pull/909) and also reordering and rewording the remaining information to make it more current (eg. no more references to "new vs legacy" but using "current vs old" instead).

This pull request also deletes all references to how the yast-storage-ng proposal partially supports the legacy format, since that support will be soon removed now that all control files are migrated to the yast-storage-ng format (something that was done already for the first release of SLE/Leap 15.0),

## Review

Better see the rendered result here: https://github.com/ancorgs/yast-storage-ng/blob/ee56865d86d234ff39840aa927e1936e294a6a47/doc/old_and_new_proposal.md

### DIFF
--- a/doc/old_and_new_proposal.md
+++ b/doc/old_and_new_proposal.md
@@ -1,270 +1,53 @@
-# Storage proposal: old (legacy) and new (ng) approach
+# Storage proposal: current (yast-storage-ng) and old (yast-storage) approach
 
-This document covers the differences between the storage proposals of
-yast-storage and yast-storage-ng, with a brief explanation of how they work and
-how both can/must be influenced by the settings coming from `control.xml` and
-other sources.
+This document describes the storage proposals of both yast-storage and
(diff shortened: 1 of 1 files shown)


dgdavid (approved):
LGTM

PR #1206, opened 2021-02-04
Partitioner: do not offer LVM pools for bcache and btrfs (bsc#1170044)

ancorgs:
## Problem

https://bugzilla.suse.com/show_bug.cgi?id=1170044

Some special kinds of LVM logical volumes (thin pools and cache pools) cannot be used as base for a bcache (nor backing or caching device) or for a Btrfs filesystem, because those logical volumes are not block devices.

But the Partitioner was not making any distinction, so they were offered in the UI just like any other logical volume. Of course, that leaded to problems at a later point if the user selected any of them to be part of any bcache or btrfs.

## Solution

In both cases, check the `Storage::BlkDevice#usable_as_blk_device?` method offered by libstorage-ng to distinguish those devices that, despite being represented as an instance of `BlkDevice`, are not really block devices in the full sense.

## Testing

Added new unit tests

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Feb  4 14:17:09 UTC 2021 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Partitioner: stop offering LVM pools as possible base devices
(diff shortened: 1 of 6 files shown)


shundhammer (approved):
LGTM

PR #1207, opened 2021-02-11
Partitioner: only collapse by default Btrfs with snapshots

ancorgs:
## Problem

To avoid the situation in which a filesystem with many subvolumes (specially snapshots) hides other devices (eg. subsequent disks), we decided to display some branches of all tables of devices collapsed by default. But the original criteria to decide whether a given branch should be initially opened or closed was pretty simplistic - just collapsing branches with more than 10 elements. That turned to be inconvenient in many cases like the one reported at [bsc#1181464](https://bugzilla.suse.com/show_bug.cgi?id=1181464): a single disk with many partitions... in which all the interesting information (the list of partitions) is hidden by default.

## Solution

By default, only collapse branches that contain Btrfs snapshots. Except for the tables in the Btrfs section of the Partitioner, in which everything is displayed as initially expanded. 

## Testing

- Tested manually with the `partitioner_testing` client
- Adapted the unit tests and added more

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Feb 11 14:16:57 UTC 2021 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Partitioner: in general, collapse branches of the tables only if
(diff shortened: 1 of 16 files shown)


joseivanlopez (approved):
LGTM

PR #1209, opened 2021-02-25
Use the driver to decide whether _netdev is needed

ancorgs:
In order to mitigate [bsc#1176140](https://bugzilla.suse.com/show_bug.cgi?id=1176140) we implemented #1198 in the SLE-15-SP2 branch.

That pull request is approved, verified by the reporter and merged. So it's time to merge this into master.

But during the code review process I agreed with @joseivanlopez that the merge into master should include an extra commit to reduce code duplication.

So this PR includes both parts, the merge and the extra commit (3a8ca23).

This is not still the final solution to the `_netdev` problem, but it will reduce to a minimum the cases in which the current sub-optimal approach hits the users. As discussed in the bug report, I think the definitive solution should not come from YaST.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Feb 25 16:26:55 UTC 2021 - Ancor Gonzalez Sosa <ancor@suse.com>
+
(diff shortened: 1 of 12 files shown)


joseivanlopez (approved):
LGTM

PR #1210, opened 2021-03-09
Partitioner: allow to define the filesystem label

joseivanlopez:
## Problem

In the Expert Partitioner, when editing the fstab options of a file system, the text field to give a label is not shown.

https://bugzilla.suse.com/show_bug.cgi?id=1183220

## Solution

Fix bug in method that checks whether a file system supports a label.

## Testing

* Added new unit tests
* Tested manually with *partitioner_testing* client.


ancorgs (approved):
LGTM

PR #1212, opened 2021-03-11
Improve LVM striped volumes support

joseivanlopez:
## Problem

When a striped logical volume is going to be created, its size should be multiple of the physical extend size. Moreover, its resulting number of physical extents should be multiple of the number of stripes. Otherwise, *lvcreate* command will round the number of extents of the new logical volume up to make it multiple of the number of stripes. This could lead to a number of extents that exceeds the number of free extents from the volume group. 

* https://bugzilla.suse.com/show_bug.cgi?id=1180723

## Solution

Now on, the size of the new striped volume is rounded-down to make the number of extents multiple of the number of stripes. This is done by the Expert Partitioner and also by the storage proposal when a new logical volume is created . Note that the size of the logical volume is also rounded when the volume is resized.

Moreover, some new validations where added in order to ensure that the user does not try to create a striped volume with a size impossible to fit according to new number of stripes and the size of the physical volumes. Take into account that the size of a striped volume is limited by the size of the physical volumes. For example, if the volume group has 3 physical volumes whose usable sizes are: 1 GiB, 2 GiB and 5 GiB:

* A volume with 2 stripes cannot be bigger than 4 GiB.
* A volume with 3 stripes cannot be bigger than 3 GiB.
* A linear volume cannot be bigger than 8 GiB.

Similarly, when the physical volumes are edited, some new validations ensure that the striped volumes can still be created with the selected physical volumes. 

Anyway, all these new validations are not 100% accurate:

* They are based on the usable size of the physical volumes but they do not consider the free extents of each physical volume (*libstorage-ng* does not provide that info yet).
* They only check whether every individual striped volume can be allocated, but they do not check whether the volume group can allocate all the striped volumes at the same time (*libstorage-ng* should calculate how the extents are assigned between the physical volumes).

## Testing

* Tested manually
* Added new unit tests

## Screenshots

<details>
<summary>Show/Hide</summary>

Screenshot from 2021-03-17 15-36-38

Screenshot from 2021-03-17 15-35-34

Screenshot from 2021-03-17 15-30-42

</details>

### DIFF
--- a/src/lib/y2partitioner/actions/controllers/lvm_vg.rb
+++ b/src/lib/y2partitioner/actions/controllers/lvm_vg.rb
@@ -218,7 +218,7 @@ module Y2Partitioner
         # Maximum number of stripes used by the striped logical volumes of the volume group
         #
         # @return [Integer] 0 if no striped volumes
-        def max_stripes
+        def lvs_stripes
           return 0 if striped_lvs.none?
 
           striped_lvs.map(&:stripes).max
--- a/src/lib/y2partitioner/actions/controllers/lvm_vg.rb
+++ b/src/lib/y2partitioner/actions/controllers/lvm_vg.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2017-2019] SUSE LLC
+# Copyright (c) [2017-2021] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -203,6 +203,27 @@ module Y2Partitioner
           BlkDeviceRestorer.new(device.plain_device).restore_from_checkpoint
         end
 
+        # Whether the volume group can allocate each striped logical volume.
+        #
+        # Note that this method only checks that the volume group has the necessary physical volumes to
+        # allocate each striped volume individually. Checking that the volume group can allocate all the
+        # striped volumes at the same time would be very tricky. Such a check requires to know how the
+        # volumes are distributed in the physical volumes.
+        #
+        # @return [Boolean]
+        def size_for_striped_lvs?
+          striped_lvs.all? { |l| vg.size_for_striped_lv?(l.size, l.stripes) }
+        end
+
+        # Maximum number of stripes used by the striped logical volumes of the volume group
+        #
+        # @return [Integer] 0 if no striped volumes
+        def max_stripes
+          return 0 if striped_lvs.none?
+
+          striped_lvs.map(&:stripes).max
+        end
+
         private
 
         # Current action to perform
@@ -379,6 +400,13 @@ module Y2Partitioner
         # @return [Boolean]
         def valid_device_for_vg?(device)
           device.is?(:disk, :multipath, :bios_raid, :md)
+        end
+
+        # Striped logical volumes from the volume group
+        #
+        # @return [Array<Y2Storage::LvmLv>]
+        def striped_lvs
+          vg.lvm_lvs.select(&:striped?)
         end
       end
     end
--- a/src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
+++ b/src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
@@ -140,14 +140,14 @@ module Y2Partitioner
       #
       # @return [String, nil]
       def striped_lvs_devices_error
-        return nil if controller.devices_in_vg.size >= controller.max_stripes
+        return nil if controller.devices_in_vg.size >= controller.lvs_stripes
 
         format(
-          # TRANSLATORS: Error message, where %{max_stripes} is replaced by a number (e.g., 3).
+          # TRANSLATORS: Error message, where %s is replaced by a number (e.g., 3).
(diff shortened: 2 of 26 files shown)


src/lib/y2partitioner/actions/controllers/lvm_vg.rb
```diff
@@ -203,6 +203,27 @@ def remove_device(device)
...
+        #
+        # @return [Boolean]
+        def size_for_striped_lvs?
+          striped_lvs.all? { |l| vg.size_for_striped_lv?(l.size, l.stripes) }
+        end
+
+        # Maximum number of stripes used by the striped logical volumes of the volume group
+        #
+        # @return [Integer] 0 if no striped volumes
+        def max_stripes
```
ancorgs, line 221: Not sure if `max_stripes` is a good name. It sounds more like a limit than as a current value. Why not simply `stripes` or something slightly more explicit like `lvs_stripes`.

src/lib/y2partitioner/widgets/lvm_vg_devices_selector.rb
```diff
@@ -92,53 +91,76 @@ def unselect(sids)
...
-          # enough. %s is replaced by a size (e.g., 4 GiB).
-          format(
-            _("The volume group size cannot be less than %s."), controller.lvs_size
-          )
+      # @return [String, nil]
+      def striped_lvs_devices_error
+        return nil if controller.devices_in_vg.size >= controller.max_stripes
+
+        format(
+          # TRANSLATORS: Error message, where %{max_stripes} is replaced by a number (e.g., 3).
```
ancorgs, line 146: NP: when there is only one placeholder I would tend to use the syntax with `%s`, which is more familiar to translators.

src/lib/y2storage/lvm_lv.rb
```diff
@@ -179,7 +186,11 @@ def overcommitted?
     # closest valid (i.e. divisible by the extent size) value, rounding down if
     # needed.
     #
-    # @param new_size [DiskSize] temptative new size of the volume, take into
+    # The resulting size is also rounded down according to the number of stripes,
+    # except when setting the minimum size. Note that lvcreate command will round
```
ancorgs, line 190: NP: I guess that "lvcreate will round it up again" only applies if the size was set to the minimum. I'm not sure if it's fully obvious from the current wording.

src/lib/y2storage/lvm_vg.rb
```diff
@@ -216,6 +216,68 @@ def potential_orphans
...
+    # example, let's say we have a volume group with 3 physical volumes: pv1 (2 GiB), pv2 (50 GiB) and
+    # pv3 (1 GiB). For a striped volume with 2 stripes, the second biggest physical volume (pv1)
+    # restricts its maximum size. In this case, the maximum size would be:
+    # 2 (stripes) * 2 GiB (pv1 size) = 4 GiB.
+    # If the number of stripes is 3, then the maximum size is limited by the third biggest physical
+    # volume (pv3), so the maximum size would be: 3 (stripes) * 1 GiB (pv3 size) = 3 GiB.
+    #
+    # @param stripes [Integer] number of stripes. It must be bigger than 1
+    #
+    # @raise [RuntimeError] when the given number of stripes is incorrect (i.e., less than 1)
```
ancorgs, line 231: Shouldn't this be an `ArgumentError`? To my taste `RuntimeError` is too generic.

src/lib/y2storage/lvm_vg.rb
```diff
@@ -216,6 +216,68 @@ def potential_orphans
...
+
+      return nil if stripes > lvm_pvs.size
+
+      lvm_pvs.map(&:blk_device).map(&:size).sort[-stripes]
+    end
+
+    # Maximum size for a new striped logical volume
+    #
+    # @note This size is not 100% accurate. It is based on the size of the block devices used as physical
+    #   volumes. But LVM probably uses some space to allocate its metadata. libstorage-ng provides some
```
ancorgs, line 244: It's not only about the LVM metadata. If I'm not mistaken, this method also assumes the stripped LV is the only one having some space in the limiting PV. In other words, this does not have into account the space already used by other planned or pe-existing LVs.

And about the LVM metadata, we already have such content elsewhere. Let me check.

ancorgs: > And about the LVM metadata, we already have some logic somewhere to deal with that. Let me check.

It's in `Y2Storage::Planned::LvmVg#useful_pv_space` and similar methods.

joseivanlopez: I have just realized that *libstorage-ng* already provides a method to get the usable size, see https://github.com/openSUSE/libstorage-ng/blob/master/storage/Devices/LvmPv.h#L85.

ancorgs (approved):
LGTM (although I have to admit I didn't go too deep into the testsuite changes).

PR #1213, opened 2021-03-30
Avoid error when using self

joseivanlopez:
## Problem

After merging https://github.com/yast/yast-storage-ng/pull/1212, unit tests started failing in IBS for SLE-15-SP3, but it worked fine in OBS for Factory. The error was produced because a private method is called over *self*, which raises an exception with ruby < 2.7, see:

* https://bugs.ruby-lang.org/issues/16123
*  https://bugs.ruby-lang.org/issues/11297
* https://blog.saeloun.com/2019/12/24/ruby-2-7-allows-calling-a-private-method-with-self.html

## Solution

Avoid direct calls of private methods over *self* to make it work in both: SLE-15-SP3 (ruby 2.5) and TW.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Mar 30 10:31:55 UTC 2021 - José Iván López González <jlopez@suse.com>
+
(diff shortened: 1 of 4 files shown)


dgdavid (approved):
LGTM

PR #1218, opened 2021-05-21
try harder matching device names (bsc#1186268)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1186268
- https://trello.com/c/gNgBB1Qh

When looking up device names from fstab, check against all possible variants (udev symlinks).


ancorgs (approved):
LGTM

PR #1219, opened 2021-06-04
ensure mount options are not doubled (bsc#1186298)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1186298
- https://trello.com/c/I08exOoV

If a filesystem is both marked as read-only and explicitly includes `ro` in its fstab option list, it should not end up with `ro` twice in its fstab entry.

## Solution

De-duplicate fstab options.

## Note

Receive counts in `include` seem to be a rather recent rspec feature (3.10, it seems); so replacing `include().once` with something weird to make the tests work in travis.


ancorgs (approved):
LGTM

PR #1223, opened 2021-06-07
ensure mount options are not doubled (bsc#1186298)

wfeldt:
## Task

Port https://github.com/yast/yast-storage-ng/pull/1219 to master branch.

## Original problem

- https://bugzilla.suse.com/show_bug.cgi?id=1186298
- https://trello.com/c/I08exOoV

If a filesystem is both marked as read-only and explicitly includes `ro` in its fstab option list, it should not end up with `ro` twice in its fstab entry.

## Solution

De-duplicate fstab options.


shundhammer (approved):
LGTM

PR #1228, opened 2021-08-04
Improve detection of candidate disks

joseivanlopez:
## Problem

The storage proposal looks for the devices that can be used for the installation (a.k.a. *candidate disks*). A device is considered a *candidate disk* if it fulfills some conditions. For example, the devices containing a repository  for the installation are discarded as *candidate disks*. The URI of the installation repositories (e.g., *hd:/?device=/dev/disk/by-id/mmc-EB1QT_0xd948672b-part2*) are analyzed to determine whether a device contains a repository. The devices indicated in the URI are not used as *candidate disks*. 

The URI of the repository can point to a device by the kernel name or by one of the udev symlinks. In the last case, the device could not be found if *libstorage-ng* does not provide that udev name (note that *libstorage-ng* filters some udev names). As result, the device is not excluded as *candidate disk*. 

* https://bugzilla.suse.com/show_bug.cgi?id=1185694

## Solution

Improve the detection of the devices indicated in the URI of an installation repository. Now, the `DeviceFinder` utility is used, which is able to find a device by any name including any symbolic link in the */dev* directory.

## Testing

- Added a new unit test
- This should be manually tested by the reported in the failing scenario.

### DIFF
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -315,11 +315,9 @@ module Y2Storage
 
     # Checks whether a device contains an installation repository
     #
-    # A device contains an installation repository if its kernel name or the kernel name of any of
-    # its descendant devices (see #{all_devices_from_device}) matches with the kernel name of the
-    # device included in the URI of an installation repository. Note that the device name indicated
-    # in the URI of a repository might not be a kernel name. The devices pointed by a repository are
-    # found by all its possible names, see #{repositories_devices}.
+    # A device contains an installation repository if the device or any of its descendant devices
+    # is included in the list of devices from the installation repository URI. For example, if the
+    # URI is "hd:/subdir?device=/dev/sda1", then "/dev/sda" contains an installation repository.
     #
     # @param device [BlkDevice]
     # @return [Boolean]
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/disk_analyzer.rb
```diff
@@ -357,13 +341,28 @@ def all_devices_from_device(device)
...
     #
     # @see #local_repositories
     #
     # @return [Array<String>]
-    def repositories_devices
-      @repositories_devices ||= local_repositories.map { |r| repository_devices(r) }.flatten
+    def repositories_device_names
+      return @repositories_device_names if @repositories_device_names
+
+      names = local_repositories.map { |r| repository_device_names(r) }.flatten.uniq
```
ancorgs, line 363: Being nitpicking, maybe flat_map is more efficient (I don't know if it really makes any difference) https://ruby-doc.org/core-3.0.2/Enumerable.html#method-i-flat_map

src/lib/y2storage/disk_analyzer.rb
```diff
@@ -315,34 +315,18 @@ def mounted_filesystem?(filesystem)
 
     # Checks whether a device contains an installation repository
     #
-    # For all possible names of the given device, it is checked if any of that
-    # names is included in the URI of an installation repository (see
-    # {#repositories_devices}). Note that the names of all devices inside the
-    # given device are considered as names of the given device (see #{device_names}),
-    # (e.g., when a disk contains a partition being used as LVM PV, the names of the
-    # LVM LVs are considered as names of the disk).
+    # A device contains an installation repository if its kernel name or the kernel name of any of
```
ancorgs, line 318: I find this documentation a bit confusing. It mentions several times the kernel names to later say that the URI can use any other kind of name. It looks to me like you are trying to describe more the internal implementation (converting everything to kernel names and then comparing) than the real effect.

ancorgs (approved):
LGTM

PR #1230, opened 2021-09-30
Fix github actions 

joseivanlopez:
## Problem

Github actions started failing because the TW docker container uses a new syscall that is not included in the whitelist of the docker version installed in the host system (ubuntu):

~~~
 /__e/node12/bin/node[18]: ../src/node_platform.cc:63:std::unique_ptr<long unsigned int> node::WorkerThreadsTaskRunner::DelayedTaskScheduler::Start(): Assertion `(0) == (uv_thread_create(t.get(), start_thread, this))' failed.
~~~

This issue would be automatically fixed if docker is updated in the host, but that is something we cannot control.

## Solution

Run the container with privileged mode. This ensures the syscall is directly passed to the host system.


ancorgs (approved):
LGTM

teclator (approved):
LGTM

PR #1231, opened 2021-09-30
Recommends libyui-qt-graph

joseivanlopez:
## Problem

The menu option "View --> Device Graphs" is not shown if *libyui-qt-graph* package is not installed. yast2-storage-ng should recommend such a package. 

* https://bugzilla.suse.com/show_bug.cgi?id=1191109 

## Solution

Recommends to install *libyui-qt-graph*, but only if *libyui-qt* is already installed.

* See https://documentation.suse.com/sbp/all/single-html/SBP-RPM-Packaging/index.html#sec-weak-dependencies


teclator (approved):
LGTM

PR #1237, opened 2021-10-19
Better management of Encryption#mount_by (jsc#SLE-20416, bsc#1181196)

ancorgs:
## Problem

The following pull request was merged into libstorage-ng in order to implement [jsc#SLE-20416](https://jira.suse.com/browse/SLE-20416) and to fix [bsc#1181196](https://bugzilla.suse.com/show_bug.cgi?id=1181196) (the bug report that originated the feature request).

https://github.com/openSUSE/libstorage-ng/pull/838

That change makes libstorage-ng use the device name by default to reference LUKS devices in the `fstab` file (instead of the global default `mount_by` method). Of course, the change does not affect the `crypttab` file, in which the default method is still used.

Unfortunately, this other pull request introduced some time ago into yast-storage-ng makes impossible to have different `mount_by` methods for the `crypttab` and the `fstab` of the same device: https://github.com/yast/yast-storage-ng/pull/1095. Having different values is a requisite for the scenario described at jsc#SLE-20416 (is is desirable in many other reasonable situations).

That pull request fixed [bsc#1165702](https://bugzilla.suse.com/show_bug.cgi?id=1165702) (which was the result of libstorage-ng not applying the default `mount_by` when probing existing encryption devices that are not in the system's `/etc/crypttab`) but in a wrong way.

## Solution

- Remove the old solution of syncing `Encryption#mount_by` with `Mountable#mount_by` (it never made any sense, since they are different concepts), so both attributes can now have different values.
- Introduce an alternative fix for the original bsc#1165702, setting the value of `Encryption#mount_by` to the global default value.
- Adapted the unit tests to the new result produced by libstorage-ng after the changed introduced for jsc#SLE-20416


src/lib/y2storage/encryption.rb
```diff
@@ -440,6 +468,19 @@ def mount_point_to_dm_name
...
 
+    # Checks whether the initial value of mount_by was forced to DEVICE by libstorage-ng due to lack
+    # of information or whether is the result of a proper initialization during normal operation
+    #
+    # @see #adjust_mount_by
+    #
+    # @return [Boolean]
+    def probed_without_crypttab?
+      return false unless exists_in_probed?
+      return !in_etc_crypttab? if in_etc_initial.nil?
```
joseivanlopez, line 479: What happens when `#assing_etc_attribute` is called before setting `in_etc_initial`?

PR #1239, opened 2021-10-28
Drop support for legacy format of ProposalSettings

ancorgs:
## Problem

The partitioning proposal can be configured via the control file as explained here: https://github.com/yast/yast-storage-ng/blob/master/doc/old_and_new_proposal.md

During the development of yast-storage-ng, apart from honoring such a configuration, we also added the ability to read proposal configurations written in the old (and way less flexible) format that was understood by the old yast-storage.

No known product/role is using that legacy format. It likely was never used in any released distribution, not even Leap 15.0 or SLE-15.

Still, we have a lot of code just to support that obsolete format.

## Solution

Remove all support for the legacy format in `ProposalSettings`. Moreover, removed all the code that was needed only to support several formats. Now everything is simpler.

## Testing

- Adapted unit tests
- Tested manually with the `proposal_testing` client
- Test coverage has decreased because we now have less code and the removed code was very well covered (even better than the current average). But coverage has not decreased on any of the kept files.


test/data/devicegraphs/output/empty_hard_disk_gpt_25GiB-sep-home.yml
```diff
@@ -10,7 +10,7 @@
         name:         /dev/sda1
         id:           bios_boot
     - partition:
-        size:         15561 MiB (15.20 GiB)
+        size:         15562 MiB (15.20 GiB)
```
joseivanlopez, line 13: why?

joseivanlopez (approved):
LGTM

test/data/devicegraphs/output/empty_hard_disk_gpt_25GiB-sep-home.yml
ancorgs: Glad you asked :-)

Is due to a difference in how the legacy and the NG devices planners used to add the partitions required for booting.

The legacy one performed the calculation taking into account only the root partition as you can see here: https://github.com/yast/yast-storage-ng/blob/381b525/src/lib/y2storage/proposal/devices_planner_strategies/legacy.rb#L56

While the former NG planner (now the only one) takes all existing planned partitions (at that point in time) into account: 
https://github.com/yast/yast-storage-ng/blob/2315bb6/src/lib/y2storage/proposal/devices_planner_strategies/ng.rb#L40

Shouldn't the result be the same anyways? Yes, basically it is... except for a tiny detail. The weight of the PlannedPartition for booting is calculated as the sum of the weights of all other known planned partitions. In the case of the legacy strategy that means the weight of the boot-related planned partitions is always equal to the root one. In the case of NG it can be different.

Is that weight really relevant if the boot-related partitions have almost-fixed sizes anyways? It does not affect the size of the boot-related partitions but it can deviate a little bit (only 1 MiB in only one of the existing tests) how the space is distributed between "/" and "/home".

PR #1240, opened 2021-11-02
Move the check for SD cards to a separate method

ancorgs:
## Problem

Commit 41d4ae7a98824290b5 introduced a check to know whether a disk is an SD/MMC card. But it did it as part of the method `#types_for_is`. That implies such circumstance is checked too often. Not only when the caller wants to know whether the disk is an SD card... but basically every time the nature of the device is checked with any purpose.

The information about the disk being an SD card or not is not present in the devicegraph, it needs an extra check with hwinfo. During normal operation, the result of the calls to hwinfo are cached, so it doesn't have a big impact. But in the testsuite, where nothing is cached, the impact is huge. It multiplies x5 the execution time of the testsuite.

## Solution

Move the check to a separate method, so the hwinfo call is only executed when the caller really wants to know if this an SD Card or not.

## Testing

Adjusted unit tests (they run in ~4 minutes again on my machine).

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Nov  2 15:46:24 UTC 2021 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Change the API to avoid the performance penalty introduced by
+  the original check for SD Cards (related to bsc#1187438).
+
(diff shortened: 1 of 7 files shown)


joseivanlopez:
> ## Problem
> 
> This commit introduced a check

What commit? Please, link it to have it at hand.

joseivanlopez (approved):
LGTM

PR #1247, opened 2021-11-16
[SLE-15-SP1] Fix duplicate PV error detection with disabled multipath

dgdavid:
## Problem

The installer has lost a specific error message when the same LVM **P**hysical **V**olume is found in several devices, which might means there are multipath devices in the system but multipath support is disabled. See screenshots below:

| SLE-15-GA | SLE-15-SP1 | openSUE TW |
| - | - | - |
| sle-15-ga-lvm-failed | sle-15-sp1-lvm-failed | openSUSE-TW-lvm-failed |
| lvm2 2.02.177-5.11.x86_64 | lvm2 2.02.180-10.16.x86_64 | lvm2 2.03.12-3.2.x86_64 |

* https://trello.com/c/jFw6d1l4/ (internal link)

## Why?

Because the lvm2 `"PV %s on %s was already found on %s."` message error used to catch the problem is not available anymore

* it was introduced [on May 6, 2016](https://github.com/lvmteam/lvm2/commit/8b7a78c728be3b9af698dae9344d01752c4cf615)  (lvm2 2.02.153) as part of the improvements of duplicate PVs handling
* **[on May 21, 2018](https://github.com/lvmteam/lvm2/commit/3c9ed33f83c90aa15e57ba6dc12d8f1a80afab6d) (lvm2 2.02.178) it changed from _warning_ to _debug cache_**. See [L2306-R2403](https://github.com/lvmteam/lvm2/commit/3c9ed33f83c90aa15e57ba6dc12d8f1a80afab6d#diff-7d86c19b8443c191b1e6bd52aec69a3a797b45c8e6a1757e2fd1324fbced42a7L2306-R2403)
* Finally, on [Aug 16, 2019](https://github.com/lvmteam/lvm2/commit/e01fddc5789b10d1bd15e8aee8985f14e81d3087) (lvm2 2.03.06) it was replaced by `Saving initial duplicate device %s previously seen on %s with PVID %s."` (but still being not a warning).

## Solution

Changing the [expected message to detect the situation](https://github.com/yast/yast-storage-ng/blob/7f15304a0b7226804f3570d8db58f20bf71af7d4/src/lib/y2storage/callbacks/libstorage_callback.rb#L121). 

[...]

PR #1250, opened 2021-11-17
[SLE-15-SP3] Fix duplicate PV error detection with disabled multipath

dgdavid:
## Problem

> The installer has lost a specific error message when the same LVM Physical Volume is found in several devices, which might means there are multipath devices in the system but multipath support is disabled

## Solution

See https://github.com/yast/yast-storage-ng/pull/1247 for reading the full story.

## Manual test

* Tested manually via driver update (SLE-15-SP3)

| Error | Details |
|-|-|
| sle15sp3-multipath-pv-fixed-dud | sle15sp3-multipath-pv-details-fixed-dud |


ancorgs (approved):
LGTM

PR #1251, opened 2021-11-18
Fix duplicate PV error detection with disabled multipath

dgdavid:
## Problem

> The installer has lost a specific error message when the same LVM Physical Volume is found in several devices, which might means there are multipath devices in the system but multipath support is disabled

## Solution

See https://github.com/yast/yast-storage-ng/pull/1247 for reading the full story.


ancorgs:
Please, rollback the version bump until we have the LUKS2 support a little bit more polished (maybe even today)

ancorgs (approved):
LGTM

PR #1252, opened 2021-11-18
Document and fix BootRequirementsChecker for LUKS2

ancorgs:
## Problem

`BootRequirementsChecker` was already adapted to deal with LUKS2 encryption at https://github.com/yast/yast-storage-ng/pull/964.

But the [boot-requirements.md](https://github.com/yast/yast-storage-ng/blob/master/doc/boot-requirements.md) document we always use to reflect how the proposals (both the Guided Proposal and the AutoYaST one) work was only updated in a couple of not-so-visible sections back then. Due to that, we even overlooked that `BootRequirementsChecker` already supported LUKS2-based scenarios.

On the other hand, while checking that everything worked as expected, I found that `BootRequirementsChecker` assumed that both the guided proposal and the AutoYaST one used always LUKS1. That is currently true for the guided proposal but not for the AutoYaST one. The AutoYaST property `<crypt_method>` can be used to specify, for example, than the root partition should be encrypted using IBM's pervasive volume encryption (that is based on LUKS2). See [AutoYaST documentation](https://documentation.suse.com/sles/15-SP3/html/SLES-all/cha-configuration-installation-options.html#ay-partition-configuration).

## Solution

- Extended `BootRequirementsChecker` unit tests to cover LUKS2 scenarios in all architectures and boot systems.
- Document `doc/boot-requirements.md` updated accordingly
- Fixed detection of planned LUKS2 encryption

## Testing

Unit tests extended

### DIFF
--- a/doc/boot-requirements.md
+++ b/doc/boot-requirements.md
@@ -281,12 +281,12 @@
 					- **requires a new GRUB partition**
 		- with an AutoYaST profile that places '/' in a LUKS2 device
 			- if there is no GRUB partition
-				- **requires new GRUB and /boot partitions (Grub2 auto-config cannot handle LUKS2**
+				- **requires new GRUB and /boot partitions (Grub2 auto-config cannot handle LUKS2)**
 			- if there is already a GRUB partition
 				- and it is on the boot disk
 					- **requires a new /boot partition (Grub2 auto-config cannot handle LUKS2)**
 				- and it is not on the boot disk
-					- **requires new GRUB and /boot partitions (Grub2 auto-config cannot handle LUKS2**
+					- **requires new GRUB and /boot partitions (Grub2 auto-config cannot handle LUKS2)**
 	- with a MS-DOS partition table
 		- if the MBR gap is big enough to embed Grub
 			- in a partitions-based proposal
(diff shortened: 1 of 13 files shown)


doc/boot-requirements.md
```diff
@@ -263,6 +279,14 @@
 					- **does not require any particular volume**
 				- and it is not on the boot disk
 					- **requires a new GRUB partition**
+		- with an AutoYaST profile that places '/' in a LUKS2 device
+			- if there is no GRUB partition
+				- **requires new GRUB and /boot partitions (Grub2 auto-config cannot handle LUKS2**
```
dgdavid, line 284: NP: Missing parenthesis

```suggestion
				- **requires new GRUB and /boot partitions (Grub2 auto-config cannot handle LUKS2)**
```

doc/boot-requirements.md
```diff
@@ -263,6 +279,14 @@
...
+			- if there is already a GRUB partition
+				- and it is on the boot disk
+					- **requires a new /boot partition (Grub2 auto-config cannot handle LUKS2)**
+				- and it is not on the boot disk
+					- **requires new GRUB and /boot partitions (Grub2 auto-config cannot handle LUKS2**
```
dgdavid, line 289:
```suggestion
					- **requires new GRUB and /boot partitions (Grub2 auto-config cannot handle LUKS2)**
```

dgdavid (approved):
LGTM, just two missing parenthesis.

dgdavid (approved):
LGTM

PR #1254, opened 2021-11-29
Adjust _netdev only in the GuidedProposal (jsc#SLE-20535)

ancorgs:
## Problem

YaST proposes the mount option "_netdev" in those scenarios when it's needed for the system to boot properly. But that's a task that comes with some challenges:

- If the "_netdev" option is needed but omitted, the system will have problems to boot.
- If the "_netdev" option is added to a mount point for which is not needed, the system will have problems to boot.
- The criteria to determine if "_netdev" is needed depends on:
  - the mount point path
  - the driver used by the device(s) supporting such a mount point
  - which other mount points are based on the same device(s)
  - what dracut is going to do regarding initrd generation

So far, YaST used to add "_netdev" if needed to the list of default options proposed for every new mount point. This is a problem because that decision is not revisited later, when the mount point is modified by the user or when other mount points are defined or deleted. For example, imagine we plan a mount point for "/var" in sda2 (with sda being one remote disk that uses one of the affected drivers). In that moment, YaST will add "_netdev" to its mount options since there is no evidence of that partition being in the root disk. If at a later point, sda3 is created and mounted as "/", the mount options of sda2 will not be re-adjusted.

See https://jira.suse.com/browse/SLE-20535

## Solution

This pull request changes two things (each one in a separate commit for easier review):

- It stops suggesting the "_netdev" option at such an early point in the workflow. Now the "_netdev" option is only added if the Guided Proposal is used and only at the very end of such a process, when all mount points are fully defined. That means that users modifying the resulting proposal or setting up the system manually are completely on their own, they must manually adjust the mount options.
- It refines the criteria for adding/removing the "_netdev" option for a given mount point, by also having the relationship with other mount points into account.

## Testing

Added new unit tests

### DIFF
--- a/src/lib/y2storage/blk_device.rb
+++ b/src/lib/y2storage/blk_device.rb
@@ -722,6 +722,14 @@ module Y2Storage
       end
     end
 
+    # Whether this device contains, directly or indirectly, a mount point returning true for the
+    # given block
+    #
+    # @return [Boolean]
+    def contain_mount_point?(&block)
+      descendants.any? { |desc| desc.is_a?(MountPoint) && block.call(desc) }
+    end
+
     # Whether the block device fulfills conditions to be used for a Windows system
     #
     # Note that only partitions can be used for installing Windows.
--- a/src/lib/y2storage/devicegraph.rb
+++ b/src/lib/y2storage/devicegraph.rb
@@ -318,6 +318,13 @@ module Y2Storage
       Filesystems::Base.all(self)
     end
 
+    # All mount points in the devicegraph, in no particular order
+    #
+    # @return [Array<MountPoint>]
+    def mount_points
+      MountPoint.all(self)
+    end
+
     # @param mountpoint [String] mountpoint of the filesystem (e.g. "/").
     # @return [Boolean]
     def filesystem_in_network?(mountpoint)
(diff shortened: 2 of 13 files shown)


src/lib/y2storage/devicegraph.rb
```diff
@@ -303,6 +303,13 @@ def filesystems
       Filesystems::Base.all(self)
     end
 
+    # All mount points in the devicegraph, in no particular order
+    #
+    # @return [Array<MountPoint>]
+    def mount_points
```
dgdavid, line 309: NP: This method does not have a unit test.

src/lib/y2storage/blk_device.rb
```diff
@@ -722,6 +722,14 @@ def systemd_remote?
       end
     end
 
+    # Whether this device contains, directly or indirectly, a mount point returning true for the
+    # given block
+    #
+    # @return [Boolean]
+    def contain_mount_point?(&block)
```
dgdavid, line 729: NP: This method does not have a unit test.

dgdavid (commented):
Do not forget to bump version and update changelog.

src/lib/y2storage/devicegraph.rb
ancorgs: Added. More than 40 lines of testcase for literally a one-liner.

dgdavid (approved):
LGTM

PR #1256, opened 2021-12-01
use libstorage-ng to determine whether efibootmgr is available (bsc#937067)

wfeldt:
## Problem

- https://trello.com/c/ZWMQnKom
- https://bugzilla.suse.com/show_bug.cgi?id=937067

EFI detection code is all over the place. Consolidate in one spot. As libstorage-ng is using this as well - rely on libstorage-ng.

## Solution

- add `Y2Storage::Arch.efibootmgr?` method
- `Y2Storage::Arch.efiboot?` takes `LIBSTORAGE_EFI` and `/etc/install.inf::efi` settings into account
- `Y2Storage::StorageManager.arch` is initialized  taking `LIBSTORAGE_EFI` and `/etc/install.inf::efi` settings into account

## Note

`#efibootmgr?` and `#efiboot?` are independent and do not rely on each other (`#efiboot? == false` does not imply `#efibootmgr? == false`, for example). This is how it works in libstorage-ng and the sematics are kept here.

## See also

- https://github.com/openSUSE/libstorage-ng/pull/846

### DIFF
--- a/src/lib/y2storage/arch.rb
+++ b/src/lib/y2storage/arch.rb
@@ -23,11 +23,29 @@ require "y2storage/storage_class_wrapper"
 module Y2Storage
   # Hardware architecture
   #
-  # This is a wrapper for Storage::Arch
+  # This is a wrapper for Storage::Arch.
   class Arch
     include StorageClassWrapper
 
     wrap_class Storage::Arch
+
+    # Wraps the Storage::Arch object passed to it or creates a new
+    # Storage::Arch object and wraps that.
+    #
+    # This also adjusts {#efiboot?} according to the `/etc/install.inf::EFI` setting.
+    #
+    # @param storage_arch [Storage::Arch] Storage::Arch object to wrap.
+    #
+    # @return [Y2Storage::Arch]
+    def initialize(storage_arch = Storage::Arch.new)
+      super(storage_arch)
+
+      Yast.import "Linuxrc"
+
+      return if Yast::Linuxrc.InstallInf("EFI").nil?
+
+      storage_arch.efiboot = Yast::Linuxrc.InstallInf("EFI") == "1"
+    end
 
     # @!method x86?
     #   @return [Boolean] whether the architecture is x86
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/arch.rb
```diff
@@ -53,6 +57,46 @@ class Arch
...
+
+    # Wraps the Storage::Arch object passed to it or creates a new
+    # Storage::Arch object and wraps that.
+    #
+    # This also adjusts {#efiboot?} according to the `/etc/install.inf::EFI` setting.
+    #
+    # @param storage_arch [Storage::Arch] Storage::Arch object to wrap.
+    #
+    # @return [Y2Storage::Arch]
+    def self.create(storage_arch = Storage::Arch.new)
```
joseivanlopez, line 90: NP: I would avoid having both: `#new` and `#create`. The only difference is whether to adjust *efiboot* or not. What about overloading the constructor?

~~~
def initialize(storage_arch = Storage::Arch.new, adjust_efiboot: true)
  super(storage_arch)

  return unless adjust_efiboot
  return uless Yast::Linuxrc.InstallInf("EFI")
  
  storage_arch.efiboot = Yast::Linuxrc.InstallInf("EFI") == "1"  
end
~~~

test/y2storage/arch_test.rb
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env rspec
+# Copyright (c) [2017] SUSE LLC
```
joseivanlopez, line 2: Please, adjust year.

src/lib/y2storage/arch.rb
wfeldt: There's no point in making the config adjustment optional. You'd just forcibly allow the `efi` boot setting and `LIBSTORAGE_EFI` to diverge. And that's what we want to avoid.

src/lib/y2storage/arch.rb
```diff
@@ -53,6 +57,46 @@ class Arch
     #   @return [Integer] the system page size
     storage_forward :page_size
 
+    # @!method self.efibootmgr?
+    #
+    # @return [Boolean] whether the UEFI boot manager can write UEFI boot entries
+    storage_class_forward :efibootmgr?
```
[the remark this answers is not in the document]
wfeldt: Yeah, I wasn't sure here.

joseivanlopez (approved):
LGTM

PR #1257, opened 2021-12-01
adapt calling named parameters to ruby3

jreidinger:
Adapt test suite for ruby3


dgdavid (commented):
I miss an specific explanation about the change itself. Is it because of https://rubyreferences.github.io/rubychanges/3.0.html#keyword-arguments-are-now-fully-separated-from-positional-arguments?

jreidinger:
@dgdavid yes, it is the reason. I plan to write summary email when I finish with all changes.

dgdavid (approved):
LTGM

PR #1258, opened 2021-12-02
Unify the Yes/No popup button IDs (bsc#1193326)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1193326
- Let's use the same button IDs in all Yes/No popups to make testing via libyui REST API easier

### DIFF
--- a/src/lib/y2partitioner/confirm_recursive_delete.rb
+++ b/src/lib/y2partitioner/confirm_recursive_delete.rb
@@ -39,7 +39,7 @@ module Y2Partitioner
       button_box = ButtonBox(
         PushButton(Id(:yes), Opt(:okButton), recursive_delete_yes_label),
         PushButton(
-          Id(:no_button),
+          Id(:no),
           Opt(:default, :cancelButton),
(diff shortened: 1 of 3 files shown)


src/lib/y2partitioner/confirm_recursive_delete.rb
```diff
@@ -39,7 +39,7 @@ def confirm_recursive_delete(devices, headline, label_before, label_after)
       button_box = ButtonBox(
         PushButton(Id(:yes), Opt(:okButton), recursive_delete_yes_label),
         PushButton(
-          Id(:no_button),
+          Id(:no),
           Opt(:default, :cancelButton),
```
jreidinger, line 43: ugh, there are no cancelButton, not?

jreidinger: ah, now I get options, it is not cancel as cancel but to pretend cancel behavior :angry:

jreidinger (approved):

PR #1248, opened 2021-11-17
Better probing errors handling

joseivanlopez:
## Problem

When the system is probed, errors are immediately notified to the user. That could be very annoying when there are multiple errors. The user has to accept each error individually. 

* https://jira.suse.com/browse/SLE-18430
* https://trello.com/c/HIIdD5gN/2736-5-better-handling-of-libstorage-ng-probing-errors

## Solution

All errors are collected and presented to the user in a single popup after probing. The *Details* buttons shows the details about all the issues in a new "fancy" way. Moreover, the Expert Partitioner adds a new menu option (*View -> System Issues*) to inspect the errors at any time.

Note: if there is only one probing issue, then the issue is reported just as before.

## Testing

- Adapted existing unit tests
- Added new unit tests
- Tested manually in a running system, during a normal installation, and during an auto installation (both, graphical and text mode)

## Screenshots

|| The issues warning | The issues report |
|-|-|-|
|Qt | installation_issues_warning | installation_issues_report
| text<br/>mode | textmode_running_system_issues_warning | textmode_running_system_issues_report |

<details>
  <summary><strong>Click HERE to show/hide MORE screenshots</strong></summary>

 ---

**:warning: Below screenshots were taken before quite small layout adjustments. So, they are not 100% accurate (maybe 98-99%?)**

---

<p align="center">The new issues warning popup</p>

| | Installation | Auto installation |
|-|-|-|
| Qt | issues_warning | autoyast_warning |
| text<br/>mode | textmode_warning | ay_textmode_warning |
| text<br/>mode<br/>80x24 | tm_80x24_warning | ay_tm_80x24_warning |

<p align="center">The new Issues report during an interactive installation</p>

  | | Error 1 | Error 2 |
  | - | - | - |
  | Qt | issues_report_e1 | issues_report_e2 |
  | text<br/>mode | textmode_report_e1 | textmode_report_e2 |
  | text<br/>mode<br/>80x24 | tm_80x24_report_e1 | tm_80x24_report_e2 |

<p align="center">The new Issues report during an auto installation</p>

 | | Error 1 | Error 2 |
 | - | - | - |
 | Qt |  autoyast_report_error_1 | autoyast_report_error_2 |
 | text<br/>mode | ay_textmode_report_error_1 | ay_textmode_report_error_2 |
 |  text<br/>mode<br/>80x24 | ay_tm_80x24_report_e1 | ay_tm_80x24_report_e2 |

[...]

PR #1259, opened 2021-12-03
Fix tests

joseivanlopez:
## Problem

Unit tests are failing in OBS when the architecture is not x86. Failing after merging https://github.com/yast/yast-storage-ng/pull/1248.

The tests are failing now because the generation of Bcache probing errors were mocked in order to avoid to indicate the architecture for every test. But such a mock became obsolete with the last chages.

* https://build.opensuse.org/request/show/935417

## Solution

Fix regression in unit tests by properly mocking the generation of Bcache errors.

## Testing

* Tests adapted.


dgdavid (approved):
LGTM

ancorgs (approved):
LGTM

PR #1245, opened 2021-11-12
Initial support for creating Luks2 devices with the Partitioner

ancorgs:
## Problem

The YaST Partitioner can be used to setup several kinds of encryptions, like explained here: https://yast.opensuse.org/blog/2019/10/09/advanced-encryption-options-land-in-the-yast-partitioner/

But something like "Regular LUKS2" was missing. That was intentional because using LUKS2 comes with many challenges, as summarized in this comment: https://bugzilla.suse.com/show_bug.cgi?id=1185291#c1

Time has come to start offering the possibility of creating LUKS2 devices in YaST, but we want to start little by little. Making it available in the Partitioner looks like a reasonable first step.

## Solution

This implements some preliminary support for LUKS2 when encrypting a block device. Such support is only available if the environment variable `YAST_LUKS2_AVAILABLE` is set. So far, it allows to set the LUKS label (apart from the password, obviously) and the PBKDF (automatically selecting the proper value would be far from trivial for YaST).

luks2b

Additionally, now the Partitioner displays the PBKDF and the LUKS2 label in the details of the encrypted device (only if LUKS2 is used). That information is available for all LUKS2 devices, even if `YAST_LUKS2_AVAILABLE` is not set.

details

## Testing

- Tested manually.
  - In a running system via the `partitioner_testing` client
  - In a full installation in a virtual machine, using LUKS2 for the `/home` file-system.
  - Opportunity taken to check what's the acceptable length and set of characters for a LUKS label. It looks very robust. It accepts up to 45 characters (and safely ignores the rest if the provided label is longer) and it accepts basically any possible printable character (safely managing several kinds of quotes).
- Added unit tests for the new code

PR #1260, opened 2021-12-06
Added YaST Console checkbox for enabling LUKS2 support

lslezak:
## Problem

Activating the experimental LUKS2 support added in #1245 is quite tricky:
- You need to remember the quite long `YAST_LUKS2_AVAILABLE` variable name
- If you forget to set it before starting YaST then you have to reboot the installation and start again :thinking: 
- Once set you cannot change it later

## Solution

Add an checkbox in the YaST expert console, that makes the setting easily accessible and it can be changed even during installation.

This is a perfect use case why I introduced the YaST expert console.

## Testing

- Tested manually
- Use the `Ctrl+Alt+Shift+C` (in Qt) or `Ctrl+D Shift+C` (in ncurses) keyboard shortcut to activate the expert console

#### Console Option

luks2

#### How it Works

- First show the default encryption options
- Then activate the YaST expert console and enable LUKS2
- Show the encryption options again, now with the LUKS2 option available

luks2

## Notes

When we add some more options in the future we could add `[Configure Partitioner]` button and group the partitioner options in a separate subdialog like the network settings.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Dec 22 17:37:23 UTC 2021 - Ladislav Slezák <lslezak@suse.cz>
+
+- Display LUKS2 configuration checkbox in the installer console
+  (related to jsc#SLE-21308)
+- 4.4.29
+
+-------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


shundhammer (approved):
LGTM

dgdavid (commented):
LGTM too, but now version and changelog are in conflict.

dgdavid (approved):
LGTM, thanks!

PR #1261, opened 2021-12-08
Ruby3

jreidinger:
Follow up of adaptation for ruby3. Finally found what was issue with OpenStruct. It is changed how it defines methods and as result, super no longer works and returns always nil. So I use access with method[]. Verified in Staging:L

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Dec  8 10:01:35 UTC 2021 - Josef Reidinger <jreidinger@suse.com>
+
+- Prepare code for ruby3 - adapt openstruct usage (bsc#1193192)
+- 4.4.21
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


ancorgs (approved):
LGTM. Good catch

dgdavid:
@jreidinger, @ancorgs 

I'm not completely sure, but after reading here and there, I think this is related to changes done at https://github.com/ruby/ostruct/pull/15 (related to https://bugs.ruby-lang.org/issues/15409), where they also updated the CAVEATS section and [encourage stopping using OpenStruct](https://github.com/ruby/ostruct/blob/5aaef97c5c6beab027b634747836203eb5621b43/lib/ostruct.rb#L67-L108).

dgdavid:
> [encourage stopping using OpenStruct](https://github.com/ruby/ostruct/blob/5aaef97c5c6beab027b634747836203eb5621b43/lib/ostruct.rb#L67-L108).

So, we could start using something similar to https://github.com/sider/goodcheck-rules/pull/56

PR #1262, opened 2021-12-09
Unit test for reusing RAID filesystem

ancorgs:
## Problem

As reported at [bsc#1193450](https://bugzilla.suse.com/show_bug.cgi?id=1193450), it currently seems to be impossible to keep the original file-system when reusing an existing software RAID with AutoYaST. AutoYaST always formats the MD again.

Turns out that a long time ago (during the development of SP1), we introduced fixes to make it possible to reuse existing RAID devices. See https://github.com/yast/yast-storage-ng/pull/835. That pull request included unit tests that verified the device was indeed reused... surprisingly they also verified that the MD is always formatted even if `<format>false</format>` was specified in the profile. It looks to me like the implementation contains an error and we have wrong tests that actually verify the error is there!

## Solution

Extend the test to verify both the format=true and format=false case. At this point in time, the scenarios with format=false fail because the error in the code has not been fixed yet.


ancorgs:
Replaced by #1266 (which targets the correct branch)

PR #1265, opened 2021-12-13
Allow abort, even from standalone modules, when probing finds issues

dgdavid:
## Problem

When a standalone client (e.g., yast2-bootloader in a running system) performs the probing and find issues, the report introduced in https://github.com/yast/yast-storage-ng/pull/1248 is shown. However, the standalone module continues anyway when the user decides to abort.

* https://bugzilla.suse.com/show_bug.cgi?id=1193749
* https://trello.com/c/9YjFdL0a/2774-3-yast2-bootloader-and-maybe-others-does-not-abort-in-probing-issues (internal link)

<details>
<summary>Click to show/hide some examples</summary>

---

<table>
<tr>
<th>yast-bootloader</th>
<th>yast-rear</th>
</tr>
<tr>
<td>

https://user-images.githubusercontent.com/1691872/146373046-3339d758-2008-45d5-8d40-dadefd812648.mp4

</td>
<td>

https://user-images.githubusercontent.com/1691872/146373104-fda7cb28-bbff-4b9f-a8ed-0d72720ee3ed.mp4

</td>
</tr>
</table>
</details>

## Solution

After several discussions, below actions were considered the best solution by now

* to introduce a `#probe!` method for not handling exceptions while probing (`#probe` handles them and simply returns `true` or `false`)
* to change `#staging` and `#probed` methods for relying on _#probe_**_!_** instead of _#probe_
* to use [the already handled `Yast::AbortException`](https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/wfm.rb#L358) when the user decides to abort instead of continue when probing issues are found.

This solution helps to avoid making changes in other YaST modules asking for either, the _staging_ or _probed_ devicegraph. So it invalidates https://github.com/yast/yast-bootloader/pull/658.

## Tests

- Added a new unit tests
- Tested manually in an installed system (by running yast-bootloader, yast-kdump, and yast-rear)
- Tested manually during an openSUSE TW installation using `yupdate patch yast/yast-storage-ng raise-error-on-probing-issues` to patch the source
- Tested manually during an openSUSE TW auto-installation using `yupdate patch yast/yast-storage-ng raise-error-on-probing-issues` to patch the source and a minimal profile.

<details>
<summary>Click to show/hide above examples after changes</summary>

---

<table>
<tr>
<th>yast-bootloader</th>
<th>yast-rear</th>
</tr>
<tr>
<td>

https://user-images.githubusercontent.com/1691872/146374125-c10793d7-d492-4365-93d8-85693c48b968.mp4

</td>
<td>

https://user-images.githubusercontent.com/1691872/146374143-26002259-d8d4-4211-b84f-8354c9709744.mp4

</td>
</tr>
</table>
</details>

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Dec 15 18:48:46 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Add support for improving the handling if the user decides to
+  abort when reporting probing issues (bsc#1193749).
+- 4.4.27
+
+-------------------------------------------------------------------
 Wed Dec 15 16:28:02 UTC 2021 - Steffen Winterfeldt <snwint@suse.com>
 
 - Partitioner: nest thin logical volumes below their thin pools (related to
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,8 +1,9 @@
 -------------------------------------------------------------------
 Wed Dec 15 18:48:46 UTC 2021 - David Diaz <dgonzalez@suse.com>
 
-- Add support for improving the handling if the user decides to
-  abort when reporting probing issues (bsc#1193749).
+- Improve probing issues handling by raising exceptions.
+- Fix aborting from standalone modules when the user decides
+  to not continue (bsc#1193749).
 - 4.4.27
 
 -------------------------------------------------------------------
--- a/src/lib/y2storage/storage_manager.rb
+++ b/src/lib/y2storage/storage_manager.rb
@@ -148,7 +148,26 @@ module Y2Storage
       @storage.deactivate
     end
 
-    # Probes all storage devices.
+    # Probes all storage devices
+    #
+    # If this method returns false, #staging and #probed could be in bad state
+    # (or not be there at all) so they should not be trusted in the subsequent
+    # code.
+    #
+    # @see #probe!
+    #
+    # @param probe_callbacks [Callbacks::Probe, nil]
+    # @return [Boolean] whether probing was successful, false if libstorage-ng
+    #   found a problem and the corresponding callback returned false (i.e. it
+    #   was decided to abort due to the error)
+    def probe(probe_callbacks: nil)
+      probe!(probe_callbacks: probe_callbacks)
+      true
+    rescue Storage::Exception, Error
+      false
+    end
+
+    # Probes all storage devices
     #
     # Invalidates the probed and staging devicegraph. Real probing is
     # only performed when the instance is not for testing.
@@ -156,15 +175,10 @@ module Y2Storage
     # With the default probe callbacks, the errors reported by libstorage-ng are stored in the
     # {#probe_issues} list.
     #
-    # If this method returns false, #staging and #probed could be in bad state
-    # (or not be there at all) so they should not be trusted in the subsequent
-    # code.
+    # @raise [Storage::Exception, Y2Storage::Error] when probe fails
     #
     # @param probe_callbacks [Callbacks::Probe, nil]
-    # @return [Boolean] whether probing was successful, false if libstorage-ng
-    #   found a problem and the corresponding callback returned false (i.e. it
-    #   was decided to abort due to the error)
-    def probe(probe_callbacks: nil)
+    def probe!(probe_callbacks: nil)
       probe_callbacks ||= Callbacks::Probe.new
 
       begin
@@ -180,9 +194,8 @@ module Y2Storage
       probe_performed
       manage_probing_issues
       DumpManager.dump(@probed_graph)
-      true
-    rescue Storage::Exception, Error
-      false
+
+      nil
     end
 
     # Probed devicegraph, after sanitizing it (see {#manage_probing_issues})
@@ -230,14 +243,13 @@ module Y2Storage
 
     # Staging devicegraph
     #
-    # @raise [Yast::AbortException] when {#probe} returns false
     # @see #staging
+    #
+    # @raise [Storage::Exception, Y2Storage::Error] when probe fails
     #
     # @return [Devicegraph]
     def staging!
-      success = probed? ? true : probe
-
-      raise Yast::AbortException unless success
+      probe! unless probed?
 
       staging
     end
--- a/src/lib/y2storage/storage_manager.rb
+++ b/src/lib/y2storage/storage_manager.rb
@@ -163,7 +163,7 @@ module Y2Storage
     def probe(probe_callbacks: nil)
       probe!(probe_callbacks: probe_callbacks)
       true
-    rescue Storage::Exception, Error
+    rescue Storage::Exception, Yast::AbortException
       false
     end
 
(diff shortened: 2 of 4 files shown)


src/lib/y2storage/storage_manager.rb
```diff
@@ -228,6 +228,20 @@ def staging
       end
     end
 
+    # Staging devicegraph
+    #
+    # @raise [Yast::AbortException] when {#probe} returns false
```
joseivanlopez, line 233: NP: I would say "when probe fails".

src/lib/y2storage/storage_manager.rb
```diff
@@ -228,6 +228,20 @@ def staging
...
+    # @see #staging
+    #
+    # @return [Devicegraph]
+    def staging!
+      success = probed? ? true : probe
+
+      raise Yast::AbortException unless success
```
joseivanlopez, line 240: TBH, raising a `Yast::AbortException` is unexpected for me. Such an exception should be raised as the result of a user action (e.g., when the user clicks abort) or as result of certain action that makes YaST abort. But I would say that calling `StorageManager#staging!` is not the case. That is a decision the caller should take. I would prefer to raise some kind of storage exception, for example, `Y2Storage::Error` or `Y2Storage::ProbeError`.

joseivanlopez: For example, [here](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/storage_manager.rb#L612) a `Yast::AbortException` is raised only if the user decides to not recover from a `Storage::LockException`.

dgdavid: I see your point. The thing is that one of the reasons the code reach that lines is precisely because the user decided to abort (`#probe` -> `#manage_probing_issues`).

But sure, we can raise a `Y2Storage::Error` and raise (or not) a `Yast::AbortException` in the callers.

dgdavid: As agreed in an offline conversation,

* changed where the exception is raised
* added #probe!
* adapted #staging!

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Dec 15 18:48:46 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Add support for improving the handling if the user decides to
```
jreidinger, line 4: is it really just support or it do real fix to respect abort from user?

src/lib/y2storage/storage_manager.rb
```diff
@@ -148,23 +148,37 @@ def deactivate
...
+    #
     # @param probe_callbacks [Callbacks::Probe, nil]
     # @return [Boolean] whether probing was successful, false if libstorage-ng
     #   found a problem and the corresponding callback returned false (i.e. it
     #   was decided to abort due to the error)
     def probe(probe_callbacks: nil)
+      probe!(probe_callbacks: probe_callbacks)
+      true
+    rescue Storage::Exception, Yast::AbortException
+      false
```
jreidinger, line 167: I expect that if it is storage exception it will print to log some details, not?

src/lib/y2storage/storage_manager.rb
```diff
@@ -192,13 +205,11 @@ def probe(probe_callbacks: nil)
     #   some errors (e.g., incomplete LVM VGs). This probed devicegraph
     #   is the result of sanitizing the initial raw probed.
     #
-    # @see #raw_probed
+    # @raise [Storage::Exception, Yast::AbortException] when probe fails
```
jreidinger, line 208: ugh, that Storage::Exception is less expected here. At first it does not have global handler, so it can lead to many reports about internal error. I think we agreed only on AbortException with API change.

src/lib/y2storage/storage_manager.rb
```diff
@@ -220,10 +231,12 @@ def raw_probed
     #   returned by libstorage-ng. This staging is initialized from the sanitized
     #   probed devicegraph (see {#manage_probing_issues}).
     #
+    # @raise [Storage::Exception, Yast::AbortException] when probe fails
```
jreidinger, line 234: same here with Storage::Exception

src/lib/y2storage/storage_manager.rb
joseivanlopez: Before these changes, such `Storage::Exception` were being masked because `#probe` simple catch them. But, most likely, next calls to the storage API would fail. We were discussing it yesterday, and that `Storage::Exception` are legitimate errors that YaST does not know how to recover from.

package/yast2-storage-ng.changes
joseivanlopez: I would change the description. All this PR is about raising exception when probing the system.

test/y2storage/storage_manager_test.rb
```diff
@@ -630,7 +630,137 @@
     end
   end
 
+  describe "#staging" do
+    context "when system has not been probed yet" do
```
joseivanlopez, line 634: I think that *#staging* and *#probed* tests are repeating the *#probe!* tests. I would keep all the checks about the exceptions in the *#probe!* tests only. And for *#staging* and *#probed* I will simply ensure that `probed!` is used.

src/lib/y2storage/storage_manager.rb
```diff
@@ -163,7 +163,9 @@ def deactivate
     def probe(probe_callbacks: nil)
       probe!(probe_callbacks: probe_callbacks)
       true
-    rescue Storage::Exception, Yast::AbortException
+    rescue Storage::Exception, Yast::AbortException => e
```
joseivanlopez, line 166: I expect libstorage-ng already logs it. If so, I would avoid to log twice.

joseivanlopez (approved):
LGTM, but I would check if libstorage-ng already logs before raising the exception.

src/lib/y2storage/storage_manager.rb
dgdavid: Ok, I didn't know. Anyway, libstorage-ng probably logs something regarding to Storage::Exception but not about Yast::AbortException. 

No strong opinion about what to do, if keep logging error or reverting that change, honestly.

joseivanlopez: If libstorage-ng already logs, then I would also log the `Yast::AbortException` before raising it.

src/lib/y2storage/storage_manager.rb
jreidinger: well, if we do not know how to recover from it, would be option to just change it to AbortException?

joseivanlopez: In that case, I would show a popup with the abort button.

ancorgs: AbortException generally means the user has decided to abort. So I think it would be incorrect to turn any other unhandled exception into that.

ancorgs: Letting the unhandled `Storage::Exception` to raise instead of masking it into a `return false` is actually consistent with our usual approach of "fail early".

PR #1157, opened 2020-10-30
Merge the new Partitioner UI into the master branch

ancorgs:
We have used a separate branch called `partitioner-ui-02` to implement most of the changes described in the ["Agreed Plan" section of the partitioner_ui document](https://github.com/yast/yast-storage-ng/blob/master/doc/partitioner_ui.md#agreed-plan-so-far).

Now it's time to merge all that into the `master` branch, so it's finally committed to openSUSE Tumbleweed.

## What changes in this new interface?

- A new application menu has been introduced. All the possible Partitioner actions are now organized into its four menus (System, Add, Device and View), as it can be seen in the following screenshot and checked in the original pull request #1139.

menu

- The contextual actions below each table have been reduced to a few simple buttons for the most common options, instead of menu-buttons including all the possibilities. The less common actions would now have to be reached through the menu. See #1137, which includes screenshots before and after that change.

- More compact layout with less headings. See #1126 and #1141, although screenshots in those pull request are not fully up-to-date (more things changed afterwards).

- Several views are now displayed as pop-ups. For example, the device details shown in the following screenshot (implemented as part of #1139), the list of bcache caching sets (see second screenshot at #1124), the "Settings" dialog, the "Device Graphs" view and the "Installation Summary" (all three extracted to pop-ups as part of #1122, see more screenshots there).

hard_disk

- Use nesting (with the possibility of expanding/collapsing) in the tables to better represent the relationship of the partitions with their corresponding block device and the relationship of the LVM logical volumes with their corresponding volume group. See #1151.

nested1b

- New default tab (called "Device Overview") to show the information of a device. It displays a table with the device itself as first element and all the dependent devices (eg. partitions) nested below. See #1141 (outdated screenshot) and #1151.

nested2b

- Unified the "Used Devices" tab (which is called "Physical Volumes" in the LVM case) for all kind of devices. Now that tab also displays a table with the device itself and its components as nested entries.

nested3

- Limited the nesting (and the labels) in the left tree to display only a first level of devices. Navigating below that level is not longer possible (eg. there is no "Device Overview" tab for an individual partition). See #1141 (outdated screenshot).

- Adapted double-click navigation. See #1147 for a description of the new behavior.

- Better wording in the "Type" column for the Bcache caching and backing devices. See #1143.

## Bonus

[...]

PR #1267, opened 2021-12-14
nest thin logical volumes below their thin pools

wfeldt:
## Task

- https://trello.com/c/fjkVmxRm

Nest thin logical volumes below their corresponding thin pools.

### Before

lvm1

### After

lvm2


src/lib/y2partitioner/widgets/device_table_entry.rb
```diff
@@ -163,6 +167,7 @@ def children(device)
             []
           end
         end
+        # rubocop:enable all
```
dgdavid, line 170: NP: Instead of disabling Rubocop, I'd extract the logic to small methods. Something like

```suggestion
        def children(device)
          return btrfs_subvolumes(device) if nesting_btrfs_subvolumes?(device)
          return lvm_lvs(device) if nesting_lvm_lvs?(device)
          
          partitions
       end
```

dgdavid (approved):
LGTM, although I'd split children method instead of disabling a Rubocop rule. Also, we could bump version and add an entry in the changelog using the same https://github.com/yast/yast-storage-ng/pull/1157/files#diff-360e4ecb172127044f3346e8e9bba17da22dcc895960b56f2f5bb5b07752818f references.

src/lib/y2partitioner/widgets/device_table_entry.rb
wfeldt: This will not help, as you'd end up with around the same number of branches and get the complexity warnings again.

You could hide the logic in the individual classes but I'm not sure there's a point.

wfeldt: If at all, then maybe something like

```
def children(device)
  btrfs_subvolumes(device) + lvm_lvs(device) + partitions(device)
end
```

dgdavid: I tested doing 

```ruby
        def children(device)
          return btrfs_subvolumes(device) if nesting_btrfs_subvolumes?(device)
          return lvm_lvs(device) if nesting_lvm_lvs?(device)

          partitions(device)
        end

        def nesting_lvm_lvs?(device)
          device.is?(:lvm_vg) || device.is?(lvm_lv)
        end

        def lvm_lvs(device)
          if device.is?(:lvm_vg)
            # All logical volumes, including thin pools but excluding thin volumes
            device.all_lvm_lvs - device.thin_lvm_lvs
          elsif device.is?(:lvm_lv)
            # Logical volumes in a thin pool, if any; empty for normal logical volumes
            device.lvm_lvs
          end
        end

        def partitions(device)
          if device.respond_to?(:partitions)
            # Logical partitions are not included because they are nested within the extended one
            device.partitions.reject { |p| p.type.is?(:logical) }
          elsif device.is?(:partition) && device.type.is?(:extended)
            # All logical partitions
            device.children
          else
            []
          end
        end

        ...
```

and Rubocop didn't complain. 

Anyway, it was an NP comment.

dgdavid: Aside note, 

> ```ruby
>           device.is?(:lvm_vg) || device.is?(:lvm_lv)
> ```

can be 

> ```ruby
>           device.is?(:lvm_vg, :lvm_lv)
> ```

wfeldt: My point was why an extra 'if' if you could just return an empty list and simply add up. That's just asking for 'if' and 'list' logic to diverge and getting surprising results.

dgdavid: > My point was why an extra 'if' if you could just return an empty list and simply add up. That's just asking for 'if' and 'list' logic to diverge and getting surprising results.

I overlooked your previous comment. Probably a browser cache problem. Fine with me.

dgdavid (approved):

dgdavid (approved):
LGTM, thanks

PR #1270, opened 2021-12-16
Improve documentation for PReP

ancorgs:
## Problem

In a PPC system with a disk that reports an optimal I/O size of 16 MiB, YaST will create a PReP partition of that size (creating an smaller one would imply a missaligned end). That's perfectly fine, since recent IBM's firmware can deal with "big" PReP partitions like that. But the firmware can be wrongly configured with some settings that limit the size of the PReP partition it accepts.

The problem was reported by some SUSE customers and the situation was clarified. The technical details and the procedure to configure the firmware were documented at [bsc#1192448](https://bugzilla.suse.com/show_bug.cgi?id=1192448), including the text of a TID (Technical Information Document) that will be soon be published through the usual SUSE channels.

## Solution

Add enough pointers in the boot requirements documentation so the relevant bug is easily found if the same problem pop-us up again. Not linking to the TID because is not yet published, but in fact the relevant information is already in the bug.

### DIFF
--- a/src/lib/y2storage/volume_specification_builder.rb
+++ b/src/lib/y2storage/volume_specification_builder.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2018] SUSE LLC
+# Copyright (c) [2018-2021] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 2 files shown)


dgdavid (approved):
LGTM

PR #1271, opened 2021-12-16
Improve LUKS activation

joseivanlopez:
## Problem

When there are several encrypted devices, the user is asked to provide the password for every one. This could be very annoying when the user is only interested on activating some of the encrypted devices.

* https://bugzilla.suse.com/show_bug.cgi?id=1162545

## Solution

Improve the *activate callback* used for requesting a LUKS password during the activation. The dialog was simplified (less text) and an option was added for skipping the activation of the rest of devices. This is a conservative change that does not cover all possible use cases:

* [Covered ] Many encrypted devices and the user wants to activate none.
* [Partially covered] Many encryption devices and the user only wants to activate some of them.
  * The user still has to skip devices one by one until reaching the devices on interest.
  * Possible improvement: use a boot option to indicate the devices to decrypt (e.g., YAST_OPEN_LUKS=/dev/sda2,/dev/sda3). 
* [Not covered] Many encrypted devices and the user wants to use the same password for all them.
  * Finding a good UI for this was challeging.
  * We decided to not support that scenario because it has not been requested yet. 

Note: The Expert Partitioner has a menu option to provide the encryption passwords. This was working only the first time the option is used because libstorage-ng was caching the activation information for the next calls. Now the cached information is reset when the Partitioner menu option is used.

## Testing

* Added new unit tests
* Tested manually

## Screenshots

<details>
<summary>Show/Hide</summary>

### Previous Dialogs

Screenshot from 2021-12-17 00-13-00

Screenshot from 2021-12-17 00-13-44

### New Dialogs

Screenshot from 2022-01-07 15-15-00

Screenshot from 2022-01-07 15-15-50

Screenshot from 2022-01-07 15-18-08

Screenshot from 2022-01-07 15-18-59

</details>

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,3 +1,13 @@
+
+-------------------------------------------------------------------
+Wed Jan 12 11:12:13 UTC 2022 - José Iván López González <jlopez@suse.com>
+
+- Allow to skip the activation of the rest of LUKS devices
+  (bsc#1162545).
+- Partitioner: always allow to provide password for closed LUKS
+  devices.
+- 4.4.32
+
 -------------------------------------------------------------------
 Wed Jan  5 21:38:56 UTC 2022 - Josef Reidinger <jreidinger@suse.com>
 
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -25,8 +25,8 @@ Url:            https://github.com/yast/yast-storage-ng
 
 Source:         %{name}-%{version}.tar.bz2
 
-# Storage::Arch.is_efibootmgr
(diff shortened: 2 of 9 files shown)


wfeldt (approved):
Much better than the old dialogs.

NP: can something be done about the bad formatting in the ncurses dialog?

joseivanlopez:
> NP: can something be done about the bad formatting in the ncurses dialog?

Yes. It is a known issue and we already have a card for it, see https://trello.com/c/lPXgke4B/4689-sles15-sp3-p3-1184784-the-new-yast2popup-removes-newline-character-in-ncurses.

BTW, I will wait for https://github.com/yast/yast-storage-ng/pull/1278 in order to ensure that rubocop is green before merging.

PR #1272, opened 2021-12-16
Warning about missing mount options (useful for _netdev)

ancorgs:
## Problem

Setting the `_netdev` mount option properly has shown to not be a trivial task. Thus, at #1254 we stopped adding it as a default option when a new mount point is created. Now the option is only added as the final step of the Guided Proposal. Users setting the system manually are at their own.

But we felt we could still do something for users that like to adjust stuff by themselves but that are used to YaST assisting them with that tricky mount option.

## Solution

This PR adds a new kind of warning to the already existing checks that are executed when the user clicks "finish" in the Expert Partitioner.

netdev_warn

## Review

Please check the current text (and also the alternatives commented in the code) and help me to present the warning in a way that is both understandable and generic (for future checks we could add for other mount options).


src/lib/y2storage/setup_checker.rb
```diff
@@ -93,6 +95,32 @@ def product_warnings
...
+
+    def mount_warning(mount_point)
+      missing = mount_point.missing_mount_options
+      return if missing.empty?
+
+      msg = n_(
+        format(
+          #"Adding the fstab option %{opt} may be needed for %{path}.",
+          "The fstab option %{opt} seems to be missing for %{path}.",
+          #"The option %{opt} may be needed in fstab for %{path} to be mounted properly.",
```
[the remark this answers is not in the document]
joseivanlopez: I would say: "the fstab option %{opt} may be needed for %{path} to be mounted properly". Basically option 3, but I would try to mention "fstab option" because the button for adding it is called so in the Expert Partitioner.

src/lib/y2storage/setup_checker.rb
```diff
@@ -153,6 +165,31 @@ def nfs?(volume)
...
+    def mount_warning(mount_point)
+      missing = mount_point.missing_mount_options
+      return if missing.empty?
+
+      # TRANSLATORS: do not translate %{opt} or %{path}. %{opt} is replaced by the name of a mount
+      # option in the singular sentence or a list of options separated by commas in the plural one.
+      # ${path} is replaced by the mount point.
+      msg = n_(
+        format(
+          "The fstab option %{opt} may be needed to properly mount %{path}.",
```
dgdavid, line 181: NP: Don't know why, but for me "The _netdev fstab option" _sounds_ better.

dgdavid (approved):
LGTM

src/lib/y2storage/setup_checker.rb
ancorgs: I don't have a strong preference, but I felt this way it would be more readable at first sight for certain option names.

PR #1274, opened 2021-12-21
Reorganize and extend clients for manual testing

ancorgs:
## Problem

We used to have two clients that where there just for the developers to do some manual testing of new features or to verify how YaST behaves in the scenario of a given bug report: 

 - `proposal_testing` to test the initial partitioning proposal made by the installer & the Guided Setup (pretty bad name, BTW, since it does not mention which of the dozens of YaST proposals it refers to)
 - `partitioner_testing` to test the Partitioner

Both clients contained duplicated code for the mocking part (ie. reading the probed devicegraph from a file).

I wanted to add just another client for manual testing. This time for manually reproducing the behavior of the AutoYaST partitioning proposal. Another separate client would have implied repeating the mocking code a third time.

## Solution

Unify all the _xxx_testing_ clients into a single one, keeping as much as possible the syntax of the original command lines. Thus:

- To test the Partitioner:
  - the old `yast2 partitioner_testing /path/to/devicegraph.xml`
  - becomes `yast2 storage_testing partitioner /path/to/devicegraph.xml`
- To test the regular proposal:
  - the old `yast2 proposal_testing /the/devicegraph.xml /the/control_file.xml`
  - becomes `yast2 storage_testing proposal /the/devicegraph.xml /the/control_file.xml`
- To test the AutoYaST proposal:
  - do `yast2 storage_testing autoinst /the/devicegraph.xml /the/profile.xml`

The first two work exactly like before. The third one looks like this:

image


src/lib/y2storage/clients/autoinst_manual_test.rb
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env ruby
+#
+# encoding: utf-8
+
+# Copyright (c) [2017-2020] SUSE LLC
```
dgdavid, line 5:
```suggestion
# Copyright (c) [2017-2021] SUSE LLC
```

src/lib/y2storage/clients/manual_test.rb
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env ruby
+#
+# encoding: utf-8
+
+# Copyright (c) [2017-2020] SUSE LLC
```
dgdavid, line 5:
```suggestion
# Copyright (c) [2017-2021] SUSE LLC
```

dgdavid (approved):
LGTM, thank you!

dgdavid (approved):
LGTM

PR #1275, opened 2021-12-23
Mock yast-installation classes and modules in tests

ancorgs:
## Problem

The manual testing client introduced at #1274 depends on yast2-installation being installed on the machine in order to execute the AutoYaST "simulation". That's perfectly fine since that client (specially when testing AutoYaST) is a developer-oriented tool.

But we don't want to introduce a hard dependency from yast2-storage-ng to yast2-installation.

On the other hand, we want the `ManualTest` client to be covered by some unit tests. Since those unit tests are executed while building the package... we have a chicken and egg problem.

## Solution

Mock all the components from yast2-installation used during the unit test for `Y2Storage::Clients::ManualTest`


dgdavid (approved):
LGTM, thank you!

Have a great Christmas :santa: :christmas_tree:

PR #1277, opened 2022-01-05
Issues reporter

jreidinger:
## Problem

Crash when reporting issue in TW due to new ruby.

- https://bugzilla.suse.com/show_bug.cgi?id=1194342

## Solution
Adapt code.

## Testing

- adapted unit test, so it failed without fix.


mvidner (approved):
LGTM

PR #1278, opened 2022-01-09
New rubocop

jreidinger:
## Problem

TW uses ruby 3 and our ancient rubocop does not work with it. So lets update rubocop to something more recent.

## Testing

- run unit tests


src/lib/y2storage/disk_analyzer.rb
```diff
@@ -374,7 +374,7 @@ def repositories_device_names
     # @param repository [Y2Packager::Repository]
     # @return [Array<String>]
     def repository_device_names(repository)
-      match_data = repository.url.to_s.match(/.*device[s]?=([^&]*)/)
+      match_data = repository.url.to_s.match(/.*devices?=([^&]*)/)
```
ancorgs, line 377: IMHO, the previous version was more explicit for eyes untrained in the perils of regexps, like mine.

.rubocop.yml
```diff
@@ -1,22 +1,34 @@
...
-Metrics/LineLength:
+Layout/LineLength:
   Max: 105
 
 Metrics/BlockLength:
   # rspec is known as DSL with big blocks
   Exclude:
     - test/**/*
 
+# TODO: fix, but for now too much work
```
ancorgs, line 20: Why exactly these three need to be disabled now? Is the new default limit more strict? Were the rules refined and now we have methods that rank higher than they used to rank? If so, how many?

src/lib/y2storage/actions_presenter.rb
```diff
@@ -110,8 +110,7 @@ def dump(file_base_name = nil)
 
     protected
 
-    attr_reader :actiongraph
-    attr_reader :collapsed_subvolumes
+    attr_reader :actiongraph, :collapsed_subvolumes
```
ancorgs, line 113: I don't see why the new version is better. Not sure if I like the rule.

Nevertheless, a better fix to calm Rubocop down (I guess) would be to document the attributes. Something like

```ruby
# @return [Actiongraph] set of actions to be presented
attr_reader :actiongraph
# @return [Boolean] whether the actions related to subvolumes are currently collapsed
attr_reader :collapsed_subvolumes
```

ancorgs (commented):
Generally I have no problem with the proposed changes. Some of them are clearly an improvement. But I have some questions (see comments).

src/lib/y2storage/disk_analyzer.rb
jreidinger: well, for me any regexp is write only and only way to verify its behavior is unit testing :)

.rubocop.yml
jreidinger: well, it counts more precisely.
mainly https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#changes-36 do better counting of conditions. and similar at https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#changes-39

So I think it is fixable, I just do not want to spend too much time in it. So if it is blocking I can check it and do refactoring.

src/lib/y2storage/actions_presenter.rb
jreidinger: agreed. This was auto correction. Not sure how to force documentation there. 
It is this cop https://msp-greg.github.io/rubocop/RuboCop/Cop/Style/AccessorGrouping.html
And it is also incompatible with sorbet and documentation, so maybe it makes sense to configure it to `separated` style. Unless someone is against.

ancorgs: I would simple disable the cop. I see no need to enforce either "grouped" or "separate". Depending on the case one style of declaration or the other may be better (although I personally prefer separate declarations).

.rubocop.yml
ancorgs: I see. Then I guess it makes sense to raise the limit. I just feel we raised them too much. The new default values are 7 and 17 and we raised ours to 12 and 32.

Let's say we keep the `rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity` directives that you deleted in some parts (those are methods somebody already considered not worth refactoring). Would we still need to go as far as 12 and 32?

jreidinger: I found that now biggest offender is https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/proposal_settings.rb#L378 as it counts each ||= as condition. I can try to refactor it, if you want. Btw also as it uses booleans and `||= true` looks quite bad from my POV.

ancorgs (approved):
LGTM. My only complain would be that the constants and methods introduced as part of the refactoring are generally not documented. That gets us away from the (unofficial) target of increasing `conf.documentation_minimal` beyond the current 91. :wink:

jreidinger:
@ancorgs documentation added :wink:

src/lib/y2storage/abstract_device_factory.rb
```diff
@@ -123,6 +123,42 @@ def build_tree_toplevel(obj)
       build_tree_recursive(nil, name, content)
     end
 
+    # Builds tree from hashe
```
ancorgs, line 126: Typo: hashe

src/lib/y2storage/planned/has_size.rb
```diff
@@ -213,6 +203,26 @@ def adjusted_size_after_ceil(device, space_size, align_grain)
         def total_weight(devices)
           devices.map(&:weight).reduce(0, :+)
         end
+
+        # Checks if there are eough space at all
```
ancorgs, line 207: "there are eough" => "there is enough"

src/lib/y2storage/planned/has_size.rb
```diff
@@ -213,6 +203,26 @@ def adjusted_size_after_ceil(device, space_size, align_grain)
...
+        # @raise RuntimeError if there is not enough space
+        def check_size(devices, space_size)
+          needed_size = DiskSize.sum(devices.map(&:min))
+          return unless space_size < needed_size
+
+          log.error "not enough space: needed #{needed_size}, available #{space_size}"
+          raise RuntimeError
+        end
+
+        # counts minimal size for each device
```
ancorgs, line 217: I wouldn't say this "counts" the minimal sizes. The name of the method is not that good. I would say it's more like:

```ruby 
# Sets sizes of the given devices to their minimal valid value
# @return [Array<Planned::Device>] modified devices with size adjusted to rounded minimum size
def init_sizes(device, rounding)
  [...]
end
```

src/lib/y2storage/proposal_settings.rb
```diff
@@ -367,12 +367,17 @@ def root_volume
...
+    # TODO: enum?
     DELETE_MODES = [:none, :all, :ondemand]
     private_constant :DELETE_MODES
 
+    # List of possible VG strategies.
+    # TODO: enum?
     LVM_VG_STRATEGIES = [:use_available, :use_needed, :use_vg_size]
     private_constant :LVM_VG_STRATEGIES
 
+    # Defaults when settings is not specified
```
ancorgs, line 380: Concordance error. What about "when a setting is not specified"?

ancorgs (commented):
Thanks for documentation, but I have some notes.

ancorgs (approved):
LGTM

PR #1282, opened 2022-01-26
Improve integration with nfs-client

joseivanlopez:
https://trello.com/c/R8S6omOH/2046-8-integrate-nfs-into-the-new-menu-based-workflow

PR #1283, opened 2022-02-02
Use `Y2NfsClient::Widgets::NfsForm` instead of `NfsClient4Part`

ancorgs:
## Problem

This pull request replaces #1282 (same content but with clear history)

Traditionally, the management of NFS shares in the YaST Partitioner has been a bit rough. The Partitioner never really provided its own UI to add/edit/delete NFS mounts. Instead, it executed an embedded version of `yast2-nfs-client`. That has always been problematic. For example, there were non-obvious problems when the same NFS share was mounted more than once in the current system in two different locations (admittedly a very uncommon scenario).

But the situation became much worse with the [revamp of the interface](https://github.com/yast/yast-storage-ng/pull/1157) done for SLE-15-SP3 (and openSUSE Leap 15.3). Since that moment, NFS management was completely inconsistent with the rest of the Partitioner interface:

 - The NFS mounts were not displayed in the left tree (like it happens with all other devices).
 - It was impossible to use the menu to access to the NFS-related actions.
 - The NFS section of the Partitioner just behaved in a completely different way compared to all other tables (different buttons ordered and activated in a different way, forms in pop-ups instead of full-screen dialogs, validations at incorrect points of the workflow...)

## Solution

This pull request goes together with https://github.com/yast/yast-nfs-client/pull/105

With this pull request and the corresponding yast-nfs-client one mentioned above, the integration between both YaST modules is done in a more sensible way. All the functionality is still available, but in a way in which the integration is completely transparent offering a consistent user experience.

## Testing

- Tested manually
- Adapted some new unit tests... **BUT** not as many as it would be desirable.

We want to merge this because of a SLE-15-SP4 beta even if the test coverage is below the standards of this repository. With the solid promise of fixing that tomorrow (literally).


joseivanlopez (approved):
LGTM

PR #1284, opened 2022-02-03
Tests for the recent code

ancorgs:
Continuation of #1283 (it adds the tests that were missing there)


test/y2partitioner/dialogs/nfs_test.rb
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env rspec
+# Copyright (c) [2020] SUSE LLC
```
dgdavid, line 2: NP: 2022 :P

dgdavid (approved):

dgdavid (commented):
Since changes are related only with unit testing, I guess version bump and changelog entry is not needed. Thus, LGTM.

PR #1286, opened 2022-02-10
Turn menu label Nfs into NFS

ancorgs:
Everywhere in the Partitioner the string "NFS" (everything uppercase) is used to refer the Network File System technology. But in the menu "Add", the string "Nfs" was used. See screenshot.

nfs

Although this is basically a cosmetic issue, it affects translations so we wanted to make sure the fix enters the SLE-15-SP4 repository before the translators do they work. Thus, the version of the package is increased even if all included changes are cosmetic and rather minimal.

I also edited the previous changelog message to make super-clear that is also a cosmetic and non-risky change.


joseivanlopez (approved):
LGTM

PR #1288, opened 2022-03-09
ensure no install media are mounted when storage proposal is run (bsc#1196061)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1196061
- https://trello.com/c/bC5yVZas

When the installation repository is on some local disk partition and the partition is included in the storage setup, the existing mountpoints will lead to confusion.

libstorage-ng

- has problems with existing mountpoints in probing data that no longer exist in the commit phase
- mixes up mountpoints as it does not know the difference between installation system (no prefix) and target system (`/mnt` prefix)
- may show popups and/or remove the mountpoint from the target system's fstab

## Solution

To avoid this, unmount all installation media before probing.

## See also

- https://github.com/openSUSE/linuxrc/issues/288

### DIFF
--- a/src/lib/y2storage/clients/inst_disk_proposal.rb
+++ b/src/lib/y2storage/clients/inst_disk_proposal.rb
@@ -50,6 +50,8 @@ module Y2Storage
       def initialize
         textdomain "storage"
 
+        detach_media_and_reprobe
+
         @devicegraph = storage_manager.staging
         @proposal = storage_manager.proposal
         # Save staging revision to check later if the system was reprobed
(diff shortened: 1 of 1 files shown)


lslezak (approved):
From the package manager POV LGTM

lslezak:
I'm just not sure about that reprobing, that's for some storage experts...

wfeldt:
This is a valid point but I didn't find a better solution.

The original probing happens in the middle of some really complex workflow (with add-ons, roles, full and online workflow) and it would make sense to not do it there at all but we need the storage data for other things. And also I'm not that sure what happens if we detach the repos at that point.

And even if we detach the repos at the first probing point people can still go for explicit reprobing in the expert partitioner and then we're in this mess again unless we at least detach the repos *again* at the proposal stage.

So I'd leave the first probing where it is to generate some storage data for whatever it is needed and at the proposal point detach the repo and probe again to get the state we need later at the commit phase.

ancorgs:
Reprobing also invalidates the staging devicegraph, throwing away all the changes done by the proposal or manually with the Expert Partitioner. And that reprobing would be silently executed every time we reach this point in the workflow. No matter if we are going forward or backwards.

I feel is too risky.

ancorgs (changes requested):
As written in my previous comment, I fair this would have too many side effects in cases that currently work. Is a too big change to be done under the hood.

wfeldt:
That leaves:

(1)
 - ensure media are detached when we do the original probing and
 - ensure media are detached at the proposal stage by reviewing our workflow

or (2)

- add a dedicated client that just does the media detaching and reprobing and insert it before the proposal; but that might raise issues as our workflow is highly configurable

PR #1290, opened 2022-03-31
Release sources before probing

joseivanlopez:
## Problem

An installation repo can be provided from a local disk by using *install=* boot option (e.g., *install=hd:/isos/tw.iso*). YaST mounts the device containing the local repository. If *libstorage-ng* has to unmount such a device (e.g., because the mount point of the repo device was edited in the Expert Partitioner), then the unmount action fails. Note that *libstorage-ng* would try to unmount the device by using a device path as the device would be mounted under the installation mount point (e.g., */mnt/var/adm/mount/AP_0xppFmPE*).

Screenshot_SLE-15-SP4_2022-03-31_12:14:25

## Solution

Always release all installation sources before probing. With that, *libstorage-ng* will never try to unmount the device. Note that YaST Packager will mount the source devices again under demand, for example, when the installation starts.  Also note that this is only needed during installation. In a running system there are no chroot paths involved.

NOTE: this PR will be merged into *pre-SLE-15-SP4* brach. Such a brach will be merged into master after branching *SLE-15-SP4*. The version should be bumped while merging *pre-SLE-15-SP4* into master.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Mar 31 12:37:01 UTC 2022 - José Iván López González <jlopez@suse.com>
+
+- Release sources before probing during installation to avoid
+  issues when unmounting devices (related to bsc#1196061).
+
+-------------------------------------------------------------------
 Thu Feb 10 10:18:23 UTC 2022 - Ancor Gonzalez Sosa <ancor@suse.com>
 
 - Partitioner: 'Add NFS' instead of 'Add Nfs' in the menu (related
--- a/src/lib/y2storage/storage_manager.rb
+++ b/src/lib/y2storage/storage_manager.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2015-2021] SUSE LLC
+# Copyright (c) [2015-2022] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -34,6 +34,7 @@ require "y2issues/list"
 
 Yast.import "Mode"
 Yast.import "Stage"
+Yast.import "Pkg"
 
 module Y2Storage
   # Singleton class to provide access to the libstorage Storage object and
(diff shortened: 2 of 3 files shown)


wfeldt:
IIRC the issue was that libstorage goes to mount the partition (as it's part of the proposal and has a mount point) - and there fails as it's already mounted.

joseivanlopez:
AFAIS in my tests, the issue happens when libstorage-ng has to unmount the mount point created for the source device (e.g., */var/adm/mount/AP_0xppFmPE*, let's call it "source mount point"). During the commit phase, libstorage-ng will try to unmount the device if the source mount point was modified. For example, editing or removing the mount path in the Expert Partitioner. To do that, libstorage-ng uses the source mount point prepended by the installation path (e.g., */mnt*). The unmount action fails because such a mount point does not exist.

Another option to make libstroage-ng to unmount the device is to delete (or format, re-use, etc) the device, but in this case unmounting is the least of our problems because the installation media would be missing.

So, relaeasing the sources before probing ensures that libstorage-ng will not try to unmount the source mount point. What scenarios can be present during the commit?

a) The source device is currently mounted in the system: libstorage-ng will not unmount it, but it will create a new mount point if needed (e.g., /mnt/data).
b) The source device is not mounted: libstorage-ng will not try to unmount either and it will mount the device if needed.
c) The source device was changed (formatted, removed): libstorage-ng will not try to unmount but YaST Package will not be able to find the sources.

wfeldt:
See a). In addition, libstorage will **mount** the sources at the mount point - and that fails as it's already mounted somewhere else.

joseivanlopez:
> See a). In addition, libstorage will **mount** the sources at the mount point - and that fails as it's already mounted somewhere else.

I have tested it, and libstorage-ng does not complain when the iso is mounted before the commit phase. Libstorage-ng mounts the device correctly and everything works as expected.

PR #1291, opened 2022-04-07
Allow quotes when matching fstab device entries

teclator:
## Problem

When an **fstab** first field entry contains quotes surrounding the **UUID** or **label** then the filesystem match fails and do not find any filesystem for the given device.

- https://bugzilla.suse.com/show_bug.cgi?id=1197692

## Solution

Modify the Regexp for matching device names using **UUID** or **labels** surrounded by quotes.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/src/lib/y2storage/filesystems/blk_filesystem.rb
+++ b/src/lib/y2storage/filesystems/blk_filesystem.rb
@@ -229,11 +229,11 @@ module Y2Storage
 
       # @see Filesystems::Base#match_fstab_spec?
       def match_fstab_spec?(spec)
-        if /^UUID=(.*)/ =~ spec
+        if /^UUID="(.*)"/ =~ spec || /^UUID='(.*)'/ =~ spec || /^UUID=(.*)/ =~ spec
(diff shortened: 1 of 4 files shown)


src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -229,11 +229,11 @@ def supports_resize?
 
       # @see Filesystems::Base#match_fstab_spec?
       def match_fstab_spec?(spec)
-        if /^UUID=(.*)/ =~ spec
+        if /^UUID="(.*)"/ =~ spec || /^UUID='(.*)'/ =~ spec || /^UUID=(.*)/ =~ spec
```
teclator, line 232: We could use also something like:

```suggestion
        if /^UUID=(?<open_quote>["'])(?<device_id>.*)(?<close_quote>["'])/ =~ spec
```

dgdavid (approved):

ancorgs (approved):
LGTM

PR #1292, opened 2022-04-08
Merge 'SLE-15-SP2' into 'SLE-15-SP3'

teclator:
It merges #1291 & #1293 into SLE-15-SP3

### DIFF
--- a/src/lib/y2storage/filesystems/blk_filesystem.rb
+++ b/src/lib/y2storage/filesystems/blk_filesystem.rb
@@ -271,12 +271,12 @@ module Y2Storage
 
       # @see Filesystems::Base#match_fstab_spec?
       def match_fstab_spec?(spec)
-        if /^UUID=(.*)/ =~ spec
-          return !Regexp.last_match(1).empty? && uuid == Regexp.last_match(1)
+        if /^UUID=(['"]?)(.*)\1$/ =~ spec
+          return !Regexp.last_match(2).empty? && uuid == Regexp.last_match(2)
         end
 
-        if /^LABEL=(.*)/ =~ spec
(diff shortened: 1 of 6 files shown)


src/lib/y2storage/filesystems/blk_filesystem.rb
```diff
@@ -406,6 +401,33 @@ def preferred_mount_by
...
+
+      # Convenience method for checking if the given STRING is a well formed label
+      #
+      # @param spec [String] content of the first column of an /etc/fstab entry
+      def label?(spec)
+        /^LABEL="(.*)"/ =~ spec || /^LABEL='(.*)'/ =~ spec || /^LABEL=(.*)/ =~ spec
+      end
+
+      # @return [Boolean]
+      def match_uuid?
```
ancorgs, line 423: I don't like the idea of using `#last_match` here if the matching actually happened in another method. This forces you to call `#match_uuid?` always right after calling `#uuid?` and only if that call succeeded. Too fragile.

teclator: Yep, have the same opinion

ancorgs: I would have expected something like this:

```ruby
def match_fstab_spec?(spec)
  return true if match_fstab_uuid?(spec)
  return true if match_fstab_label?(spec)

  named_device = devicegraph.find_by_any_name(spec)
  blk_devices.include?(named_device)
end
```

teclator: After modifying the Regexp we do not need any of these changes at all.

dgdavid (approved):

PR #1297, opened 2022-04-14
Fix empty help in some Partitioner dialogs (bsc#1194274)

mvidner:
## Problem

Some dialogs in the partitioner have a *Help* button that pops up an empty window.

- https://bugzilla.suse.com/show_bug.cgi?id=1194274
- https://trello.com/c/22Mos13d

## Solution

These fixes work via a method of a base class of all these dialogs, [Y2Partitioner::Dialogs::Popup#set_help_text](https://github.com/yast/yast-storage-ng/blob/SLE-15-SP4/src/lib/y2partitioner/dialogs/popup.rb#L100).

A better fix would be to use a common mechanism of CWM+Wizard:
Wizard.SetHelpText relying on existence of a widget with Id(:WizardDialog)
but that is risky for SLE15-SP4

## Testing

Tested manually.
For easier testing, the dialogs are:

- MkfsOptions "Format Options"
  Add Partition, Next, Next, Options, Help
- FstabOptions "Fstab Options"
  Add Partition, Next, Next, Mount Device, Fstab Options, Help
- ImportMountPoints "Import Mount Points from Existing System"
  System menu, Import Mount Points (in installation mode)
- PartitionTableClone "Clone partition layout of /dev/sdx"
  Device menu, Clone Partitions to Another Device, Help

## Screenshots

Working help popups with this fix:

nohelp-format-options
nohelp-fstab-options
nohelp-import-mount-points
nohelp-clone-partition-table

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Apr 18 09:01:42 UTC 2022 - Martin Vidner <mvidner@suse.com>
+
+- Fix empty help in some Partitioner dialogs (bsc#1194274)
+- 4.4.38
+
+-------------------------------------------------------------------
 Thu Apr  7 16:29:00 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
 
 - Fix fstab entry filesystem matching allowing the use of quotes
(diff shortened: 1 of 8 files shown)


dgdavid (approved):
LGTM, thank you!

PR #1298, opened 2022-04-19
Fix empty help in some Partitioner dialogs (bsc#1194274) master

mvidner:
See #1297, merging it from SLE-15-SP4 to master...

> These fixes work via a method of a base class of all these dialogs, [Y2Partitioner::Dialogs::Popup#set_help_text](https://github.com/yast/yast-storage-ng/blob/SLE-15-SP4/src/lib/y2partitioner/dialogs/popup.rb#L100).
> 
> A better fix would be to use a common mechanism of CWM+Wizard:
> Wizard.SetHelpText relying on existence of a widget with Id(:WizardDialog)
> but that is risky for SLE15-SP4

so here I'm replacing all of it with a better fix :smile: 

> 
> CWM.MergeHelps is the focal point, see where its result goes to
> understand the following fix:
> 
> The critical piece is having a widget with ID(:WizardDialog) because
> that is the widget to which CWM will assign the merged help texts of the
> individual widgets, using Wizard.SetContentsButtons or
> Wizard.RestoreHelp.
> 

Once this is in place, we can remove many methods of Y2Partitioner::Dialogs::*

Tested manually.

### DIFF
--- a/src/lib/y2partitioner/dialogs/popup.rb
+++ b/src/lib/y2partitioner/dialogs/popup.rb
@@ -26,7 +26,6 @@ module Y2Partitioner
     class Popup < Base
       def wizard_create_dialog(&block)
         Yast::UI.OpenDialog(layout)
-        set_help_text
         block.call
       ensure
         Yast::UI.CloseDialog()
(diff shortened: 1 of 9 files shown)


src/lib/y2partitioner/dialogs/popup.rb
```diff
@@ -41,7 +40,8 @@ def layout
           1, # left / right
           0.45, # top / bottom
           VBox(
-            Id(:help_text_container),
+            # This Id makes the Help texts work, see Wizard.SetHelpText
```
dgdavid, line 43: NP: Maybe we can explain a bit more here, what do you think?

```suggestion
            # This Id makes the Help texts work because CWM will assign the merged help texts
            # of the individual widgets to that one, see Wizard.SetHelpText
```

dgdavid (approved):
LGTM, just a minor suggestion.

Thanks!

PR #1304, opened 2022-05-31
Y2Partitioner: Allow 4 KiB Min Chunk Size for RAID0 / RAID10 [master]

shundhammer:
## Target Branch

This is the merge to _master_ / _Factory_ of https://github.com/yast/yast-storage-ng/pull/1303.

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1200018

## Trello

https://trello.com/c/YN8fSMy3

## Problem

A customer wishes to use 4 KiB (page size) chunks for creating a RAID0 on NVME, but the YaST partitioner does not offer that selection; it starts with 64 KiB.

yast-raid-chunk-size

## Cause

Back when we introduced that feature, we decided on a minimum chunk size of 64 KiB unless the default chunk size for that RAID type would suggest something smaller. For RAID0, there was no different default, so the minimum was 64 KiB.

There was no strict technical limitation behind this minimum, only our belief back at the time (in 2017) that smaller values would not be useful.

## Fix

Use a more elaborate minimum, depending on RAID type:

- RAID0: At least 4 KiB or the block size, whichever is larger.
- RAID10: At least 4 KiB or the block size or the page size, whichever is largest. Notice that some architectures have a pretty large page size (64 KiB for ppc64).
- Other RAID levels use their default chunk size as the minimum, but no more than 64 KiB.

yast-raid-chunk-size-4k

Now the minimum chunk size for RAID0 is 4 KiB, but the default is still 64 KiB which might is probably better for performance in most scenarios.

## Related PR

- This raised the minimum:
  https://github.com/yast/yast-storage-ng/pull/426

- PR for SLE-15-SP4: https://github.com/yast/yast-storage-ng/pull/1303

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue May 31 13:04:11 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Partitioner: Allow min chunk size of 4 KiB (page size) for RAID0 /
+  RAID10 (bsc#1200018)
+- 4.5.7
+
+-------------------------------------------------------------------
 Mon May 23 13:12:58 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
 
 - Fixed failing unit test: Added translatable message for new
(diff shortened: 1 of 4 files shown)


ancorgs (approved):
LGTM

PR #1306, opened 2022-08-16
New mountby

jreidinger:
## Problem

Libstorage-ng introduces new types for mount_by which causes failure in test suites and also other code is not adapted for it.

- [bsc#1202225](https://bugzilla.suse.com/show_bug.cgi?id=1202225)
- https://trello.com/c/Xz60ksqP/3042-p1-yast-storage-ng-broken-due-to-new-mount-bys-on-libstorage-ng
- https://github.com/openSUSE/libstorage-ng/pull/893

## Solution

Introduce list of supported mount by and skip newly added ones as there is no request to support them.

## Testing

- *Adapted unit tests*

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug 16 17:42:49 UTC 2022 - Josef Reidinger <jreidinger@suse.com>
+
+- Adapt to new types of mount by in libstorage-ng. Skipped by now
+  as there is no request to support it. (bsc#1202225)
+- 4.5.8
+
+-------------------------------------------------------------------
 Tue May 31 13:04:11 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
 
 - Partitioner: Allow min chunk size of 4 KiB (page size) for RAID0 /
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-storage-ng
-Version:        4.5.7
+Version:        4.5.8
 Release:        0
 Summary:        YaST2 - Storage Configuration
 License:        GPL-2.0-only OR GPL-3.0-only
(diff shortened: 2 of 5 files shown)


ancorgs:
I would prefer adapting yast2-storage-ng in a way that actually makes the new undesired types invisible to the users of `Y2Storage`.

This actually implies changing all the calls everywhere from `MountByType.all` into calls to `MountByType.all_supported`, including the calls in the test suite. That's a modification in the API.

In my opinion, the correct fix should imply no changes in the testsuite.

In other words, I would modify `MountByType.all` to keep returning the same result that it returned before the change in libstorage-ng.

mvidner:
> In other words, I would modify `MountByType.all` to keep returning the same result that it returned before the change in libstorage-ng.

I tried this but that alone did not fix the tests:
```diff
diff --git a/src/lib/y2storage/filesystems/mount_by_type.rb b/src/lib/y2storage/filesystems/mount_by_type.rb
index d3f64ebe..a50278e6 100644
--- a/src/lib/y2storage/filesystems/mount_by_type.rb
+++ b/src/lib/y2storage/filesystems/mount_by_type.rb
@@ -36,6 +36,10 @@ module Y2Storage
 
       wrap_enum "MountByType"
 
+      def self.all
+        super - [PARTLABEL, PARTUUID]
+      end
+
       # Hash with the properties for each mount by type.
       #
       # Keys are the symbols representing the types and values are hashes that

```

jreidinger:
@mvidner I think that you cannot use super for wrapped class.

@ancorgs well, it can be done that way. What I do not like about it is that it is not consistent as for other classes all is all what is supported by libstorage-ng and here it will be reduced. But if you are OK with it, I can redo it.

ancorgs:
Actually what I found inconsistent with other classes is using `WhateverEnum.all` (eg. `DasdType.all`, `EncryptionType.all`, LvType.all`, etc.) in all of them and then having to use `MountByType.all_supported`.

jreidinger:
resolved in https://github.com/yast/yast-storage-ng/pull/1307

PR #1307, opened 2022-08-17
New mountby2

jreidinger:
## Problem

Libstorage-ng introduces new types for mount_by which causes failure in test suites and also other code is not adapted for it.

- [bsc#1202225](https://bugzilla.suse.com/show_bug.cgi?id=1202225)
- https://trello.com/c/Xz60ksqP/3042-p1-yast-storage-ng-broken-due-to-new-mount-bys-on-libstorage-ng
- https://github.com/openSUSE/libstorage-ng/pull/893
- original pr with alternative  solution https://github.com/yast/yast-storage-ng/pull/1306

## Solution

Adapt a code to skip newly added ones as there is no request to support them.

## Testing

- *Adapted unit tests*

### DIFF
--- a/src/lib/y2storage/blk_device.rb
+++ b/src/lib/y2storage/blk_device.rb
@@ -791,7 +791,7 @@ module Y2Storage
     #
     # @return [Filesystems::MountByType]
     def preferred_mount_by
-      Filesystems::MountByType.best_for(self, possible_mount_bys)
+      Filesystems::MountByType.best_for(self, possible_mount_bys & Filesystems::MountByType.all)
     end
 
     # @see Device#is?
--- a/src/lib/y2storage/blk_device.rb
+++ b/src/lib/y2storage/blk_device.rb
@@ -289,13 +289,20 @@ module Y2Storage
     #   @return [Encryption] nil if the device is not encrypted
     storage_forward :encryption, as: "Encryption", check_with: :has_encryption
 
-    # @!method possible_mount_bys
+    # @!method storage_possible_mount_bys
     #   Possible mount-by methods to reference the block device itself, regardless of its content
     #
     #   @see #preferred_name
     #
     #   @return [Array<Filesystems::MountByType>]
-    storage_forward :possible_mount_bys, as: "Filesystems::MountByType"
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/blk_device.rb
```diff
@@ -791,7 +791,7 @@ def encrypt_with_method(method, dm_name, **method_args)
     #
     # @return [Filesystems::MountByType]
     def preferred_mount_by
-      Filesystems::MountByType.best_for(self, possible_mount_bys)
+      Filesystems::MountByType.best_for(self, possible_mount_bys & Filesystems::MountByType.all)
```
ancorgs, line 794: I find this inconsistent with the way it's done for `MountPoint`.  In `MountPoint` you redefined `#possible_mount_bys` to return a filtered list. Here, `#possible_mount_bys` still returns an unfiltered list and the filter is done at `preferred_mount_by`.

src/lib/y2storage/filesystems/mount_by_type.rb
```diff
@@ -90,6 +90,14 @@ def udev_name(value)
       end
 
       class << self
+
+        alias_method :all_nonfiltered, :all
```
ancorgs, line 94: Inconsistent use of `nonfiltered` vs `unfiltered` (below, in the `MountPoint` class).

src/lib/y2storage/blk_device.rb
jreidinger: well, reason here is that it is private method and this is its only usage.

src/lib/y2storage/filesystems/mount_by_type.rb
jreidinger: and which one you prefer?

ancorgs: Well, actually what we always do when we override a list or value coming from libstorage-ng is having `storage_whatever` for the raw value and then `whatever` for the converted one.

For consistency I guess we should have `MountByType.storage_all` vs `MountByType.all` and `MountPoint#storage_possible_mount_bys` vs `MountPoint@possible_mount_bys`.

ancorgs (approved):
LGTM

PR #1308, opened 2022-08-23
Security policy checks

joseivanlopez:
## Problem

YaST installer is now able to validate whether a setup fulfills the installation requirements of the DISA STIG security policy, see https://github.com/yast/yast-security/pull/128. Some of those checks affect to the storage setup. But neither Guided Setup nor Expert Partitioner are performing checks for the enabled security policy.  

* Part of
  * https://trello.com/c/SqFGWTyB/3044-build-a-basic-initial-api-for-the-stig-checker
  * https://jira.suse.com/browse/PM-3526
* Requires: https://github.com/yast/yast-security/pull/128

## Solution

Perform policy checks and show issues in both the storage proposal dialog and the Expert Partitioner. Note that a policy can be enabled by default with the `YAST_SECURITY_POLICY` boot parameter, for example `YAST_SECURITY_POLICY=stig`.

NOTE: this will be merged after https://github.com/yast/yast-security/pull/128.

## Testing

* Added new unit tests
* Tested manually

## Screenshots

Screenshot from 2022-10-03 16-55-32

Screenshot from 2022-10-03 16-56-13

### DIFF
--- a/src/lib/y2storage/setup_checker.rb
+++ b/src/lib/y2storage/setup_checker.rb
@@ -115,8 +115,8 @@ module Y2Storage
 
       return @security_policies_warnings if @security_policies_warnings
 
-      policies_warnings = security_policies_issues.map do |policy, issues|
-        warnings = issues.map { |i| SetupError.new(message: i.message) }
+      policies_warnings = security_policies_failing_rules.map do |policy, failing_rules|
+        warnings = failing_rules.map { |r| SetupError.new(message: "#{r.id} #{r.description}") }
         [policy, warnings]
       end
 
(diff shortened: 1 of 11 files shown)


src/lib/y2storage/setup_checker.rb
```diff
@@ -105,8 +107,51 @@ def mount_warnings
...
+
+        policies_manager.issues(scope).to_h
+      end
+
+      issues || {}
+    end
+
+    # Ensures security polices are correctly loaded
+    #
+    # Package yast2-security is not added as dependency to avoid cyclic dependencies.
```
mvidner, line 146: Please list the known use cases, and whether yast2-security is present/absent
Like
1. Installation time: yes, because yast2-security is included
2. RPM build time: no, avoiding cyclic build dependency
3. ... ?

mvidner:
OK, only when seeing this PR do I understand the purpose of *scopes* in https://github.com/yast/yast-security/pull/131 : 
so that a YaST module can run only the relevant part of the security policy which it can configure

mvidner (changes requested):
LGTM, except the dependencies:

This depends on https://github.com/yast/yast-security/pull/131 and it should be expressed as a RPM dependency.

We can't do a `Require` on yast2-security because that would make a circular dependency, but this should work:
```
Conflicts: yast2-security < 4.4.15
```

joseivanlopez:
> LGTM, except the dependencies:
> 
> This depends on [yast/yast-security#131](https://github.com/yast/yast-security/pull/131) and it should be expressed as a RPM dependency.
> 
> We can't do a `Require` on yast2-security because that would make a circular dependency, but this should work:
> 
> ```
> Conflicts: yast2-security < 4.4.15
> ```

Actually it does not conflict, policy issues are not shown in that case.

mvidner (approved):
LGTM now.

OK, my fault, I was basically repeating my request *after* you have fulfilled it by the comment in commit ee3e55586a8e57a3b4334b96d008955bea494ff1.

I guess it's unsettling to me to see a dependency handled in code and omitted in RPM spec 
:laughing:

imobachgs (approved):

PR #1309, opened 2022-09-26
Add user probe callbacks

imobachgs:
## Problems

* Indirectly, `StorageManager` is coupled with UI. The [IssuesManager#report_probing_issues](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/issues_manager.rb) is called if an issue is registered during the probing phase (see [manage_probing_issues method](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/storage_manager.rb#L501).
* While analyzing the problem, we discovered that the Devicegraph contains a [IssuesManager](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/issues_manager.rb)  instance. This class just adds some UI logic on top of an [IssuesList](https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/y2issues/list.rb), so holding a reference in our data model looks tricky.

## Solution

To decouple the UI from the probing process, we decided to rely on the callbacks mechanism. However, we found out that there are two use-cases for those callbacks:

* Use them to augment our data model, as we do currently (adding the issues to the Devicegraph).
* Redefine the UI interaction.

So instead of just extending the `Callbacks::Probe`, we decided to keep this class as an implementation details and inject a [new callback class](https://github.com/yast/yast-storage-ng/blob/c4c8afffb8796899181a55721115ab97eac17a86/src/lib/y2storage/callbacks/user_probe.rb).

~~At this point, it only implements a `#report_probing_issues` method and the API is still under discussion. Event the method name looks redundant (why about just `report_issues`?).~~

Last but not least, we decided to get rid of the `IssuesManager` class, as it is not needed by the `Devicegraph` class anymore.

## To do

- [x] Discuss the current API
- [x] Should we consider doing the same for the activation? Perhaps we could open a separate PBI. By now, let's keep as they are.


src/lib/y2storage/storage_manager.rb
```diff
@@ -180,9 +181,9 @@ def probe(probe_callbacks: nil)
     #
     # @raise [Storage::Exception, Yast::AbortException] when probe fails
     #
-    # @param probe_callbacks [Callbacks::Probe, nil]
-    def probe!(probe_callbacks: nil)
-      probe_callbacks ||= Callbacks::Probe.new
+    # @param callbacks [Callbacks::Probe, nil]
+    def probe!(callbacks = nil)
+      probe_callbacks ||= Callbacks::Probe.new(user_callbacks: callbacks)
```
joseivanlopez, line 186: probe_callbacks = Callbacks::Probe.new(user_callbacks: callbacks)

src/lib/y2storage/callbacks/user_probe.rb
```diff
@@ -0,0 +1,59 @@
...
+  module Callbacks
+    # Class to implement callbacks used during probing
+    #
+    # This classes can be inherited to customize the behavior during
+    # the probing phase.
+    class UserProbe
+      include Yast::Logger
+
+      def initialize
+        textdomain "storage"
```
joseivanlopez, line 30: NP: I would not use textdomain here because there are no translations.

src/lib/y2storage/callbacks/yast_probe.rb
```diff
@@ -0,0 +1,79 @@
...
+    class YastProbe < UserProbe
+      include Yast::I18n
+      include Yast::Logger
+
+      def initialize
+        super
+        textdomain "storage"
+      end
+
+      # Report activation and probing issues through IssuesReporter
```
joseivanlopez, line 41: NP: Reports

joseivanlopez (approved):

src/lib/y2storage/storage_manager.rb
```diff
@@ -180,9 +181,9 @@ def probe(probe_callbacks: nil)
     #
     # @raise [Storage::Exception, Yast::AbortException] when probe fails
     #
-    # @param probe_callbacks [Callbacks::Probe, nil]
-    def probe!(probe_callbacks: nil)
-      probe_callbacks ||= Callbacks::Probe.new
+    # @param callbacks [Callbacks::Probe, nil]
+    def probe!(callbacks = nil)
```
joseivanlopez, line 185: Maybe we could take the opportunity for moving `Yast::Pkg.SourceReleaseAll if Yast::Mode.installation` to the callback.

joseivanlopez (approved):
LGTM

PR #1310, opened 2022-10-18
Adaptations needed for a richer DBus interface

ancorgs:
## Problem

To implement the [storage management we envision](https://github.com/yast/d-installer/blob/master/doc/storage_ui.md) for the D-Installer we need the ability to run the `GuidedProposal` with a couple of customizations:

- Trying only once with one of the target sizes (we don't want the double attempt with `desired` and `min` in D-Installer because it can be confusing).
- Having the possibility to force the min and max sizes of all volumes, disabling the mechanisms that adjust them (like the modifier related to activating snapshots)

## Solution

Implemented a new subclass `MinGuidedProposal` that tries only once using the `min_size` as starting point and ignoring the `desired_size`. Surely inheritance is not the nicest mechanism but it works perfectly here and is the mechanism that was chosen to customize the GuidedProposal (eg. `InitialGuidedProposal`).

Added two new flags to `VolumeSpecification` to ignore the automatic size adjustments.

Added the possibility to store a failed proposal in the `StorageManager`. That makes no sense in the context of YaST but is used in D-Installer (since the proposal is the only mechanism there to produce an staging devicegraph).

## Testing

Added unit tests for the new functionality


joseivanlopez (approved):
LGTM

PR #1315, opened 2022-11-17
Support for LUKS2 in the GuidedProposal

ancorgs:
## Problem

The `GuidedProposal` (and its variants) is the component responsible of calculating a good storage layout to install the system. So far it's used:

- By YaST to calculate the initial storage proposal
- By YaST if the option "Guided Setup" is used in the partitioning screen
- By AutoYaST if any of the options for [automatic partitioning](https://documentation.suse.com/sles/15-SP4/html/SLES-all/cha-configuration-installation-options.html#CreateProfile-Automatic-Partitioning) or for [guided partitioning](https://documentation.suse.com/sles/15-SP4/html/SLES-all/cha-configuration-installation-options.html#CreateProfile-Guided-Partitioning) is used
- By the current prototypes of D-Installer

For many reasons outlined in [this bugzilla comment](https://bugzilla.suse.com/show_bug.cgi?id=1185291#c1) and several other places, the `GuidedProposal` always uses LUKS1 to encrypt the devices.

But recent versions of openSUSE Tumbleweed and the ALP prototypes include a heavily patched version of Grub2 that can handle LUKS2 reasonably well as long as PBKDF2 is used as password-based key derivation function.

When installing any ALP-based product with D-Installer we want to make it possible to use TPM-based encryption. For that, LUKS2 is the only option, it cannot work with LUKS1.

## Solution

**Important:** this pull request also includes some previous commits to do some cleanup and refactor some parts. Review commit by commit.

Implement support for LUKS2 (with a configurable PBKDF) in the `GuidedProposal` so it can be used by D-Installer to implement TPM-based encryption.

There are no plans to use that capability from (Auto)YaST. Everything remains full backwards compatible using LUKS1 by default for any encryption done by the proposal.

The code assumes Grub2 can handle LUKS2 devices using PBKDF2. At the moment of writing, that is not true in the case of SLE-15-SP5 or any previous version of SLE or Leap. That implies we need to revisit that logic in the `BootRequirementsChecker` in the close future. On the bright side, Michael Chang commented he could make SLE-15-SP5 work just like TW in that regard. If that happens we would just need to remove the warning comments. :-)

`GuidedProposal` will always propose a separate unencrypted partition for `/boot` if LUKS2 is used with Argon2i or Argon2id, since Grub2 cannot handle those key derivation functions, not even in Tumbleweed or ALP-based systems. Due to a current bug, that's not true in the following case:  LVM with the physical volumes encrypted with LUKS2 using Argon2i or Argon2id. The plan is to fix that in an upcoming pull request, to not block the current one (LVM+LUKS2 with Argon is out of the scope of both YaST and D-Installer for now).

## Testing

- Tested manually. It works in a current Factory (and Grub2 indeed opens the LUKS2 device with PBKDF2)
- Created new unit tests and adapted the existing ones

## Pending for the future (after merging)

- Fix the LVM+Argon2 case - `BootRequirementsChecker` is not proposing a separate /boot
- As explained above, we need to re-evaluate the SLE-15-SP5 case in the following months. Very likely SP5 will be adapted and we will only need to delete the comments.
- As the number of parameters to configure encryption grows, it would make sense to reconsider how those parameters are passed `GuidedProposal`. We could likely add some kind of new class (eg. `EncryptionSettings`) to encapsulate that information instead of using three fields in the already overloaded `ProposalSettings`.

### DIFF
--- a/src/lib/y2storage/pbkd_function.rb
+++ b/src/lib/y2storage/pbkd_function.rb
@@ -36,15 +36,18 @@ module Y2Storage
       @name = name
     end
 
+    # Instance of the function to be always returned by the class
+    # TRANSLATORS: name of a key derivation function used by LUKS
+    ARGON2ID = new("argon2id", N_("Argon2id"))
+    # Instance of the function to be always returned by the class
+    # TRANSLATORS: name of a key derivation function used by LUKS
+    ARGON2I = new("argon2i", N_("Argon2i"))
+    # Instance of the function to be always returned by the class
+    # TRANSLATORS: name of a key derivation function used by LUKS
+    PBKDF2 = new("pbkdf2", N_("PBKDF2"))
+
     # All possible instances
-    ALL = [
-      # TRANSLATORS: name of a key derivation function used by LUKS
-      new("argon2id", N_("Argon2id")),
-      # TRANSLATORS: name of a key derivation function used by LUKS
-      new("argon2i",  N_("Argon2i")),
-      # TRANSLATORS: name of a key derivation function used by LUKS
-      new("pbkdf2",   N_("PBKDF2"))
-    ].freeze
+    ALL = [ARGON2ID, ARGON2I, PBKDF2].freeze
     private_constant :ALL
 
     # Sorted list of all possible roles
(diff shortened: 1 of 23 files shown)


imobachgs (approved):

src/lib/y2storage/pbkd_function.rb
```diff
@@ -68,5 +70,34 @@ def self.find(value)
     def name
       _(@name)
```
joseivanlopez, line 71: Does a variable translation work?

joseivanlopez (commented):
LGTM

PR #1316, opened 2022-11-29
[Proposal] New setting to not reuse LVM volume groups

ancorgs:
## Problem

For historical reasons, YaST tries to reuse existing LVM volume groups when making a proposal.

That behavior can be very confusing in many situations. Some of the associated problems and historical factors are described here:
https://trello.com/c/Ib7QC298/19-proposal-strategy-for-reusing-stuff-in-proposal-bug1091998-and-others

Now `Y2Storage::GuidedProposal` is also used by D-Installer and we don't want to follow the YaST approach for reusing stuff. Since D-Installer represents a new playground, we don't need to play by the old rules and expectations.

 ## Solution

Introduce a new setting into `Y2Storage::ProposalSettings` that will be used only by D-Installer (for the time being) and that will inhibit the mechanism to reuse existing LVM volume groups.

## Testing

Added unit tests. Backwards compatibility ensured because all the previous tests keep working with no modification.

## Note about the version

Version of the package not bumped because we are not in a hurry to submit this to Factory and @jreidinger is already working on another change for yast-storage-ng.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Nov 29 15:07:25 UTC 2022 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Proposal: new setting to prevent reusing LVM volume groups
+  (related to gh#yast/d-installer#264).
+
+-------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


imobachgs (approved):
LGTM. There is no need to bump the version but we need to make sure that the change is available in the YaST:Head:D-Installer project too (it is linked, so it should be available).

PR #1317, opened 2022-12-07
Release version 4.5.15

dgdavid:
## Problem

https://github.com/yast/yast-storage-ng/pull/1316 introduced a small change in the API allowing to reuse the LVM volume groups. However, the version was not bumped and, accordingly, the code has not been _released_ yet to Factory.

## Solution

Bump version from 4.5.14 to 4.5.15

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Dec  7 14:26:51 UTC 2022 - David Diaz <dgonzalez@suse.com>
+
+- Allow reusing LVM volume groups
+- 4.5.15
+
+-------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


shundhammer (approved):
LGTM

package/yast2-storage-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Dec  7 14:26:51 UTC 2022 - David Diaz <dgonzalez@suse.com>
+
+- Allow reusing LVM volume groups
+- 4.5.15
```
joseivanlopez, line 5: I wouldn't have added a new changelog entry but simply adding the version to the lastest one. In fact, now the last two entries are kind of contradictory :)

dgdavid: You're right :( I didn't realize about it.

PR #1321, opened 2023-02-21
Ruby 3.2 yast2-storage-ng sometimes fails to build, diff in Marshal.dump 

mvidner:
## Problem

Ruby 3.2 yast2-storage-ng sometimes fails to build, diff in Marshal.dump

- https://bugzilla.suse.com/show_bug.cgi?id=1208259
- https://trello.com/c/jRe5bWVE/3249-ostumbleweed-1206419-staging-ruby-keyword-parameters-yast-stack-fails-to-build-against-ruby-32

The failing test means to check that a new `ProposalSettings` is in fact equal to the original one, and does so by dumping and reloading the two objects with `Marshal`. It fails in case the two objects differ in the order in which their instance variables were assigned.

## Solution

Instead of using `Marshal`, add `ProposalSettings#==`.

Do it by introducing a simple module, `EqualByInstanceVariables`, including it in that class and a couple of others that are used by it.

Gory details: everything I've learned about equality in Ruby along the way, very messy so far: https://gist.github.com/mvidner/eb10bf4d70f0df5c1073a2229fd4491e

## Testing

- Added new unit tests

## Screenshots

N/A

### DIFF
--- a/src/lib/y2storage/equal_by_instance_variables.rb
+++ b/src/lib/y2storage/equal_by_instance_variables.rb
@@ -1,0 +1,38 @@
+# Copyright (c) [2023] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Y2Storage
+  # Define `==` by having the same class and same (==) instance variables.
+  #
+  # The strict comparison (`eql?`) is unchanged and will be `false`
+  # for different objects. See Yast2::Equatable which can also
+  # specify the instance variables used for the comparison:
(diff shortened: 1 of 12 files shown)


mvidner:
Today's findings:

The `Marshal` dumps are different because some otherwise equal objects objects have *different ordering of instance variables*, namely in `VolumeSpecification` objects. Why, yes, I _did_ spend a stupid amount of time to find the reason for this difference. Unsuccessfully.

My plan is to accept the reality of the nondeterministic ordering and instead remove the gorilla from the room: `Marshal`. We are using it because `ProposalSettings#==` does not work properly

WIP:
```rb
irb(main):001:0> class C; def initialize(attrs); attrs.each { |a| instance_variable_set(a.to_sym, a.to_s) } ; end ; end
=> :initialize
irb(main):004:0> a = C.new([])
=> #<C:0x0000563aa62d27b0>
irb(main):005:0> b = C.new([])
=> #<C:0x0000563aa618b848>
irb(main):006:0> a == b
=> false
irb(main):007:0> module EqualByIVs; def ==(other); return if other.class != self.class; instance_variables.all? { |iv| instance_variable_get(iv) == other.instance_variable_get(iv) }; end; end
=> :==
irb(main):008:0> class C; include EqualByIVs; end
=> C
irb(main):009:0> a == b
=> true
```

src/lib/y2storage/equal_by_instance_variables.rb
```diff
@@ -0,0 +1,38 @@
...
+
+module Y2Storage
+  # Define `==` by having the same class and same (==) instance variables.
+  #
+  # The strict comparison (`eql?`) is unchanged and will be `false`
+  # for different objects. See Yast2::Equatable which can also
+  # specify the instance variables used for the comparison:
+  # https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/yast2/equatable.rb
+  module EqualByInstanceVariables
+    # TODO: move to base library?
```
jreidinger, line 28: :+1:

jreidinger (approved):

PR #1322, opened 2023-03-10
[SLE15-SP5] Removed unnecessary executable flags from files (bsc#1209094)

lslezak:
## Problem

- Some library files have unnecessary executable file flag set
- Although this is harmless (you will either get errors or the script does nothing if you try running them) it might make problems if you run some security scanner on the system.
- See https://bugzilla.suse.com/show_bug.cgi?id=1209094

## Notes

It turned out that the executable flags were added by Rubocop when it found a shebang (`#!/usr/bin/env ruby`) at the beginning of the file. See https://github.com/yast/yast-storage-ng/commit/381b52516e3425dc4ad1107b16c4abada2a485e6#diff-6fd7d4d70024d603ba6a0d36f66051086247b4c3a1845dff08420dc5e1c1f5c4. Later the shebang has been removed, but the executable flag stayed untouched.

## Solution

- Remove the shebang and the exec flag
- Additionally remove the UTF-8 encoding flag, it is the Ruby default for quite some time already
- The exec flag is kept where it makes sense (unit tests and examples)


joseivanlopez (approved):
LGTM

PR #1325, opened 2023-03-31
bsc#1209697: fix widgets titles in the dialog to select a partitioning schema

imobachgs:
## Problem

Although strings are marked for translation, they are not properly translated later.

https://bugzilla.suse.com/show_bug.cgi?id=1209697

## Solution

Fix the calls to `_` to act on the marked test.

## Testing

- Tested manually

## Screenshots

translated-partitioning-scheme

### DIFF
--- a/src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
+++ b/src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
@@ -109,10 +109,8 @@ module Y2Storage
               CheckBox(
                 Id(:separate_vgs),
                 format(
-                  # TRANSLATORS: %{widget_label} refers to the label of the widget. %{paths} is a
-                  # comma separated list of paths
-                  _("%{widget_label}\n(%{paths})"),
-                  widget_label: WIDGET_LABELS[:use_separate_vgs],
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/dialogs/guided_setup/select_scheme.rb
```diff
@@ -201,7 +199,7 @@ def base_help_text
               "so make sure not to lose it!</i>" \
               "</p>"
             ),
-            disk_encryption_label: WIDGET_LABELS[:use_disk_encryption]
+            disk_encryption_label: _(WIDGET_LABELS[:use_disk_encryption])
```
ancorgs, line 202: Hmmm, reviewing this I just noticed `:use_disk_encryption` is not a key on that hash. Can you verify how the help text looks?

imobachgs: Oh, good point, I did not check the keys. Now it looks OK.

ancorgs (approved):
LGTM.

I assume this also fixes the "separate LVMs" case (which is only used for SUMA).

PR #1327, opened 2023-03-31
Ensure thin LVM volumes are exported when cloning the system

ancorgs:
## Problem

Fix for [bsc#1209725](https://bugzilla.opensuse.org/show_bug.cgi?id=1209725)

When exporting the AutoYaST profile (ie. when cloning the system), LVM thin volumes were ignored.

According to AutoYaST documentation, the thin pool and all its thin volumes must be included as `<partition>` elements within the `<drive>` that represents the LVM volume group.

In the devicegraph, the thin pool and its thin volumes are not represented at the same hierarchy level. The volumes are nested into the pool. That was overlooked when implementing the code to clone the system.

## Solution

Ensure all logical volumes are taken into consideration, despite the nesting. With that, the exported profile includes the thin pool as well as all its thin volumes.

## Testing

- Added a regression unit test.
- Fix verified by the bug reporter in the real affected system.


dgdavid (approved):
LGTM

PR #1329, opened 2023-04-11
[SP5] Improve display of Dell BOSS

ancorgs:
## Problem

Some things has changed regarding BOSS devices manufactured by Dell. That leaded to the creation of [bug#1200975](https://bugzilla.suse.com/show_bug.cgi?id=1200975).

From the discussions in the bug report, we identified two things that could be improved in the YaST side.

1. The regexp to detect a Dell BOSS disk needs to be updated. Originally Dell stated that all BOSS devices would contain the string "DELLBOSS" as part of the model (https://jira.suse.com/browse/SLE-17578). Looks like that's not true anymore. According to a comment at the mentioned [bug#1200975](https://bugzilla.suse.com/show_bug.cgi?id=1200975), the new criteria is "_the model string will always have both Dell and BOSS in some order_".
2. For a typical disk, the column "Type" in the Expert Partitioner is computed as `device.vendor + device.model` (both fields taken from hwinfo). But that's redundant because "model" in hwinfo is a combination of vendor + device with quite some added heuristics to make the result nice looking.

## Solution

- Extend the BOSS criteria to check for the presence of both "Dell" (checked in a case-insensitive way) and "BOSS" (checked case sensitive) in any order. That covers all the known models like "DELLBOSS VD" or "Dell BOSS-N1 Modular".
- Adjust the Type column. See screenshots below.

## Testing

Added new unit tests

## Screenshots

Type column before the change.

before

Type column after the change.

after

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Apr 11 13:38:25 UTC 2023 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Adjusted detection of Dell BOSS devices (bsc#1200975).
+- Partitioner: improved column Type for disks (bsc#1200975).
+- 4.5.20
+
+-------------------------------------------------------------------
 Fri Mar 31 11:21:44 UTC 2023 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Fix the translation of widgets titles in the dialog to select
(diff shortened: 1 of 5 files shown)


shundhammer (approved):
LGTM

Let's hope they won't change their mind again about "BOSS" always in capital letters. ;-)

ancorgs:
BOSS is an acronym and has always being represented in capital letters in absolutely all previous conversations. So let's hope Dell keeps being consistent here. :wink:

I decided to use IGNORECASE for "dell" to support both "DELL" and "Dell", but I don't think the same justifies for the "BOSS" case (I don't foresee any alternative capitalization there).

shundhammer (approved):
Now even better with change log and version bump.

PR #1335, opened 2023-04-28
Adaptations to have more flexible proposal (useful for Agama)

ancorgs:
## Problem

We are re-using the YaST `GuidedProposal` at Agama (to be precise, we use `MinGuidedProposal` which is identical but ignores `desired_size` and performs only one attempt with the minimum).

The `GuidedProposal` can find the best location for every requested volumes within a set of candidate devices. It also allows to specify that a given volume must be placed in a specific disk. But turns out it was not able to do both things at the same time!

If a setting called `allocate_volume_mode` was set to `auto` (which is the default), assignations of volume to specific disks were ignored and the proposal freely allocated all the volumes within the designated candidate devices. If `allocate_volume_mode` was set to `device` then the proposal calculated the candidate devices as the sum of all devices already assigned to a particular volume.

In other words, the proposal was not able to handle requirements like "allocate all volumes at sda except for volume `/home` that should go alone to sdb".

## Solution

Now when the allocate mode is `auto`, the proposal still honors when a volume is assigned to a specific device, no matter if that device is part of `candidate_devices` or not.

That allows us to offer the following options at Agama proposal configuration.

### Agama's partition-based proposal

- All partitions are created in the boot disk by default.
- It's possible to override that for a particular volume and specify a target disk for it

That can be achieved by setting the following `Y2Storage::Settings`

- root_device -> boot disk
- candidate_devices -> [boot_disk]
- Assign `device` to volumes that don't go to the boot disk

### Agama's LVM-based proposal

- The system VG is created by default in the boot disk
- It's possible to specify a set of several disks if we want the system VG to (potentially) extend over them (will do it only if necessary). It's even possible to select a disk or set of disks that do not include the boot disk.
- All LVs are created by default in the system VG
- It's possible to override that for a particular volume and specify an alternative VG name and a target disk (only one) for it

That can be achieved by setting the following `Y2Storage::Settings`

- root_device -> boot disk
- candidate_devices -> disks to be used to allocate the system VG
- separate_vgs -> true
- Assign `separate_vg_name` and `device` to volumes that don't go to the system VG

## Testing

Added unit tests to demonstrate all that is possible now

## Review

Better review commit by commit. This also includes a couple of clean-up commits.

### DIFF
--- a/test/y2storage/proposal_agama_basis_test.rb
+++ b/test/y2storage/proposal_agama_basis_test.rb
@@ -1,0 +1,181 @@
+#!/usr/bin/env rspec
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require_relative "spec_helper"
+require "storage"
+require "y2storage"
+require_relative "#{TEST_PATH}/support/proposal_examples"
+require_relative "#{TEST_PATH}/support/proposal_context"
+
+describe Y2Storage::MinGuidedProposal do
+  describe "#propose with settings in the Agama style" do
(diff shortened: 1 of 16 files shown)


test/y2storage/proposal_agama_basis_test.rb
```diff
@@ -0,0 +1,179 @@
...
+        let(:expected_scenario_filename) { "agama-basis-separate_boot" }
+
+        before { settings.volumes.each { |v| v.device = "/dev/sda" } }
+
+        # Creates all partitions at sda except the bios_boot one (at sdb)
+        include_examples "proposed layout"
+      end
+    end
+
+    xcontext "when ProposalSettings#lvm is set to true" do
```
ancorgs, line 107: This is a mistake. Let me re-enable these.

joseivanlopez (approved):
LGTM

joseivanlopez (changes requested):
Changelog entry and version bump are missing.

joseivanlopez (approved):

PR #1336, opened 2023-05-04
Add commit callbacks to inst_prepdisk

joseivanlopez:
## Problem

Agama uses the *inst_prepdisk* client to perform the required storage actions. That client calls to the storage commit action, which uses the default commit callbacks from YaST to report the errors.

Agama needs to provide its own callbacks, see https://github.com/openSUSE/agama/pull/558.

## Solution

Allows to pass commit callbacks to the *inst_prepdisk* client.


ancorgs (approved):
LGTM

PR #1337, opened 2023-05-16
New strategy for SpaceMaker

ancorgs:
## Problem

The traditional behavior of `Y2Storage::Proposal::SpaceMaker` is coupled to the options offered by YaST in that regard. Somehow summarized in this screenshot.

auto

But for Agama we want to offer other options.

First of all, we want to stop making a difference for partitions of type Windows/Linux/other. The criteria for classifying a partition in one group or the other is not so clear.

On the other hand, we want to make it possible for Agama to specify any of the four modes described at [storage_ui.md](https://github.com/openSUSE/agama/blob/master/doc/storage_ui.md). That is:

 - Delete everything in the disk(s).
 - Resize existing partition(s).
 - Do not modify existing partition(s).
 - Custom, with the user specifying what to do with every individual partition in the affected disks: resize it, delete it or keep it as it is.

## Solution

This adds to the `ProposalSetting` the concept of different strategies for making space.

The default strategy is the classic one, now called `auto` and based on the known settings `resize_windows`, `windows_delete_mode`, `linux_delete_mode` and `other_delete_mode`.

A new strategy called `bigger_resize` is added. This strategy ignores the settings previously mentioned and works with a explicit list of actions specified in the proposal settings. Thus, the caller is responsible for deciding what to do with each partition or disk (`:force_delete`, `:delete` or `:resize`). Omitted devices are kept without modification. The only decision taken by `SpaceMaker` itself is the order in which those actions must be applied while looking for valid layout. As the name of the strategy suggests, it first executes all the `:force_delete` actions, then the `:resize` ones (sorted by "recoverable size") and finally the `:delete` ones.

With this solution, all the mentioned Agama modes can be easily implemented.

 - Delete everything in the disk(s). Settings with `:force_delete` for all partitions in the affected disks.
 - Resize existing partition(s).  Settings with `:resize` for all partitions in the affected disks.
 - Do not modify existing partition(s). Settings with an empty list of actions.
 - Custom. Settings describing the exact actions decided by the user.

## Implementation

The pull request is divided into several commits because the code was refactored into small steps before implementing the strategy support at SpaceMaker. Our comprehensive battery of unit tests survived all those reorganizations.

## Testing

Added new unit tests.

### DIFF
--- a/src/lib/y2storage/proposal/space_maker.rb
+++ b/src/lib/y2storage/proposal/space_maker.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2015] SUSE LLC
+# Copyright (c) [2015-2023] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -19,13 +19,11 @@
 
 require "fileutils"
 require "storage"
-require "y2storage/planned"
-require "y2storage/partition"
+require "y2storage/partitionable"
 require "y2storage/disk_size"
 require "y2storage/free_disk_space"
 require "y2storage/proposal/partitions_distribution_calculator"
-require "y2storage/proposal/partition_killer"
-require "y2storage/proposal/space_maker_prospects"
+require "y2storage/proposal/space_maker_actions"
 
 module Y2Storage
   module Proposal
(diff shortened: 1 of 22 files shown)


src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -306,95 +291,89 @@ def force_ptables(planned_partitions)
...
-        prospect.available = false
+      # @return [Array<Integer>] sids of the devices that has been deleted as a consequence
+      #   of the action
+      def execute_action(action, graph, planned_partitions = nil, disk_name = nil)
+        case action
+        when SpaceMakerActions::Shrink
+          execute_shrink(action, graph, planned_partitions, disk_name)
+        when SpaceMakerActions::Delete
+          execute_delete(action, graph)
+        else
```
joseivanlopez, line 327: NP: I would explicitly check by `SpaceMakerActions::Shrink`, and loging/raising an exception in else.

src/lib/y2storage/proposal/space_maker_actions/shrink.rb
```diff
@@ -0,0 +1,58 @@
...
+          []
+        end
+
+        protected
+
+        # @param partition [Partition]
+        def shrink_partition(partition)
+          target = shrink_size.unlimited? ? DiskSize.zero : partition.size - shrink_size
+          # Explicitly avoid alignment to keep current behavior (to be reconsidered)
+          partition.resize(target, align_type: nil)
```
joseivanlopez, line 53: Does it prevent to resize beyond the available free space?

src/lib/y2storage/proposal_space_settings.rb
```diff
@@ -0,0 +1,100 @@
...
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2storage/equal_by_instance_variables"
+
+module Y2Storage
+  # Class to encapsulate all the GuidedProposal settings related to the process of making space
+  # to allocate the new operating system
+  class ProposalSpaceSettings
+    include EqualByInstanceVariables
```
joseivanlopez, line 27: Just as note, we have the `Equatable` mixin for custom comparison: https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/yast2/equatable.rb

joseivanlopez (commented):
Version bump and changlog?

src/lib/y2storage/proposal_space_settings.rb
joseivanlopez: Yes, only mentioning to promote it for other cases :)

joseivanlopez (approved):
LGTM

src/lib/y2storage/proposal/space_maker.rb
ancorgs: I fear that would only lead to a line of code not covered by tests. This is quite internal, not something receiving direct input from the caller. I wouldn't expect a `SpaceMakerActions::List` to return an action from an unknown type.

PR #1340, opened 2023-06-06
Prevent setting the Volume Label for a Mounted Btrfs or Swap

shundhammer:
## Target Branch

**This is for _SLE-15-SP4_**. The merge to _SLE-15-SP5_ and _master_ / _Factory_ will follow.

## Bugzilla

**L3:** https://bugzilla.suse.com/show_bug.cgi?id=1211337

## Trello

https://trello.com/c/0ASTpHRy

## Problem

Trying to change the volume label of a mounted Btrfs or swap partition in the YaST partitioner fails with an error message

  _Setting label of btrfs on /dev/sda2 (57.50 GiB) to root Unexpected situation found in the system._

The "Details" button shows

  _Caught signal #127: "/sbin/btrfs filesystem label '/dev/sda2' 'root'_

which is scary and confusing and does not give the user any hint what's actually wrong.

Unexpected_situation

Caught_signal

In the case of a root Btrfs, it even hangs later because subvolumes cannot be mounted.

## Cause

Even though some filesystem types do support changing the volume label while mounted, in our context this has implications with the subvolumes. Handling the subvolumes properly very likely requires a very intrusive code change which involves quite some risk.

## Fix

So we chose the less intrusive approach to disable it on the UI level: When we detect that the filesystem or swap space is actively mounted, we are now disabling the "Volume Label" field in the "Fstab Options" dialog.

## Screenshots

partitioner-btrfs
_Partitioner with a mounted Btrfs on /dev/sdb1_

partitioner-edit
_Editing /dev/sdb1_

btrfs-mounted
_"Volume Label" field disabled because the filesystem is mounted, so changing it would fail._

partitioner-edit-after-unmount
_After unmounting, saving the changes and entering the partitioner again, the volume label can now be changed._

## Test

Extensive manual tests in a VM with a second virtual disk where partitions can be easily added and changed.

## Related PRs

- Merge to _SLE-15-SP5_: TBD
- Merge to _master_: TBD

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jun  7 08:03:52 UTC 2023 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Prevent setting the volume label for a mounted btrfs or swap
+  (bsc#1211337)
+- 4.4.43
+
+-------------------------------------------------------------------
 Tue Jan 17 14:32:36 UTC 2023 - Stefan Hundhammer <shundhammer@suse.com>
 
 - Extended regexp to identify Dell BOSS storage devices (bsc#1200975)
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-storage-ng
-Version:        4.4.42
+Version:        4.4.43
 Release:        0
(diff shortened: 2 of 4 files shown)


shundhammer:
I am not sure if we should really restrict that to Btrfs and swap, or if we shouldn't generally disable that "Volume Label" field for any mounted filesystem.

shundhammer:
@aschnell FYI

ancorgs (approved):
LGTM

PR #1344, opened 2023-06-28
Honor proposal encryption from ProductFeatures

ancorgs:
## Problem

We want to make it possible for the Common Criteria role to provide an encryption password, so the partitioning proposal uses encryption by default.

And we want to do so in the less intrusive possible way, since we want the change to land at already released branches like 15.4 and 15.5.

## Solution

On the one hand, we implemented https://github.com/yast/system-role-common-criteria/pull/10, to make the Common Criteria role dialog ask for the encryption password and store it at the field `encryption` of the `ProductFeatures`.

On this pull request, we make sure the storage `GuidedProposal` honors the mentioned field. The storage proposal already took all its default settings from the `ProductFeatures`, but encryption was not previously considered because it makes no sense to define a given password for all the installations of a product. But `ProductFeatures` is actually a very handy mechanism to store this value without affecting any other part of the code.

## Testing

- Added a new unit test
- Tested manually together with https://github.com/yast/system-role-common-criteria/pull/10

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jun 29 11:42:21 UTC 2023 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Honor encryption settings if they are set into ProductFeatures
+  by the Common Critera role (jsc#PED-4166, jsc#PED-4474).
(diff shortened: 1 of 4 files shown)


joseivanlopez (commented):
Missing changelog entry and version bump.

joseivanlopez (approved):

PR #1345, opened 2023-07-04
WIP: Storage Package Handling

shundhammer:
## Problem

For MicroOS which uses another storage proposal client, storage packages were not added to the package proposal, so features like multipath didn't work.

## Fix

Factored out the storage package handling from inst_disk_proposal.rb to some common classes.

Now using a class method in PackageHandler to separate required and optional storage packages.


shundhammer:
This is now **superseded by #1346 **. Closing.

PR #1350, opened 2023-07-10
Ensure adding Storage Support Software Packages for MicroOS [SLE-15-SP5]

shundhammer:
## Target Branch / Product

_This is the merge to **SLE-15-SP5** of #1346 ._

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1212452

## Trello

https://trello.com/c/T9nqET80/

## Problem

On an system with a multipath storage setup, the reboot after the installation failed. It turned out that the _multipath-tools_ package was not installed.

## Cause 

A regular SLE-15-SPx or Leap 15.x or Tumbleweed installation uses the _inst_disk_proposal_ client in the installation workflow which at its end determines which storage features (e.g. Btrfs, XFS, LVM, multipath) are used and adds the software packages that are needed for those technologies to the pool of packages that are going to be installed.

But MicroOS has a different installation workflow, and it doesn't use that default _inst_disk_proposal_ client at all; instead, it uses a custom _partitions_proposal_ client to handle the storage proposal. And in that client, the packages handling was missing.

## Fix

Added the packages handling to that _partitions_proposal_ client.

To avoid code duplication, that handling was factored out from the _inst_disk_proposal_ client: It is now done by a new `PackageHandler::set_proposal_packages_for(devicegraph)` method which is now called from both clients.

## Test

- New unit tests for that new method and for fringe cases

- Manual test in a Leap 15.5 VM with the changed files to test against regressions

- Manual test in a MicroOS 5.3 VM to check if the support packages are added; changed the storage setup between the default Btrfs to a nonstandard XFS and observed the log file if the packages were added correctly.

## Installer Self-Update / Maintenance Update

- No self-update needed for SLE-15-SP5 / Leap 15.5

- The change can wait for the next QU for SLE-15-SP5 / Leap 15.5

## Related PRs

- Original PR for _SLE-15-SP4:_ #1346 
- Merge to _master_ / _Factory_: TBD

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -4,7 +4,7 @@ Wed Jul  5 13:42:12 UTC 2023 - Stefan Hundhammer <shundhammer@suse.com>
 - Ensure adding storage support software packages for MicroOS
   which uses its custom partitions_proposal client, not the 
   standard inst_disk_proposal client (bsc#1212452)
-  https://github.com/yast/yast-storage-ng/pull/1347
+  https://github.com/yast/yast-storage-ng/pull/1350
 - 4.5.24
 
 -------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


joseivanlopez (approved):
LGTM

I think CC commits are safe. Nothing changes until CC role is adapted.

PR #1361, opened 2023-10-18
Speed-up tests a bit in systems with slow hwinfo

ancorgs:
## Problem

The testsuite used to run in around 2-4 minutes in my personal system.

We have always known that we should be more aggressive mocking calls to `hwinfo` and some libstorage-ng operations (mostly those that result in calls to `udevadm --settle`). But even without that mocking, the situation was still acceptable.

Recently things became MUCH worse. Now the testsuite is much slower. I was not able to find what caused it. No luck bisecting the changes in the repository, so I guess the problem is somewhere else (some combination of changes in libstorage-ng or another YaST package, Ruby, the Tumbleweed kernel, udev and whatnot).

## Solution

Even if the source of the recent slow-down is still there, I decided to introduce the long-overdue mocking for `hwinfo` in order to mitigate the problem.

Now the execution time is acceptable again, even if there is still room for improvement (if someone can find why/how at some point we went from a couple of minutes to ~10mins).

### DIFF
--- a/test/spec_helper.rb
+++ b/test/spec_helper.rb
@@ -136,6 +136,9 @@ RSpec.configure do |c|
       end
     end
 
+    allow(Y2Storage::HWInfoReader.instance).to receive(:for_device)
+      .and_return Y2Storage::HWInfoDisk.new
+
     # Bcache is only supported for x86_64 architecture. Probing the devicegraph complains if Bcache is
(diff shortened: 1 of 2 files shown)


mvidner (approved):
LGTM

BTW have you tried `rspec --profile`?

joseivanlopez (approved):
LGTM

lslezak (approved):
I can confirm that on my system it is more than twice faster.

PR #1362, opened 2023-10-19
Add internal setting to configure swap reusing in GuidedProposal

ancorgs:
## Problem

Currently YaST’s guided storage proposal reuses any available swap, even if that swap is not in the candidate devices.

That’s the traditional behavior inherited from the SLE-12 times. Although it has been reported as confusing at https://bugzilla.suse.com/show_bug.cgi?id=1175535 and https://bugzilla.suse.com/show_bug.cgi?id=1215639 .

It’s still not decided which possibilities we want to use regarding re-use of partitions at Agama. But while we decide what to do in Agama, the current behavior of reusing swap is confusing because it’s not clear what is happening looking at the UI.

## Solution

This implements a new setting at `ProposalSettings` to fine-tune swap reusing at the `GuidedProposal`. It has three possible values:

- `:any` Current behavior. For keeping compatibility in YaST.
- `:none` Simply don’t reuse any swap. This will be used by Agama while we decide what to do.
- `:candidate` Only reuse swaps that are in the candidate disks.

YaST keeps using `:any` and there is no plan to make that configurable in AutoYaST or in the YaST UI, although the modes `:none` and `:candidate` can easily be part of a future solution for the mentioned bug reports.

## Testing

Added new unit tests. Not tested with YaST in a real system, but the change looks safe enough.

Tested with a modified version of Agama that sets the new setting to `:none`. Works as expected.

### DIFF
--- a/src/lib/y2storage/proposal/devices_planner.rb
+++ b/src/lib/y2storage/proposal/devices_planner.rb
@@ -93,19 +93,38 @@ module Y2Storage
       # @param required_size [DiskSize]
       # @return [Partition]
       def reusable_swap(required_size)
-        return nil if settings.use_lvm || settings.use_encryption
+        return nil if skip_swap_reuse?
 
         partitions = available_swap_partitions
-        partitions.select! { |part| part.size >= required_size }
+        partitions.select! { |part| can_be_reused?(part, required_size) }
         # Use #name in case of #size tie to provide stable sorting
         partitions.min_by { |part| [part.size, part.name] }
       end
 
-      # Returns all avaiable swap partitions
+      # Returns all available and acceptable swap partitions
       #
       # @return [Array<Partition>]
       def available_swap_partitions
         devicegraph.partitions.select(&:swap?)
+      end
+
+      # Whether reusing swap partitions is pointless
+      #
+      # @return [Boolean]
+      def skip_swap_reuse?
+        settings.use_lvm || settings.use_encryption || settings.swap_reuse == :none
+      end
+
+      # Whether it is acceptable to reuse the given swap partition
+      #
+      # @param partition [Partition]
+      # @param required_size [DiskSize]
+      # @return [Boolean]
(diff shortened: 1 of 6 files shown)


joseivanlopez (approved):
LGTM

src/lib/y2storage/proposal/devices_planner.rb
```diff
@@ -93,21 +93,40 @@ def planned_boot_devices(planned_devices)
...
       #
       # @return [Array<Partition>]
       def available_swap_partitions
         devicegraph.partitions.select(&:swap?)
       end
 
+      # Whether reusing swap partitions is pointless
+      #
+      # @return [Boolean]
+      def skip_swap_reuse?
```
jreidinger, line 114: I would probably use shorter possitive `reuse_swap?` and above have `return nil unless reuse_swap?`

src/lib/y2storage/proposal/devices_planner.rb
```diff
@@ -93,21 +93,40 @@ def planned_boot_devices(planned_devices)
...
+      end
+
+      # Whether it is acceptable to reuse the given swap partition
+      #
+      # @param partition [Partition]
+      # @param required_size [DiskSize]
+      # @return [Boolean]
+      def can_be_reused?(partition, required_size)
+        return false if partition.size < required_size
+        return true unless settings.swap_reuse == :candidate
```
jreidinger, line 125: here I think it is confusing to match against candidate, where in fact you want to check if it is any so I would use `return true if settings.swap_reuse == :any`. Third `:none` cannot happen and here it is more easy to read like reuse it if anything is good enough.

src/lib/y2storage/proposal/devices_planner.rb
ancorgs: Method renamed

jreidinger (approved):

PR #1363, opened 2023-10-25
WIP: New EncryptionMethod using fde-tools for TPM-based unlocking

ancorgs:
## Problem

Both ALP and openSUSE Tumbleweed include a package called `fde-tools` that allow to setup encrypted devices (using LUKS2) to be automatically unlocked during boot without user intervention based on information stored and validated in the TPM of the system.

The process has its limitations but it certainly works as proven by the preliminary (and rather hacky) support present at Agama. So it's time to move that support from Agama to the core of YaST.

## Some more-or-less related links

- https://github.com/yast/yast-installation/issues/1088
- https://trello.com/c/5x8G7xj8/89-d-installer-better-support-for-tpm-unlocking-of-fde
- https://jira.suse.com/browse/PM-2766
- https://jira.suse.com/browse/PM-2395
- https://bugzilla.suse.com/show_bug.cgi?id=1210512

## Solution

This introduces a new encryption method (`TPM_FDE`) in yast2-storage-ng. Thus, TPM unlocking based on fde-tools can be configured by both AutoYaST and Agama.

If the system meets all the technical requirements to use the new method, it will be available in AutoYaST. That introduce the need to properly document all the implications such as the need to  boot the system in a very specific way after installation (so second stage CANNOT be used) or the inability to manually tweak the grub2 config.

Even if the mentioned technical requirements are met, the new method will still not be displayed in the UI for interactive YaST unless the boot parameter `YAST_LUKS2_AVAILABLE ` is passed. The reason is that explaining the feature properly is quite complex. Even more difficult than regular LUKS2 (which is already hidden by the same boot parameter). Apart from the usual LUKS2 implications like memory consumption or limited Grub2 support, this adds new requirements like the ones briefly mentioned above.

## Testing

In progress. Just like development itself.


src/lib/y2storage/encryption_method/tpm_fde.rb
```diff
@@ -0,0 +1,151 @@
...
+      def used_for?(_encryption)
+        # One candidate criteria (still waiting for some conversations with fde-toold developers)
+        # could be:
+        # encryption.type.is?(:luks2) && key_file == EncryptionProcesses::TpmFdeTools.key_file_name
+        false
+      end
+
+      # @see Base#available?
+      #
+      # In this initial implementation this always returns false because there important
```
joseivanlopez, line 62: becasuse there **are**

src/lib/y2storage/encryption_method/tpm_fde.rb
```diff
@@ -0,0 +1,151 @@
...
+      # @param blk_device [Y2Storage::BlkDevice]
+      # @param dm_name [String]
+      #
+      # @return [Y2Storage::Encryption]
+      def create_device(blk_device, dm_name, label: "")
+        encryption_process.create_device(blk_device, dm_name, label: label)
+      end
+
+      # Special method used in the tests to reset all memoizations
+      def reset
```
joseivanlopez, line 103: NP: Is this strictly needed? Having it only because tests is not good. Maybe we can use some metaprograming in tests to set `@tpm_present` instance variable.

src/lib/y2storage/encryption_processes/fde_tools.rb
```diff
@@ -0,0 +1,104 @@
...
+      def tpm_present
+        Yast::Execute.on_target!(FDECTL, "tpm-present")
+        log.info "FDE: TPMv2 detected"
+        true
+      rescue Cheetah::ExecutionFailed
+        log.info "FDE: TPMv2 not detected"
+        false
+      end
+
+      alias_method :tpm_present?, :tpm_present
```
joseivanlopez, line 58: NP: why not to directly call it `#tpm_present?`

src/lib/y2storage/encryption_processes/fde_tools_config.rb
```diff
@@ -0,0 +1,122 @@
+# Copyright (c) [2018-2020] SUSE LLC
```
joseivanlopez, line 1: NP: 2023

src/lib/y2storage/encryption_processes/tpm_fde_tools.rb
```diff
@@ -0,0 +1,142 @@
...
+      private_constant :CRYPT_OPTIONS
+
+      # Content of the third column of crypttab for all involved devices
+      KEY_FILE_NAME = "/.fde-virtual.key".freeze
+      private_constant :KEY_FILE_NAME
+
+      # Content of the third column of crypttab for all involved devices
+      #
+      # @return [String]
+      def self.key_file_name
```
joseivanlopez, line 50: NP: I would move this method inside the `class << self` block.

joseivanlopez (commented):
I left some comments, but I haven't finished with the review yet :)

src/lib/y2storage/encryption_processes/tpm_fde_tools.rb
```diff
@@ -0,0 +1,142 @@
...
+      end
+
+      # @see Base#encryption_type
+      def encryption_type
+        EncryptionType::LUKS2
+      end
+
+      # @see Base#post_commit
+      #
+      # TODO: this implemented only for the installation case. In an installed system the procedure
```
joseivanlopez, line 87: This **is**

src/lib/y2storage/encryption_processes/tpm_fde_tools.rb
```diff
@@ -0,0 +1,142 @@
...
+      # would be completely different and will likely include steps like configuring the fde-tools,
+      # calling "fdectl regenerate-key" and regenerating initrd and the bootloader configuration.
+      #
+      # @param device [Encryption] encryption that has just been created in the system
+      def post_commit(device)
+        return unless Yast::Mode.installation
+
+        self.class.devices ||= []
```
joseivanlopez, line 95: BTW: why is `#devices` a class instance?

src/lib/y2storage/encryption_processes/fde_tools.rb
ancorgs: No strong reason. I just wanted to have the method names equivalent to the `fdectl` commands (obviously turning hyphens into underscores). Since the `fdectl` subcommands are `tpm-present`, `add-secondary-password` and `add-secondary-key`...

src/lib/y2storage/encryption_processes/fde_tools_config.rb
ancorgs: Of course, you don't miss one of these.

test/y2storage/encryption_method/tpm_fde_test.rb
```diff
@@ -0,0 +1,162 @@
+# Copyright (c) [2021] SUSE LLC
```
joseivanlopez, line 1: NP: 2023

src/lib/y2storage/encryption_method/tpm_fde.rb
ancorgs: I did it that way because that's the approach we followed with the memoization of available packages at `StorageFeatures` (a code I had to touch recently). But I'm fine with changing that.

test/y2storage/encryption_processes/fde_tools_config_test.rb
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env rspec
+# Copyright (c) [2018] SUSE LLC
```
joseivanlopez, line 2: NP: 2023

src/lib/y2storage/yast_feature.rb
```diff
@@ -0,0 +1,61 @@
...
+    # Sorted list of all features defined by YaST
+    def self.all
+      ALL.dup
+    end
+
+    # Drop the cache of packages for all known YaST features
+    #
+    # This is only ever needed if the available packages might have changed
+    # since the last use of this class.
+    def self.drop_cache
```
joseivanlopez, line 50: NP: maybe this could be moved to the base class.

src/lib/y2storage/yast_feature.rb
```diff
@@ -0,0 +1,61 @@
...
+      all.each(&:drop_cache)
+    end
+
+    # @return [Boolean]
+    def ==(other)
+      other.class == self.class && other.id == id
+    end
+
+    alias_method :eql?, :==
```
joseivanlopez, line 59: NP: just a remainder, we have `Equatable` mixin, see https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/yast2/equatable.rb

joseivanlopez:
Just for the records: failing tests on leap is expected. Changes in this PR require a new version of *libstorage-ng* which is not submitted to leap (only to Tumbleweed).

In general, running unit tests on leap for the *master* branch is useless. Note that *SLE-15-SPX* branches have already diverged.

src/lib/y2storage/encryption_processes/tpm_fde_tools.rb
ancorgs: Because there is a single list of devices for all the processes. We need to build a list with all the affected devices (which happens during `post_commit`) and then configure all those devices at once after chrooting. I will document that better.

ancorgs: Improved the documentation to say basically the same than my previous comment.

src/lib/y2storage/encryption_method/tpm_fde.rb
ancorgs: I realized this is actually not needed because the tests create a new instance of the encryption method on each example. So there is no caching affecting the tests.

joseivanlopez (approved):
LGTM

imobachgs (approved):
The changelog looks good (the rest was already reviewed).

PR #1364, opened 2024-01-04
Experimental rubocop

jreidinger:
Trying changes from new rubocop. Review per commit to distinguish what was done manually and what automatic.

see also https://github.com/yast/yast-yast2/pull/1298

### DIFF
--- a/src/lib/y2storage/autoinst_profile/drive_section.rb
+++ b/src/lib/y2storage/autoinst_profile/drive_section.rb
@@ -619,7 +619,7 @@ module Y2Storage
       #   nil if the root filesystem is not applicable.
       def enabled_snapshots?(filesystems)
         root_fs = filesystems.find(&:root?)
-        return nil if root_fs.nil? || (root_fs.multidevice? && !btrfs_drive_section?)
+        return false if root_fs.nil? || (root_fs.multidevice? && !btrfs_drive_section?)
 
         root_fs.respond_to?(:snapshots?) && root_fs.snapshots?
       end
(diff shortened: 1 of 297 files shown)


src/lib/y2storage/autoinst_profile/drive_section.rb
```diff
@@ -619,7 +619,7 @@ def use_value_from_string(use)
       #   nil if the root filesystem is not applicable.
       def enabled_snapshots?(filesystems)
         root_fs = filesystems.find(&:root?)
-        return nil if root_fs.nil? || (root_fs.multidevice? && !btrfs_drive_section?)
+        return false if root_fs.nil? || (root_fs.multidevice? && !btrfs_drive_section?)
```
jreidinger, line 622: and this cause several test failures. I am not sure if it is correct change. On one hand it is `?` method, but nil is documented there. Also not sure if it is common in storage or one single occurrence.

src/lib/y2storage/proposal/partitions_distribution_calculator.rb
```diff
@@ -247,7 +247,7 @@ def partition_candidate_spaces(partition, candidate_spaces, extra_spaces)
       #
       # @return [Boolean]
       def compatible_disk?(partition, space)
-        return true unless partition.disk && partition.disk != space.disk_name
+        true unless partition.disk && partition.disk != space.disk_name
```
lslezak, line 250: It says `@return [Boolean]` but it actually returns `true` or `nil`... :thinking: 

Yard strictly defines `Boolean` as either `true` or `false` values, nothing else.

PR #1365, opened 2024-02-01
Handling of YAST_REUSE_LVM linuxrc option

mchf:
## Problem

https://jira.suse.com/browse/PED-6407

The Storage Proposal in SLE 15 allows for creating a layout based on LVM. For historical reasons, it tries to reuse the existing LVM volume group for new installations. This sometimes produces a non-optimal, non-logical or even confusing disk setup.

## Solution

linuxrc option to disable lvm reuse

## Testing

- *Tested manually*

### DIFF
--- a/src/lib/y2storage/proposal/lvm_helper.rb
+++ b/src/lib/y2storage/proposal/lvm_helper.rb
@@ -21,7 +21,6 @@ require "fileutils"
 require "y2storage/disk_size"
 require "y2storage/exceptions"
 require "y2storage/planned/lvm_vg"
-require "y2storage/storage_env"
 
 module Y2Storage
   module Proposal
--- a/src/lib/y2storage/proposal/lvm_helper.rb
+++ b/src/lib/y2storage/proposal/lvm_helper.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2015] SUSE LLC
+#eCopyright (c) [2015] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -21,6 +21,7 @@ require "fileutils"
 require "y2storage/disk_size"
 require "y2storage/exceptions"
 require "y2storage/planned/lvm_vg"
+require "y2storage/storage_env"
 
 module Y2Storage
   module Proposal
@@ -166,9 +167,9 @@ module Y2Storage
       # @return [Boolean]
       def try_to_reuse?
         # Setting introduced to completely avoid LVM reusing in D-Installer.
-        # It's a new playground, so no need to carry YaST behaviors that have
-        # proven to be confusing.
-        return false if settings.lvm_vg_reuse == false
+        # However it happened to be handy even in old yast later on. Preset
+        # settings can get overwritten by Linuxrc boot param.
+        return false if StorageEnv.instance.reuse_lvm? == false || settings.lvm_vg_reuse == false
 
         # We introduced this when we still didn't have a mechanism to activate
         # existing LUKS. Now such mechanism exists but this check has never been
--- a/src/lib/y2storage/proposal/lvm_helper.rb
+++ b/src/lib/y2storage/proposal/lvm_helper.rb
@@ -167,9 +167,7 @@ module Y2Storage
       # @return [Boolean]
       def try_to_reuse?
         # Setting introduced to completely avoid LVM reusing in D-Installer.
-        # However it happened to be handy even in old yast later on. Preset
-        # settings can get overwritten by Linuxrc boot param.
-        return false if StorageEnv.instance.reuse_lvm? == false || settings.lvm_vg_reuse == false
+        return false if settings.lvm_vg_reuse == false
 
         # We introduced this when we still didn't have a mechanism to activate
         # existing LUKS. Now such mechanism exists but this check has never been
--- a/src/lib/y2storage/proposal/lvm_helper.rb
+++ b/src/lib/y2storage/proposal/lvm_helper.rb
@@ -1,4 +1,4 @@
-#eCopyright (c) [2015] SUSE LLC
+# Copyright (c) [2015] SUSE LLC
 #
 # All Rights Reserved.
 #
--- a/src/lib/y2storage/proposal_settings.rb
+++ b/src/lib/y2storage/proposal_settings.rb
@@ -29,6 +29,7 @@ require "y2storage/volume_specifications_set"
 require "y2storage/encryption_method"
 require "y2storage/equal_by_instance_variables"
 require "y2storage/proposal_space_settings"
+require "y2storage/storage_env"
 
 module Y2Storage
   # Class to manage settings used by the proposal (typically read from control.xml)
--- a/src/lib/y2storage/proposal_settings.rb
+++ b/src/lib/y2storage/proposal_settings.rb
@@ -424,7 +424,6 @@ module Y2Storage
       load_feature(:proposal, :other_delete_mode)
       load_feature(:proposal, :delete_resize_configurable)
       load_feature(:proposal, :lvm_vg_strategy)
-      load_feature(:proposal, :lvm_vg_reuse)
       load_feature(:proposal, :allocate_volume_mode)
       load_feature(:proposal, :multidisk_first)
       load_size_feature(:proposal, :lvm_vg_size)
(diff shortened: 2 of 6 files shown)


src/lib/y2storage/proposal_settings.rb
```diff
@@ -414,6 +414,7 @@ def load_features
       load_feature(:proposal, :other_delete_mode)
       load_feature(:proposal, :delete_resize_configurable)
       load_feature(:proposal, :lvm_vg_strategy)
+      load_feature(:proposal, :lvm_vg_reuse)
```
ancorgs, line 417: This `load_feature` is for reading values from the YaST control file. It makes no sense to add `lv_vg_reuse` if we are not adding it to the list of settings that can be configured through the control file.

src/lib/y2storage/proposal/lvm_helper.rb
```diff
@@ -166,9 +167,9 @@ def new_volume_group
       # @return [Boolean]
       def try_to_reuse?
         # Setting introduced to completely avoid LVM reusing in D-Installer.
-        # It's a new playground, so no need to carry YaST behaviors that have
-        # proven to be confusing.
-        return false if settings.lvm_vg_reuse == false
+        # However it happened to be handy even in old yast later on. Preset
+        # settings can get overwritten by Linuxrc boot param.
+        return false if StorageEnv.instance.reuse_lvm? == false || settings.lvm_vg_reuse == false
```
ancorgs, line 172: I would not short-circuit things here. We already have a well defined `settings.lvm_vg_reuse` that should be honored. This makes things hard to debug (inconsistency between the setting and the behavior).

What I would do instead is relying on `StorageEnv.instance.reuse_lvm?` to set the default value of `settings.lvm_vg_reuse`. Unsurprisingly defaults are applied at `ProposalSettings#apply_defaults`.

src/lib/y2storage/proposal_settings.rb
mchf: i know ... i just took the opportunity when already touching it. It might happen that we will want to defer the behavior in the future products

src/lib/y2storage/proposal/lvm_helper.rb
mchf: I've relocated of ```lvm_vg_reuse``` default setup.

Idea is that it should
1) be set by default (```DEFAULT```) to true or
2) set according to what is in control file (added in this PR, if we decide to use it in the future) or
3) set it to value explicitly requested by a user if user set something, keep it as is if user set nothing

src/lib/y2storage/proposal/lvm_helper.rb
```diff
@@ -21,6 +21,7 @@
 require "y2storage/disk_size"
 require "y2storage/exceptions"
 require "y2storage/planned/lvm_vg"
+require "y2storage/storage_env"
```
ancorgs, line 24: Leftover

src/lib/y2storage/proposal_settings.rb
```diff
@@ -401,6 +402,14 @@ def apply_defaults
       end
     end
 
+    # Some values can be explicitly enforced by user
+    # This setting should have precendence over everything else
+    def apply_user_enforced
+      value = StorageEnv.instance.reuse_lvm?
```
ancorgs, line 408: Now I miss the `require "y2storage/storage_env"` at the top of this file.

src/lib/y2storage/proposal_settings.rb
ancorgs: So you propose to document the new value in the control file documentation and all that. Isn't it?

Because adding support in the code without making it fully official (mentioning it in the feature, adding it to the control file config, etc.) doesn't seem the way to go.

ancorgs (changes requested):
I have some doubts a

ancorgs:
Review was sent when I was still working on the comment (I clicked the wrong button, I guess).

I was saying that I like the idea of `apply_user_enforced` although I think the current implementation does not actually sets the value for the setting.

That being said, I have doubts about adding the setting to the control file. I'm not saying is a bad idea, I literally mean I have doubts whether is a good idea. On the bright side, we will be prepared for configuring the default value per-product in the future without code changes. On the other hand, adding things to the control file definition is a live-lasting contract signed in blood. :wink:

src/lib/y2storage/proposal_settings.rb
mchf: usual consequence of moving code here and there :-/ I have to test run it again.

src/lib/y2storage/storage_env.rb
```diff
@@ -89,6 +92,15 @@ def luks2_available?
       active?(ENV_LUKS2_AVAILABLE, default: false)
     end
 
+    # Whether YaST should reuse existing LVM
+    #
+    # see jsc#PED-6407 or jsc#IBM-1315
+    #
+    # @return [Boolean, nil] boolean as explicitly set by user, nil if user set nothing
+    def reuse_lvm?
+      active?(ENV_REUSE_LVM, default: nil)
```
ancorgs, line 101: One more concern. I had totally overlooked this.

A method with a name ending with a question mark but not always returning a boolean (`nil` is possible) looks totally unexpected and anti-ruby to me.

mchf:
> Review was sent when I was still working on the comment (I clicked the wrong button, I guess).
> 
> I was saying that I like the idea of `apply_user_enforced` although I think the current implementation does not actually sets the value for the setting.
> 
> That being said, I have doubts about adding the setting to the control file. I'm not saying is a bad idea, I literally mean I have doubts whether is a good idea. On the bright side, we will be prepared for configuring the default value per-product in the future without code changes. On the other hand, adding things to the control file definition is a live-lasting contract signed in blood. 😉

I had an idea that we can easily switch default behavior later on if we want to. Question is if it pays of to do it in a creepy way or we should hope in big bang. Adding a feature into control file even changing default proposal, both is radical. Which way brings us more benefits? I'm not sure ... but with control file it at least won't require our assistance in that decision (hope).

src/lib/y2storage/storage_env.rb
mchf: ok ... I'm not ruby expert. I'll rename that

src/lib/y2storage/storage_env.rb
```diff
@@ -89,6 +92,15 @@ def luks2_available?
       active?(ENV_LUKS2_AVAILABLE, default: false)
     end
 
+    # Whether YaST should reuse existing LVM
+    #
+    # see jsc#PED-6407 or jsc#IBM-1315
+    #
+    # @return [Boolean, nil] boolean as explicitly set by user, nil if user set nothing
+    def requested_lvm_reuse
+      active?(ENV_REUSE_LVM, default: nil)
```
ancorgs, line 101: Sorry to be picky about this, but the problem already mentioned about boolean and nil is here just moved one level down. :-)

Still, there is a method `active?` that returns something that is not a boolean. You are actually deviating from the original meaning of that method which explicitly declares both the return value and the parameter `default` to be booleans.

If you want to introduce support for nil values, please do it at `requested_lvm_reuse`, but don't twist `active?`.

mchf: ok, i tried and it didn't pass ... it happens ;-)

So another try is here.

src/lib/y2storage/proposal_settings.rb
mchf: Do not continue with that idea

src/lib/y2storage/storage_env.rb
```diff
@@ -89,6 +92,19 @@ def luks2_available?
...
+    #
+    # see jsc#PED-6407 or jsc#IBM-1315
+    #
+    # @return [Boolean, nil] boolean as explicitly set by user, nil if user set nothing
+    def requested_lvm_reuse
+      value = read(ENV_REUSE_LVM)
+
+      return env_str_to_bool(value) if value
+
+      nil
```
ancorgs, line 105: Nitpicking comment: I think this would do the same and it's nicer to my Ruby eyes

```ruby
def requested_lvm_reuse
  value = read(ENV_REUSE_LVM)
  return unless value

  env_str_to_bool(value)
```

How could you easily check if both implementations are equivalent? Well, I you would have unit tests... :wink:

ancorgs (approved):
In general it's good. But I really miss unit tests.

We have a `mock_env` helper that makes it easy to test how things behave depending on the presence / absence of some boot parameters.

https://github.com/search?q=repo%3Ayast%2Fyast-storage-ng%20mock_env&type=code

ancorgs (approved):
Thanks for the unit tests

PR #1366, opened 2024-02-08
Console: LUKS2 support it no longer experimental

cwickert:
## Problem

While working on the LUKS2 documentation ([PED-3882](https://jira.suse.com/browse/PED-3882)), I noticed the checkbox in the YaST console still reads "Enable Experimental LUKS2 Encryption Support". I guess with the latest work around [PED-3878](https://jira.suse.com/browse/PED-3878), we can remove the 'experimental' state.

### DIFF
--- a/src/lib/installation/console/plugins/luks2_checkbox.rb
+++ b/src/lib/installation/console/plugins/luks2_checkbox.rb
@@ -36,7 +36,7 @@ module Installation
 
         def label
           # TRANSLATORS: check box label
-          _("Enable Experimental LUKS2 Encryption Support")
+          _("Enable LUKS2 Encryption Support")
         end
 
         def store


ancorgs:
Thanks. It was very kind of you to open the pull request. But...

In fact, for interactive installation nothing has really changed since the string was introduced. Unless you are using AutoYaST, you still need either a special boot parameter or tweaking a setting in the advanced console, so the feature is still experimental (ie. we only want people to use it if they know what they are doing).

I expect changes in this area in the short term, anyways. So the string may disappear soon (if we remove the need to use the boot parameter).

Last but not least, changing user-visible strings has many implications because it affects translations. The new string wouldn't be translated so it would be displayed in English until the translators of the corresponding languages have the time to react, commit and integrate the corresponding translations. So we only change user-visible strings when they are clearly wrong or lead to misunderstandings.

PR #1370, opened 2024-02-20
Set different encryption types (luks1, luks2) via product proposal/encryption settings

schubi2:
Due
https://github.com/yast/yast-installation/pull/1109
the user can set systemd-boot und luks2 for the proposal.
The communication will be done via the product description (partitioning/proposal/encryption)
This has the advantage that the product can also define partitions which are encrypted.

Other associated PRs :
https://github.com/yast/yast-installation-control/pull/125
https://github.com/yast/yast-bootloader/pull/694

### DIFF
--- a/src/lib/y2storage/proposal_settings.rb
+++ b/src/lib/y2storage/proposal_settings.rb
@@ -30,6 +30,7 @@ require "y2storage/encryption_method"
 require "y2storage/equal_by_instance_variables"
 require "y2storage/proposal_space_settings"
 require "y2storage/storage_env"
+require "y2storage/pbkd_function"
 
 module Y2Storage
   # Class to manage settings used by the proposal (typically read from control.xml)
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/proposal_settings.rb
```diff
@@ -441,20 +442,39 @@ def load_features
...
+    # expert dialog
     def load_encryption
       enc = feature(:proposal, :encryption)
-
       return unless enc
-      return unless enc.respond_to?(:password)
 
-      passwd = enc.password.to_s
-      return if passwd.nil? || passwd.empty?
+      if enc.is_a?(Hash)
```
ancorgs, line 466: I have just started the review, but this smells bad. The fact that `feature(:proposal, :encryption)` can return different data types depending on who decided the settings looks quite hacky.

src/lib/y2storage/proposal_settings.rb
```diff
@@ -441,20 +442,39 @@ def load_features
       load_encryption
     end
 
+    def load_encryption_hash(enc)
+      passwd = enc["password"]
+      self.encryption_password = passwd if !passwd.nil? && !passwd.empty?
+
+      type = enc["type"]
```
ancorgs, line 449: Being a bit nitpicking, I would not use "type" as key for that entry of the hash. You are using it to set the value of `Encryption#encryption_method`, which is fully correct. But there is another attribute `Encryption#type` which has a different set of values and can be easily be confused.

src/lib/y2storage/proposal_settings.rb
ancorgs: After a closer inspection, now I see why you did it. The way in which the Common Criteria role injects the password into the settings is certainly hacky.

That being said, I fear injecting the password as a plain value is also quite a bad idea.

schubi2:
It is obsolete. Closing.

PR #1372, opened 2024-02-26
[GuidedProposal] Support for more Agama-based scenarios

ancorgs:
## Problem

At #1335, we introduced changes to make it possible to force the location of certain volumes, even when the `allocate_volume_mode` was set to `auto`.

But there was still one combination that didn't work. When LVM was used to group the system volumes it was not possible to allocate one of the volumes in a separate partition without creating a dedicated volume group for it.

On the other hand, on recent tests we found mandatory actions to make space were not being properly executed in all cases. Mandatory actions were only applied to the disks listed at `candidate_devices`, but that was not enough after the changes introduced in the mentioned pull request #1335.

Last but not least, we wanted to make it possible for Agama to specify that a given volume should reuse an existing block device instead of creating a new partition and/or logical volume, similar to what the AutoYaST profile can do.

## Solution

The `GuidedProposal` algorithm is modified to pre-calculate an initial list of planned devices before executing the mandatory space-making actions. Thus, additional disks are not longer overlooked.

That change is performed as several steps in separate commits for easier review.

Additionally, this pull request introduces new fields `reuse_name` and `reformat` for the volumes, making it possible to reuse existing block devices or even file-systems. That depends on the previous changes in the algorithm already mentioned.

## Testing

Added new unit tests for all the Agama combinations and adapted existing ones.

## To be done

The behavior about what to keep / delete while making space is a bit surprising in some scenarios. That's not a consequence of the changes introduced in this pull request but on how the `:bigger_resize` strategy of the SpaceMaker is defined.

We need to reconsider those cases and decide how do we want SpaceMaker to act. On the bright side, now we have many test cases marked with FIXME and pointing to a Trello card that will be soon handled by the YaST Team.


src/lib/y2storage/proposal/devices_planner.rb
```diff
@@ -155,16 +155,28 @@ def planned_device(volume)
...
         end
       end
 
+      # @see #planned_device
+      #
+      # @param volume [VolumeSpecification] volume that is NOT assigned to a separate volume group
+      # @return [Boolean]
+      def planned_lv?(volume)
+        # Exceptional case: although the mode is auto, a concrete device has been specified for a
+        # volume that is not a assigned to a separate VG
```
joseivanlopez, line 151: Typo: leftover "a".

src/lib/y2storage/proposal/space_maker.rb
```diff
@@ -298,21 +295,33 @@ def force_ptables(planned_partitions)
         end
       end
 
+      # Identifiers of all the devices that should be kept as-is, either because their are
+      # part of #protected_sids or because they would affect them indirectly
+      #
+      # @param devicegraph [Devicegraph] graph used to calculate the relationship between sids
+      # @return [Array<Integer>]
+      def all_protected_sids(devicegraph)
```
joseivanlopez, line 303: Question: Could not a PV of a re-used VG be resized to make space for the booting partition?

src/lib/y2storage/volume_specification.rb
```diff
@@ -236,6 +254,20 @@ def fs_type_configurable?
       fs_types.size > 1
     end
 
+    # @see #reformat
+    #
+    # @return [Boolean]
+    def reformat?
+      reformat
```
joseivanlopez, line 261: NP: Why not an alias method?

joseivanlopez (approved):
LGTM

Thanks a lot for splitting it in separate meaningful commits. It makes the review much easier!

src/lib/y2storage/proposal/space_maker.rb
ancorgs: I added your comment to https://trello.com/c/88lh9u52/341-agama-redefine-the-biggerresize-approach-for-non-partitioned-disks where similar topics are discussed.

src/lib/y2storage/volume_specification.rb
ancorgs: No strong reason, I copied it from `Planned::CanBeFormatted`, I can change it.

src/lib/y2storage/proposal/devices_planner.rb
ancorgs: Fixed

src/lib/y2storage/volume_specification.rb
ancorgs: Fixed anyways.

joseivanlopez (approved):

PR #1373, opened 2024-03-07
Extract device description and filesystem label

joseivanlopez:
## Problem

The partitioner implements some logic to generate the content of the columns (e.g., for generating the descrition of the device, the label of a file system, etc). Such information is useful for other projects like Agama, but reusing that logic is not possible as it is because it is directly implemented in the partitioner columns (UI stuff).

## Solution

Extract the logic for generating useful information to separate classes, making possible to reuse it. 

## Testing

* Unit tests adapted.
* Tested manually

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Mar  7 16:26:04 UTC 2024 - José Iván López González <jlopez@suse.com>
+
+- Share the logic for generating the description of a device and
+  the label of a file system.
+- Needed for (gh#openSUSE/agama#1079).
+- 5.0.8
+
+-------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


test/y2partitioner/device_description_test.rb
```diff
@@ -0,0 +1,210 @@
...
+          expect(description).to include("bcache0")
+        end
+      end
+
+      context "and it is used as caching device in a bcache" do
+        let(:scenario) { "bcache1.xml" }
+        let(:device_name) { "/dev/vdb" }
+
+        it "returns 'Bcache cache'" do
+          expect(description).to include("Bcache caching")
```
dgdavid, line 205: NP: I know you just moved the code, but the line above said "returns 'Bcache cache'"

dgdavid (approved):
LGTM

ancorgs (approved):
LGTM

PR #1374, opened 2024-03-12
Don't pass -1 to unsigned long long in swig bindings (bsc#1221222)

shundhammer:
## Bugzilla

## Problem

## Cause

## Fix

If `DiskSize::unlimited` (-1) is detected in the `FakeDeviceFactory` for that function, use a very large `unsigned long long` value instead (4 EiB).

### DIFF
--- a/src/lib/y2storage/fake_device_factory.rb
+++ b/src/lib/y2storage/fake_device_factory.rb
@@ -769,7 +769,7 @@ module Y2Storage
       size = args["size"] || DiskSize.zero
       raise ArgumentError, "\"size\" missing for lvm_lv #{lv_name}" unless args.key?("size")
 
-      size = DiskSize.EiB(4) if size.unlimited? # bsc#1221222
+      size = DiskSize.EiB(4) if size.unlimited? # (2^62 bytes) bsc#1221222
 
       lvm_lv = parent.create_lvm_lv(lv_name, size)
       create_lvm_lv_stripe_parameters(lvm_lv, args)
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/fake_device_factory.rb
```diff
@@ -769,6 +769,8 @@ def create_lvm_lv(parent, args)
       size = args["size"] || DiskSize.zero
       raise ArgumentError, "\"size\" missing for lvm_lv #{lv_name}" unless args.key?("size")
 
+      size = DiskSize.EiB(4) if size.unlimited? # bsc#1221222
```
mvidner, line 772:
```suggestion
      size = DiskSize.EiB(4) if size.unlimited? # bsc#1221222, 2**62 B
```

Just making explicit the size of the number, I'm unfamiliar with such large prefixes

mvidner (changes requested):
Thinking about this, I think this is the wrong fix.
SWIG will throw an exception again once we try passing a `DiskSize.unlimited` in a different way.

Why don't we change it here, to the maximum unsigned value of `2**64 - 1` ?
https://github.com/yast/yast-storage-ng/blob/b3e7d0d6ab334f7f6484646a615fe1ad5e66744d/src/lib/y2storage/disk_size.rb#L297

shundhammer:
I think that is the new example that Arvin just wrote for this bug. Is that relevant here?

aschnell:
> I don't know; he wrote about it on Slack. Maybe it's not checked in yet; maybe it was just a quick and dirty test to see if our theory with -1 and `unsigned long long` was correct.

Yes, it was just a quick check and not checked into git.

shundhammer:
Superseded by PR #1375

PR #1375, opened 2024-03-12
Fix unlimited-sized fake device graphs (bsc#1221222)

shundhammer:
## Bugzilla

## Problem

## Cause

## Fix

Don't use _unlimited_ for LVM LVs in the YAML fake device graphs; that's unrealistic and not possible in real life anyway.

This fix is limited purely to unit test data. There is no code change.

## Alternative PR

See also PR #1374 which did not find agreement.


aschnell (approved):
LGTM

shundhammer:
Submitted manually to Factory:

https://build.opensuse.org/request/show/1157596

PR #1377, opened 2024-04-03
Fix total size

ancorgs:
## Problem

When the `GuidedProposal` uses the `:bigger_resize` strategy for making space, it receives a list of potential resize actions.

Combining LVM with a separate boot disk, it may happen that the list of resize actions include actions on partitions of the boot disks that are not needed, because there is already enough free space in the booting disk without resizing any partition.

Sometimes, that leaded to an exception in the algorithm of `Proposal::PartitionsDistributionsCalculator`.

## Solution

Take the corner case into account and return a zero in the value for "missing size" (there is no need to resize).

## Testing

- Added a new unit test
- Tested manually


joseivanlopez (approved):
LGTM

PR #1378, opened 2024-04-03
Fix unlimited-sized fake device graphs (bsc#1221222)  [SLE-15-SP6 Backport]

shundhammer:
## Target Branch

This is a backport of PR #1375 to **SLE-15-SP6**.

## Bugzilla

## Problem

## Cause

## Fix

Don't use _unlimited_ for LVM LVs in the YAML fake device graphs; that's unrealistic and not possible in real life anyway.

This fix is limited purely to unit test data. There is no code change.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Apr  3 11:23:01 UTC 2024 - Stefan Hundhammer <shundhammer@suse.com>
+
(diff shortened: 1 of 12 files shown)


shundhammer:
The CI / Tests check is red with "_NameError:   uninitialized constant Storage::UF_BCACHEFS_", which means that the docker image for this test does not yet have the latest _libstorage-ng_ with that new enum value; probably because there was a chicken-egge problem with both requiring each other. See also PR #1371 which also failed in Jenkins for that reason.

jreidinger (approved):

PR #1379, opened 2024-04-17
Optionally include the encryption status in the device description

lslezak:
## Problem

- The device labels in Agama do not include the encryption status
- In YaST it is OK because it has a dedicated column with encryption status
- In Agama we would like to have the status included directly in the label

## Solution

- Adapt the code to include the encryption status in the label

## Notes

- The encryption status is included only when explicitly requested. That allows to keep the backward compatibility with YaST and show it only in Agama.
- See related https://github.com/openSUSE/agama/pull/1155

## Testing

- Tested manually

## Screenshots

agama_encrypted_disk

### DIFF
--- a/src/lib/y2storage/device_description.rb
+++ b/src/lib/y2storage/device_description.rb
@@ -100,7 +100,7 @@ module Y2Storage
 
     # Filesystem name
     #
-    # @param filesystem [Y2Storage::BlkFilesystem]
+    # @param filesystem [Y2Storage::Filesystems::Base]
     # @return [String]
     def filesystem_label(filesystem)
       label = filesystem.type.to_human_string
@@ -223,14 +223,14 @@ module Y2Storage
           if lvm_snapshot.encrypted? && include_encryption
             # TRANSLATORS: %{origin} is replaced by an LVM logical volume name
             # (e.g., /dev/vg0/user-data)
-            _("Encrypted thin Snapshot of %{origin}")
+            _("Encrypted Thin Snapshot of %{origin}")
           else
             # TRANSLATORS: %{origin} is replaced by an LVM logical volume name
             # (e.g., /dev/vg0/user-data)
             _("Thin Snapshot of %{origin}")
           end
         elsif lvm_snapshot.encrypted? && include_encryption
-          _("Encrypted snapshot of %{origin}")
+          _("Encrypted Snapshot of %{origin}")
         # TRANSLATORS: %{origin} is replaced by an LVM logical volume name
         # (e.g., /dev/vg0/user-data)
         else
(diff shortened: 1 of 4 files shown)


lslezak:
## Questions

- Is this the right approach?
- Which devices can be actually encrypted?
- Should we check the encryption status for all kinds of devices or are there some cases where it does not make sense or is not supported? It could save some work for the translators...

joseivanlopez:
> ## Questions
> * Is this the right approach?

I think it is a valid approach.

> * Which devices can be actually encrypted?

All block devices could be encrypted, even logical volumes (Disk, Partition, Md, LvmPv, LvmLv, etc).

> * Should we check the encryption status for all kinds of devices or are there some cases where it does not make sense or is not supported? It could save some work for the translators...

In `DeviceDescription` there are two "branches": `#formatted_device_type_label` and `unformatted_device_type_label`. This PR covers the first branch, but I think we should add "Encryption" also in the cases of unformatted devices (e.g., "Encrypted PV of %s").

lslezak:
> In `DeviceDescription` there are two "branches": `#formatted_device_type_label` and `unformatted_device_type_label`. This PR covers the first branch, but I think we should add "Encryption" also in the cases of unformatted devices (e.g., "Encrypted PV of %s").

Yes, this first commit was just a PoC and first I wanted to get the answer for the question about which devices can be encrypted. I'll add it in the following commit(s).

dgdavid:
> * In YaST it is OK because it has a dedicated column with encryption status
> * In Agama we would like to have the status included directly in the label

Just thinking aloud: YaST could benefit from this approach too.

lslezak:
Possibly yes, but that's currently out of scope...

src/lib/y2storage/device_description.rb
```diff
@@ -176,11 +220,21 @@ def lvm_pv_type_label(lvm_pv)
     def snapshot_type_label(lvm_snapshot)
       label =
         if lvm_snapshot.is?(:lvm_thin_snapshot)
-          # TRANSLATORS: %{origin} is replaced by an LVM logical volumme name
-          # (e.g., /dev/vg0/user-data)
-          _("Thin Snapshot of %{origin}")
+          if lvm_snapshot.encrypted? && include_encryption
+            # TRANSLATORS: %{origin} is replaced by an LVM logical volume name
+            # (e.g., /dev/vg0/user-data)
+            _("Encrypted thin Snapshot of %{origin}")
```
joseivanlopez, line 226: I am not sure about using capital letter for "Snapshots". If so, I would use it for both Thin and Snapshots.

src/lib/y2storage/device_description.rb
```diff
@@ -176,11 +220,21 @@ def lvm_pv_type_label(lvm_pv)
...
+          else
+            # TRANSLATORS: %{origin} is replaced by an LVM logical volume name
+            # (e.g., /dev/vg0/user-data)
+            _("Thin Snapshot of %{origin}")
+          end
+        elsif lvm_snapshot.encrypted? && include_encryption
+          _("Encrypted snapshot of %{origin}")
```
joseivanlopez, line 233: Snapshots does not start with capital letter here.

joseivanlopez (approved):
LGTM

PR #1383, opened 2024-05-03
LUKS2 enablement: alternative take 

ancorgs:
## Problem

We have been discussing for some time how to make LUKS2 easily configurable (and the default option) when installing Tumbleweed with YaST.

This [relatively old comment](https://github.com/yast/yast-installation/issues/1088#issuecomment-2044736341) summarizes my proposal, which was:

  - Make LUKS2 always available in the Expert Partitioner.
  - Make the storage proposal (a.k.a. Guided Setup) more configurable via control.xml. So Tumbleweed can set the encryption method and the key derivation function to be used by the proposal (LUKS2 + PBKDF2 is the most-compatible configuration right now).
  - Always use LUKS2 + PBKDF2 as default value in the Expert Partitioner form for encryption (same rationale about compatibility).

## Solution

This pull request implements exactly that. One commit per item.

There are no user-visible changes in the installation workflow or in the Guided Setup. YaST will now simply use LUKS2 and PBKDF2 if configured to do so at control.xml and the user decided to encrypt the system.

The default encryption step at the Expert partitioner now looks like this (for all systems, it does not depend on control.xml).

default_partitioner

## Testing

I just did some preliminary manual tests of the Partitioner (see screenshot above).


lnussel:
sounds good. who can put an official review on this?

joseivanlopez (commented):
I only miss changelog and new version.

ancorgs:
> I only miss changelog and new version.

Added.

joseivanlopez (approved):
LGTM

lnussel:
how does the control.xml snippet have to look like for that?

schubi2:
E.g.:

    <partitioning>
        <expert_partitioner_warning config:type="boolean">false</expert_partitioner_warning>

        <proposal>
            <lvm config:type="boolean">false</lvm>
            <encryption_method>luks2</encryption_method>
            <encryption_pbkdf>argon2i</encryption_pbkdf>
        </proposal>

PR #1384, opened 2024-06-13
Fix device description of encrypted PV

joseivanlopez:
## Problem

The description of an encrypted physical volume is wrong: *PV of system* instead of *Encrypted PV of system*.

## Solution

Fix `ProductDescription`.

## Testing

* Added new unit test
* Tested manually

## Screenshots

Screenshot from 2024-06-13 16-20-10


lslezak (approved):
Thanks for the fix!

ancorgs (approved):
LGTM

PR #1385, opened 2024-06-14
Fix a bug when formatting a full disk as Btrfs with snapshots

ancorgs:
## Problem

This bug at Agama was reported https://github.com/openSUSE/agama/issues/1339

Turns out the root of the error was an uncaught exception at the GuidedProposal caused because YaST was trying to adjust the sizes of the root device to fulfill the Btrfs-related settings... but such a device was a disk instead of an LV or partition.

## Solution

Added the same guard we have in other parts of `DevicesPlanner` (eg. `adjust_sizes`).

## Testing

- Added a new unit test
- Tested manually with Agama. It solves the reported bug.


joseivanlopez (approved):
LGTM

PR #1386, opened 2024-07-23
Tiny internal code reorganization to ease Agama development

ancorgs:
https://github.com/openSUSE/agama/pull/1448 introduces a new approach to manage storage devices at Agama. Instead of relying on the Y2Storage proposals living in this repository, that PR introduces a new `Y2Storage::AgamaProposal` class living at the Agama repository.

That new proposal needs to:

 - Reuse the logic to detect shadowed volumes
 - Ensure the `SpaceMaker` does not clean up the attribute `#forced_ptable_type` for all devices just for the sake of it (it doesn't seem to be a clear need for such initialization).

All that could be fixed on the Agama side:

- In the first case it would be relatively easy to just duplicate the code, since it's only 5 lines (including logging).
- In the second case, the corresponding `SpaceMaker` method can be redefined by the Agama proposal to leave out the excessive initialization.

But this PR offers an alternative by modifying the code directly at the yast2-storage-ng repository, which may be considered cleaner.


joseivanlopez (approved):
LGTM

PR #1387, opened 2024-08-26
Fix drive section

joseivanlopez:
## Problem

Agama calculates an AutoYaST proposal from a given JSON config file. Such a config file is a direct conversion from XML to JSON. This implies that some symbol attributes (e.g., [type](https://doc.opensuse.org/projects/autoyast/#id-1.9.5.2.10.8.7.4) from *drive* section) are converted to string (JSON does not support symbols).

The AutoYaST proposal code works fine with strings instead of symbols, for example for the fileystem type, encryption method, etc. Except for the `type` attribute, which must be a symbol.

## Solution

Ensure the *type* attribute is a symbol when importing a hash. 

Required by https://github.com/openSUSE/agama/pull/1471.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Aug 26 14:47:40 UTC 2024 - José Iván López González <jlopez@suse.com>
+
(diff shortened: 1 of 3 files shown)


ancorgs (approved):
LGTM

PR #1392, opened 2024-09-26
Changes to allow automatic generation of LVM PVs at the AgamaProposal

ancorgs:
We want to offer new possibilities with the so-called AgamaProposal (that will be used to implement all the features described at [this document](https://github.com/agama-project/agama/blob/master/doc/auto_storage.md).

For that we need a more powerful SpaceMaker that can automatically calculate physical volumes for an arbitrary number of LVM volume groups extended over different sets of disks.

This pull request generalizes the existing SpaceMaker to gain that capability and to be less coupled with the traditional YaST GuidedProposal.

Original API

```ruby
class Y2Storage::Proposal::SpaceMaker
  # @param settings [ProposalSettings] Note the class
  def initialize(disk_analyzer, settings); end

  # @param planned_partitions [Array<Planned::Partition>]
  # @param lvm_helper [Proposal::LvmHelper]
  def provide_space(original_graph, planned_partitions, lvm_helper); end

   # @param planned_partitions [Array<Planned::Partition>]
   def prepare_devicegraph(original_graph, planned_partitions = []); end
end
```

New API.

Does not use LvmHelper or ProposalSettings. Accepts different sets of partitions and/or volume groups. Many arguments can be omitted on the calls. The caller is more explicit about what to do instead of relying on SpaceMaker to take decisions (like which disks to clean) based on the proposal settings.

```ruby
class Y2Storage::Proposal::SpaceMaker
  # @param settings [ProposalSpaceSettings] Note the class
  def initialize(disk_analyzer, settings); end

  # @param partitions [Array<Planned::Partition>]
  # @param volume_groups [Array<Planned::LvmVg>]
  # @param default_disks [Array<String>]
  def provide_space(original_graph, partitions: [], volume_groups: [], default_disks: [])
  end

  # @param disks [Array<String>]
   def prepare_devicegraph(original_graph, disks); end
end
```

The behavior is unchanged except one corner case related to how it calculates the space needed to subtract from existing partitions in order to make space. The new behavior looks more correct. Anyways, that corner case should not affect YaST.

## Bonus

This also modifies the mentioned calculation of how much a partition must be resized.

The previous algorithm was quite aggressive when LVM is involved. A perfect size would be computationally too expensive to find, but now we do two attempts, one with the original logic and another one with a logic that can still succeed but is less aggressive when there are several spaces in the disk.

## Notes

As usual, this is structured in separate logical commits to easy review.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct  8 12:21:38 UTC 2024 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- SpaceMaker: make it possible to generate physical volumes for
+  several volume groups (needed by gh#agama-project/agama#1655).
+- 5.0.20
+
+-------------------------------------------------------------------
 Mon Sep 23 11:27:55 UTC 2024 - Stefan Hundhammer <shundhammer@suse.com>
 
 - Use the newer exfatprogs instead of exfat-utils (bsc#1187854)
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-storage-ng
-Version:        5.0.19
+Version:        5.0.20
 Release:        0
 Summary:        YaST2 - Storage Configuration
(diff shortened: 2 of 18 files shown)


joseivanlopez (approved):
LGTM

PR #1395, opened 2024-11-29
explicitelly mount /mnt/run as private to avoid mount point propagations

jreidinger:
## Problem

In agama where is systemd root is mounted as sharable and it affects also default for nested mounts. Result is that default mount bind is also sharable. As consequence systemd credentials creates under /run its own mount point when user login and this mount point is as long busy as long user is logged in. Result is that /mnt cannot be clearly unmounted.

trello card: https://trello.com/c/4YxqRmZF/3944-3-agama-could-not-umount-mnt-run-after-the-installation-systemd-creds
lwn article about topic: https://lwn.net/Articles/689856/

## Solution

Make mount --bind explcitely private to avoid propagation of mount points.

## Testing

- *Tested manually*

### DIFF
--- a/src/lib/y2storage/clients/inst_prepdisk.rb
+++ b/src/lib/y2storage/clients/inst_prepdisk.rb
@@ -79,9 +79,9 @@ module Y2Storage
         mount_in_target("/proc", "proc", "-t proc")
         mount_in_target("/sys", "sysfs", "-t sysfs")
         mount_in_target(EFIVARS_PATH, "efivarfs", "-t efivarfs") if mount_efivars?
-        # systemd makes default for mount bind sharable, but it cause trouble with
-        # systemd credentials that created another mount point undef /run when someone
-        # login. Result is that later /run cannot be unmounted. There is no benefit now
+        # systemd makes default for mount bind sharable, but it causes troubles with
+        # systemd credentials that created another mount point under /run when someone
+        # logs in. Result is that later /run cannot be unmounted. There is no benefit now
         # to have /mnt/run sharable mounted with propagation of mount points.
         # For defaults see https://lwn.net/Articles/689856/
         mount_in_target("/run", "/run", "--bind --make-private")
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/clients/inst_prepdisk.rb
```diff
@@ -79,7 +79,12 @@ def commit
         mount_in_target("/proc", "proc", "-t proc")
         mount_in_target("/sys", "sysfs", "-t sysfs")
         mount_in_target(EFIVARS_PATH, "efivarfs", "-t efivarfs") if mount_efivars?
-        mount_in_target("/run", "/run", "--bind")
+        # systemd makes default for mount bind sharable, but it cause trouble with
+        # systemd credentials that created another mount point undef /run when someone
```
imobachgs, line 83:
```suggestion
        # systemd credentials that created another mount point under /run when someone
```

src/lib/y2storage/clients/inst_prepdisk.rb
```diff
@@ -79,7 +79,12 @@ def commit
...
+        # login. Result is that later /run cannot be unmounted. There is no benefit now
```
imobachgs, line 84:
```suggestion
        # logs in. Result is that later /run cannot be unmounted. There is no benefit now
```

src/lib/y2storage/clients/inst_prepdisk.rb
```diff
@@ -79,7 +79,12 @@ def commit
         mount_in_target("/proc", "proc", "-t proc")
         mount_in_target("/sys", "sysfs", "-t sysfs")
         mount_in_target(EFIVARS_PATH, "efivarfs", "-t efivarfs") if mount_efivars?
-        mount_in_target("/run", "/run", "--bind")
+        # systemd makes default for mount bind sharable, but it cause trouble with
```
imobachgs, line 82:
```suggestion
        # systemd makes default for mount bind sharable, but it causes troubles with
```

imobachgs (approved):

ancorgs (approved):
LGTM

PR #1396, opened 2024-12-13
Using BLS bootloader as default.

schubi2:
## Problem

We need BLS bootloader in the future as the default bootloader:
https://news.opensuse.org/2024/10/08/grub2-bls
BLS suggests 1GiB for boot partition:
https://github.com/yast/yast-bootloader/pull/708
So, in that case we are reserving this space.

This can be disabled with YAST_NO_BLS_BOOT=1

## Testing

- *Added a new unit test*
- *Tested manually*


src/lib/y2storage/boot_requirements_strategies/bls.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      private
+
+      # @return [VolumeSpecification]
+      def efi_volume
+        if @efi_volume.nil?
+          @efi_volume = volume_specification_for("/boot/efi")
+          # BLS needs 1GiB for boot at least
+          # https://uapi-group.org/specifications/specs/boot_loader_specification/
+          @efi_volume.min_size = DiskSize.GiB(1)
```
aplanas, line 39: only a suggestion, but maybe we can put it as 512MB for a minimal size

src/lib/y2storage/boot_requirements_checker.rb
```diff
@@ -148,6 +150,17 @@ def arch_strategy_class
       end
     end
 
+    # Whether BLS boot will be used for booting
+    #
+    # @return [Boolean]
+    def bls?
+      Yast.import "Linuxrc"
```
ancorgs, line 157: We prefer to use environment variables instead of directly reading Linuxrc arguments. For several reasons that include from easier testing and development to avoiding coupling the library to Linuxrc (for example, in SLFO this library is still used but Linuxrc is not there).

See how some env variables are recognized at https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/storage_env.rb and then used in other parts of the code. There is even a helper that will turn the variable into boolean for you checking for "1", "on" and all the usual Linuxrc ways of handling that.

If I'm not mistaken, if you use a variable name that starts with `YAST_` that will automagically be available to YaST as environment variable and it will not be persisted in the installed system. So no modification to linuxrc would be needed.

src/lib/y2storage/boot_requirements_strategies/uefi.rb
```diff
@@ -64,6 +64,17 @@ def warnings
         res + res_new
       end
 
+      private
+
+      # @return [VolumeSpecification]
```
ancorgs, line 69: Why was did moved from `protected` into `private`? The reason is not obvious to me.

test/y2storage/boot_requirements_checker_x86_test.rb
```diff
@@ -290,14 +290,31 @@
       end
 
       context "when proposing a new EFI partition" do
+
```
ancorgs, line 293: I'm surprised Rubocop didn't complain about this extra blank line. :-)

test/y2storage/boot_requirements_checker_x86_test.rb
```diff
@@ -290,14 +290,31 @@
...
         let(:efi_part) { find_vol("/boot/efi", checker.needed_partitions(target)) }
         let(:desired_efi_part) { find_vol("/boot/efi", checker.needed_partitions(:desired)) }
         # Default values to ensure proposal of EFI partition
         let(:efiboot) { true }
         let(:efi_partitions) { [] }
 
         include_examples "proposed EFI partition basics"
-        include_examples "flexible size EFI partition"
+
+        context "and BLS bootloader as default" do
```
ancorgs, line 302: To make the document more understandable I would probably use something like: "if BLS installation is not explicitly disabled". Or would simply skip this whole context and add the `include_examples` directly, since this is now the default behavior.

test/y2storage/boot_requirements_checker_x86_test.rb
```diff
@@ -290,14 +290,31 @@
...
+          before do
+            allow(Yast::Linuxrc).to receive(:InstallInf).with("NO_BLS_BOOT").and_return(0)
+          end
+
+          include_examples "EFI partition for BLS bootloaders"
+        end
+
+        context "and no BLS bootloader as default" do
```
ancorgs, line 310: Similar to the previous one. Of course, this context needs to be here (since it explain a non-default behavior) but I would phrase it as "and BLS installation is explicitly disabled".

ancorgs:
Something very important. You modified `boot_requirements_checker_x86_test.rb` but I'm pretty sure there are more architectures affected. Please update all the relevant tests and re-generate the [boot requirements document](https://github.com/yast/yast-storage-ng/blob/master/doc/boot-requirements.md). It's crucial to keep that document in sync with the code.

You can do it by running

```
rake doc:bootspecs
```

And please then read the content to verify the behavior in all architectures is what you expect. That document is our contract signed with blood (and it will also contain your blood once this PR is merged) :wink:

ancorgs:
> Something very important. You modified `boot_requirements_checker_x86_test.rb` but I'm pretty sure there are more architectures affected.

Hmmm, I was checking the document now and I found the size of the EFI partition for aarch is indeed not mentioned there. We should add it, similar to the x86 and Raspberry Pi cases which both have a subsection saying "when proposing a new EFI partition...".

BTW, as you can see in that document the EFI size for Raspberry Pi is not modified as part of this PR. Is that intentional?

ancorgs:
Another consideration. I see this new strategy reuses the original criteria of `BootRequirementsStrategies::UEFI` to reuse an existing EFI partition (if the user chose a non-aggressive approach to make space). As you can see [here](https://github.com/yast/yast-storage-ng/blob/00c076d27ef8186398cd47cccde95588a90601e5/src/lib/y2storage/boot_requirements_strategies/uefi.rb#L215), such criteria do not include the size. EFI partitions will be re-used (eg. shared with a pre-existing operating system) if they look like a sane EFI, no matter the size.

src/lib/y2storage/boot_requirements_strategies/bls.rb
schubi2: I have set it.

src/lib/y2storage/boot_requirements_checker.rb
schubi2: Good point. Espl. regarding the future of linuxrc. Fixed.

src/lib/y2storage/boot_requirements_strategies/uefi.rb
schubi2: bls.rb is derived from uefi.rb and has its own efi_volume function which should be used. Is this wrong ?

src/lib/y2storage/boot_requirements_checker.rb
```diff
@@ -127,6 +127,8 @@ def strategy_class
         BootRequirementsStrategies::NfsRoot
       elsif raspberry_pi?
         BootRequirementsStrategies::Raspi
+      elsif bls?
+        BootRequirementsStrategies::BLS
```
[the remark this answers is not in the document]
schubi2: Yes, looks better. Fixed.

test/support/proposed_partitions_examples.rb
```diff
@@ -148,6 +148,26 @@
...
+  using Y2Storage::Refinements::SizeCasts
+
+  context "when aiming for the recommended size" do
+    let(:target) { :desired }
+
+    it "requires /boot/efi to use FAT32" do
+      expect(efi_part.mkfs_options).to include "-F32"
+    end
+
+    it "requires it to be at least 1 GiB" do
```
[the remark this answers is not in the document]
schubi2: Wow, what a great call "rake doc:bootspecs" ! I have fixed it.

test/y2storage/boot_requirements_checker_x86_test.rb
schubi2: Me too :-) Fixed.

src/lib/y2storage/boot_requirements_strategies/uefi.rb
ancorgs:
```ruby
class Foo
  def initialize(name)
    @name = name
  end

  def speak
    puts "I'm #{decorated_name}"
  end

  protected

  def decorated_name
    "**#{@name}**"
  end
end

class Bar < Foo
  protected

  def decorated_name
    "--#{@name}--"
  end
end

foo = Foo.new("John")
foo.speak

bar = Bar.new("John")
bar.speak
```

ancorgs:
Another general question. What's the plan for Raspberry Pi?

schubi2:
> Another general question. What's the plan for Raspberry Pi?

Currently only x86 and aarch64 will be supported.

src/lib/y2storage/boot_requirements_strategies/uefi.rb
schubi2: Yes, you are right. Thanks for the lesson. ;-)  Fixed.

test/y2storage/boot_requirements_checker_x86_test.rb
schubi2: Hm, not sure if I have done the right thing now. I have renamed it at least.

test/y2storage/boot_requirements_checker_x86_test.rb
schubi2: I have renamed it.

schubi2:
> > Something very important. You modified `boot_requirements_checker_x86_test.rb` but I'm pretty sure there are more architectures affected.
> 
> Hmmm, I was checking the document now and I found the size of the EFI partition for aarch is indeed not mentioned there. We should add it, similar to the x86 and Raspberry Pi cases which both have a subsection saying "when proposing a new EFI partition...".
> 

I have added it for aarch64.

> BTW, as you can see in that document the EFI size for Raspberry Pi is not modified as part of this PR. Is that intentional?

For the first step we are supporting x86 and aarch64 only.

schubi2:
> Another consideration. I see this new strategy reuses the original criteria of `BootRequirementsStrategies::UEFI` to reuse an existing EFI partition (if the user chose a non-aggressive approach to make space). As you can see [here](https://github.com/yast/yast-storage-ng/blob/00c076d27ef8186398cd47cccde95588a90601e5/src/lib/y2storage/boot_requirements_strategies/uefi.rb#L215), such criteria do not include the size. EFI partitions will be re-used (eg. shared with a pre-existing operating system) if they look like a sane EFI, no matter the size.

Hm, I am not sure what's the right way here. I would allow it. But I am not an expert here. What's your opinion ?

schubi2:
I am not sure, if I have fit all your requirements. Could you please check it again ?

schubi2:
@ancorgs Ancor, how to go on here ? Do I have to do anything ?

ancorgs (approved):
Code-wise it looks good to me. Let's release this and see what happens.

PR #1397, opened 2024-12-30
Convert timestamp to UTC

bmwiedemann:
Convert timestamp to UTC
to make parsing unambiguous.

Without this change, osc build mis-parsed this as another date
which breaks reproducible builds

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -2288,7 +2288,7 @@ Fri Oct 26 11:37:48 UTC 2018 - snwint@suse.com
   partitioner
 
 -------------------------------------------------------------------
-Sat Oct 20 22:10:48 WEST 2018 - igonzalezsosa@suse.com
+Sat Oct 20 21:10:48 UTC 2018 - igonzalezsosa@suse.com
 
 - Fixes and improvements to AutoYaST partitioning:
   - Improve support to reuse a disk as a PV (bsc#1107298).


dgdavid (commented):
Hi!

Thanks a lot for the requested change.

I'm not against of merging it, but I'd like to know a bit more about the rationale and/or problem before going ahead :sweat_smile:

Probably due to my knowledge of OBS internals, but according to the PR description it looks a bit weird to me that only the changed WEST entry is the problematic one since there are a bunch of CEST entries in the changes file.

Moreover, there are (not too much) WEST and CEST entries across the YaST modules. See  https://github.com/search?q=org%3Ayast+%22+WEST+%22+path%3A*.changes&type=code and https://github.com/search?q=org%3Ayast+%22+CEST+%22+path%3A*.changes&type=code

I've tried to find some information about the UTC _requirement_ at [OBS documentation](https://openbuildservice.org/help/) without success. So, I wonder if it is about a bug in the OBS side or just I'm missing something.

May I ask you to provide a bit more of information for a better understanding of the problem?

Thanks a lot!

jreidinger:
Well, we usually create such entries with `osc vc` so I kind of hope that osc build should parse it correctly. Maybe it should be fixed on osc build side?
But of course this PR is ok as it is and there is not problem to merge it. Just do not expect we will actively change it everywhere.

jreidinger (approved):

bmwiedemann:
I'm not sure how exactly it happened. What I observed was that on one machine (with `LC_CTYPE=en_DK.utf8; LANG=en_US.utf8`), `osc build` produced .src.rpms with .spec files that had a date that was one day earlier than what the .changes file said.

And I observed it for timestamps that had an hour >12 , including one with [`18:00 PM UTC`](build.opensuse.org/request/show/1233574)

Some other yast packages in ring1 were also affected:
autoyast2, libyui, yast2, yast2-caasp, yast2-network, yast2-nfsserver, yast2-registration

bmwiedemann:
Here is some more detail on how it failed:
`/usr/lib/build/changelog2spec` uses `/usr/lib/perl5/vendor_perl/5.40.0/Time/Zone.pm` which knows about CEST, but not WEST or IST.

```
perl -e 'use Time::Zone; print tz_offset("CEST"),"\n"'
7200
perl -e 'use Time::Zone; print tz_offset("WEST"),"\n"'

```

I made a simple toolchain patch at https://github.com/openSUSE/obs-build/pull/1047 so we don't need to patch any more .changes entries.

dgdavid:
@bmwiedemann 

Thanks a ton! For both, the input about the problem and the patch to fix it! :heart:

PR #1401, opened 2025-02-12
Require libstorage bindings for the current Ruby version (bsc#1235598)

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1235598

## Problem

Package or dist upgrades may lead to confusion what Ruby version to use and break YaST.

## Fix

Require bindings for the current Ruby version

### DIFF
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-storage-ng
-Version:        5.0.24
+Version:        5.0.25
 Release:        0
 Summary:        YaST2 - Storage Configuration
 License:        GPL-2.0-only OR GPL-3.0-only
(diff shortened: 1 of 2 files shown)


package/yast2-storage-ng.spec
```diff
@@ -49,6 +49,8 @@ BuildRequires:  rubygem(%{rb_default_ruby_abi}:parallel_tests)
 Requires:       findutils
 # Encryption#use_key_file_in_commit
 Requires:       libstorage-ng-ruby >= 4.5.144
```
arvidjaar, line 51: Isn't it redundant now when it requires the exact version anyway?

shundhammer: https://bugzilla.suse.com/show_bug.cgi?id=1235598#c8

lslezak (approved):

PR #1402, opened 2025-02-14
AutoYaST part of pervasive encryption

mvidner:
## Problem

Recently was added support for PAES-encrypted volumes but only selecting the apqns and key type to be used manually.  Therefore, would be nice to be able to select the apqns and key type through the AutoYaST profile.

- https://trello.com/c/Ffc1JCwe

## Solution

Added AutoYaST support for selecting in the profile the APQNs and the pervasive encryption key type to be used.

## Testing

- *Tested manually*
- Added unit test

### DIFF
--- a/src/lib/y2storage/planned/can_be_encrypted.rb
+++ b/src/lib/y2storage/planned/can_be_encrypted.rb
@@ -57,6 +57,16 @@ module Y2Storage
       # @return [String, nil] nil or empty string to use the default cipher
       attr_accessor :encryption_cipher
 
+      # Selected APQNs to generate a new security key for pervasive encryption
+      #
+      # @return [Array<String>]
+      attr_accessor :encryption_pervasive_apqns
+
+      # Pervasive key key_type
+      #
+      # @return [String, nil] nil or empty string to use the default key type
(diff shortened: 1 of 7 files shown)


src/lib/y2storage/planned/can_be_encrypted.rb
```diff
@@ -102,7 +114,12 @@ def final_device!(plain_device)
         result = super
         if create_encryption?
           method = encryption_method || EncryptionMethod.find(:luks1)
-          result = plain_device.encrypt(method: method, password: encryption_password)
+          args = {}
+          if method.is?(:pervasive_luks2)
```
ancorgs, line 118: Making this difference is indeed the only way to make it work. But that's IMHO due to a sub-optimal implementation of the classes handling pervasive encryption (my fault, I was too focused on the Partitioner case and didn't consider the planned devices).

So please add a comment so I will try to fix it in the future. Something like "FIXME: For pervasive_luks2 the arguments need to be passed directly at #encrypt instead of being able to assign them afterwards. That's a defect on the API of that encryption method that should be fixed."

imobachgs (approved):

PR #1404, opened 2025-02-24
Discard RAM disks as candidates for installation

ancorgs:
## Problem

When executing Agama via PXE, many RAM disks are offered as candidates for installation.

See https://github.com/agama-project/agama/issues/2042

## Solution

After some offline discussion, we do not foresee any situation in which anyone would like to install on a RAM disk neither using Agama or YaST, so this pull request discards those disks as installation candidates (based on the device name, since there is nothing at the devicegraph identifying them beyond that).

## Testing

Tested manually with Agama at a PXE installation.


kobliha:
@aschnell Is there any way to identify these disks in libstorage-ng?

imobachgs (approved):

aschnell:
> @aschnell Is there any way to identify these disks in libstorage-ng?

There is an internal function is_brd but it is not exported. I could export it but in the end it also just looks at the device name.

PR #1405, opened 2025-02-27
Fine tuning BLS boot strategy proposal

mchf:
## Problem

Folowup of https://github.com/yast/yast-storage-ng/pull/1396/files#diff-d0c8ea3d63135679325c6c61b13d11f85edc06bbb141d983037c55846699e125R143

The above proposal intoduced BLS boot proposal. Only limitation was target architecture, but it doesn't fit for conservative products like SLE for now.

## Solution

Added product based condition into BLS decision proposal. BLS is proposed only for tumbleweed (including Slowroll)

## Testing

- TBD: *Added a new unit test*
- *Tested manually*

### DIFF
--- a/src/lib/y2storage/boot_requirements_checker.rb
+++ b/src/lib/y2storage/boot_requirements_checker.rb
@@ -23,6 +23,7 @@ require "y2storage/storage_manager"
 require "y2storage/storage_env"
 
 Yast.import "Arch"
+Yast.import "Product"
 
 module Y2Storage
   #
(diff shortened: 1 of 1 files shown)


ancorgs:
I'm not so sure. We usually do not hardcode product/distribution names in the source code. If something like that is needed, we usually add a configuration parameter or something like that, so each product can decide how to act.

ancorgs:
Or thinking again about it, it may be time to separate Tumbleweed and Agama-based distributions on separate development branches. Just thinking aloud.

joseivanlopez:
I would go with "First way". The code of *yast-storage-ng* is already prepared to use the `no_bls_bootloader` env variable, so let's use it. I think we can set the ENV variable in the "startup" phase of Agama.

mchf:
> I would go with "First way". The code of _yast-storage-ng_ is already prepared to use the `no_bls_bootloader` env variable, so let's use it. I think we can set the ENV variable in the "startup" phase of Agama. And I would read the value from the product config (yaml file).

yes that is my current plan. I'm just searching for nice place where to put initialization of the env variable. I have some promising place already ... but discuss it there, as it doesn't have any sense to keep this one open.

Thanks for exhausting commentary.

PR #1406, opened 2025-03-14
Supporting LUKS2 encryption by TPM2 device (jsc#PED-10703)

schubi2:
## Problem

We would like to give the user an easy way to use TPM2 devices for LUKS2 encrypted
partitions.  (jsc#PED-10703)

## Solution

The user can select the TPM2 support for encryption in the guided setup if TPM2 is available on that machine.
This decision will be shown in the "Suggested Partitioning" frame where you can change your decision regarding TPM2 again. This line will only be shown if there is at least one partition which will be encryption with LUKS2 and TPM2 is available on the machine. This is needed because the user can also manually setup the partition without guided setup.
If TPM2 support has been selected grub2-bls will be selected for the bootloader by YAST (in yast-bootloader)

The TPM2 support will be set in yast-bootloader:
https://github.com/yast/yast-bootloader/pull/713

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

tpm2_1
tpm2_2

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Mar 12 14:00:32 UTC 2025 - Stefan Schubert <schubi@suse.de>
+
+- Supporting LUKS2 encryption by TPM2 device (jsc#PED-10703).
+- 5.0.28
+
+-------------------------------------------------------------------
 Mon Feb 24 15:11:04 UTC 2025 - Ancor Gonzalez Sosa <ancor@suse.com>
 
 - Discarded RAM disks as candidate for installation
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,24 @@
 -------------------------------------------------------------------
+Mon Feb 24 15:11:04 UTC 2025 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Discarded RAM disks as candidate for installation
+  (gh#agama-project/agama#2042).
+- 5.0.27
+
+-------------------------------------------------------------------
+Fri Feb 21 13:30:05 UTC 2025 - Knut Anderssen <kanderssen@suse.com>
+
+- Added AutoYaST support for selecting the APQNs and pervasive
+  encryption key type (jsc#PED-10950).
+- 5.0.26
+
+-------------------------------------------------------------------
+Wed Feb 12 11:01:03 UTC 2025 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Require libstorage bindings for the current Ruby version (bsc#1235598)
+- 5.0.25
+
+-------------------------------------------------------------------
 Tue Jan 21 10:00:00 UTC 2025 - Stefan Schubert <schubi@suse.de>
 
 - Reserve 1GiB boot partition for grub2-bls bootloader (jsc#PED-10703).
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-storage-ng
-Version:        5.0.27
+Version:        5.0.28
 Release:        0
 Summary:        YaST2 - Storage Configuration
 License:        GPL-2.0-only OR GPL-3.0-only
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -16,7 +16,7 @@
 #
 
 Name:           yast2-storage-ng
-Version:        5.0.24
+Version:        5.0.27
 Release:        0
 Summary:        YaST2 - Storage Configuration
 License:        GPL-2.0-only OR GPL-3.0-only
@@ -49,6 +49,8 @@ BuildRequires:  rubygem(%{rb_default_ruby_abi}:parallel_tests)
 Requires:       findutils
 # Encryption#use_key_file_in_commit
 Requires:       libstorage-ng-ruby >= 4.5.144
+# Require libstorage bindings for the current Ruby version (bsc#1235598)
+Requires:       libstorage-ng-ruby-%{rb_ver}
 # Arch.has_tpm2
 Requires:       yast2 >= 5.0.13
 # Y2Packager::Repository
--- a/package/yast2-storage-ng.spec
+++ b/package/yast2-storage-ng.spec
@@ -28,8 +28,8 @@ Source:         %{name}-%{version}.tar.bz2
 # Encryption#use_key_file_in_commit
 BuildRequires:	libstorage-ng-ruby >= 4.5.144
 BuildRequires:  update-desktop-files
-# Replace PackageSystem with Package
-BuildRequires:  yast2 >= 4.4.38
+# Arch.has_tpm2
+BuildRequires:  yast2 >= 5.0.13
 BuildRequires:  yast2-devtools >= 4.2.2
 # yast/rspec/helpers.rb
 BuildRequires:  yast2-ruby-bindings >= 4.4.7
@@ -49,8 +49,8 @@ BuildRequires:  rubygem(%{rb_default_ruby_abi}:parallel_tests)
 Requires:       findutils
 # Encryption#use_key_file_in_commit
 Requires:       libstorage-ng-ruby >= 4.5.144
-# Replace PackageSystem with Package
-Requires:       yast2 >= 4.4.38
+# Arch.has_tpm2
+Requires:       yast2 >= 5.0.13
 # Y2Packager::Repository
 Requires:       yast2-packager >= 3.3.7
 # for AbortException and handle direct abort
--- a/src/lib/y2partitioner/actions/controllers/encryption.rb
+++ b/src/lib/y2partitioner/actions/controllers/encryption.rb
@@ -52,7 +52,7 @@ module Y2Partitioner
 
         # Selected APQNs to generate a new secure key for pervasive encryption
         #
-        # @return [Array<Y2Storage:.EncryptionProcesses::Apqn>]
+        # @return [Array<Y2Storage::EncryptionProcesses::Apqn>]
         attr_accessor :apqns
 
         # @return [String] Type for the new secure key for pervasive encryption
--- a/src/lib/y2storage/autoinst_profile/partition_section.rb
+++ b/src/lib/y2storage/autoinst_profile/partition_section.rb
@@ -54,6 +54,8 @@ module Y2Storage
         { name: :crypt_label },
         { name: :crypt_cipher },
         { name: :crypt_key_size },
+        { name: :crypt_pervasive_apqns },
+        { name: :crypt_pervasive_key_type },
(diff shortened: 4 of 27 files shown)


joseivanlopez:
Some initital comments:

* The infrastructure for encrypting with TPM FDE is already there and it is used by Agama, see https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/encryption_method/tpm_fde.rb#L40.
* The proposal settings already has an attribute to indicate the encryption method to use, see https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/proposal_settings.rb#L178. Setting it to `EncryptionMethod::TPM_FDE` makes the proposal to encrypt all the partitons using TPM.
* The Expert Partitioner does not offer TMP encryption yet. I would begin with adding TPM to the Expert Partitioner and then to the Guided. Otherwise, what happens when you edit a TPM encrypted partition in the Expert Partitioner?
* I am not sure if adding that switcher to the "Suggested Partitioning" summary is a good idea. For example, let say that sda1 uses LUKS1 and sda2 TPM. What happens when the option is disabled? And what about if it is enabled again?
* I left a comment in the yast2 PR because yast-storage-ng already has logic to detect whether TPM is available, see https://github.com/yast/yast-yast2/pull/1317#issuecomment-2728282221.
* There are some important limitations we must consider, see https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/encryption_method/tpm_fde.rb#L68-L73

IMHO, we need @ancorgs point of view. This feature is not that straightforward and it implies some risk.

thkukuk:
> Some initital comments:
> 
> * The infrastructure for encrypting with TPM FDE is already there and it is used by Agama, see https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/encryption_method/tpm_fde.rb#L40.

fd-tools is not upstream and not used by anything except SLES 16.0 Beta. All other distros use already or working on using the upstream systemd tools for this, like this PR or Tumbleweed/MicroOS. It requires a heavy patched grub2 bootloader, thus is not portable and limited to what a bootloader can do, especially with the crypto stuff. Customers don't like it because of the negative impacts during first boot, see business cases in https://jira.suse.com/browse/PED-10703, customers want the "upstream" solution.

> * The proposal settings already has an attribute to indicate the encryption method to use, see https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/proposal_settings.rb#L178. Setting it to `EncryptionMethod::TPM_FDE` makes the proposal to encrypt all the partitons using TPM.

Correct, you want to encrypt all partitions using TPM, everything else doesn't make any sense and does not fulfill data protection laws (see business case from the above 4 big customers).

> * The Expert Partitioner does not offer TPM encryption yet. I would begin with adding TPM to the Expert Partitioner and then to the Guided. Otherwise, what happens when you edit a TPM encrypted partition in the Expert Partitioner?

Like with several other cases e.g. for SL Micro, my suggestion would be to print a warning, that this feature is not available with Expert Partitioner.

aplanas:
>     * The infrastructure for encrypting with TPM FDE is already there and it is used by Agama, see https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/encryption_method/tpm_fde.rb#L40.

Should be noted that this is not about TPM2 FDE, is about systemd based FDE. And the code in Agama is about GRUB2 based FDE.

Both share that TPM2 can be used, but the systemd can use TPM2+PIN, FIDO2, certificates, etc.  Also one is running in user space, and other at bootloader level.  Another very important difference is that one requires BLS (grub2-bls or systemd-boot)

Sadly the name TPM2 is not descriptive.
 
>     * The proposal settings already has an attribute to indicate the encryption method to use, see https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/proposal_settings.rb#L178. Setting it to `EncryptionMethod::TPM_FDE` makes the proposal to encrypt all the partitons using TPM.

Should we need EncryptionMethod::SYSTEMD_FDE?

schubi2:
> * The Expert Partitioner does not offer TPM encryption yet. I would begin with adding TPM to the Expert Partitioner and then to the Guided. Otherwise, what happens when you edit a TPM encrypted partition in the Expert Partitioner?
> * I am not sure if adding that switcher to the "Suggested Partitioning" summary is a good idea. For example, let say that sda1 uses LUKS1 and sda2 TPM. What happens when the option is disabled? And what about if it is enabled again?
There are two cases where the TPM2 switcher is shown in the summary:
- The user has selected encryption and TPM2 support in the "guided setup" (LUKS2 encryption has been used automatically)
- The user has gone into the expert partitioner and has selected one or more partitions for LUKS2 encryption and with the same password. In all other cases (LUKS1 encryption, different passwords, no encryption) the TPM2 switcher will not be shown in the summary at all.

In all other cases all the TPM2 switcher will not be shown in the summary and TPM support is disabled.

schubi2:
This PR does not use fde-tools at all. It will be checked, if TPM2 support makes sense at all and the user can select it.
This flag will be given to yast-bootloader which suggests grub2-bls for installation. The rest will be done via sdbootutil in yast-bootloader.

schubi2:
> > Should we need EncryptionMethod::SYSTEMD_FDE?
> 
> Yes, I think so. We should add other encryption method instead of patching `ProposalSettings` and `StorageManager`. Moreover, I think the check added to yast2 should be here in yast-storage-ng, probably as part of that new SYSTEMD_FDE.

I will have a look on it, but I am not sure if this works in all cases. But let's see.....

joseivanlopez:
> * The user has gone into the expert partitioner and has selected one or more partitions for LUKS2 encryption and with the same password. In all other cases (LUKS1 encryption, different passwords, no encryption) the TPM2 switcher will not be shown in the summary at all.

So, let say the following steps:

* Go to Expert Partitioner and encrypt sda1 and sda2 using LUKS2 and the same password.
* Go back to the summary and click on enabling TPM2.
* Then click on disabling TPM2.

I think the result is sda1 and sda2 using LUKS2 encryption instead of SYSTEMD_FDE, right? In that case, I think the switcher is safe enough.

schubi2:
> > * The user has gone into the expert partitioner and has selected one or more partitions for LUKS2 encryption and with the same password. In all other cases (LUKS1 encryption, different passwords, no encryption) the TPM2 switcher will not be shown in the summary at all.
> 
> So, let say the following steps:
> 
> * Go to Expert Partitioner and encrypt sda1 and sda2 using LUKS2 and the same password.
> * Go back to the summary and click on enabling TPM2.
> * Then click on disabling TPM2.
> 
> I think the result is sda1 and sda2 using LUKS2 encryption instead of SYSTEMD_FDE, right? In that case, I think the switcher is safe enough.

Yes, that's how it is working :-)

joseivanlopez:
Thanks guys for your feedback. In short, I suggest to do the following:

* Add a new encryption method `EncryptionMethod::SYSTEMD_FDE`. I would take `TPM_FDE` as inspiration.
* Do not offer it in the Expert Partitioner. Maybe setting `#available?` to `false` could be enough, see https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/encryption_method/base.rb#L97.
* In the guided proposal, I think the suggested approach is ok (adding TPM option in the guided wizard and adding the switcher in the summary).

BTW, what about AutoYaST?

aplanas:
> BTW, what about AutoYaST?

Maybe for later? How would be AutoYaST in the context of Agama? IIUC the work done in this PR (part of it at least) can be used in Agama, but I am not sure how much of the work that we can do in AutoYaST will be reused.

joseivanlopez:
> Maybe for later? How would be AutoYaST in the context of Agama? IIUC the work done in this PR (part of it at least) can be used in Agama, but I am not sure how much of the work that we can do in AutoYaST will be reused.

Agama does not use AutoYaST, but I guess we need to add AutoYaST support for the products that are still using YaST.

schubi2:
> * Add a new encryption method `EncryptionMethod::SYSTEMD_FDE`. I would take `TPM_FDE` as inspiration.
> * Do not offer it in the Expert Partitioner. Maybe setting `#available?` to `false` could be enough, see https://github.com/joseivanlopez/yast-storage-ng/blob/master/src/lib/y2storage/encryption_method/base.rb#L97.

I fear that it does not work and I am assuming that TPM_FDE will have the same problems if it will used by YAST:
As far I have understood the encryption information will be stored in the devicegraph. The encryption object is on top of the device which has to be encrypted. The regarding EncryptionMethod will be generated here:
https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/blk_device.rb#L785 .
So, the EncryptionMethod belongs to a device. 
In our case the information to use TPM2 or not is independent from any device. (Sure, the devices have to fulfill the requirements of the TPM2 usage) So, where to store the using TPM2 information in the devicegraph ?
If I put it to each LUKS2 encrypted device this would produce problems in the the expert partitioner where SYSTEMD_FDE are not shown due the  `#available?` = `false` flag.
Somehow the data structure does not fit to our requirements.

I must admit that your suggestion, storing the information more or less in the devicegraph, has been my first approach and I failed. That's why I have set the information to use TPM2 or not on an upper level of storage-ng:

proposal_settings.rb:    attr_accessor :encryption_use_tpm2 

and / or (if no proposal) :

storage_manager.rb:    attr_accessor :encryption_use_tpm2

The bootloader will check this flag and decide how to setup TPM2 and which bootloader has to be taken for it.

Maybe I am wrong, but up to now for me that has been the clearest way.
But I am open for suggestions ;-)

> 
> BTW, what about AutoYaST?

OK, up to now we have not taken into consideration but I can add it too. 
But first of all we should be sure, which way we are going here.

aplanas:
> In our case the information to use TPM2 or not is independent from any device. (Sure, the devices have to fulfill the requirements of the TPM2 usage)

Please, forgive if I am misunderstanding this.

It is partially true that to use TPM2 / FIDO2 it is independent from any device, but this is something that we can change.

Today `sdbootutil enroll --method=XXX` will apply the same method to all the devices.  This is most of the time what the user expect.  If there are devices that sdbootutil should ignore, because they are managed by the user, then they are marked with x-sdbootutil.ignore in the /etc/crypptab

We can extend this, and support `sdbootutil enroll --method=XXX --devices=X1,X2..` so instead of enrolling all the devices using the same method, it will enroll only a subset.

Would that help?

schubi2:
Alberto and  I  have discussed this. We will investigating the next days how "sdbootutil enroll --method=XXX --devices=X1,X2.." can be done. If yes, selecting TPM2 in the expert partitioner and therefore  EncryptionMethod::SYSTEMD_FDE would makes sense.

schubi2:
> Alberto and I have discussed this. We will investigating the next days how "sdbootutil enroll --method=XXX --devices=X1,X2.." can be done. If yes, selecting TPM2 in the expert partitioner and therefore EncryptionMethod::SYSTEMD_FDE would make sense.

We have implemented it in https://github.com/yast/yast-storage-ng/pull/1409
So, we can close this PR

PR #1407, opened 2025-04-03
Add env variables to enforce test mode

ancorgs:
## Problem

To ease development of Agama, it would be really useful to have a way to fake the storage setup using a devicegraph stored in a file.

## Solution

Introduce two new environment variables that will be honored by yast-storage-ng:

- YAST_TEST_MODE limits the access to the system. YaST will create a test libstorage-ng instance instead of real one and will not try to read information from the system when asked for resize information or when searching a device.
- YAST_DEVICEGRAPH_FILE allows to specify a mocked devicegraph.

## Testing

- Tested manually during Agama development
- Unit tests adapted and expanded


dgdavid (approved):
LGTM

Thanks a ton for finally build that _little_ yet helpful _helper_ :heart:

PR #1408, opened 2025-04-10
Manage /etc/lvm/devices

ancorgs:
## Problem

It was decided that lvm2 should use system/devices feature by default in all (open)SUSE distributions (openSUSE Tumbleweed, SLES 16 and Leap 16).

Libstorage-ng was already adapted to handle that, but some adaptations were still missing in the installer.

- https://jira.suse.com/browse/PED-7355
- https://jira.suse.com/browse/PED-12241

## Solution

- Ensure libstorage-ng generates the corresponding files if it considers they are missing.
- Copy the files to the target system at the end of installation.

## Testing

- Tested manually at Tumbleweed (YaST) :heavy_check_mark: 
- Tested manually at SLES 16 (Agama) :heavy_check_mark: 
- Added unit tests

## Bonus

[This change](https://github.com/openSUSE/libstorage-ng/pull/1022) that was very recently introduced at libstorage-ng and needed the corresponding adaptation at YaST (we have unit tests to ensure we keep both lists in sync). That's what's the last commit of this pull request is about.

The libstorage-ng change is at the devel project but is still not in Factory or SLFO. But better to fix it now than to open another separate PR just for that. And our CI runs against the devel project, anyways.

### DIFF
--- a/test/y2storage/clients/finish_test.rb
+++ b/test/y2storage/clients/finish_test.rb
@@ -1,5 +1,5 @@
 #!/usr/bin/env rspec
-# Copyright (c) [2018-2019] SUSE LLC
+# Copyright (c) [2018-2025] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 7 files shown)


test/y2storage/clients/finish_test.rb
```diff
@@ -149,6 +154,40 @@
           end
         end
       end
+
+      context "if there are files at /etc/lvm/devices" do
+        around do |example|
+          @orig_root = Dir.mktmpdir
```
joseivanlopez, line 160: Are instance variables needed?

joseivanlopez (approved):
LGTM

PR #1411, opened 2025-05-23
New public function to check whether a software RAID is candidate for installation

ancorgs:
## Problem

Agama needs to check whether a software RAID can be used as candidate for installation according to the current YaST heuristics, but the method is private.

Needed for https://github.com/agama-project/agama/pull/2388

## Solution

Extract some logic to a new `DiskAnalyzer#candidate_software_raid?` public method.

### DIFF
--- a/src/lib/y2storage/disk_analyzer.rb
+++ b/src/lib/y2storage/disk_analyzer.rb
@@ -159,6 +159,23 @@ module Y2Storage
         !device.name.match?(/^\/dev\/ram\d+$/)
     end
 
+    # Checks whether the given software RAID can be considered a valid candidate for a Linux
+    # installation
+    #
+    # Apart from matching conditions of #candidate_device?, a valid software RAID candidate must
+    # either, have a partition table or do not have children.
+    #
+    # See {#candidate_disks} for extra explanations (e.g. the relevance of EFI) and for
+    # Fate/Bugzilla references.
+    #
+    # @param md [Md]
+    # @return [Boolean]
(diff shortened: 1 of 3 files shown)


src/lib/y2storage/disk_analyzer.rb
```diff
@@ -159,6 +159,23 @@ def candidate_device?(device)
...
+    #
+    # Apart from matching conditions of #candidate_device?, a valid software RAID candidate must
+    # either, have a partition table or do not have children.
+    #
+    # See {#candidate_disks} for extra explanations (e.g. the relevance of EFI) and for
+    # Fate/Bugzilla references.
+    #
+    # @param md [Md]
+    # @return [Boolean]
+    def candidate_software_raid?(md)
```
joseivanlopez, line 173: I think that having `#candidate_device?` and `#candidate_software_raid?` is a bit weird. A md could be a `#candidate_device?` but it returns false for `#candidate_software_raid?`, which makes the "candidate" meaning a bit confusing. I would propose something like this:

* `#available_device?(device)`: whether the device is visible for installation (no repos, nothing mounted).
* `#support_boot?(device)`: whether the device supports booting (true for drives, for md it depends).
* `#candidate_device?(device)`: wheter the device is available for installation and supports boot.

What do you think?

joseivanlopez (approved):
LGTM

src/lib/y2storage/disk_analyzer.rb
ancorgs: Done

PR #1412, opened 2025-07-16
Fix iSCSI driver name

joseivanlopez:
## Problem

The name of the iSCSI driver is wrong, which makes the method `#systemd_remote?` to wrongly report false for an iSCSI device. As side effect, the _netdev option is never added.

Related to https://bugzilla.suse.com/show_bug.cgi?id=1244935. 

## Solution

Use correct driver name.

## Testing

- Tested manually


ancorgs (approved):
LGTM

PR #1413, opened 2025-07-16
Fixed grub2 error messages and proposals.

schubi2:
## Problem

Storage-ng makes some checks regarding boot partition and the usage of grub2 and grub2-efi in every
case. It also suggests subvolumes especially for grub2 and grub2-efi.
Now we have two BLS bootmanager for which these actions are wrong or not needed at least.

- https://bugzilla.opensuse.org/show_bug.cgi?id=1244755

## Solution

These checks/suggestions should not be done if the preferred bootloader is systemd-boot or grub2-bls

## Testing

- Tested manually
- Added units tests


src/lib/y2storage/boot_requirements_strategies/base.rb
```diff
@@ -84,7 +85,8 @@ def needed_partitions(target)
       def warnings
         res = []
 
-        if !boot_readable_by_grub?
+        if !boot_readable_by_grub? &&
+            !Y2Storage::BootRequirementsStrategies::BLS.bls_bootloader_proposed?
```
joseivanlopez, line 89: It is strange to have a base class with knowledge about its derivated classes. I would add a new method in the base class (e.g., `#grub_warning?`), which can be redefined in `BLS` class to always return false.

src/lib/y2storage/boot_requirements_strategies/bls.rb
```diff
@@ -28,6 +31,22 @@ def initialize(*args)
         super
       end
 
+      def self.bls_bootloader_proposed?
```
joseivanlopez, line 34: As commented above, I would move this method to `Analyzer`.

src/lib/y2storage/setup_checker.rb
```diff
@@ -24,6 +24,8 @@
 require "y2storage/boot_requirements_checker"
 require "y2storage/proposal_settings"
 require "y2storage/with_security_policies"
+# memory size detection
+require "yast2/hw_detection"
```
joseivanlopez, line 28: There is a method in yast-stoage-ng to get the RAM size: `StorageManager.instance.arch.ram_size`

src/lib/y2storage/setup_checker.rb
```diff
@@ -78,7 +80,32 @@ def errors
...
+    end
+
+    # All encryption warnings detected in the setup
+    #
+    # Argion2* needs at least 4GByte momory. cryptsetup could crash with less than 4GByte
+    # RAM.(bsc#1246876)
+    #
+    # @return [Array<SetupError>]
+    def encryption_warnings
+      return [] if Yast2::HwDetection.memory >= ((4000 - 64) << 20)
```
joseivanlopez, line 93: Please, use DiskSize:


```suggestion
      return [] if DiskSize.new(StorageManager.instance.arch.ram_size) >= DiskSize.GiB(4)
```

src/lib/y2storage/volume_specification.rb
```diff
@@ -408,6 +410,19 @@ def load_features(volume_features)
     # the architecture does not support to resume from swap (i.e., for s390).
     def adjust_features
       self.adjust_by_ram = false if swap? && !resume_supported?
+
+      return unless Y2Storage::BootRequirementsStrategies::BLS.bls_bootloader_proposed?
+
+      # Removing grub2/grub2-efi specific subvolumes because they are not needed.
+      @subvolumes.delete_if do |subvol|
```
joseivanlopez, line 417: Why not to avoid adding these subvolumes in `SubvolSpecification.fallback_list` method?

joseivanlopez (commented):
First round of comments. I miss some tests about BLS, for example, to check that the grub subvolumes are not added, etc.

Anyway, I would prefer @ancorgs has a look too. He knows very well all this about booting.

src/lib/y2storage/boot_requirements_strategies/bls.rb
schubi2: Moved...

src/lib/y2storage/setup_checker.rb
schubi2: Wow, that's nice !!!!

src/lib/y2storage/setup_checker.rb
```diff
@@ -78,7 +80,32 @@ def errors
...
+        @devicegraph.encryptions
+          .select(&:supports_pbkdf?)
+          .each do |e|
+          next unless ["argon2id", "argon2i"].include?(e.pbkdf.value)
+
+          @encryption_warnings << SetupError.new(
+            message: format(_("Using %s for %s but this needs 4 GByte RAM at least."),
+              e.pbkdf.name, e.blk_device.name)
+          )
+        end
```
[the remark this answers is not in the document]
schubi2: Rubocop has set it this way. I must admit, I do not like that way too ;-)

src/lib/y2storage/volume_specification.rb
schubi2: Hm, I am not sure if this is the right way. 
fallback_list will be used if there has not been defined a suggestion in the control.xml file:
https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/subvol_specification.rb#L239
There are grub2 defines subvolumes e.g. in the TW control.xml file. These will be currently dropped with this PR.
So, it is no problem to switch the default bootloader in the control.xml file e.g. back to grub2 without taking care about the subvolumes.
But again....I could be wrong

test/y2storage/setup_errors_presenter_test.rb
```diff
@@ -49,6 +50,8 @@
 
   let(:mount_errors) { [] }
 
+  let(:encryption_errors) { [] }
```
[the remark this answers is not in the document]
schubi2: Yes, I have added them

test/y2storage/volume_specification_test.rb
```diff
@@ -411,6 +411,32 @@
...
+        end
+
+        context "and a BLS bootloader is default" do
+          before do
+            allow(Y2Storage::BootRequirementsStrategies::Analyzer).to receive(
+                                                                   :bls_bootloader_proposed?
+                                                                 ).and_return(true)
+          end
+
+          it "subvolumes contains grub2 specific entries" do
```
joseivanlopez, line 434:
```suggestion
          it "does not contain grub2 specific subvolumes" do
```

src/lib/y2storage/setup_checker.rb
```diff
@@ -78,7 +80,32 @@ def errors
...
     # @return [Array<SetupError>]
     def warnings
-      boot_warnings + product_warnings + mount_warnings + security_policy_warnings
+      boot_warnings + product_warnings + mount_warnings + security_policy_warnings + encryption_warnings
+    end
+
+    # All encryption warnings detected in the setup
+    #
+    # Argion2* needs at least 4GByte momory. cryptsetup could crash with less than 4GByte
+    # RAM.(bsc#1246876)
```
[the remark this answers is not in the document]
joseivanlopez: Sorry, I don't see the fix.

src/lib/y2storage/volume_specification.rb
joseivanlopez: I see. Probably the subvolumes related to booting should not defined by the control.xml file but by the boot strategy. So, each strategy would have a list of subvolumes, and then it adds the subvolumes if needed (similar to adding a separate mount point). But this is probably out of the scope of this PR. Could you please add a comment in that line? Explaining this option for the future.

src/lib/y2storage/volume_specification.rb
```diff
@@ -408,6 +410,22 @@ def load_features(volume_features)
     # the architecture does not support to resume from swap (i.e., for s390).
     def adjust_features
       self.adjust_by_ram = false if swap? && !resume_supported?
+
+      return unless Y2Storage::BootRequirementsStrategies::Analyzer.bls_bootloader_proposed?
+
+      # Removing grub2/grub2-efi specific subvolumes because they are not needed.
+      # Often these subvolumes are defined in the control.xml file. So, currently it makes
+      # no sense to define it in the SubvolSpecification.fallback_list method because this
+      # will be called only if no entry is in the control.xml file.
```
joseivanlopez, line 419: I meant adding a comment to indicate how to avoid removing subvolumes here and  to improve the subvolumes proposal. Something like:

```suggestion
      # Currently, the subvolumes needed for booting are directly defined in the control.xml file (or
      # provided  by the fallback list).  But such subvolumes depend on the selected boot strategy.
      # In the future, each strategy could provide its own list of subvolumes (similar to what happens now
      # with the required partitions for booting). With that, control files do not have to provide the
      # subvolumes for booting and there is no need for removing subvolumes here. 
```

PR #1414, opened 2025-07-24
Prevent failure when encrypting a partitionable (bsc#1246970)

ancorgs:
## Problem

Agama fails to calculate a storage proposal if the configuration specifies that a disk that initially contained partitions must be directly encrypted and formatted as a whole.

Do doubt pre-existing partitions should be destroyed as part of the process, but Agama struggles with that.

The problem does not happen if the disk is formatted but not encrypted.

## Solution

Handle the situation in `CanBeEncrypted` in the same what that it was already handled at `CanBeFormatted`. That is, removing descendants of the devicegraph before proceeding.

## Testing

- Added a new unit test
- Tested manually in an Agama instance


joseivanlopez (approved):
LGTM

PR #1420, opened 2026-01-29
Simplify check for TpmFde encryption method

ancorgs:
## Problem

When the TpmFde encryption method was introduced in Agama, it was only supported in a few distribution since it needed the experimental package `fde-tools`. Hence, the availability of that package was considered a condition for the availability of the encryption method.

That has been always problematic since the availability of a certain package is something that can change over time during installation. But it became a serious issue recently, because newer versions of Agama (based on the new APIv2) parallelize more operations than before. So it can happen that the availability of those package is checked even before the software stack has configured all repositories. That makes it impossible to use the TpmFde encryption method in an unattended installation.

## Solution

Stop checking for the package availability. That's something we had already agreed to do some time ago.

On the one hand, the package is not longer experimental and it should be in all (open)SUSE distributions.

On the other hand, we already have a mechanism (the storage features) to enforce the installation of a package if a certain feature is used. That is honored for TpmFde since the introduction of #1363. That mechanism is good enough to manage package dependencies for all storage technologies, so there is not reason to use a different mechanism for the TpmFde availability.

## Extra

This pull request also introduces a fix for the unit tests. They got broken by #1413, which short-circuits the volumes configuration in a rather hacky way. After that change, the unit tests only worked if executed in a certain order and if no `/etc/YaST2/ProductFeatures` file was present in the system running the tests.

Unit tests should work properly now... until we have a chance to fix that code to use a cleaner approach.

### DIFF
--- a/package/yast2-storage-ng.changes
+++ b/package/yast2-storage-ng.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jan 29 09:00:40 UTC 2026 - Ancor Gonzalez Sosa <ancor@suse.com>
+
+- Adjusted the criteria to check whether TPM-based full-disk
+  encryption is available at Agama (bsc#1257315).
+- 5.0.40
+
+-------------------------------------------------------------------
 Thu Jan 15 08:26:47 UTC 2026 - Ladislav Slezák <lslezak@suse.com>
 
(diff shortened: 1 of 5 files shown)


joseivanlopez (approved):
LGTM

PR #1423, opened 2026-05-04
Improve encryption processes

joseivanlopez:
### Problem

The encryption process methods like `#pre_commit` or `#post_commit` have a `device` parameter, allowing to access to the encryption device data. But the method `#finish_installation` has no parameters.

### Solution

Add a `device` parameter to the `#finish_installation` method. The encryption device now passes itself to the encryption process during the finish installation phase, allowing implementations to access device information if needed.

Needed for Agama, see https://github.com/agama-project/agama/pull/3448.


ancorgs (approved):
LGTM

PR #1424, opened 2026-05-06
Add TPM BLS encryption method

joseivanlopez:
Adds support for BLS-compliant encryption with TPM unlocking.

Changes

* New encryption method: `TpmBls` - BLS-compliant LUKS2 encryption with TPM unlocking via sdbootutil
* New `Tpm` class: Provides TPM detection and capability checking (replaces `Yast::Arch` calls)
* New `Sdboot` encryption process: Handles device encryption setup and TPM enrollment using sdbootutil

Currently `TmpBls` is marked as unavailable in YaST (`available? = false`) as it's primarily implemented for Agama, but may eventually replace `SystemdFde`.

### DIFF
--- a/src/lib/y2storage/encryption_method/tpm_bls.rb
+++ b/src/lib/y2storage/encryption_method/tpm_bls.rb
@@ -21,15 +21,17 @@
 
 require "y2storage/encryption_method/base"
 require "y2storage/encryption_processes/sdboot"
+require "y2storage/tpm"
 require "yast"
-
-Yast.import "Arch"
+require "yast2/execute"
 
 module Y2Storage
   module EncryptionMethod
     # BLS compliant encryption method that allows to encrypt a device using LUKS2 and configure the
     # unlocking process via the system TPM.
     class TpmBls < Base
+      include Yast
+
       def initialize
         textdomain "storage"
 
@@ -51,7 +53,7 @@ module Y2Storage
       #
       # @return [Boolean]
       def possible?
-        arch.efiboot? && Yast::Arch.has_tpm2
+        arch.efiboot? && tpm.policy_authorize_nv?
       end
 
       # Creates an encryption device for the given block device.
@@ -70,12 +72,19 @@ module Y2Storage
       private
 
       # @see Base#encryption_process
+      # @return [EncryptionProcesses::Sdboot]
       def encryption_process
         EncryptionProcesses::Sdboot.new(self)
       end
 
+      # @return [Y2Storage::Arch]
       def arch
         StorageManager.instance.arch
+      end
+
+      # @return [Y2Storage::Tpm]
+      def tpm
+        Tpm.instance
       end
     end
   end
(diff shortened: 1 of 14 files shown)


src/lib/y2storage/encryption_method/tpm_bls.rb
```diff
@@ -0,0 +1,82 @@
...
+      def available?
+        false
+      end
+
+      # Whether the target system meets the requisites to setup devices using this encryption
+      # method.
+      #
+      # @return [Boolean]
+      def possible?
+        arch.efiboot? && Yast::Arch.has_tpm2
```
ancorgs: That `has_tmp2` method is actually a check about this encryption mechanism, not a general check for the presence of TPM. We should consider moving it here.

src/lib/y2storage/encryption_processes/sdboot.rb
```diff
@@ -0,0 +1,151 @@
...
+
+require "y2storage/encryption_authentication"
+require "y2storage/encryption_processes/base"
+require "y2storage/encryption_type"
+require "yast"
+require "yast2/execute"
+
+module Y2Storage
+  module EncryptionProcesses
+    # Encryption process that allows to setup device unlocking based on the sdbootutil provied by
```
ancorgs: provied

src/lib/y2storage/encryption_processes/sdboot.rb
```diff
@@ -0,0 +1,151 @@
...
+      # @see EncryptionProcesses::Base#encryption_type
+      def encryption_type
+        EncryptionType::LUKS2
+      end
+
+      # @see Base#finish_installation
+      # @param device [Encryption]
+      def finish_installation(device)
+        # Only TPM unlucking is supported by Agama so far.
+        return unless authentication&.is?(:tpm2)
```
ancorgs: Do we need an Authentication object at all?

joseivanlopez: The idea is to reuse the encryption process `Sdboot` for fido, etc.

ancorgs: In the future. But at this point I'm not even sure if those Authentication objects introduced for the initial BLS implementation of YaST will be the way to go in the future. I would not introduce a dependency at this point with no gain.

src/lib/y2storage/encryption_processes/sdboot.rb
```diff
@@ -0,0 +1,141 @@
...
+      # @see EncryptionProcesses::Base#encryption_type
+      def encryption_type
+        EncryptionType::LUKS2
+      end
+
+      # @see Base#finish_installation
+      # @param device [Encryption]
+      def finish_installation(device)
+        # Only TPM unlocking is supported by Agama so far.
+        return unless method.is?(:tpm_bls)
```
ancorgs: Not sure if that checks makes much sense at this point. The process is only used by tpm_bls anyways. And we usually don't have guard clauses in that direction (ie. guards to abort process steps based on the method).

ancorgs (approved):
LGTM

PR #1425, opened 2026-05-08
Fix the BLS architectural mess and add BootRequirementStrategy::BlsEfi

ancorgs:
## Problem

Recently, some contributions introduced support for BLS bootloaders in YaST. But the implementation does not really comply to the [BLS specification](https://uapi-group.org/specifications/specs/boot_loader_specification/). Instead of using the partitions and mount points recommended in that document, it uses an ESP partition mounted at `/boot/efi`.

Moreover, the mentioned contributions took many shortcuts from the architectural point of view. That resulted in convoluted code that makes it very hard to modify how the boot partitions are managed in YaST and in Agama.

Now we want to introduce support for BLS bootloaders in Agama, but we want to do it in a way that is compliant with the BLS specification.

## Solution

This pull request introduces a new strategy `BootRequirementStrategy::BlsEfi` that can be used by Agama to setup the boot-related partitions in a way that is compliant with the BLS specification.

It keeps the current YaST behavior.

The pull request also includes some preliminary commits to sanitize the codebase.

## Testing

- Added new unit tests
- Tested manually with an adapted version of Agama


joseivanlopez (approved):
LGTM

PR #1427, opened 2026-05-14
Fix BLS LEGACY bootloader type

joseivanlopez:
## Problem

The bootloader type BLS LEGACY was not initialized as a BLS bootloader.

## Solution

Properly intitialize BLS LEGACY bootloader.


ancorgs (approved):
LGTM

PR #1428, opened 2026-05-15
Mount securityfs

ancorgs:
## Problem

We found yet another trace of the architectural mayhem caused during the initial BLS implementation. See #1425.

## Solution

Stop accessing the staging devicegraph and checking its content from places that should never have done that.

## Testing

- Tested manually in a system with TPM.
- Not tested in a system without TPM but that securityfs seems to be there in all modern Linux systems. And, if not, I would only expect a warning in the logs as the biggest "problem".


joseivanlopez (approved):
LGTM

PR #1429, opened 2026-05-15
Use session keyring to pass keys to sdbootutil 

ancorgs:
## Problem

During manual testing it was detected that `sdbootutil` is not able to find the keys stored in the keyring `@u` when running in Agama, even though that's the keyring that YaST is using (and supposedly is working).

## Solution

Use the keyring `@s`, which seems to work according to manual testing.

## Testing

Tested manually


joseivanlopez (approved):
LGTM

PR #1430, opened 2026-06-19
Provide more info about packages

joseivanlopez:
## Problem

Agama is going to assume all packages are available in order to avoid calling to YaST software. But Agama needs to know what packages are optional. Right now there is no way to know what packages are optional. 

Needed for https://github.com/agama-project/agama/pull/3649.

## Solution

Extend the class `StorageFeatureList`. Now it provides the method `#packages` which returns a list of `Package` objects. Agama can check if each package is optional.

NOTE: rubocop failure is not related to this PR.

### DIFF
--- a/src/lib/y2storage/feature.rb
+++ b/src/lib/y2storage/feature.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2023] SUSE LLC
+# Copyright (c) [2023-2026] SUSE LLC
 #
 # All Rights Reserved.
 #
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/feature.rb
```diff
@@ -31,6 +31,42 @@ module Y2Storage
...
+      def optional?
+        !!@optional
+      end
+
+      # Check if a package is an optional package that is unavailable.
+      # See also bsc#1039830
+      #
+      # @return [Boolean] true if package is optional and unavailable,
+      #                   false if not optional or if available.
+      def unavailable_optional?
```
jreidinger, line 65: I would prefer more user friendly name of method like `skip?`

joseivanlopez: Oh, that code is not new, I only moved it. If you don't mind, I would prefer to keep it and discuss it with the author (@ancorgs).

jreidinger: ok

jreidinger (approved):

PR #1431, opened 2026-08-11
Avoid generating too long crypt names

joseivanlopez:
## Problem

In some cases the udev ids are too long, and basing the crypt name on such ids ends up in a too long name (more than 128 characters), making `cryptsetup` to fail.

https://bugzilla.suse.com/show_bug.cgi?id=1247173

## Solution

Validate the automatic generated crypt name, ensuring the proposed name is valid. A valid crypt name ("cr_device_X") is used as fallback if none of the candidate names are valid.

### DIFF
--- a/src/lib/y2storage/encryption.rb
+++ b/src/lib/y2storage/encryption.rb
@@ -573,7 +573,7 @@ module Y2Storage
       candidates = []
       candidates << blk_device.dm_table_name unless blk_device.dm_table_name.empty?
       candidates << mount_point_to_dm_name unless mount_point.nil?
-      candidates += blk_device.udev_ids if blk_device.udev_ids.any?
+      candidates += blk_device.udev_ids
       candidates << blk_device.basename
 
       candidates.map { |c| generate_auto_dm_table_name(c) }
--- a/src/lib/y2storage/encryption.rb
+++ b/src/lib/y2storage/encryption.rb
@@ -53,6 +53,11 @@ module Y2Storage
     #   @return [Array<Encryption>] all the encryption devices in the given devicegraph
     storage_class_forward :all, as: "Encryption"
 
+    # @!method self.valid_dm_table_name?(dm_table_name)
+    #   @param dm_table_name [String]
+    #   @return [Boolean]
+    storage_class_forward :valid_dm_table_name?
+
     # @!method in_etc_crypttab?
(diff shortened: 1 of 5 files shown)


src/lib/y2storage/encryption.rb
```diff
@@ -53,6 +53,11 @@ class Encryption < BlkDevice
     #   @return [Array<Encryption>] all the encryption devices in the given devicegraph
     storage_class_forward :all, as: "Encryption"
 
+    # @!method self.valid_dm_table_name?(dm_table_name)
```
ancorgs: Shouldn't this be at `BlkDevice`?

src/lib/y2storage/encryption.rb
```diff
@@ -193,36 +198,16 @@ def dm_table_name=(name)
...
-          mount_point_to_dm_name
-        elsif blk_device.udev_ids.any?
-          blk_device.udev_ids.first
-        else
-          blk_device.basename
-        end
-
-      self.class.ensure_unused_dm_name(devicegraph, "cr_#{name}")
+      name = candidate_auto_dm_table_names.find { |c| self.class.valid_dm_table_name?(c) }
+      name || generate_auto_dm_table_name("device")
```
ancorgs: Please make `"device"` a constant with an explanation

src/lib/y2storage/encryption.rb
joseivanlopez: Yes :)

src/lib/y2storage/encryption.rb
```diff
@@ -570,6 +555,38 @@ def propagate_mount_option(fstab_option, crypttab_option, mount_points)
...
+    # a problem if they are presented to the user in the UI.
+    #
+    # Note that any change to the encryption name generation could affect to the pervasive encryption
+    # key generation, specially when probed encryption names are modified. Right now, probed names are
+    # not touched.
+    def candidate_auto_dm_table_names
+      candidates = []
+      candidates << blk_device.dm_table_name unless blk_device.dm_table_name.empty?
+      candidates << mount_point_to_dm_name unless mount_point.nil?
+      candidates += blk_device.udev_ids if blk_device.udev_ids.any?
```
ancorgs: I guess the check for `any?` is not actually needed. If none exist then none is added.

ancorgs (approved):
LGTM

